PR #16210, opened 2022-10-20
Fixed #34063 -- Made AsyncClient populate request.POST from form data body. 

shalgrim:
[Trac issue](https://code.djangoproject.com/ticket/34063)

We found that when `FakePayload` tries to `read` and is given a number of bytes that is larger than the available content, it hits an assert statement and fails.

When we (@kevswanberg @carltongibson) compared this to the WSGI `TestClient`, we noticed that it converted its `_stream` member to a `LimitedStream`, which has a more robust `read` method.

When we made this change to `asgi.py` we found that the `test_client` tests still passed

### DIFF
--- a/django/test/client.py
+++ b/django/test/client.py
@@ -14,7 +14,7 @@ from asgiref.sync import sync_to_async
 from django.conf import settings
 from django.core.handlers.asgi import ASGIRequest
 from django.core.handlers.base import BaseHandler
-from django.core.handlers.wsgi import WSGIRequest
+from django.core.handlers.wsgi import LimitedStream, WSGIRequest
 from django.core.serializers.json import DjangoJSONEncoder
 from django.core.signals import got_request_exception, request_finished, request_started
 from django.db import close_old_connections
@@ -198,7 +198,8 @@ class AsyncClientHandler(BaseHandler):
             sender=self.__class__, scope=scope
         )
         request_started.connect(close_old_connections)
-        request = ASGIRequest(scope, body_file)
+        # Wrap FakePayload body_file to allow large read() in test environment.
+        request = ASGIRequest(scope, LimitedStream(body_file, len(body_file)))
         # Sneaky little hack so that we can easily get round
         # CsrfViewMiddleware. This makes life easier, and is probably required
         # for backwards compatibility with external tests against admin views.
@@ -598,7 +599,10 @@ class AsyncRequestFactory(RequestFactory):
             body_file = request.pop("_body_file")
         else:
             body_file = FakePayload("")
-        return ASGIRequest(self._base_scope(**request), body_file)
+        # Wrap FakePayload body_file to allow large read() in test environment.
+        return ASGIRequest(
+            self._base_scope(**request), LimitedStream(body_file, len(body_file))
+        )
 
     def generic(
         self,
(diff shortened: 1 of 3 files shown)


shalgrim:
I looked at this a bit more today, and I don't know if it's the right approach. Let me try to explain.

Now, the key error occurs if we add in the new test in this PR without the other change. And what happens is you end up at [this assert statement](https://github.com/django/django/blob/5ec64fa481892747ee5ce7ec13584cc4fe53b857/django/test/client.py#L82-L84) with `num_bytes` being sent to 65_536 and so the assert fails.

However, I noticed (and wasn't able to get out of my head) that if you put a breakpoint [here](https://github.com/django/django/blob/5ec64fa481892747ee5ce7ec13584cc4fe53b857/django/test/client.py#L202) and then, in the debugger, access `request.body`, it calls [this `body` property](https://github.com/django/django/blob/5ec64fa481892747ee5ce7ec13584cc4fe53b857/django/http/request.py#L337) and the test passes.

This bothered me because in both cases, the `read` on `FakePayload` is being called, but one time it passes because it is called without a `num_bytes` value (so it is set to `None` and later gets set to `_len`) and passes, but the other time it is called with the value of 65_536.

When it gets called without `num_bytes`, it's coming from [here](https://github.com/django/django/blob/5ec64fa481892747ee5ce7ec13584cc4fe53b857/django/http/request.py#L421) in `HTTPRequest` where `self._stream` is the `FakePayload` object.

However, when you run the test as is without the breakpoint or accessing `request.body`, then `FakePayload`'s `read` gets called from [this point in `ChunkIter`](https://github.com/django/django/blob/5ec64fa481892747ee5ce7ec13584cc4fe53b857/django/http/multipartparser.py#L507). At that point `self.flo` is the `FakePayload`, so we're calling it with `num_bytes` set to 65_536. (If you've set the breakpoint and accessed `request.body` then you still reach this point except `self.flo` is `IOBytes`, which doesn't have a problem handling `num_bytes` being bigger than what is available (same as with `LimitedStream`).

So maybe the solution is to make `FakePayload` not have a problem when `num_bytes` is too big, because it seems like `ChunkIter` expects the things it uses not to have that kind of problem

carltongibson (approved):
Hi @shalgrim @kevswanberg — thanks for picking this up at DjangoCon. All I can say is what a monster! 😈 :) 

The problem occurs in two places — once with the async client and once with the request factory. (The client reimplements `request()` so a lone fix in the factory isn't sufficient.) 

I've added test cases for both GET and POST paths for both the client and the request factory. 

I've then moved the `LimitedStream` wrapper to the point where we instantiate the `ASGIRequest`. This most closely maps to `WSGIRequest`, which already itself uses `LimitedStream`. (For `WSGIRequest` that's to stop a malicious client reading the input stream beyond its own request, a concept with doesn't apply in the ASGI case.) 

I looked at moving the fix into `FakePayload`, but there's a test that depends on that raising the error from `read()`, so there'd be an amount of work more to see if that's a feasible approach. (Using `LimitedStream` here **is** fine, so I'm inclined to stick.) 

Let's have this now. Good work both: there was a lot of digging to do there! 

Hope you had fun at the sprints and will stick with us. Reach out if I can help at all 🎁

PR #16216, opened 2022-10-21
Refs #32873 -- Adjusted formatting note on USE_L10N default.

boxed:
The paragraph 

> The formatting system is enabled by default. To disable it, it's necessary to set :setting:`USE_L10N = False <USE_L10N>` in your settings file. 

was correct in Django 3, but from Django 4 and later USE_L10N is set to True by default.

I guess there will be bigger changes needed to this document in the 5.x branch.

### DIFF
--- a/docs/topics/i18n/formatting.txt
+++ b/docs/topics/i18n/formatting.txt
@@ -13,8 +13,8 @@ When it's enabled, two users accessing the same content may see dates, times and
 numbers formatted in different ways, depending on the formats for their current
 locale.
 
-The formatting system is disabled by default. To enable it, it's
-necessary to set :setting:`USE_L10N = True <USE_L10N>` in your settings file.
+The formatting system is enabled by default. To disable it, it's
+necessary to set :setting:`USE_L10N = False <USE_L10N>` in your settings file.
 
 .. note::
 


docs/topics/i18n/formatting.txt
```diff
@@ -13,8 +13,10 @@ When it's enabled, two users accessing the same content may see dates, times and
 numbers formatted in different ways, depending on the formats for their current
 locale.
 
-The formatting system is disabled by default. To enable it, it's
-necessary to set :setting:`USE_L10N = True <USE_L10N>` in your settings file.
+The formatting system is enabled by default. To disable it, it's
+necessary to set :setting:`USE_L10N = False <USE_L10N>` in your settings file. 
+The ability to turn off the formatting system is deprecated and will be removed in
+Django 5.
```
shangxiao, line 19: Personally I think it'd better to remove the paragraph altogether 🙂

boxed: I thought about it for sure. But that would cause other problems throughout the document. There is a lot of "if enabled" talk. 

But yea, for 5.0 it needs those changes.

shangxiao: oh ic

It's up to the mergers but I wonder if all "if enabled" references should just also be removed 🤔

oh well I'll let them decide 🤷‍♂️

shangxiao: Thanks for your input btw ♥️

PR #16219, opened 2022-10-21
Refs #31920 -- Added async await support for SimpleLazyObject

lirontb:
https://code.djangoproject.com/ticket/31920


django/utils/functional.py
```diff
@@ -416,7 +423,23 @@ def __init__(self, func):
...
+        else:
+            self._wrapped = self._setupfunc()
+
+    def __await__(self):
+        async def _():
+            if iscoroutinefunction(self._wrapped):
+                self._wrapped = await self._wrapped()
+            elif callable(self._wrapped):
+                self._wrapped = await sync_to_async(self._wrapped)()
+            return self._wrapped
```
andrewgodwin, line 437: I would love to see a set of comments on this code explaining exactly what it's doing and why - it took long enough to work out at the sprints how this was going to work that we should save others the trouble.

tests/utils_tests/test_simplelazyobject.py
```diff
@@ -14,3 +18,41 @@ def test_pickle(self):
         pickle.dumps(x, 0)
         pickle.dumps(x, 1)
         pickle.dumps(x, 2)
+
+    async def test_async_retrieval(self):
```
shangxiao, line 22: Just FYI the Django coding standard asks that we don't write test docstrings with "Ensure that x does y" or "Test that foo does bar" 🙂  Usually the test method should be enough to describe the conditions but if it isn't enough some comments to elaborate are welcomed.

Also writing comments that describe what the next line does are discouraged unless it's unclear (as per the document-the-why-not-the-what guidelines). I think if the code is unclear what it's testing then some elaboration in the docstring may be helpful 👍

django/utils/functional.py
lirontb: I've added some comments as requested.

PR #16220, opened 2022-10-21
Improved "rebase" example for upstream changes in working with Git docs.

SmailBestybay:
In [after upstream has changed](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/working-with-git/#after-upstream-has-changed) section, the code example does not explicitly instruct to `git rebase upstream/main`.
Although, it is mentioned in the next sentence after the code section, for someone new to Git, it is not obvious that `git rebase` alone is not enough. Hopefully this change will save some time for the next newcomer learning to contribute to Django.


shangxiao (approved):
Thanks 👍 I wholeheartedly agree 🙂

PR #16222, opened 2022-10-23
Completed test coverage for contrib.auth.forms.

mgaligniana:
Add tests and small readability improvement to complete `contrib.auth.forms` coverage.

image
image

image

image

### DIFF
--- a/django/contrib/auth/forms.py
+++ b/django/contrib/auth/forms.py
@@ -378,12 +378,11 @@ class SetPasswordForm(forms.Form):
     def clean_new_password2(self):
         password1 = self.cleaned_data.get("new_password1")
         password2 = self.cleaned_data.get("new_password2")
-        if password1 and password2:
-            if password1 != password2:
-                raise ValidationError(
-                    self.error_messages["password_mismatch"],
-                    code="password_mismatch",
-                )
+        if password1 and password2 and password1 != password2:
+            raise ValidationError(
+                self.error_messages["password_mismatch"],
+                code="password_mismatch",
+            )
         password_validation.validate_password(password2, self.user)
         return password2
 
--- a/tests/auth_tests/test_forms.py
+++ b/tests/auth_tests/test_forms.py
@@ -590,6 +590,14 @@ class AuthenticationFormTest(TestDataMixin, TestCase):
                     form.fields[field_name].widget.attrs["autocomplete"], autocomplete
                 )
 
+    def test_no_password(self):
+        data = {"username": "username"}
+        form = AuthenticationForm(None, data)
+        self.assertIs(form.is_valid(), False)
+        self.assertEqual(
+            form["password"].errors, [Field.default_error_messages["required"]]
+        )
+
 
 class SetPasswordFormTest(TestDataMixin, TestCase):
     def test_password_verification(self):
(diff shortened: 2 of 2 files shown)


django/contrib/auth/forms.py
```diff
@@ -378,12 +378,11 @@ def __init__(self, user, *args, **kwargs):
     def clean_new_password2(self):
         password1 = self.cleaned_data.get("new_password1")
         password2 = self.cleaned_data.get("new_password2")
-        if password1 and password2:
-            if password1 != password2:
-                raise ValidationError(
-                    self.error_messages["password_mismatch"],
-                    code="password_mismatch",
-                )
+        if password1 and password2 and password1 != password2:
```
shangxiao, line 381: password1 and password2 are required form fields, so I'm wondering whether the empty/None check here is superfluous? 🤔

tests/auth_tests/test_forms.py
```diff
@@ -655,6 +665,17 @@ def test_validates_password(self):
...
+        user = User.objects.get(username="testclient")
+        data = {
+            "new_password1": "new-password",
+        }
+        form = SetPasswordForm(user, data)
+        self.assertFalse(form.is_valid())
+        self.assertEqual(
+            form["new_password2"].errors, [Field.default_error_messages["required"]]
+        )
+
```
shangxiao, line 678: It may be worth also including a test for the opposite.

django/contrib/auth/forms.py
mgaligniana: Hi shangxiao! Thanks for the review.

I thought the same!

But when I ran the tests I realized that we need the two variables to raise the "didn't match error".

image

`UserCreationForm`, `SetPasswordForm` and now `AdminPasswordChangeForm` have the same condition!

tests/auth_tests/test_forms.py
```diff
@@ -1201,6 +1222,7 @@ def test_success(self, password_changed):
         self.assertEqual(password_changed.call_count, 0)
         form.save()
         self.assertEqual(password_changed.call_count, 1)
+        self.assertEqual(form.changed_data, ["password"])
```
shangxiao, line 1225: Just curious what was the rationale for including these assertions on changed_data?

mgaligniana: We are on the same page shangxiao!

At first, I just added the assertion to cover the lines inside of the `changed_data` property (as I show in the third screenshot) 

I'm still a bit confused about the use/proposed of the property but then, when I added to every `test_` inside of `AdminPasswordChangeFormTest` I realized that the prop returns `['password']` when both fields were completed, if one or two of them are empty/None it will return `[]`

tests/auth_tests/test_forms.py
mgaligniana: Of course! I like that!

What do you think about a `test_validate_fields` that would be a combination between  `AdminPasswordChangeFormTesttest_missing_passwords` and `AdminPasswordChangeFormTest.test_one_password`?

tests/auth_tests/test_forms.py
```diff
@@ -590,6 +590,16 @@ def test_html_autocomplete_attributes(self):
                     form.fields[field_name].widget.attrs["autocomplete"], autocomplete
                 )
 
+    def test_one_of_fields_was_not_given(self):
+        data = {
+            "username": "username",
+        }
+        form = AuthenticationForm(None, data)
+        self.assertFalse(form.is_valid())
+        self.assertTrue(
```
felixxm:
```suggestion
        self.assertEqual(
```

tests/auth_tests/test_forms.py
felixxm: I added an assertion when data is empty. IMO there is no need for more tests.

felixxm (commented):
@mgaligniana Thanks :+1:

mgaligniana:
Thank you @felixxm !

PR #16235, opened 2022-10-28
Fixed #34112 -- Added async interface to model methods

DevilsAutumn:
[Ticket #34112](https://code.djangoproject.com/ticket/34112)


docs/releases/4.2.txt
```diff
@@ -26,6 +26,15 @@ and only officially support the latest release of each series.
...
 ========================
 
+Asynchronous Model interface
+----------------------------
+
+``Model`` now provides an asynchronous interface for some model methods that
+user database. These are named as -per the existing synchronous operations but
+with an ``a`` prefix, for example ``asave()``, ``adelete()``, and so on.
+
+see :ref:`async-model-methods` for details.
```
adamchainz, line 36: This is a much smaller change than the async QuerySet API, so I think it should be a list item under the Models heading (line 215). Separate sections are reserved for headline features.

docs/topics/async.txt
```diff
@@ -76,7 +76,6 @@ corruption.
 Queries & the ORM
 -----------------
 
-.. versionadded:: 4.1
```
adamchainz, line 79: Undo, removing such annotations is managed by the fellows

docs/topics/async.txt
```diff
@@ -91,10 +90,23 @@ Detailed notes can be found in :ref:`async-queries`, but in short:
 * ``async for`` is supported on all QuerySets (including the output of
   ``values()`` and ``values_list()``.)
 
+Django also supports some asynchronous model methods that use the database::
+
+    async def updateBook(...):
+        book = await author.books.afirst()
+        book.pages = 20
+        await book.asave()
```
adamchainz, line 98: Always use snake_case for Python function names as per pep 8: https://peps.python.org/pep-0008/#function-and-variable-names

Also your example was not transaction-safe, so we should not recommend it. Let’s stick with a simple create (passing `using` to make it a case that `Model.objects.acreate` cannot handle)

```suggestion
    async def make_book(...):
    	book = Book(...)
    	await book.asave(using="secondary")
```

tests/async_model/models.py
```diff
@@ -0,0 +1,6 @@
+from django.db import models
```
adamchainz, line 1: I don't think we need a new test module (`async_model`) for these few tests, it’s just creating sprawl. (I know you copied from `async_queryset` - I don't think that needs its own test module either, really.)

Better to rename `async_queryset` to `async_models` - or perhaps just merge the queryset and model tests into the `async` folder, with separate test files. That test module is also small.

docs/topics/http/middleware.txt
```diff
@@ -312,7 +312,7 @@ If your middleware has both ``sync_capable = True`` and
 ``async_capable = True``, then Django will pass it the request without
 converting it. In this case, you can work out if your middleware will receive
 async requests by checking if the ``get_response`` object you are passed is a
-coroutine function, using :py:func:`asyncio.iscoroutinefunction`.
+coroutine function, using ``asyncio.iscoroutinefunction``.
```
adamchainz, line 315: Please undo. The previous form linked to the asyncio documentation. Also this file is unrelated to your change.

docs/topics/db/models.txt
```diff
@@ -861,6 +861,26 @@ code will automatically support those arguments when they are added.
...
+code - it will block up the event loop (or, more likely, Django will
+notice and raise a ``SynchronousOnlyOperation`` to stop that from
+happening).
+
+Fortunately, you can use some asynchronous model methods that use the database:
+
+- ``asave()`` works asynchronously for :meth:`~Model.save`.
+- ``adelete()`` works asynchronously for :meth:`~Model.delete`.
+- ``arefresh_from_db()`` works asynchronously for :meth:`~Model.refresh_from_db`.
+
```
adamchainz, line 883: I don't think we need this section, we already have the async topic docs to link to the reference.

Also it is inserted in the wrong place - the below heading “Executing custom SQL” is part of the previous section. The different underline characters `=` and `-` signify different heading levels: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections

adamchainz (changes requested):
Thank you, this is a great start. I have some comments below.

tests/async_model/models.py
DevilsAutumn: what name should i keep for the separate test files? Will ``test_async_model_methods.py`` and ``test_async_queryset.py`` work ?

adamchainz: Yup!

adamchainz:
The docs build is failing due to missing ref - see logs on GitHub actions.

DevilsAutumn:
> The docs build is failing due to missing ref - see logs on GitHub actions.

Fixed.

docs/topics/http/middleware.txt
adamchainz: Ah, I guess you changed this due to a failing reference with the Python 3.11 docs? This was fixed in #16234.

docs/releases/4.2.txt
```diff
@@ -236,6 +236,10 @@ Models
 * :class:`~django.db.models.functions.Now` now supports microsecond precision
   on MySQL and millisecond precision on SQLite.
 
+* ``Model`` now provides an asynchronous interface for some model methods that
+  user database. These are named as -per the existing synchronous operations
+  but with an ``a`` prefix, for example ``asave()``, ``adelete()``, and so on.
```
adamchainz, line 241: The `QuerySet` release note in 4.1 didn’t list every method because it would have been too long. But we’re only adding three methods here, so I think we should list them. Also we should use `:meth:` for references so they work as links, which is a lot more useful for readers.

I suggest:

```suggestion
* ``Model`` now provides asynchronous versions of some model methods that use
  the database, using an ``a`` prefix: :meth:`~Model.adelete`,
  :meth:`~Model.arefresh_from_db`, and :meth:`~Model.asave`.
```

tests/async/test_async_model_methods.py
```diff
@@ -0,0 +1,33 @@
...
+        cls.s2 = Book.objects.create(
+            title="The Great Gatsby",
+        )
+
+    async def test_asave(self):
+        book = await Book.objects.aget(title="David Copperfield")
+        book.pages = 10
+        await book.asave()
+        getBook = await Book.objects.aget(title="David Copperfield")
+        self.assertEqual(book.pages, getBook.pages)
```
adamchainz, line 21: Avoid refetching the instance by using its reference in `self`.

Avoid camelcase as per PEP 8.

```suggestion
        self.s1.field = 10
        await self.s1.asave()
        refetched = await SimpleModel.objects.aget()
        self.assertEqual(refetched.field, 10)
```

tests/async/test_async_model_methods.py
```diff
@@ -0,0 +1,33 @@
...
+
+class AsyncModelOperationTest(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.s1 = Book.objects.create(
+            title="David Copperfield",
+        )
+        cls.s2 = Book.objects.create(
+            title="The Great Gatsby",
+        )
```
adamchainz, line 14: Only one model instance is needed in the below tests, so let’s avoid creating two.

```suggestion
        cls.s1 = SimpleModel.objects.create(field=0)
```

tests/async/test_async_model_methods.py
```diff
@@ -0,0 +1,33 @@
...
+        book.pages = 10
+        await book.asave()
+        getBook = await Book.objects.aget(title="David Copperfield")
+        self.assertEqual(book.pages, getBook.pages)
+
+    async def test_adelete(self):
+        book = await Book.objects.aget(title="The Great Gatsby")
+        await book.adelete()
+        count = await Book.objects.acount()
+        self.assertEqual(count, 1)
```
adamchainz, line 27: As above, reuse instance from `self.s1`
```suggestion
        await self.s1.adelete()
        count = await SimpleModel.objects.acount()
        self.assertEqual(count, 0)
```

tests/async/test_async_model_methods.py
```diff
@@ -0,0 +1,33 @@
...
+
+    async def test_arefresh_from_db(self):
+        book = await Book.objects.aget(title="The Great Gatsby")
+        await Book.objects.filter(pk=book.pk).aupdate(pages=20)
+        await book.arefresh_from_db()
+        self.assertEqual(book.pages, 20)
```
adamchainz, line 33:
```suggestion
        await SimpleModel.objects.filter(pk=self.s1.pk).aupdate(field=20)
        await self.s1.arefresh_from_db()
        self.assertEqual(self.s1.field, 20)
```

tests/async/test_async_model_methods.py
```diff
@@ -0,0 +1,33 @@
+from django.test import TestCase
+
+from .models import Book
```
adamchainz, line 3:
```suggestion
from .models import SimpleModel
```

tests/async/models.py
```diff
@@ -1,5 +1,16 @@
...
 
 
 class SimpleModel(models.Model):
     field = models.IntegerField()
+    created = models.DateTimeField(default=timezone.now)
+
+
+class Book(models.Model):
+    title = models.CharField(max_length=200)
+    pages = models.IntegerField(default=0)
```
adamchainz, line 16: with the tests merged, there’s no need to create a new model, we can reuse `SimpleModel`. Each model increases the runtime a little bit of the test suite as its table needs to be created. We try to avoid this where possible.

```suggestion
```

adamchainz (changes requested):
Thanks for the update, good work on merging the modules. Here’s another round of review.

You can just use GitHub’s [Add suggestions to batch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request) and commit my suggestions in one. Then locally squash the commit together.

I haven’t checked my test changes—hopefully there are no bugs.

docs/releases/4.2.txt
DevilsAutumn: Thanks for the review @adamchainz , but the ref is missing for these methods.

DevilsAutumn:
Thanks for the reviews @adamchainz but i think i kind of messed up in squashing, I'll just close this PR and open a new one with all the updated changes. Apologies.

adamchainz:
Okay. There’s never a need to create a new PR on GitHub. You can always fix your local git however you need (even shuffling around branches temporarily, with renames) and then force push to the branch on GitHub. But that’s for next time.

PR #16250, opened 2022-11-03
Fixed #33701 -- Added fine-grained error locations to the technical 500 debug page.

Giebisch:
Solves Ticket [33701](https://code.djangoproject.com/ticket/33701)

With Python 3.11 the technical_500.html debug page now shows, where in the line the error occurred.

traceback

### DIFF
--- a/django/views/debug.py
+++ b/django/views/debug.py
@@ -1,4 +1,5 @@
 import functools
+import itertools
 import re
 import sys
 import types
@@ -15,7 +16,7 @@ from django.utils.datastructures import MultiValueDict
 from django.utils.encoding import force_str
 from django.utils.module_loading import import_string
 from django.utils.regex_helper import _lazy_re_compile
-from django.utils.version import get_docs_version
+from django.utils.version import PY311, get_docs_version
 
 # Minimal Django templates engine to render the error templates
 # regardless of the project's TEMPLATES setting. Templates are
@@ -546,6 +547,24 @@ class ExceptionReporter:
                 pre_context = []
                 context_line = "<source code not available>"
                 post_context = []
+
+            colno = tb_area_colno = ""
+            if PY311:
+                _, _, start_column, end_column = next(
+                    itertools.islice(
+                        tb.tb_frame.f_code.co_positions(), tb.tb_lasti // 2, None
+                    )
+                )
+                if start_column and end_column:
+                    underline = "^" * (end_column - start_column)
+                    spaces = " " * (start_column + len(str(lineno + 1)) + 2)
+                    colno = f"\n{spaces}{underline}"
+                    tb_area_spaces = " " * (
+                        4
+                        + start_column
+                        - (len(context_line) - len(context_line.lstrip()))
+                    )
+                    tb_area_colno = f"\n{tb_area_spaces}{underline}"
             yield {
                 "exc_cause": exc_cause,
                 "exc_cause_explicit": exc_cause_explicit,
@@ -562,6 +581,8 @@ class ExceptionReporter:
                 "context_line": context_line,
                 "post_context": post_context,
                 "pre_context_lineno": pre_context_lineno + 1,
+                "colno": colno,
+                "tb_area_colno": tb_area_colno,
             }
             tb = tb.tb_next
 
--- a/docs/releases/4.2.txt
+++ b/docs/releases/4.2.txt
@@ -161,7 +161,8 @@ Email
 Error Reporting
 ~~~~~~~~~~~~~~~
 
-* The debug page now shows :pep:`exception notes <678>` on Python 3.11+.
+* The debug page now shows :pep:`exception notes <678>` and
+  :pep:`fine-grained error locations <657>` on Python 3.11+.
 
 File Storage
 ~~~~~~~~~~~~
--- a/tests/view_tests/tests/test_debug.py
+++ b/tests/view_tests/tests/test_debug.py
@@ -766,6 +766,79 @@ class ExceptionReporterTests(SimpleTestCase):
         self.assertIn(implicit_exc.format("<p>Second exception</p>"), text)
         self.assertEqual(3, text.count("<p>Final exception</p>"))
 
+    @skipIf(
+        sys._xoptions.get("no_debug_ranges", False)
+        or os.environ.get("PYTHONNODEBUGRANGES", False),
+        "Fine-grained error locations are disabled.",
+    )
+    @skipUnless(PY311, "Fine-grained error locations were added in Python 3.11.")
+    def test_highlight_error_position(self):
+        request = self.rf.get("/test_view/")
+        try:
+            try:
+                raise AttributeError("Top level")
+            except AttributeError as explicit:
+                try:
+                    raise ValueError(mark_safe("<p>2nd exception</p>")) from explicit
+                except ValueError:
+                    raise IndexError("Final exception")
+        except Exception:
+            exc_type, exc_value, tb = sys.exc_info()
+
+        reporter = ExceptionReporter(request, exc_type, exc_value, tb)
+        html = reporter.get_traceback_html()
+        self.assertIn(
+            "<pre>                raise AttributeError(&quot;Top level&quot;)\n"
+            "                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^</pre>",
+            html,
+        )
+        self.assertIn(
+            "<pre>                    raise ValueError(mark_safe("
+            "&quot;&lt;p&gt;2nd exception&lt;/p&gt;&quot;)) from explicit\n"
+            "                         "
+            "^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^</pre>",
+            html,
+        )
+        self.assertIn(
+            "<pre>                    raise IndexError(&quot;Final exception&quot;)\n"
+            "                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^</pre>",
+            html,
+        )
+        # Pastebin.
+        self.assertIn(
+            "    raise AttributeError(&quot;Top level&quot;)\n"
+            "    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n",
+            html,
+        )
+        self.assertIn(
+            "    raise ValueError(mark_safe("
+            "&quot;&lt;p&gt;2nd exception&lt;/p&gt;&quot;)) from explicit\n"
+            "    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n",
+            html,
+        )
+        self.assertIn(
+            "    raise IndexError(&quot;Final exception&quot;)\n"
+            "    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n",
+            html,
+        )
+        # Text traceback.
+        text = reporter.get_traceback_text()
+        self.assertIn(
+            '    raise AttributeError("Top level")\n'
+            "    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n",
+            text,
+        )
+        self.assertIn(
+            '    raise ValueError(mark_safe("<p>2nd exception</p>")) from explicit\n'
+            "    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n",
+            text,
+        )
+        self.assertIn(
+            '    raise IndexError("Final exception")\n'
+            "    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n",
+            text,
+        )
+
     def test_reporting_frames_without_source(self):
         try:
             source = "def funcName():\n    raise Error('Whoops')\nfuncName()"
--- a/AUTHORS
+++ b/AUTHORS
@@ -802,6 +802,7 @@ answer newbie questions, and generally made Django that much better:
     Rachel Tobin <rmtobin@me.com>
     Rachel Willmer <http://www.willmer.com/kb/>
     Radek Švarz <https://www.svarz.cz/translate/>
+    Rafael Giebisch <rafael@giebisch-mail.de>
     Raffaele Salmaso <raffaele@salmaso.org>
     Rajesh Dhawan <rajesh.dhawan@gmail.com>
     Ramez Ashraf <ramezashraf@gmail.com>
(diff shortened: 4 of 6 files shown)


django/views/debug.py
```diff
@@ -544,6 +545,13 @@ def get_exception_traceback_frames(self, exc_value, tb):
                 pre_context = []
                 context_line = "<source code not available>"
                 post_context = []
+            spaces = "\n" + " " * (
+                len(context_line)
+                - len(context_line.lstrip())
+                + len(str(lineno + 1))
+                - 2
+            )
+            colno = spaces + traceback.format_tb(tb, limit=1)[0].split("\n")[-2]
```
felixxm, line 554: I'd prefer to use `tb.tb_frame.f_code.co_positions()` instead of cutting the traceback on Python 3.11+ .

We need to take into account that `start_column` and `end_column` are `None` when `-X no_debug_ranges` is used.

docs/releases/4.2.txt
```diff
@@ -259,7 +259,8 @@ Signals
 Templates
 ~~~~~~~~~
 
-* ...
+* The ``technical_500.html`` debug page now also highlights the columns in the 
+  line in which an error occurred.
```
felixxm, line 263: We should mention that it's only on Python 3.11+.

felixxm (commented):
@Giebisch Thanks :+1: I left comments. Also, a regression test is required.

django/views/debug.py
adamchainz: Yes please use the API rather than parsing the text that Python happens to output right now

Giebisch: Thanks for the feedback! Agreed, I'd rather use the API, but had some difficulties.
If I access `co_positions()` there are way to many tuples being returned, how do I know which one is the one I need ?

adamchainz: I don't know - you'll need to read the feature description and PEP as linked in ticket: https://docs.python.org/3.11/whatsnew/3.11.html#pep-657-fine-grained-error-locations-in-tracebacks

Giebisch: Yes, I've read through them several times, but it's still not completely clear to me. The "reference implementation" link is also broken. So any help is greatly appreciated!

ammaraskar: The relevant bits from the `co_positions` docs are:

> The iterator returns tuples containing the (start_line, end_line, start_column, end_column). The i-th tuple corresponds to the position of the source code that compiled to the i-th instruction. Column information is 0-indexed utf-8 byte offsets on the given source line.

And 

> ...
> When this occurs, some or all of the tuple elements can be None.

Giebisch: @ammaraskar Thanks for your reply! This is the part that I get. But after getting these disassembled instructions, how do I know in which one the error occured ? For any given function potentially hundreds of tuples/instructions will be returned.

ammaraskar: Aah, that is given by `tb.tb_lasti`. See this code in the traceback module: https://github.com/python/cpython/blob/f09da28768b77713566e932e912f107b6b57e8fd/Lib/traceback.py#L354

The higher level api to get this info would be to do `StackSummary.extract()` with whatever `limit` you want and then you can iterate over the summary to get `FrameSummary` objects which have a `frame_summary.colno` and `frame_summary.end_colno` fields.

Giebisch: Thanks so much! `tb.tb_lasti` is probably what I was missing. I'll get right back to work and update the PR in the next days!

django/views/debug.py
```diff
@@ -544,6 +545,17 @@ def get_exception_traceback_frames(self, exc_value, tb):
                 pre_context = []
                 context_line = "<source code not available>"
                 post_context = []
+            if PY311 and not sys._xoptions.get("no_debug_ranges", False):
```
felixxm, line 548: There is no need to check `no_debug_ranges`. `start_column` and `end_column` are `None` when `-X no_debug_ranges` is used.

django/views/debug.py
```diff
@@ -544,6 +545,17 @@ def get_exception_traceback_frames(self, exc_value, tb):
...
+                last_instr = next(
```
felixxm, line 549: Is it possible that `last_instr` is None?

django/views/debug.py
```diff
@@ -544,6 +545,17 @@ def get_exception_traceback_frames(self, exc_value, tb):
...
+                    itertools.islice(
+                        tb.tb_frame.f_code.co_positions(), tb.tb_lasti // 2, None
+                    )
+                )
+                if last_instr is not None and last_instr[2] is not None:
```
felixxm, line 554: I would encapsulate `last_instr`:
```python
_, _, start_column, end_column = last_instr
```

django/views/debug.py
Giebisch: I'm pretty sure it is not. CPython uses the exact same implementation, so that should be safe.

docs/releases/4.2.txt
```diff
@@ -259,7 +259,8 @@ Signals
 Templates
 ~~~~~~~~~
 
-* ...
+* For Python 3.11+ the ``technical_500.html`` debug page now also highlights the columns in the 
+  line in which an error occurred.
```
adamchainz, line 263: I think this should be categorized under “Error Reporting”

tests/view_tests/tests/test_debug.py
```diff
@@ -731,6 +732,31 @@ def test_reporting_of_nested_exceptions(self):
...
+                    raise ValueError(mark_safe("<p>Second exception</p>")) from explicit
+                except ValueError:
+                    raise IndexError(mark_safe("<p>Final exception</p>"))
+        except Exception:
+            # Custom exception handler, just pass it into ExceptionReporter
+            exc_type, exc_value, tb = sys.exc_info()
+
+        reporter = ExceptionReporter(request, exc_type, exc_value, tb)
+        html = reporter.get_traceback_html().replace(" ", "")
+        if PY311 and not sys._xoptions.get("no_debug_ranges", False):
```
adamchainz, line 751: Move this condition to a `unittest.skipIf` decorator on the test, because the other branch isn’t testing anything new on Python <3.11?

adamchainz (changes requested):
I tried this out locally and it’s good that it’s highlighting the general error location, but it’s not pointing at the exact error operator.

I ran a mini-app with this code:

```py
import html
import os
import sys

from django.core.wsgi import get_wsgi_application
from django.http import HttpResponse
from django.urls import path

settings.configure(
    # Dangerous: disable host header validation
    ALLOWED_HOSTS=["*"],
    # Use DEBUG=1 to enable debug mode
    DEBUG=(os.environ.get("DEBUG", "") == "1"),
    # Make this module the urlconf
    ROOT_URLCONF=__name__,
    SECRET_KEY="123",
)


def index(request):
    name = request.GET.get("name", "World")
    1 / 0
    return HttpResponse(f"Hello, {html.escape(name)}!")


urlpatterns = [
    path("", index),
]

app = get_wsgi_application()

if __name__ == "__main__":
    from django.core.management import execute_from_command_line

```

Visiting the index page errors with `ZeroDivisionError`. In the terminal, Python’s traceback shows the division operator as where the problem occurred:

```
  File "/Users/chainz/Documents/Projects/django/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/chainz/Documents/Projects/django/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/chainz/tmp/errtest/app.py", line 23, in index
    1 / 0
    ~~^~~
ZeroDivisionError: division by zero
```

The current code correctly highlights the general locations, but for the bottom frame, it doesn’t show the exact position:

Xnapper-2022-11-09-10 48 05

I think we should be able to match that?

adamchainz:
Also, the values in `co_positions()` are byte offsets, not character offsets. So we’re currently not highlighting correctly when non-ASCII chars are used on a given line:

Xnapper-2022-11-09-10 56 13

I guess you already saw the appropriate code in CPython’s `traceback.c`: https://github.com/python/cpython/blob/0124b5dd28eff7bb80eb7244e97e402a036db13b/Python/traceback.c#L812

tests/view_tests/tests/test_debug.py
felixxm: According to PEP-657, positions can also be deactivated using the `PYTHONNODEBUGRANGES` environment variable, so we should check both.

Giebisch:
@adamchainz Unfortunately it's not possible to determine the `left_end_offset` and the `right_start_offset` solely with `co_positions()`, so the code is now a bit more complicated than before. This fixes both of your issues though :)

django/views/debug.py
```diff
@@ -544,6 +545,54 @@ def get_exception_traceback_frames(self, exc_value, tb):
...
+                        tb.tb_frame.f_code.co_positions(), tb.tb_lasti // 2, None
+                    )
+                )
+                if start_column and end_column:
+                    from traceback import (
+                        StackSummary,
+                        _byte_offset_to_character_offset,
+                        _extract_caret_anchors_from_line_segment,
+                        _walk_tb_with_full_positions,
+                    )
```
adamchainz, line 561: move imports to module level

docs/releases/4.2.txt
```diff
@@ -155,7 +155,8 @@ Email
 Error Reporting
 ~~~~~~~~~~~~~~~
 
-* ...
+* For Python 3.11+ the ``technical_500.html`` debug page now also highlights the columns in the 
+  line in which an error occurred.
```
adamchainz, line 159: Re-reading this, I think we can use the same language as Python, and there’s no need to mention the template name which I imagine most users don’t know about.

Plus, we show wrap at 80 chars

```suggestion
* The debug page now shows Python 3.11’s fine-grained error locations.
```

adamchainz (commented):
Great work finding the functions to call. I guess there’s a concern that we’re using Python’s private API now, which won’t necessarily be available on other interpreters (PyPy etc.).

I’d like to see tests for the unicode handling and matching the `~~^~~` for the final line.

django/views/debug.py
felixxm: IMO it's too much, it's not worth the additional complexity and adding so many Python's internals to Django.

Giebisch: Yea I agree with @felixxm.
This is a case that shouldn't even occur on a production server, so it is better to add as few imports as possible.

Giebisch:
@adamchainz Sure, I'll add more tests soon.
I've not used PyPy or similar. What do you suggest, reimplement all functions ? I thought the standard modules are something someone can rely on.

django/views/debug.py
felixxm: @adamchainz Does it work for you? I think we should revert recent changes and only use `start_column`, `end_column` to highlight the general error location. At least until an official API is added to Python.

adamchainz: I agree. Perhaps we could open an upstream github issue to ask about this?

adamchainz:
> I've not used PyPy or similar. What do you suggest, reimplement all functions ? I thought the standard modules are something someone can rely on.

Technically only the documented parts of the standard library are what you can rely on. Practically I think PyPy and other interpreters tend to copy-paste the stdlib code, but wherever a C extension is invloved they may re-implement it.

Moreover undocumented functions may change in any release, perhaps even in a bugfix release.

django/views/debug.py
Giebisch: @felixxm @adamchainz Alright, no worries. Now only the general error location gets highlighted with `start_column` and `end_column`. I've also fixed unicode behaviour. Please check again :) Let's wait until the official API gets updated.

felixxm:
> Also, the values in `co_positions()` are byte offsets, not character offsets. So we’re currently not highlighting correctly when non-ASCII chars are used on a given line:
> 
> Xnapper-2022-11-09-10 56 13
> 
> I guess you already saw the appropriate code in CPython’s `traceback.c`: https://github.com/python/cpython/blob/0124b5dd28eff7bb80eb7244e97e402a036db13b/Python/traceback.c#L812

I don't think we can do much more for non-ASCII characters :thinking: , e.g.:
- Python's traceback:
  ```
  File "/tickets_projects/ticket_33723/ticket_33723/urls.py", line 24, in myview2
    return "✅✅✅✅✅✅ / 0"[334]
           ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^
  ```
- our traceback without the ["Unicode workaround"](https://github.com/django/django/pull/16250/files#diff-eb948fbc65f2117752334d2fc61d04bfbb635aa3d91f7e1b652376efea04e658R557-R559) (underline with the same length as in Python):
  ```
  File "/tickets_projects/ticket_33723/ticket_33723/urls.py", line 24, in myview2
    return "✅✅✅✅✅✅ / 0"[334]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ```
- our traceback with the ["Unicode workaround"](https://github.com/django/django/pull/16250/files#diff-eb948fbc65f2117752334d2fc61d04bfbb635aa3d91f7e1b652376efea04e658R557-R559):
  ```
  File "/tickets_projects/ticket_33723/ticket_33723/urls.py", line 24, in myview2
    return "✅✅✅✅✅✅ / 0"[334]
           ^^^^^^^^^^^^^^^^^
  ``` 
I'd remove the "Unicode workaround".

Giebisch:
> - our traceback with the ["Unicode workaround"](https://github.com/django/django/pull/16250/files#diff-eb948fbc65f2117752334d2fc61d04bfbb635aa3d91f7e1b652376efea04e658R557-R559):
> 
>   ```
> 
>   File "/tickets_projects/ticket_33723/ticket_33723/urls.py", line 24, in myview2
> 
>     return "✅✅✅✅✅✅ / 0"[334]
> 
>            ^^^^^^^^^^^^^^^^^
> 
>   ``` 
> 
> I'd remove the "Unicode workaround".

I would argue that this solution still looks the best, but of course it's not perfect either, so I can see why we can leave it out. I will update shortly

felixxm:
@adamchainz What do you think? :point_up:

adamchainz:
@felixxm You’re right, it does seem that Python’s unicode handling is incorrect. I hope it can be corrected.

But yes I think we can proceed without for now. Some level of location highlighting is better than none!

Giebisch:
@adamchainz Alright, I have removed it. Then that should be all for now 👍

felixxm (commented):
@Giebisch Thanks :+1: We should also add error locations to the "Pastebin" output and to the `txt` template. It's a bit tricky because they use a different indentation.  I added tests for both.

Giebisch:
@felixxm True! I fixed the docs and added error locations for pastebin and the txt template. I had to change the tests a bit, now everything works perfectly for me. But please double check if everything is correct for you too.

felixxm (commented):
@Giebisch Thanks :+1: I pushed tiny final edits.

PR #16292, opened 2022-11-12
Refs #32339 -- Updated Form API docs to prefer as_div() output style.

smithdc1:
The `as_div()` output style is recommended, and will become the default in Django 5.0. 

This PR updates the form api docs to use this style. 

I've not yet updated the outputs which just call `str` as that will still default to the table output until the switch is made in 5.0. 

A couple of items were maybe a little surprising. 

- The override of `label` doesn't work anymore (it's stored on the BoundField when the class is created). So I switched to changing the suffix. 
- `->` is escaped. Is this right?

### DIFF
--- a/docs/ref/forms/api.txt
+++ b/docs/ref/forms/api.txt
@@ -342,23 +342,23 @@ attribute::
     >>> f.fields['name']
     <django.forms.fields.CharField object at 0x7ffaac6324d0>
 
-You can alter the field of :class:`Form` instance to change the way it is
-presented in the form::
+You can alter the field and :class:`.BoundField` of :class:`Form` instance to
+change the way it is presented in the form::
 
-    >>> f.as_table().split('\n')[0]
-    '<tr><th>Name:</th><td><input name="name" type="text" value="instance" required></td></tr>'
-    >>> f.fields['name'].label = "Username"
-    >>> f.as_table().split('\n')[0]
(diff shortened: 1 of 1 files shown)


carltongibson (commented):
Hey @smithdc1  — thanks for this. It looks good. 

> The override of label doesn't work anymore (it's stored on the BoundField when the class is created). So I switched to changing the suffix.

OK, so this has been the case since the bound field cache was introduced in 65441b (2015) as far as I can see. 

It does work on a fresh instance: 

```python
>>> f = ContactForm()
>>> f.fields["subject"].label = "Testing"
>>> f.as_table().replace("\n", "").split("</tr>")[0]
'<tr><th><label for="id_subject">Testing:</label></th><td><input type="text" name="subject" maxlength="100" required id="id_subject"></td>'
```

(That's using Django 3.2, since I was checking for a regression, but you get the point.)

I wonder if it's worth showing the bound field version: 

```python
>>> f["subject"].label = "Testing Again"
>>> f.as_table().replace("\n", "").split("</tr>")[0]
'<tr><th><label for="id_subject">Testing Again:</label></th><td><input type="text" name="subject" maxlength="100" required id="id_subject"></td>'
```

I kind of feel setting the `label` makes a better example than the `label_suffix`; how about you? (Happy to go with your considered opinion.)

> -> is escaped. Is this right?

Yes. 👍


Good change. 🎁

carltongibson (approved):
Yep, great. Thanks @smithdc1! 🎁

PR #16293, opened 2022-11-12
Refs #25307 -- Replaced SQLQuery.rewrite_cols by replace_expressions.

charettes:
The latter offers a more generic interface that doesn't require specialized expression types handling.


django/db/models/sql/query.py
```diff
@@ -381,60 +381,6 @@ def _get_col(self, target, field, alias):
...
-        # Resolving this expression results in a join to author, but there
-        # is no guarantee the awards column of author is in the select clause
-        # of the query. Thus we must manually add the column to the inner
-        # query.
-        orig_exprs = annotation.get_source_expressions()
-        new_exprs = []
-        for expr in orig_exprs:
-            # FIXME: These conditions are fairly arbitrary. Identify a better
-            # method of having expressions decide which code path they should
-            # take.
```
charettes, line 401: We've got this method now :)

django/db/models/sql/query.py
```diff
@@ -508,20 +454,32 @@ def get_aggregation(self, using, added_aggregate_names):
...
-                    expression, col_cnt = inner_query.rewrite_cols(expression, col_cnt)
-                    outer_query.annotations[alias] = expression.relabeled_clone(
-                        relabels
-                    )
-                    del inner_query.annotations[alias]
-                    annotation_select_mask.remove(alias)
+                replacements = {}
+                for col in self._gen_cols([expression], resolve_refs=False):
+                    if not (col_ref := col_refs.get(col)):
+                        index = len(col_refs) + 1
```
charettes, line 471: The +1 here was to have the exact same SQL as `rewrite_cols` generated (starting `__col%d` at 1 instead of 0) but is purely cosmetic.

charettes:
buildbot, test on oracle.

PR #16310, opened 2022-11-18
Fixed #29049 -- Added Slicing Notation to F Expressions.

SirAbhi13:
[Ticket 29049](https://code.djangoproject.com/ticket/29049)

Referenced PR: #14731 

Continued the previous PR from where it was left of.
Fixed the wrong SQL query that was being generated for OuterRef.
I removed the test `test_slicing_of_f_len` because even if I was testing with sliced arrays generated through [Slice Transformations](https://docs.djangoproject.com/en/4.1/ref/contrib/postgres/fields/#slice-transforms), the error persisted, perhaps need to open another ticket for it?

Other than that PR looks good to me.


ngnpope:
Please look at my comments regarding use of `None` instead of `""` for `self.end` on the previous PR. This hasn't been considered yet.

SirAbhi13:
> Please look at my comments regarding use of `None` instead of `""` for `self.end` on the previous PR. This hasn't been considered yet.

Hi Nick,
I actually did try to implement replacing `""` with `None` for `self.end`, the way you proposed in the previous PR, but it ended up breaking the code later in the execution. So my question is should I just do it like this
```       
if self.end is None:
          return f"{lhs}[%s:%s]", params + [self.start, self.end]
      else:
          return f"{lhs}[%s:]", params + [self.start]
```
because None is not expected by compiler.py later and ends up not working.

SirAbhi13:
I did end up getting the `None` value working, turns out I was making a mistake during testing on my part.
May I know which part in the suggested changes you have reservations about?

The tests that are failing are for mysql and maria db, because they are evaluating the queries in case insensitive mode, hence returning values in testcases where empty queryset is expected. I Will have to research on how to prevent that.

ngnpope:
Hey @SirAbhi13. I think quite some time has passed so I've lost track of where I was going with that thought.

I did find a local branch with a bunch of fixes though, so I've polished that off and put up #17596 to get this over the line.

Thanks for the reminder and your efforts to help push this forward. Quite the number of co-authors on this one!

PR #16325, opened 2022-11-24
[WIP] Fixes #29574 and #34151

DevilsAutumn:
This PR solves [ticket #29574](https://code.djangoproject.com/ticket/29574) and [ticket #34151](https://code.djangoproject.com/ticket/34151) . Tested the code for cross app fk as well as reverse migration and it was working as expected for both the tickets . 

But i am unclear at some points:
- Is this the right approach to get the referenced models(and to solve the issues) ? I tried using ```model._meta.related_objects``` but it was not working .
- writing tests


DevilsAutumn:
@felixxm @charettes if you could please take a look. 🙇

naranjo-it:
Hello @DevilsAutumn . If I understand correctly this should fix the issues but only some extra tests were missing, right? It sounds promising. I'll take a look and try to solve it from here. Hoewever, if you have received feedback through other channels or have more info regarding this issue I'd be happy if you can share the updates here. Thanks!

PR #16329, opened 2022-11-25
Fixed #21604 -- Embed raw queries as subqueries

wiltso:
Added functionallity so RawQueries can be used as subquries when using __in filter


django/db/models/sql/query.py
```diff
@@ -82,7 +82,11 @@ def get_children_from_q(q):
 class RawQuery:
     """A single raw SQL query."""
 
-    def __init__(self, sql, using, params=()):
+    has_select_fields = True
+    contains_aggregate = []
+    contains_over_clause = False
+
+    def __init__(self, sql, using="default", params=()):
```
charettes, line 89: Lets keep this PR focused on ticket-21604; making `using` default to `"default"` should be tackled somewhere else.

```suggestion
    def __init__(self, sql, using, params=()):
```

tests/raw_query/tests.py
```diff
@@ -413,7 +420,14 @@ def test_bool(self):
...
     def test_len(self):
-        self.assertEqual(len(Book.objects.raw("SELECT * FROM raw_query_book")), 4)
+        self.assertEqual(len(Book.objects.raw("SELECT * FROM raw_query_book")), 5)
         self.assertEqual(
             len(Book.objects.raw("SELECT * FROM raw_query_book WHERE id = 0")), 0
         )
+
+    def test_as_subquery(self):
+        author_id_sub_query = RawQuery(
+            "SELECT id FROM raw_query_author WHERE last_name LIKE %s", params=["Smith"]
```
charettes, line 430: This should be passing `using="default"`

tests/raw_query/tests.py
```diff
@@ -413,7 +420,14 @@ def test_bool(self):
...
+        )
+        print(author_id_sub_query)
+        self.assertEqual(len(Book.objects.filter(author_id__in=author_id_sub_query)), 4)
```
charettes, line 433: This happens to work in two queries because `RawQuery` defines an `__iter__`.

django/db/models/sql/query.py
```diff
@@ -151,6 +155,9 @@ def _execute_query(self):
         self.cursor = connection.cursor()
         self.cursor.execute(self.sql, params)
 
+    def resolve_expression(self, query, *args, **kwargs):
```
charettes, line 158: The `resolve_expression` method of the _resolvable_ interface must return a _compilable_ which is an object with an `as_sql` method as mentioned in the ticket.

If the test currently pass it must be because `__iter__` is implicitly called by `In(Lookup)` and you are not actually executing the `RawQuery`'s SQL as a subquery like it should be.

What I believe should happen here is for `resolve_expression` to return copy of `self` that has a flag set to denote that it is used as a subquery `e.g. `clone = self.clone(self.using); clone.subquery = True` and an `as_sql` method that return something like `f"({self.sql})" if self.subquery else self.sql, params` which the params tweaks of `_execute_query`.

From there `_execute_query` could be adapted to call `self.as_sql` to pass it to the cursor

django/db/models/sql/query.py
```diff
@@ -1284,7 +1300,6 @@ def build_lookup(self, lookups, lhs, rhs):
             lookup_class = lhs.get_lookup(lookup_name)
             if not lookup_class:
                 return
-
```
charettes, line 1287: Revert this line change.

docs/releases/4.2.txt
```diff
@@ -355,6 +355,9 @@ Miscellaneous
 * The undocumented ``negated`` parameter of the
   :class:`~django.db.models.Exists` expression is removed.
 
+* Add functionality so :class:`~django.db.models.sql.RawQuery`
+  can be used as a subquery
```
charettes, line 359:
```suggestion
  can be used as a subquery.
```

tests/raw_query/tests.py
```diff
@@ -413,7 +420,15 @@ def test_bool(self):
...
             len(Book.objects.raw("SELECT * FROM raw_query_book WHERE id = 0")), 0
         )
+
+    def test_as_subquery(self):
+        author_id_sub_query = RawQuery(
+            "SELECT id FROM raw_query_author WHERE last_name LIKE %s",
+            using="default",
+            params=["Smith"],
+        )
+        self.assertEqual(len(Book.objects.filter(author_id__in=author_id_sub_query)), 4)
```
charettes, line 434:
```suggestion
        with self.assertNumQueries(1):
            self.assertEqual(
                len(Book.objects.filter(author_id__in=author_id_sub_query)), 4
            )
```

charettes (commented):
Given `sql.query.RawQuery` is not part of the documented API and the issue was initially reported against `models.RawQuerySet` I wonder if we should just add `RawQuerySet.resolve_expression` as well so `RawQuerySet` can be used for `__in` lookup as well as `QuerySet.raw`.

The implementation should pretty be pretty straightforward; 

```python
class RawQuerySet:
    ...

    def resolve_expression(self, *args, **kwargs):
        return self.query.resolve_expression(self, *args, **kwargs)
```

jacobtylerwalls:
Thanks for being so responsive to the reviews. This pull request languished outside the patch review queue because the ticket's "needs improvement" flag was never unset. I'm unsetting it right now.

docs/releases/4.2.txt
```diff
@@ -483,6 +483,9 @@ Miscellaneous
 * The undocumented ``negated`` parameter of the
   :class:`~django.db.models.Exists` expression is removed.
 
+* Add functionality so :class:`~django.db.models.sql.RawQuery`
+  can be used as a subquery.
```
sarahboyce, line 487: `RawQuery` is not part of the public API. I think instead we can add a new section to https://docs.djangoproject.com/en/5.2/topics/db/sql/#performing-raw-queries which says that the raw queries can be used in subquerys (with a `.. versionadded:: 6.0` note) then we can add a release note in the 6.0 release notes that refer to this

django/db/models/sql/query.py
```diff
@@ -82,6 +82,11 @@ def get_children_from_q(q):
 class RawQuery:
     """A single raw SQL query."""
 
+    has_select_fields = True
+    contains_aggregate = []
```
sarahboyce, line 86:
```suggestion
    contains_aggregate = False
```
All other `contains_aggregate` attributes are boolean

tests/raw_query/tests.py
```diff
@@ -413,7 +420,18 @@ def test_bool(self):
         )
 
     def test_len(self):
-        self.assertEqual(len(Book.objects.raw("SELECT * FROM raw_query_book")), 4)
+        self.assertEqual(len(Book.objects.raw("SELECT * FROM raw_query_book")), 5)
         self.assertEqual(
             len(Book.objects.raw("SELECT * FROM raw_query_book WHERE id = 0")), 0
         )
+
+    def test_as_subquery(self):
```
sarahboyce, line 428: As `RawQuery` is not part of the public API, and from what I can see, not tested directly. Instead I think we should have:
```python
    def test_as_subquery(self):
        author_id_sub_query = Book.objects.raw(
            "SELECT id FROM raw_query_author WHERE first_name LIKE %s",
            using="default",
            params=["Bob"],
        )
            self.assertEqual(
                len(Book.objects.filter(author_id__in=author_id_sub_query)), 1
            )
```

I tweaked the query so that we don't need to create a new book object and edit the other tests :+1:

sarahboyce (commented):
Thank you for the PR :+1: 
I have minor comments, please can you also rebase onto main and have a single commit :+1:

PR #16354, opened 2022-12-03
Improved test coverage for django.db.transaction.

Polandia94:
Better coverage of the django/db/transactions.py with new test.


felixxm (commented):
@Polandia94 Thanks :+1: Welcome aboard :boat:

PR #16355, opened 2022-12-03
Fixed #34172 -- Improved ModelAdmin.get_urls docs.

sephii:
The new docs don’t show unsecure code anymore, to prevent security issues created by copy-paste.

[Trac ticket](https://code.djangoproject.com/ticket/34172)

### DIFF
--- a/docs/ref/contrib/admin/index.txt
+++ b/docs/ref/contrib/admin/index.txt
@@ -1575,7 +1575,7 @@ templates used by the :class:`ModelAdmin` views:
             def get_urls(self):
                 urls = super().get_urls()
                 my_urls = [
-                    path('my_view/', self.my_view),
+                    path('my_view/', self.admin_site.admin_view(self.my_view))
                 ]
                 return my_urls + urls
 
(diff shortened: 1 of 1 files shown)


docs/ref/contrib/admin/index.txt
```diff
@@ -1609,35 +1609,14 @@ templates used by the :class:`ModelAdmin` views:
...
-                urls = super().get_urls()
-                my_urls = [
-                    path('my_view/', self.admin_site.admin_view(self.my_view))
-                ]
-                return my_urls + urls
-
-    Notice the wrapped view in the fifth line above::
-
-        path('my_view/', self.admin_site.admin_view(self.my_view))
+    Notice how the ``self.my_view`` function is wrapped in
```
shangxiao, line 1612: felixx/carlton: seems like this may need to be in the Note box as well… thoughts?

docs/ref/contrib/admin/index.txt
```diff
@@ -1575,7 +1575,7 @@ templates used by the :class:`ModelAdmin` views:
             def get_urls(self):
                 urls = super().get_urls()
                 my_urls = [
-                    path('my_view/', self.my_view),
+                    path('my_view/', self.admin_site.admin_view(self.my_view))
```
shangxiao, line 1578: One possible improvement could be to add a comment above this line (similar to the explanatory comments below) that indicates the necessity of `admin_view()` with a "see note below" type thing.

shangxiao:
I like this as I greatly appreciate it when reference docs are as succinct as possible - or at least gets to the point _first_ then offers explanation afterwards/links to explanations.  (also see diátaxis recommendations)

carltongibson (commented):
Hi @sephii — thanks for this, looks good. 

I pushed some possible edits in https://github.com/django/django/pull/16355/commits/b0ebb242608ae8e5de0d8110701ecc4a3c084fdc based on @shangxiao's comments: 

* Made a reference to `AdminSite.admin_view()` **before** the example. 
* Moved the explanation into a `note` callout. (I put this before the existing note, since it's security sensitive.) 
* Adjusted the wording slightly. 

Let me know what you think.

sephii:
Thanks for the changes @carltongibson. Looks good to me!

PR #16392, opened 2022-12-19
Renamed Technical Board to Steering Council wherever mentioned

andrewgodwin:
This is for the implementation of DEP 12, which has now been fully adopted.

I additionally fixed a few references to where votes happen, as we're preferring to have them in the forum now for the SC at least.

### DIFF
--- a/docs/internals/contributing/bugs-and-features.txt
+++ b/docs/internals/contributing/bugs-and-features.txt
@@ -125,9 +125,10 @@ How we make decisions
 =====================
 
 Whenever possible, we strive for a rough consensus. To that end, we'll often
-have informal votes on |django-developers| about a feature. In these votes we
-follow the voting style invented by Apache and used on Python itself, where
-votes are given as +1, +0, -0, or -1. Roughly translated, these votes mean:
+have informal votes on |django-developers| or the Django Forum about a feature.
+In these votes we follow the voting style invented by Apache and used on Python
+itself, where votes are given as +1, +0, -0, or -1.
+Roughly translated, these votes mean:
 
 * +1: "I love the idea and I'm strongly committed to it."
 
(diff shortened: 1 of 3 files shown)


felixxm:
@andrewgodwin Thanks :+1: Should we change it now :thinking:? or maybe after the Django 4.2 feature freeze, when we will vote for the Steering Council for 5.X release cycle.

felixxm:
We only have 4 weeks to the feature freeze. I think it would be better to wait with rename until the Steering Council is elected as the Steering Council and not as the Technical Board. My 2 cents.

andrewgodwin:
@felixxm The DEP was written to impose an immediate change on the name and do-the-job requirements rather than wait for an election, plus I already instituted it on the teams page on djangoproject. I can of course roll that back, but I'd rather this go in place immediately if we can.

(Plus, we'll be inviting a feature proposal vote right after the feature freeze, but before the new election, as that's how the DEP 10 timing works out)

felixxm:
@django/steering-council This is a [DEP-0012](https://github.com/django/deps/blob/main/accepted/0012-steering-council.rst)'s implementation milestone, so according to [DEP-0010](https://github.com/django/deps/blob/main/accepted/0010-new-governance.rst#how-django-is-developed) I'd like to inform the Steering Council of an intent to merge it. Please hold a vote :ballot_box: Voting will end on December 26th, 2022 :calendar:

carltongibson:
Really?

andrewgodwin:
My reading of DEP 10 is:

> "Major Change" means any change to Django's codebase of scope significant enough to require the use of the DEP process.

DEP 12 was not a *codebase* change, and was instead a governance change, and so I don't think it's in scope under this rule. Especially as we just had two votes on it.

felixxm:
I would argue that docs are part of the codebase, but fine, I'm a little tired of being the only one trying to enforce DEP-10 rules. We had the same discussion in previous DEP implementations (#15387, #14670) :shrug: For example, I always wait for approval to merge my typo fixes in docs.

carltongibson:
You do great @felixxm! 

But if DEP 10 really does require yet another vote here, then we need another change to DEP 10. 

Both the TB/SC and DSF Board have already held two votes. That's plenty, especially that there's no actual significant implementation to vote on. 

(What's the spirit of the relevant clause? How can we rephrase it --- if we need to -- so that we capture that but that we're not getting in our own way all the time?)

felixxm:
> Really?

Feel-free to ignore my request and move it forward.

 :wink:

carltongibson:
@felixxm My comment wasn't meant at you, rather at the rules. I realise I should have made that clearer. Sorry. 💝 

(Long day — not your fault. _mea culpa_)

andrewgodwin:
The spirit of the clause, in my interpretation, is "the technical board/steering council should review every big code change milestone to ensure it matches the DEP's proposed direction". The idea being, to ensure that as a Major Change DEP gets implemented over time, that it is still working down the path proposed by the original DEP.

My reasoning here is that this isn't a Major Change in the technical sense; there's no implementation time gap here, so another vote does not seem necessary. I'm unsure if we can clarify further without going and doing _another_ DEP to fix it, in the vein of DEP 44, but if that makes sense to everyone here for now that's a start.

(and @felixxm, please keep on enforcing DEP 10, especially now the Steering Council is actually required to come review stuff when you ask, or we get sent to an election!)

carltongibson:
It’s a principle of interpretation that you have to look for what makes most sense (in what you're interpreting, here the DEP 10 text). Another vote here isn't that, so there's strong evidence it's not the correct reading. 

Then there's a curious _"initially"_ in the _Changing this governance_ section:

> Changes to this governance process shall be treated **initially** as Major Changes to Django... 
> ([src](https://github.com/django/deps/blob/main/accepted/0010-new-governance.rst#changing-this-governance-process), emphasis added.)

Given that the DEP is so carefully worded, I have to take that as deliberate. 

All changes to DEP 10 are essentially text changes. It can’t make sense to mean to require a vote on a change of wording, followed by another vote to actually adjust the text. To agree to one, is to agree to the other. A PR review sure, did an error slip in? — but that's not a substantive matter.

This is in contrast to the Black changes in particular, where it’s substantive matter of fact whether the implementation is acceptable. There’s (conceptual) room for a difference there, and the vote to confirm the implementation makes sense. 

I take the **initially** as meaning that governance changes should be treated, seriously, and formally, and require vote for adoption. But that the process is then that outlined in the section there. Once that's done though, the implementation it **moot**. It doesn't make sense to continue to treat that part of it as a major change. Hence **initially**.

I will proceed to merge, taking responsibility for doing so, on that basis. If anyone feels that is in error, they may request a Steering Council review post facto.

carltongibson (approved):

PR #16417, opened 2023-01-02
Refs #31300 -- Add function-based virtual fields

jpnauta:
As proposed in [this forum](https://groups.google.com/g/django-developers/c/9Mf7YqDA4bg/m/wb07E71oAAAJ), [this forum](https://groups.google.com/g/django-developers/c/ADSuUUuZp3Q/m/eZGYZv74AQAJ) and [this ticket](https://code.djangoproject.com/ticket/31300). Kudos to Paolo for the thorough research on backend support for this feature.

Code samples:

```
from django.db import models

# Basic usage
class MyModel(models.Model):
    a = models.IntegerField()
    b = models.IntegerField()
    a_plus_b = models.GeneratedField(expression=F("a") + F("b"))

# Option to customize how field is persisted
class MyModel(models.Model):
    name = models.CharField(max_length=10)
    lower_name = models.GeneratedField(
        expression=Lower("name"),
        output_field=TextField(
            db_collation="fr_FR"
        ),
    )

# Option to not store as column in database (not supported in Postgres)
class MyModel(models.Model):
    name = models.CharField(max_length=10)
    lower_name = models.GeneratedField(
        expression=Lower("name", output_field=TextField()), db_persist=False
    )
```

Implementation notes:
- The feature is implemented for Postgres, SQLite, and MySQL/MariaDB
- This feature has limitations on MySQL/MariaDB and older versions of SQLite because the generated field value won't be set on `save()` since `INSERT...RETURNING` isn't implemented for this DB backend
- Postgres does not support `GENERATED ALWAYS ... VIRTUAL` fields 
- SQLite doesn't support constants being injected into the `CREATE TABLE` statement
- It should be easy to implement this feature for Oracle, I just didn't bother quite yet since it's tedious to run Oracle locally

Please let me know what you think! This is a tentative first draft, so feedback and ideas are greatly appreciated 😄

### DIFF
--- a/django/db/backends/base/features.py
+++ b/django/db/backends/base/features.py
@@ -353,6 +353,13 @@ class BaseDatabaseFeatures:
     # Does the backend support column comments in ADD COLUMN statements?
     supports_comments_inline = False
 
+    # Does the backend support generated columns?
+    supports_generated_columns = False
+    # Does the backend allow parameters in generated column definitions?
+    supports_generated_columns_params = False
+    # Does the backend support virtual generated columns?
+    supports_virtual_generated_columns = False
+
     # Does the backend support the logical XOR operator?
     supports_logical_xor = False
 
--- a/django/db/backends/base/schema.py
+++ b/django/db/backends/base/schema.py
@@ -287,7 +287,14 @@ class BaseDatabaseSchemaEditor:
     # Field <-> database mapping functions
 
     def _iter_column_sql(
-        self, column_db_type, params, model, field, field_db_params, include_default
+        self,
+        column_db_type,
+        params,
+        model,
+        field,
+        field_db_params,
+        include_default,
+        generated_sql,
     ):
         yield column_db_type
         if collation := field_db_params.get("collation"):
@@ -332,7 +339,9 @@ class BaseDatabaseSchemaEditor:
             and self.connection.features.interprets_empty_strings_as_nulls
         ):
             null = True
-        if not null:
+        if generated_sql:
+            yield generated_sql
+        elif not null:
             yield "NOT NULL"
         elif not self.connection.features.implied_column_null:
             yield "NULL"
@@ -361,20 +370,26 @@ class BaseDatabaseSchemaEditor:
         if column_db_type is None:
             return None, None
         params = []
-        return (
-            " ".join(
-                # This appends to the params being returned.
-                self._iter_column_sql(
-                    column_db_type,
-                    params,
-                    model,
-                    field,
-                    field_db_params,
-                    include_default,
-                )
-            ),
-            params,
+        generated_sql = None
+        if generated_parameters := field_db_params.get("generated_parameters"):
+            generated_expression_sql, persisted, generated_params = generated_parameters
+            generated_sql = self._column_generated_sql(
+                generated_expression_sql, persisted
+            )
+            params += generated_params
+        sql = " ".join(
+            # This appends to the params being returned.
+            self._iter_column_sql(
+                column_db_type,
+                params,
+                model,
+                field,
+                field_db_params,
+                include_default,
+                generated_sql,
+            )
         )
+        return sql, params
 
     def skip_default(self, field):
         """
@@ -421,6 +436,11 @@ class BaseDatabaseSchemaEditor:
             default_sql %= tuple(self.prepare_default(p) for p in params)
             params = []
         return sql % default_sql, params
+
+    def _column_generated_sql(self, expression_sql, persisted):
+        """Return the SQL to use in a GENERATED ALWAYS clause."""
+        persistency_sql = "STORED" if persisted else "VIRTUAL"
+        return "GENERATED ALWAYS AS (%s) %s" % (expression_sql, persistency_sql)
 
     @staticmethod
     def _effective_default(field):
@@ -811,6 +831,8 @@ class BaseDatabaseSchemaEditor:
         old_type = old_db_params["type"]
         new_db_params = new_field.db_parameters(connection=self.connection)
         new_type = new_db_params["type"]
+        old_generated_parameters = old_db_params.get("generated_parameters")
+        new_generated_parameters = new_db_params.get("generated_parameters")
         if (old_type is None and old_field.remote_field is None) or (
             new_type is None and new_field.remote_field is None
         ):
@@ -847,6 +869,11 @@ class BaseDatabaseSchemaEditor:
                 "Cannot alter field %s into %s - they are not compatible types "
                 "(you cannot alter to or from M2M fields, or add or remove "
                 "through= on M2M fields)" % (old_field, new_field)
+            )
+        elif old_generated_parameters != new_generated_parameters:
+            raise ValueError(
+                "Modifying GeneratedFields is not supported - the field %s must "
+                "be removed and re-added with the new definition." % new_field
             )
 
         self._alter_field(
--- a/django/db/backends/postgresql/features.py
+++ b/django/db/backends/postgresql/features.py
@@ -70,6 +70,9 @@ class DatabaseFeatures(BaseDatabaseFeatures):
     supports_update_conflicts = True
     supports_update_conflicts_with_target = True
     supports_covering_indexes = True
+    supports_generated_columns = True
+    supports_generated_columns_params = True
+    supports_virtual_generated_columns = False
     can_rename_index = True
     test_collations = {
         "non_default": "sv-x-icu",
--- a/django/db/backends/sqlite3/features.py
+++ b/django/db/backends/sqlite3/features.py
@@ -40,6 +40,9 @@ class DatabaseFeatures(BaseDatabaseFeatures):
     supports_json_field_contains = False
     supports_update_conflicts = True
     supports_update_conflicts_with_target = True
+    supports_generated_columns = Database.sqlite_version_info >= (3, 31, 0)
+    supports_generated_columns_params = False
+    supports_virtual_generated_columns = True
     test_collations = {
         "ci": "nocase",
         "cs": "binary",
--- a/django/db/backends/sqlite3/schema.py
+++ b/django/db/backends/sqlite3/schema.py
@@ -135,7 +135,7 @@ class DatabaseSchemaEditor(BaseDatabaseSchemaEditor):
             # Choose a default and insert it into the copy map
             if (
                 create_field.db_default is NOT_PROVIDED
-                and not create_field.many_to_many
+                and not (create_field.many_to_many or create_field.generated)
                 and create_field.concrete
             ):
                 mapping[create_field.column] = self.prepare_default(
--- a/django/db/models/fields/generated.py
+++ b/django/db/models/fields/generated.py
@@ -1,0 +1,137 @@
+from django.core import checks
+from django.db import connections
+from django.db.models.sql import Query
+
+from . import Field
+
+__all__ = ["GeneratedField"]
+
+
+class GeneratedField(Field):
+    generated = True
+    db_returning = True
+
+    _query = None
+    _resolved_expression = None
+    output_field = None
+
+    def __init__(self, *, expression, output_field=None, db_persist=True, **kwargs):
+        self.expression = expression
+        self._output_field = output_field
+        self.db_persist = db_persist
+
+        assert "editable" not in kwargs, "Cannot set `editable` for GeneratedField"
+        assert "blank" not in kwargs, "Cannot set `blank` for GeneratedField"
+        assert "default" not in kwargs, "Cannot set `default` for GeneratedField"
+
+        kwargs["editable"] = False
+        kwargs["blank"] = True
+        kwargs["default"] = None
+
+        super().__init__(**kwargs)
+
+    def contribute_to_class(self, *args, **kwargs):
+        super().contribute_to_class(*args, **kwargs)
+
+        self._query = Query(model=self.model, alias_cols=False)
+        self._resolved_expression = self.expression.resolve_expression(
+            self._query, allow_joins=False
+        )
+        self.output_field = (
+            self._output_field
+            if self._output_field is not None
+            else self._resolved_expression.output_field
+        )
+
+    def _expression_sql(self, connection):
+        return self._resolved_expression.as_sql(
+            compiler=connection.ops.compiler("SQLCompiler")(
+                self._query, connection=connection, using=None
+            ),
+            connection=connection,
+        )
+
+    def check(self, **kwargs):
+        databases = kwargs.get("databases") or []
+        return [
+            *super().check(**kwargs),
+            *self._check_sql_expression(databases),
+            *self._check_persistence(databases),
+        ]
+
+    def _check_sql_expression(self, databases):
+        errors = []
+        for db in databases:
+            connection = connections[db]
+
+            if not (
+                connection.features.supports_generated_columns
+                or "supports_generated_columns" in self.model._meta.required_db_features
+            ):
+                errors.append(
+                    checks.Error(
+                        "%s does not support GeneratedFields."
+                        % connection.display_name,
+                        obj=self,
+                        id="fields.E220",
+                    )
+                )
+
+            if connection.features.supports_generated_columns:
+                _, params = self._expression_sql(connection=connection)
+                if params and not (
+                    connection.features.supports_generated_columns_params
+                    or "supports_generated_columns_params"
+                    in self.model._meta.required_db_features
+                ):
+                    errors.append(
+                        checks.Error(
+                            "%s does not support GeneratedFields with "
+                            "parameters." % connection.display_name,
+                            obj=self,
+                            id="fields.E221",
+                        )
+                    )
+
+        return errors
+
+    def _check_persistence(self, databases):
+        errors = []
+
+        for db in databases:
+            connection = connections[db]
+
+            if not self.db_persist and not (
+                connection.features.supports_virtual_generated_columns
+                or "supports_virtual_generated_columns"
+                in self.model._meta.required_db_features
+            ):
+                errors.append(
+                    checks.Error(
+                        "%s does not support non-persisted GeneratedFields."
+                        % connection.display_name,
+                        obj=self,
+                        id="fields.E222",
+                        hint="remove the persisted=False argument",
+                    )
+                )
+
+        return errors
+
+    def deconstruct(self):
+        name, path, args, kwargs = super().deconstruct()
+        del kwargs["blank"]
+        del kwargs["default"]
+        del kwargs["editable"]
+        kwargs["expression"] = self.expression
+        if self._output_field is not None:
+            kwargs["output_field"] = self._output_field
+        if self.db_persist is not True:
+            kwargs["db_persist"] = self.db_persist
+        return name, path, args, kwargs
+
+    def db_parameters(self, connection):
+        db_params = self.output_field.db_parameters(connection)
+        expression_sql, params = self._expression_sql(connection=connection)
+        db_params["generated_parameters"] = (expression_sql, self.db_persist, params)
+        return db_params
--- a/django/db/models/query.py
+++ b/django/db/models/query.py
@@ -1829,6 +1829,9 @@ class QuerySet(AltersData):
         Insert a new record for the given model. This provides an interface to
         the InsertQuery class and is how Model.save() is implemented.
         """
+        if self.model._meta.generated_fields and fields:
+            fields = [f for f in fields if not f.generated]
+
         self._for_write = True
         if using is None:
             using = self.db
--- a/docs/ref/checks.txt
+++ b/docs/ref/checks.txt
@@ -237,6 +237,9 @@ Model fields
 * **fields.W907**: ``django.contrib.postgres.fields.CITextField`` is
   deprecated. Support for it (except in historical migrations) will be removed
   in Django 5.1.
+* **fields.E220**: ``<database>`` does not support ``GeneratedField``\s.
+* **fields.E221**: ``<database>`` does not support ``GeneratedField``\s with constant parameters.
+* **fields.E222**: ``<database>`` does not support ``GeneratedField``\s with no persistence.
 
 File fields
 ~~~~~~~~~~~
--- a/docs/ref/models/fields.txt
+++ b/docs/ref/models/fields.txt
@@ -1185,6 +1185,49 @@ when :attr:`~django.forms.Field.localize` is ``False`` or
     information on the difference between the two, see Python's documentation
     for the :mod:`decimal` module.
 
+``GeneratedField``
+------------------
+
+.. versionadded:: 5.0
+
+.. class:: GeneratedField(expression, output_field=None, db_persist=True, **kwargs)
+
+A field which is computed based on other fields in the model. Similar to a
+database view, this field is managed and updated by the database itself. Uses
+the ``GENERATED`` SQL syntax.
+
+There are two kinds of generated columns: stored and virtual. A stored generated
+column is computed when it is written (inserted or updated) and occupies storage
+as if it were a normal column. A virtual generated column occupies no storage and
+is computed when it is read. Thus, a virtual generated column is similar to a view
+and a stored generated column is similar to a materialized view.
+
+``GeneratedField`` is supported on MariaDB, MySQL, PostgreSQL, and SQLite. SQLite
+does not support parameters in generated column definitions.
+
+.. attribute:: GeneratedField.expression
+
+    An :class:`Expression` used by the database to automatically set the field
+    value whenever the model is changed.
+
+    The expressions should be deterministic and only reference fields within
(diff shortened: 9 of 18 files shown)


django/db/backends/base/schema.py
```diff
@@ -826,6 +853,11 @@ def alter_field(self, model, old_field, new_field, strict=False):
                 "(you cannot alter to or from M2M fields, or add or remove "
                 "through= on M2M fields)" % (old_field, new_field)
             )
+        elif old_computed_parameters != new_computed_parameters:
+            raise ValueError(
+                "Modifying ComputedFields is not supported - the field %s must "
+                "be dropped and recreated." % new_field
+            )
```
jpnauta, line 860: AFAICT Postgres does not support the concept of modifying `GENERATED ALWAYS` columns, so it felt easiest to disable the ability to modify these columns altogether

django/db/backends/sqlite3/features.py
```diff
@@ -42,6 +42,9 @@ class DatabaseFeatures(BaseDatabaseFeatures):
     supports_json_field_contains = False
     supports_update_conflicts = Database.sqlite_version_info >= (3, 24, 0)
     supports_update_conflicts_with_target = supports_update_conflicts
+    supports_computed_field = True
+    supports_computed_field_params = False
```
jpnauta, line 46: SQLite does not allow params to be passed into the `CREATE TABLE` SQL statement

django/db/backends/postgresql/features.py
```diff
@@ -70,6 +70,9 @@ class DatabaseFeatures(BaseDatabaseFeatures):
     supports_update_conflicts = True
     supports_update_conflicts_with_target = True
     supports_covering_indexes = True
+    supports_computed_field = True
+    supports_computed_field_params = True
+    supports_computed_field_virtual = False
```
jpnauta, line 75: Postgres only supports `GENERATED ALWAYS ... STORED` columns, not `GENERATED ALWAYS ... VIRTUAL`

django/db/backends/sqlite3/schema.py
```diff
@@ -233,7 +233,10 @@ def is_self_referential(f):
         if create_field:
             body[create_field.name] = create_field
             # Choose a default and insert it into the copy map
-            if not create_field.many_to_many and create_field.concrete:
+            if (
+                not (create_field.many_to_many or create_field.computed)
+                and create_field.concrete
+            ):
```
jpnauta, line 239: When cloning a SQLite database to add a new column, computed columns should not be included or they will cause an error

django/db/models/query.py
```diff
@@ -1792,6 +1792,9 @@ def _insert(
         Insert a new record for the given model. This provides an interface to
         the InsertQuery class and is how Model.save() is implemented.
         """
+        if self.model._meta.computed_fields and fields:
+            fields = [f for f in fields if f not in self.model._meta.computed_fields]
```
jpnauta, line 1796: When performing a `save()`, `create()` or `bulk_create()`, values for compute fields should not be specified

docs/ref/checks.txt
```diff
@@ -234,6 +234,9 @@ Model fields
 * **fields.W907**: ``django.contrib.postgres.fields.CITextField`` is
   deprecated. Support for it (except in historical migrations) will be removed
   in Django 5.1.
+* **fields.E220**: ``<database>`` does not support ``ComputedField``\s.
+* **fields.E221**: ``<database>`` does not support ``ComputedField`` with constant parameters\s.
+* **fields.E222**: ``<database>`` does not support ``ComputedField`` with no persistence\s.
```
jpnauta, line 239: This is my first P/R, so I am not sure what the protocol is for adding new error checks. Any advise would be greatly appreciated!

docs/ref/checks.txt
```diff
@@ -234,6 +234,9 @@ Model fields
 * **fields.W907**: ``django.contrib.postgres.fields.CITextField`` is
   deprecated. Support for it (except in historical migrations) will be removed
   in Django 5.1.
+* **fields.E220**: ``<database>`` does not support ``ComputedField``\s.
+* **fields.E221**: ``<database>`` does not support ``ComputedField``\s with constant parameters.
+* **fields.E222**: ``<database>`` does not support ``ComputedField``\s with no persistence.
```
jpnauta, line 239: This is my first P/R, so I am not sure what the protocol is for adding new error checks. Any advise would be greatly appreciated!

pauloxnet:
Jeremy I'm glad to see your PR based on my proposal.

I'll try to review it in few days.

Have see the comment from Simon in the mailing list about a nulla le base field to use custom expression?
https://groups.google.com/g/django-developers/c/9Mf7YqDA4bg/m/wb07E71oAAAJ

jpnauta:
> Jeremy I'm glad to see your PR based on my proposal.
> 
> I'll try to review it in few days.
> 
> Have see the comment from Simon in the mailing list about a nulla le base field to use custom expression? https://groups.google.com/g/django-developers/c/9Mf7YqDA4bg/m/wb07E71oAAAJ

Thanks Paolo, I appreciate it! Yes, I have included Simon's suggestion to have an optional `output_field`, which by default is calculated using the given expression. I have added usages example to the P/R description accordingly :smile:

jpnauta:
Hi @pauloxnet any thoughts on this? Just wanted to check-in 😄

rmortes:
Hey! I've been waiting for this functionality for months, and I've just seen it's being developed right now. Great work!!

I have a question about this feature: Would it support subqueries? Say I'm building a bank, and I take a `Snapshot` of my `Users` money daily. This `Snapshot` has a float `money` field and a FK to User. Would it be possible to add a new computed field `delta` with the value `F("money") - Snapshot.objects.filter(user=F("user")).order_by("-created_date").first().money`?

Thanks!

jpnauta:
Hi @rmortes, thanks for the interest in this P/R! Sadly I suspect your use case will indeed not work. As explained in [Postgres' docs](https://www.postgresql.org/docs/current/ddl-generated-columns.html), *the generation expression can only use immutable functions and cannot use subqueries or reference anything other than the current row in any way*. I'm not sure of the exact limitations for other DBs, but I suspect they are similar to Postgres.

pauloxnet:
> Hi @pauloxnet any thoughts on this? Just wanted to check-in smile

Sorry for the late reply. I'll try to take a proper look to this PR before the weekend.

jpnauta:
@pauloxnet (or anyone else) are you willing to review this?

django/db/models/query.py
adamchainz: Using `in <list>` leads to O(n²) runtime, can you not use the flag on the field?

```suggestion
        if self.model._meta.computed_fields and fields:
            fields = [f for f in fields if not f.computed]
```

tests/invalid_models_tests/test_ordinary_fields.py
```diff
@@ -1057,3 +1058,61 @@ class Meta:
...
+                    id="fields.E220",
+                ),
+            ],
+        )
+
+    @skipUnlessDBFeature("supports_computed_field")
+    @skipIfDBFeature("supports_computed_field_params")
+    def test_params_not_supported(self):
+        class Model(models.Model):
+            field = models.ComputedField(Value("value", output_field=TextField()))
```
adamchainz, line 1086: You shouldn't need `output_field` here, as `Value` can determine it automatically:

```suggestion
            field = models.ComputedField(Value("value"))
```

django/db/models/fields/computed.py
```diff
@@ -0,0 +1,131 @@
...
+            connection=connection,
+        )
+
+    def check(self, **kwargs):
+        databases = kwargs.get("databases") or []
+        return [
+            *super().check(**kwargs),
+            *self._check_sql_expression(databases),
+            *self._check_persistence(databases),
+        ]
```
adamchainz, line 56: IMO we could merge these into the `check()` function here, it would save a lot of temporary lists and looping over the databases list.

django/db/models/fields/computed.py
```diff
@@ -0,0 +1,131 @@
...
+    output_field = None
+
+    def __init__(self, expression, output_field=None, persisted=True, *args, **kwargs):
+        self.expression = expression
+        self._output_field = output_field
+        self.persisted = persisted
+
+        kwargs["editable"] = False
+        kwargs["blank"] = True
+        kwargs["default"] = None
```
adamchainz, line 25: I think we should error if any of these are passed in `**kwargs`

django/db/models/fields/computed.py
```diff
@@ -0,0 +1,131 @@
...
+
+class ComputedField(Field):
+    computed = True
+    db_returning = True
+
+    _query = None
+    _resolved_expression = None
+    output_field = None
+
+    def __init__(self, expression, output_field=None, persisted=True, *args, **kwargs):
```
adamchainz, line 18: `persisted` is a slightly odd param name, since it’s in the past tense. We don’t have any other past tense names that I can find in other fields.

I think `persist` would be a better name, or maybe `db_persist`, since it's a database-level thing, like `db_column` ?

Also taking `args` allows duplicate arguments for e.g. `editable`. Let's make this class [keyword-only](https://lukeplant.me.uk/blog/posts/keyword-only-arguments-in-python/):

```suggestion
    def __init__(self, *, expression, output_field=None, persisted=True, **kwargs):
```

(`Field` really should be keyword-only, but it's too late to fix that.)

adamchainz (changes requested):

adamchainz:
Also I would prefer we used the name `GeneratedField`, sticking with the SQL syntax, as previously discussed on the mailing list. I don’t see why we would copy SQL Alchemy, and “computed” isn’t any clearer IMO.

And I would like to say thanks for the work so far!

django/db/models/fields/computed.py
jpnauta: I would tend to disagree here - every other `Field` in the django project follows this convention, so I recommend we do the same.

Screen Shot 2023-03-25 at 6 23 50 PM
Screen Shot 2023-03-25 at 6 23 21 PM
Screen Shot 2023-03-25 at 6 24 11 PM

django/db/models/fields/computed.py
jpnauta: Changed to `db_persist`

jpnauta:
Thank you for the helpful feedback @pauloxnet and @adamchainz! I believe I have addressed all your comments - please review once more and let me know if I missed anything.

pauloxnet:
As I wrote before the change looks good to me, I just think it's missing a section added in the model fields documentation, together with other fields.

jpnauta:
Thanks @pauloxnet, missed that part. I have added field documentation now, let me know if there's any other sections I missed.

pauloxnet:
> Thanks @pauloxnet, missed that part. I have added field documentation now, let me know if there's any other sections I missed.

Thanks.

I've unset the "Patch needs improvement" flag from the ticket.

Now we await the reviews of other Django developers.

docs/ref/models/fields.txt
```diff
@@ -1150,6 +1150,44 @@ when :attr:`~django.forms.Field.localize` is ``False`` or
     information on the difference between the two, see Python's documentation
     for the :mod:`decimal` module.
 
+``GeneratedField``
+------------------
+
```
felixxm, line 1155: `versionadded` annotation is missing.

django/db/backends/base/features.py
```diff
@@ -339,6 +339,13 @@ class BaseDatabaseFeatures:
     # Does the backend support column comments in ADD COLUMN statements?
     supports_comments_inline = False
 
+    # Does the backend support generated columns?
+    supports_generated_columns = False
+    # Does the backend allow parameters in generated column definitions?
+    supports_generated_columns_params = False
+    # Does the backend support virtual generated columns?
+    supports_virtual_generated_columns = False
```
felixxm, line 347: Base feature flags don't match feature flags in database backends.

tests/invalid_models_tests/test_ordinary_fields.py
```diff
@@ -1057,3 +1058,62 @@ class Meta:
...
+        self.assertEqual(
+            Model._meta.get_field("field").check(databases={"default"}),
+            [
+                Error(
+                    "%s does not support GeneratedFields." % connection.display_name,
+                    obj=Model._meta.get_field("field"),
+                    id="fields.E220",
+                ),
+            ],
+        )
```
felixxm, line 1080: This should be checked for both cases:
```suggestion
    def test_field_not_supported(self):
        class Model(models.Model):
            name = models.IntegerField()
            field = models.GeneratedField(expression=F("name"))

        self.assertEqual(
            Model._meta.get_field("field").check(databases={"default"}),
            [
                Error(
                    "%s does not support GeneratedFields." % connection.display_name,
                    obj=Model._meta.get_field("field"),
                    id="fields.E220",
                ),
            ] if not connection.features.supports_generated_field else [],
        )
```
The same changes are needed for other feature flags. Also, we should check that `required_db_features` is respected (see for example `test_db_comment_required_db_features`).

django/db/backends/sqlite3/features.py
```diff
@@ -42,6 +42,9 @@ class DatabaseFeatures(BaseDatabaseFeatures):
     supports_json_field_contains = False
     supports_update_conflicts = Database.sqlite_version_info >= (3, 24, 0)
     supports_update_conflicts_with_target = supports_update_conflicts
+    supports_generated_field = True
+    supports_generated_field_params = False
+    supports_generated_field_virtual = True
```
felixxm, line 47: Is it true for all supported versions of SQLite (i.e. 3.21+)? I tried with SQLite 3.26 and tests crash:
```
...
django.db.utils.OperationalError: near "AS": syntax error
```

pauloxnet: As I already reported in the developer mailing list [1], SQLite added support for Generated Columns in version 3.31.0 (2020-01-22) [2].

- [1] https://groups.google.com/g/django-developers/c/9Mf7YqDA4bg/m/D5a5jUCACwAJ
- [2] https://www.sqlite.org/releaselog/3_31_0.html

pauloxnet:
> ... Support for virtual fields on Oracle is missing.

Reiterating my very little knowledge of Oracle, I had investigated a year ago there seemed to be support for "Virtual Columns" from Oracle Database 11g Release 1
https://oracle-base.com/articles/11g/virtual-columns-11gr1

docs/ref/models/fields.txt
```diff
@@ -1150,6 +1150,44 @@ when :attr:`~django.forms.Field.localize` is ``False`` or
...
+.. attribute:: GeneratedField.expression
+
+    An :class:`Expression` used by the database to automatically set the field
+    value whenever the model is changed.
+
+    Several restrictions apply to generated columns, depending on the database.
+    A :class:`GeneratedField` should be immutable and only reference
+    concrete columns inside the model itself. References to other models' fields
+    is not supported. See your database's documentation for more information.
```
adamchainz, line 1170: This paragraph was mixing the concepts of columns and fields. Plus I think "deterministic" is the right word, rather than "immutable" (the input fields into the expression are allowed to mutate). My attempt to improve:

```suggestion
    The expressions should be deterministic and only reference fields within
    the model (in one database table). Database backends can impose further
    restrictions.
```

The "(in one database table)" is because a model can span multiple tables, in the case of multi-table inheritance.

docs/ref/models/fields.txt
```diff
@@ -1150,6 +1150,44 @@ when :attr:`~django.forms.Field.localize` is ``False`` or
...
+.. attribute:: GeneratedField.db_persist
+
+    Determines if the database column should occupy storage as if it were a real
+    column. If ``False``, the column acts as a virtual column and does
+    not occupy database storage space.
+
+    Default is ``True``. PostgreSQL only supports persisted columns.
+
+``GeneratedField`` is supported on MariaDB, MySQL, PostgreSQL, and SQLite. SQLite
+does not support parameters in generated column definitions.
```
adamchainz, line 1189: I think this should go before the attribute description?

docs/releases/5.0.txt
```diff
@@ -279,6 +279,8 @@ Models
   :ref:`Choices classes <field-choices-enum-types>` directly instead of
   requiring expansion with the ``choices`` attribute.
 
+* Added ``django.models.fields.GeneratedField``.
```
adamchainz, line 282: 1. link to the appropriate docs
2. summarize the feature

Without these, the release note is not going to surface your hard work and make it easy for others to discover :)

```suggestion
* The new :class:`~django.db.models.fields.GeneratedField` allows creation of
  database generated columns.
```

docs/ref/models/fields.txt
```diff
@@ -1150,6 +1150,44 @@ when :attr:`~django.forms.Field.localize` is ``False`` or
...
     for the :mod:`decimal` module.
 
+``GeneratedField``
+------------------
+
+.. class:: GeneratedField(expression, output_field=None, db_persist=True, **kwargs)
+
+A field which is computed based on other fields in the model. Uses the database's
+``GENERATED ALWAYS AS`` syntax. Similar to a database view, this field is managed
+and updated by the database itself.
```
adamchainz, line 1160: Slight reordering of details, since SQL is a bit more of an implementation detail.

```suggestion
A field which is computed based on other fields in the model. Similar to a
database view, this field is managed and updated by the database itself. Uses
the ``GENERATED`` SQL syntax.
```

Also could we add an example after this paragraph? Just a basic use case, maybe from the tests, would really help users see what this does.

docs/ref/models/fields.txt
```diff
@@ -1150,6 +1150,44 @@ when :attr:`~django.forms.Field.localize` is ``False`` or
...
+    concrete columns inside the model itself. References to other models' fields
+    is not supported. See your database's documentation for more information.
+
+.. attribute:: GeneratedField.output_field
+
+    An optional :class:`Field` used to explicitly define the model's underlying
+    column type. This can be used to customize the field's DB collation or other
+    field-specific properties.
+
+    By default, this field is derived from the given ``expression``.
```
adamchainz, line 1178: Tried to tighten this up. "Properties" isn't quite the right word - that can lead to confusion with `@property`. "attribute" is what we really talk about in Python.

```suggestion
    An optional :class:`Field` to define the field’s data type. This can be
    used to customize attributes like the field's collation. By default, the
    output field is derived from ``expression``.
```

docs/ref/models/fields.txt
pauloxnet: I agree

docs/releases/5.0.txt
pauloxnet: LGTM

docs/ref/models/fields.txt
jpnauta: I pulled "immutable" from the Postgres docs. That said, "deterministic" works for me and I'm good with these changes.

jpnauta:
Thanks all for the feedback! I believe I implemented addressed all your comments.

I haven't implemented Oracle because I haven't found any way to run it locally. Any tips? I'd be happy to implement the features, but I currently have no way of verifying the changes work as expected.

piw2266:
하늘2023. 3. 29. 12:00에 Jeremy Nauta ***@***.***>님이 작성:
Thanks all for the feedback! I believe I implemented addressed all your comments.
I haven't implemented Oracle because I haven't found any way to run it locally. Any tips? I'd be happy to implement the features, but I currently have no way of verifying the changes work as expected.

piw2266:
하늘2023. 3. 29. 12:44에 이 포도 ***@***.***>님이 작성:하늘2023. 3. 29. 12:00에 Jeremy Nauta ***@***.***>님이 작성:
Thanks all for the feedback! I believe I implemented addressed all your comments.
I haven't implemented Oracle because I haven't found any way to run it locally. Any tips? I'd be happy to implement the features, but I currently have no way of verifying the changes work as expected.

piw2266:
하늘2023. 3. 29. 12:45에 이 포도 ***@***.***>님이 작성:하늘2023. 3. 29. 12:44에 이 포도 ***@***.***>님이 작성:하늘2023. 3. 29. 12:00에 Jeremy Nauta ***@***.***>님이 작성:
Thanks all for the feedback! I believe I implemented addressed all your comments.
I haven't implemented Oracle because I haven't found any way to run it locally. Any tips? I'd be happy to implement the features, but I currently have no way of verifying the changes work as expected.

jpnauta:
@felixxm @pauloxnet @adamchainz any comments on this P/R, or tips for implementing Oracle?

pauloxnet:
> @felixxm @pauloxnet @adamchainz any comments on this P/R, or tips for implementing Oracle?

LGTM

I've no experience with Oracle , so I can't help you on that point :-(

adamchainz:
Personally I'm fine with us leaving out Oracle support for new features like this, if it's troublesome. I am not sure we should be chasing all of its quirks for all features at this point. [Only 2%](https://lp.jetbrains.com/django-developer-survey-2022/#text-403) of survey respondents use it.

That said, it may be worth a try. The django-docker-box project helps you run multiple database backends - it has [instructions for setting up Oracle](https://github.com/django/django-docker-box#oracle). Try that if you have the time and energy.

Otherwise @felixxm loves Oracle (😜) and may be able to help.

pauloxnet:
> Personally I'm fine with us leaving out Oracle support for new features like this, if it's troublesome.

I agree with you.

jpnauta:
I attempted to use [django-docker-box](https://github.com/django/django-docker-box/) but ran into issue after issue. Seems to be a [known issue](https://github.com/django/django-docker-box/issues/4). I got close to running it using `truevoly/oracle-12c`, only to realize that Oracle 19 is required. 

Unless anyone can help implement this feature in Oracle, I'm going to consider this P/R ready to merge!

FYI @felixxm

jpnauta:
Thanks for the info @LilyFoote! Interested in working together to implement Oracle support? If you're willing to review this P/R and QA it for Oracle, then I'd be happy to take a "best guess" at what the implementation would be. How does that sound?

LilyFoote:
@jpnauta Actually, I should have a PR with a few commits for you soon adding a first draft of Oracle support.

adamchainz:
Hero!

jpnauta:
Thanks @LilyFoote for implementing Oracle support! I have merged the latest from `main` now, I believe this is ready to go.

@felixxm @pauloxnet are we able to get this P/R merged so we can move forward towards Lily's P/R?

pauloxnet (approved):

pauloxnet:
This PR LGTM

pauloxnet:
@jpnauta can you rebase this PR so that can be easier to review for others?

jpnauta:
@pauloxnet done! FYI @LilyFoote you will need to rebase as well 😄

LilyFoote:
@jpnauta rebased.

schinckel:
I'm pretty late to the party, but I only just came across this today.

> a_plus_b = models.GeneratedField(expression=F("a") + F("b"))

Will this field update within the context of a model instance if the values of a and/or b are changed?

jpnauta:
> I'm pretty late to the party, but I only just came across this today.
> 
> > a_plus_b = models.GeneratedField(expression=F("a") + F("b"))
> 
> Will this field update within the context of a model instance if the values of a and/or b are changed?

Yes @schinckel, there are a few tests in this P/R that do exactly that! Regardless of when/how the model's fields are changed, the generated field will always be `a` plus `b`.

schinckel:
Thanks for the reply. I've looked at the tests, and it seems to me that a refresh from the database is required to trigger the value to update.

(I'm asking because I implemented https://django-shared-property.readthedocs.io/en/latest/ specifically to handle hybrid properties. I'm going to look deeper into how you implemented this to see if I can make live-updating-properties work on top of this PR).

schinckel:
Yeah, that's not quite what I meant.

This was more about an update to a dependent column affecting the generated column's value without having _any_ database access required.

jpnauta:
@schinckel gotcha - in that case, no, this feature requires a DB update for the feature to work

schinckel:
Coolio. I think I can still make my stuff (which already works) build on this PR when it's merged anyway. I implemented allowing related objects' fields as part of the interface, but I think it's probably "better" to restrict to only local fields.

django/db/models/fields/generated.py
```diff
@@ -0,0 +1,137 @@
...
+    output_field = None
+
+    def __init__(self, *, expression, output_field=None, db_persist=True, **kwargs):
+        self.expression = expression
+        self._output_field = output_field
+        self.db_persist = db_persist
+
+        assert "editable" not in kwargs, "Cannot set `editable` for GeneratedField"
+        assert "blank" not in kwargs, "Cannot set `blank` for GeneratedField"
+        assert "default" not in kwargs, "Cannot set `default` for GeneratedField"
```
felixxm, line 25: This is a user-facing error so we shouldn't use `assert` (check out ticket-32508).

felixxm:
Closing in favor or #16860.

PR #16480, opened 2023-01-18
Refs #34233 -- Used @functools.cache.

ngnpope:
Python 3.9+ supports `@functools.cache` as an alias for `@functools.lru_cache(maxsize=None)`.


hugovk (approved):
Looks good, thanks!

Docs ref: https://docs.python.org/3/library/functools.html#functools.cache

PR #16481, opened 2023-01-18
Refs #34233 -- Used types.NoneType.

ngnpope:
Available since Python 3.10 where it was reintroduced.


felixxm (approved):
@ngnpope Thanks :+1:

PR #16495, opened 2023-01-21
Fixed #32539 -- Added toggleable facet filters to ModelAdmin.

sarahboyce:
Ticket: https://code.djangoproject.com/ticket/32539

Facet counts can be added to all filters on a ModelAdmin by either setting ``show_facets`` to ``True`` or by adding the query parameter `?...&_facets=1`


emarhin (approved):
👍

felixxm:
@simonw What do you think?

carltongibson (commented):
Hey @sarahboyce — thanks for this. It's looks really quite good. 

I uploaded a test project (with a DB, you can just `runserver` on) here: https://github.com/carltongibson/django_ticket_32539

It has a couple of the requisite filters, and sure enough: 

Screenshot 2023-02-02 at 10 02 28

So 💃. 

I think the docs don't really pull-out what's on offer here, or exactly what's going on when defining the filter. (First coffee and all but, I knew what I was hoping for this morning and was still a bit 🤔 as I was putting it together.) I wonder if we can make it clearer. 

* Can we make this easier still? (The lookup implementation feels a bit clunky)
* Is faceting going to get expensive (too suddenly — is there a cliff?) Do we need to be able to toggle it? 

I will see if I can ask @simonw if he had a cycle to look and advise. 

This is pretty cool though. A couple of defensive questions (which are no more than that): 

* Are there gotchas to this pattern we need to handle before we just ship it?
* This is entirely internal to `SimpleListFilter` — should it live in an external package first pass to iron out issues?


**Update**: I edited my comment on the ticket but, could we make this work by default (with perhaps a `?...&facets=1` toggle) for even the simple filters from a field name? 🤔

sarahboyce:
Hi @carltongibson thank you for testing :tada: glad it kinda works 

I think it should be possible to have the facet counts magically appear everywhere. I really like the toggle idea 
I have updated the `SimpleListFilter` one, you can have a look if you like but otherwise I will every now and again push up for the other filters. I will mark the PR as draft

I'm thinking we can have a warning that there will be a performance hit, especially for the `SimpleListFilter`, with the other ones, should be able to do some fancy aggregates to reduce the impact. :thinking: 

Thank you again :star:

sarahboyce:
Alright, I think that is ready for another look :+1: 
I think I could add a picture similar to at the top of the filter docs to show the filters, is there a demo site the images are coming from? 

> Are there gotchas to this pattern we need to handle before we just ship it?

If anyone has a filter with a parameter name of "facets" I think they will lose their filter, I could make the name really obscure, but it is optional

simonw:
> If anyone has a filter with a parameter name of "facets" I think they will lose their filter, I could make the name really obscure, but it is optional

How about using `?_facets=1` instead? That's a pattern I use a lot in Datasette - having an underscore to avoid accidental collisions with existing columns - but I don't think it's a pattern Django has used before, so it might be weird to add it for this.

simonw:
I'm trying this out against an existing Django application. It looks _really_ good!

I installed the branch with:

    pip install https://github.com/sarahboyce/django/archive/refs/heads/ticket_32539_lookups.zip

Now when I add `?facets=1` to an admin URL for anything that uses `list_filter=...` I get this:

image

Django Debug Toolbar shows me that the SQL query looks like this:

image

```sql
SELECT COUNT("feedstats_subscribercount"."id") FILTER (WHERE "feedstats_subscribercount"."path" = '/atom/entries/') AS "0__c",
       COUNT("feedstats_subscribercount"."id") FILTER (WHERE "feedstats_subscribercount"."path" = '/atom/everything/') AS "1__c",
       COUNT("feedstats_subscribercount"."id") FILTER (WHERE "feedstats_subscribercount"."path" = '/atom/links/') AS "2__c"
  FROM "feedstats_subscribercount"
```
This adds a `count / filter where` clause for every displayed value. That's fine for things with only a few unique values, but it does mean you end up with a pretty huge SQL query for columns with a large number of values. Arguably those aren't a great fit for the `list_filter=` function though, so I'm not sure if that's something we need to worry about.

I like the idea of a toggle to turn these on and off. I would want them on by default for most of my projects - I wonder if the Django `admin.ModelAdmin` subclasses could grow a `show_facets=True` option that sets that default?

simonw:
I noticed that it's not currently sorting the listed filters by the number of results:

image

When I'm working with facet interfaces generally I always like to sort them with the largest counts at the top, as then the feature doubles as a "what's most popular" feature.

This makes more sense for things like column value facets than it does for date filters - this filter interface here makes more sense sticking to its time order than sorting by those numbers:

image

It looks to me like Django defaults to sorting these filters alphabetically - as illustrated by this larger example that facets/filters by user agent:

image

I don't have a particularly strong opinion about this. I feel that sorting by numerical count makes more sense that sorting alphabetically for most cases, but that's just an intuition based on how I've used these kinds of interfaces in the past - I'm not ready to confidently state that every Django project in existence would benefit more from numerical than from alphabetical sorting here.

simonw:
Just spotted this: clicking on one of those filter links currently does _not_ persist the `?facets=1` in the URL:

facets

This feels wrong to me: when I'm interacting with facets I like to do things like drill down using one to see how that affects the numeric counts for the others.

Ideally the URLs generated for the facets should persist the `?facets=1` query string parameter.

simonw:
I'm really impressed with how slim this implementation is - love how much new functionality we get from so little code, and the tests look good to me too.

kezabelle:
> How about using `?_facets=1` instead? [...] having an underscore to avoid accidental collisions with existing columns - but I don't think it's a pattern Django has used before [...]

👍 There is some prior precedent -- `_popup` and `_to_field` keys are already used in `contrib.admin.options`.

sarahboyce:
Thank you @simonw for the review and testing :star: 
- `facets` has been updated to `_facets`
- adding `_facets=1` to the url should now be persisted in filter links
- added an attribute `show_facets` to the ModelAdmin to allow facets to be turned on by default without using the query parameter
- I did an update for getting the counts for `SimpleListFilter` to achieve this in a single query
- Updated the docs and tried to add a note around that the queries may cause performance issues

I need to think a bit more about the ordering, I can see the value of having it. Maybe the `SimpleListFilter` as it's set up with a specific order _maybe_ doesn't want to be ordered (but maybe does) and the date one mostly will look ok with the count order just sometimes 7 days ago is bigger than this year. Choosing to add it to some and not others feels like it might look like a bug. So I feel like it should be all or nothing.


I think this is ready for another set of testing and review, happy for anyone else to weigh in their thoughts on whether they are for or against the ordering also.

pauloxnet:
@sarahboyce congratulations for the functionality you propose and for its implementation. I can only suggest to use more f-strings in some places of the code because it's more readable and faster. Also, I think the Django version should be specified in the two additions to the documentation.

docs/ref/contrib/admin/filters.txt
```diff
@@ -210,3 +210,14 @@ It is possible to specify a custom template for rendering a list filter::
...
+
+
+Facets
+======
+
+.. versionadded:: 5.0
+
+Counts can be appended in brackets to each of the filter option labels by
+adding ``_facets=1`` to the URL query. These counts will update according to the
+current applied filters. To have these applied by default for a model admin,
```
smithdc1, line 222:
```suggestion
currently applied filters. To have these applied by default for a model admin,
```

docs/ref/contrib/admin/index.txt
```diff
@@ -965,6 +965,22 @@ subclass::
     editing, or deleting an object. You can have filters cleared by setting
     this attribute to ``False``.
 
+.. attribute:: ModelAdmin.show_facets
+
+    .. versionadded:: 5.0
+
+    Set ``show_facets`` to ``True`` to have all filter options appended with
+    the count of results that match that filter option in brackets. These
```
smithdc1, line 973:
```suggestion
    the count of results that match that filter option. These
```

It took me a few attempts to understand the 'in brackets' part. Maybe we can drop that part? 🤷‍♂️

docs/ref/contrib/admin/index.txt
```diff
@@ -965,6 +965,22 @@ subclass::
...
+    counts will update in line with currently applied filters. By default, this
+    is set to ``False``.
+
+    .. admonition:: Performance considerations with facets
```
smithdc1, line 977:
```suggestion
    .. admonition:: Performance considerations with facets.
```
Does this need a full stop?

docs/ref/contrib/admin/index.txt
sarahboyce: Yeah no problem :+1:

docs/ref/contrib/admin/index.txt
sarahboyce: A lot of the other admonitions don't have full stops so I guess maybe it's like a title and not needed? I'm not sure though, happy to add it in

docs/ref/contrib/admin/filters.txt
```diff
@@ -210,3 +210,14 @@ It is possible to specify a custom template for rendering a list filter::
 
 See the default template provided by Django (``admin/filter.html``) for a
 concrete example.
+
```
felixxm, line 213: Chop unnecessary blank line.
```suggestion
```

tests/admin_filters/tests.py
```diff
@@ -1315,6 +1333,307 @@ def test_simplelistfilter_with_queryset_based_lookups(self):
...
 
+    def test_filters_with_facets(self):
+        modeladmin = DecadeFilterBookAdmin(Book, site)
+
+        # Facets with no filters applied
+        request = self.request_factory.get("/", {"_facets": "1"})
+        request.user = self.alfred
+        changelist = modeladmin.get_changelist_instance(request)
+        queryset = changelist.get_queryset(request)
+        self.assertEqual(list(queryset), list(Book.objects.order_by("-id")))
```
felixxm, line 1344:
```suggestion
        self.assertSequenceEqual(queryset, Book.objects.order_by("-id"))
```

carltongibson (commented):
This is looking great. 

> I need to think a bit more about the ordering...

Unless you have a solution come to mind @sarahboyce (which you might! 😅) I'd be inclined towards taking ordering as a possible follow-up ticket, rather than blocking on it. (Most of the filters I use don't have too many values, since I generally switch to search once it gets out of hand, and this is already a massive win — headline feature for the release likely.)

django/contrib/admin/views/main.py
```diff
@@ -103,6 +111,7 @@ def __init__(
             self.page_num = 1
         self.show_all = ALL_VAR in request.GET
         self.is_popup = IS_POPUP_VAR in request.GET
+        self.add_facets = model_admin.show_facets or IS_FACETS_VAR in request.GET
```
carltongibson, line 114: I just wonder if this should be an enum `{ALWAYS|ALLOW|NEVER}` (or something) — with `NEVER` disabling the faceting entirely. 

(_"One of our admin users brought down production by adding the `IS_FACETS_VAR` query string arg..."_)


🤔

sarahboyce: Ooooh this is a clever idea, with `NEVER` as the default?

carltongibson: I thought `ALLOW` as the default.

sarahboyce: That's updated, though might not be documented in the clearest of ways. If there is a nice setting based off an enum I can look at, let me know. Happy to update :)

sarahboyce:
> Unless you have a solution come to mind @sarahboyce (which you might! sweat_smile) 

I have _something_ for the ordering [9c0f1f5](https://github.com/django/django/pull/16495/commits/9c0f1f5ba4361fad14c60c62ea11a9a7772176d9), I have put it in a seperate commit and we can still decide whether to include it or not. :+1:

docs/ref/contrib/admin/filters.txt
```diff
@@ -210,3 +210,14 @@ It is possible to specify a custom template for rendering a list filter::
...
+Facets
+======
+
+.. versionadded:: 5.0
+
+Counts can be appended in brackets to each of the filter option labels by
+adding ``_facets=1`` to the URL query. These counts will update according to
```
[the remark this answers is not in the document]
sarahboyce: You're right :+1: updated

django/contrib/admin/options.py
```diff
@@ -69,6 +69,13 @@
 
 IS_POPUP_VAR = "_popup"
 TO_FIELD_VAR = "_to_field"
+IS_FACETS_VAR = "_facets"
+
+
+class FacetFilterSetting(models.TextField):
+    NEVER = "NEVER"
+    ALLOW = "ALLOW"
+    ALWAYS = "ALWAYS"
```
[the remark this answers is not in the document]
sarahboyce: > [0]: I think. Maybe folks would want it separately documented to search for/link to. (Not sure.)

I think I know what you're after but not sure the most natural place to put it/how best to document it :thinking: 
I have made the other updates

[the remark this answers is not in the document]
carltongibson: Ok, no stress. Let me take a pop tomorrow AM with a concrete suggestion. 👍

carltongibson (commented):
Hi @sarahboyce. 

OK I pushed some suggested docs edits. Please take a look, and use as you think. (The main bit was documenting the `ShowFacets` enum as nested in the `show_facets` section, since that's the only place it applies — but it's targetable with rst roles this way.) 

I'm not sure if `change-list` is meant to have a hyphen or not. Sorry 😅


Then I just wonder — please feel free to shout "NO!" at me :) — if a UI toggle to add/remove the `_facets` query parameter just at the top of the filter list might be a nice finishing touch? 🤔 (For me, it's a **long way** until 5.0, and we could add that as an easy picking ticket for someone to follow-up with...)

docs/releases/5.0.txt
```diff
@@ -46,6 +46,10 @@ Minor features
 * The new :meth:`.AdminSite.get_log_entries` method allows customizing the
   queryset for the site's listed log entries.
 
+* Facet counts are now shown for applied filters in admin change-list UI. By
+  default, facet counts can be toggled via a ``_facets`` query parameter, but
+  this behavior can be configured via the new :attr:`.ModelAdmin.show_facets`.
+
```
carltongibson, line 52: This is a headline feature for me.

django/contrib/admin/views/main.py
```diff
@@ -103,6 +112,9 @@ def __init__(
             self.page_num = 1
         self.show_all = ALL_VAR in request.GET
         self.is_popup = IS_POPUP_VAR in request.GET
+        self.add_facets = model_admin.show_facets == ShowFacets.ALWAYS or (
+            model_admin.show_facets == ShowFacets.ALLOW and IS_FACETS_VAR in request.GET
```
carltongibson, line 116: Enum comparisons should use `is`: https://docs.python.org/3/howto/enum.html#comparisons

django/contrib/admin/filters.py
```diff
@@ -312,25 +399,48 @@ def __init__(self, field, request, params, model, model_admin, field_path):
...
             ),
             "display": _("All"),
+            "count": sum(facet_counts.values()) if add_facets else None,
         }
         none_title = ""
-        for lookup, title in self.field.flatchoices:
+        none_count = None
+        for i, (lookup, title) in enumerate(self.field.flatchoices):
+            count = facet_counts[f"{i}__c"] if add_facets else None
+            title = f"{title} ({count})" if add_facets else title
```
felixxm, line 432:
```suggestion
            if add_facets:
                count = facet_counts[f"{i}__c"]
                title = f"{title} ({count})"
            else:
                count = None
```

sarahboyce:
> OK I pushed some suggested docs edits. Please take a look, and use as you think. (The main bit was documenting the `ShowFacets` enum as nested in the `show_facets` section, since that's the only place it applies — but it's targetable with rst roles this way.)

It looks good to me! Thank you, the enum especially is much clearer :star: 

> Then I just wonder — please feel free to shout "NO!" at me :) — if a UI toggle to add/remove the `_facets` query parameter just at the top of the filter list might be a nice finishing touch? thinking (For me, it's a **long way** until 5.0, and we could add that as an easy picking ticket for someone to follow-up with...)

Ahhhh I see what you mean, it's definitely nice for feature discovery :thinking: probably quite important for it to be used
Similar to the pretty toggle for the admin light mode/dark mode. I will have a quick look later and if I'm struggling, I'll push to another ticket :+1:

django/contrib/admin/views/main.py
```diff
@@ -114,7 +126,6 @@ def __init__(
             del self.params[PAGE_VAR]
         if ERROR_FLAG in self.params:
             del self.params[ERROR_FLAG]
-
```
felixxm, line 117: Please revert unrelated formatting change.

docs/releases/5.0.txt
felixxm: Agreed.

django/contrib/admin/filters.py
```diff
@@ -454,34 +581,59 @@ def __init__(self, field, request, params, model, model_admin, field_path):
...
             ),
             "display": _("All"),
+            "count": sum(facet_counts.values()) if add_facets else None,
         }
         include_none = False
-        for val in self.lookup_choices:
+        empty_title = self.empty_value_display
+        empty_count = None
+        for i, val in enumerate(self.lookup_choices):
+            count = facet_counts[f"{i}__c"] if add_facets else None
```
felixxm, line 614: I would format this as in `ChoicesFieldListFilter` to avoid so many ternaries.

django/contrib/admin/filters.py
```diff
@@ -376,37 +487,37 @@ def __init__(self, field, request, params, model, model_admin, field_path):
             (
                 _("Today"),
                 {
-                    self.lookup_kwarg_since: str(today),
-                    self.lookup_kwarg_until: str(tomorrow),
+                    self.lookup_kwarg_since: today,
+                    self.lookup_kwarg_until: tomorrow,
```
felixxm, line 491: Moving `str()` calls to the `DateFieldListFilter.get_choices()` sound like a cleanup that can be done in advance :thinking: What do you think?

felixxm (commented):
@sarahboyce Many thanks :star: I left small suggestions.

django/contrib/admin/filters.py
sarahboyce: Tried to reduce the ternaries :+1:

django/contrib/admin/filters.py
```diff
@@ -111,19 +126,45 @@ def lookups(self, request, model_admin):
...
+            "count": sum(facet_counts.values()) if add_facets else None,
         }
-        for lookup, title in self.lookup_choices:
+        for i, (lookup, title) in enumerate(self.lookup_choices):
+            if add_facets:
+                count = facet_counts.get(f"{i}__c", -1)
+                if count != -1:
+                    title = f"{title} ({count})"
+                else:
+                    title = f"{title} (-)"
```
felixxm, line 160: Maybe walrus :smiley: Sorry for nitpicking
```suggestion
                if (count := facet_counts.get(f"{i}__c", -1)) != -1:
                    title = f"{title} ({count})"
                else:
                    title = f"{title} (-)"
```

django/contrib/admin/filters.py
sarahboyce: [fbed3f1](https://github.com/django/django/pull/16495/commits/fbed3f1283ba45259a6fe252defc8e9cc5b55d66) pulled it out into it's own commit :+1:

django/contrib/admin/filters.py
sarahboyce: No worries :+1: updated

sarahboyce:
Toggle off
image

Toggle on
image

:shrug: maybe?

akshaykedar1996:
I'd be happy to provide more information on the change made in Django's codebase!

In Django, the ModelAdmin class is used to define the admin interface for a particular model. Prior to the change made in ticket #16495, ModelAdmin did not have built-in support for facet filters, which allow users to filter data based on a set of predefined options.

In ticket #32539, support for toggleable facet filters was added to ModelAdmin. This change allows developers to specify a list of facets for a particular model, along with the options for each facet. Users can then toggle the facets to filter the data based on their preferences.

To add facet filters to a ModelAdmin class, developers can use the list_filter attribute and specify the facets as a list of strings. For example, to add facet filters for a Book model based on the author and genre fields, the following code can be used:

python

Models.py file
-----------------------------------------------------------------------------------------------------------------------
class BookAdmin(admin.ModelAdmin):
    list_filter = ('author', 'genre')
------------------------------------------------------------------------------------------------------------------------
By default, the facet filter options are displayed in a list, but developers can customize the appearance and behavior of the facet filters using the list_filter_template attribute.

Overall, the addition of toggleable facet filters to ModelAdmin provides a more user-friendly interface for filtering data in the Django admin, which is an important feature for many Django applications.

tests/admin_filters/tests.py
```diff
@@ -1150,7 +1172,7 @@ def test_simplelistfilter(self):
 
         # Make sure the correct queryset is returned
         queryset = changelist.get_queryset(request)
-        self.assertEqual(list(queryset), list(Book.objects.order_by("-id")))
+        self.assertSequenceEqual(list(queryset), list(Book.objects.order_by("-id")))
```
felixxm, line 1175: `list()` is unnecessary when you use `assertSequenceEqual()`:
```suggestion
        self.assertSequenceEqual(queryset, Book.objects.order_by("-id"))
```

tests/admin_filters/tests.py
```diff
@@ -1315,6 +1337,390 @@ def test_simplelistfilter_with_queryset_based_lookups(self):
         self.assertIs(choices[2]["selected"], False)
         self.assertEqual(choices[2]["query_string"], "?publication-decade=the+00s")
 
+    def test_filters_with_facets(self):
+        modeladmin = DecadeFilterBookAdmin(Book, site)
+
+        # Facets with no filters applied
```
felixxm, line 1343:
```suggestion
        # Facets with no filters applied.
```

tests/admin_filters/tests.py
```diff
@@ -1315,6 +1337,390 @@ def test_simplelistfilter_with_queryset_based_lookups(self):
...
+        request = self.request_factory.get("/?_facets")
+        request.user = self.alfred
+        changelist = modeladmin.get_changelist_instance(request)
+        queryset = changelist.get_queryset(request)
+        self.assertSequenceEqual(list(queryset), list(Book.objects.order_by("-id")))
```
felixxm, line 1348:
```suggestion
        self.assertSequenceEqual(queryset, Book.objects.order_by("-id"))
```

tests/admin_filters/tests.py
```diff
@@ -1315,6 +1337,390 @@ def test_simplelistfilter_with_queryset_based_lookups(self):
...
+            self.assertIn("_facets", choice["query_string"])
+
+        # When filtered for books in the 1980s, counts are updated
+        request = self.request_factory.get(
+            "/", {"_facets": "", "publication-decade": "the 80s"}
+        )
+        request.user = self.alfred
+        changelist = modeladmin.get_changelist_instance(request)
+        queryset = changelist.get_queryset(request)
+        self.assertEqual(list(queryset), [])
```
felixxm, line 1438:
```suggestion
        self.assertSequenceEqual(queryset, [])
```

tests/admin_filters/tests.py
```diff
@@ -1315,6 +1337,390 @@ def test_simplelistfilter_with_queryset_based_lookups(self):
...
+        request = self.request_factory.get(
+            "/", {"author__id__exact": self.alfred.pk, "_facets": ""}
+        )
+        request.user = self.alfred
+        changelist = modeladmin.get_changelist_instance(request)
+        queryset = changelist.get_queryset(request)
+        self.assertEqual(
+            list(queryset),
+            list(Book.objects.filter(author=self.alfred).order_by("-id")),
+        )
```
felixxm, line 1517:
```suggestion
        self.assertSequenceEqual(
            queryset, Book.objects.filter(author=self.alfred).order_by("-id")
        )
```

tests/admin_filters/tests.py
```diff
@@ -1315,6 +1337,390 @@ def test_simplelistfilter_with_queryset_based_lookups(self):
...
+
+    def test_facets_disallowed(self):
+        modeladmin = DecadeFilterBookAdminDisallowFacets(Book, site)
+
+        # Facets not their even when in url query
+        request = self.request_factory.get("/?_facets")
+        request.user = self.alfred
+        changelist = modeladmin.get_changelist_instance(request)
+        queryset = changelist.get_queryset(request)
+        self.assertSequenceEqual(list(queryset), list(Book.objects.order_by("-id")))
```
felixxm, line 1665:
```suggestion
        self.assertSequenceEqual(queryset, Book.objects.order_by("-id"))
```

tests/admin_views/tests.py
```diff
@@ -807,6 +807,36 @@ def test_limited_filter(self):
             msg_prefix="Changelist filter not correctly limited by limit_choices_to",
         )
 
+    def test_change_list_facet_toggle(self):
+        # Toggle visible when show_facet is the default of admin.ShowFacets.ALLOW
```
felixxm, line 811: Wrap comments at 79 chars
```suggestion
        # Toggle is visible when show_facet is the default of
        # admin.ShowFacets.ALLOW.
```

tests/admin_views/tests.py
```diff
@@ -807,6 +807,36 @@ def test_limited_filter(self):
...
+            '<a href="?_facets=True" class="viewlink">Add Facets</a>',
+            msg_prefix="Expected facet filter toggle not found in changelist view",
+        )
+        response = self.client.get(f"{admin_url}?_facets=True")
+        self.assertContains(
+            response,
+            '<a href="?" class="hidelink">Remove Facets</a>',
+            msg_prefix="Expected facet filter toggle not found in changelist view",
+        )
+        # Toggle not visible when show_facet is admin.ShowFacets.ALWAYS
```
felixxm, line 825:
```suggestion
        # Toggle is not visible when show_facet is admin.ShowFacets.ALWAYS.
```

tests/admin_views/tests.py
```diff
@@ -807,6 +807,36 @@ def test_limited_filter(self):
...
+        response = self.client.get(reverse("admin:admin_views_workhour_changelist"))
+        self.assertNotContains(
+            response,
+            "Facets",
+            msg_prefix="Expected not to find facet toggle in changelist view",
+        )
+        # Toggle not visible when show_facet is admin.ShowFacets.NEVER
```
felixxm, line 832:
```suggestion
        # Toggle is not visible when show_facet is admin.ShowFacets.NEVER.
```

tests/admin_filters/tests.py
```diff
@@ -1315,6 +1337,390 @@ def test_simplelistfilter_with_queryset_based_lookups(self):
...
+        self.assertEqual(choices[2]["display"], "bob (1)")
+        for choice in choices:
+            self.assertIn("_facets", choice["query_string"])
+
+        filterspec_simplelistfilter = changelist.get_filters(request)[0][1]
+        choices = list(filterspec_simplelistfilter.choices(changelist))
+        self.assertEqual(choices[1]["display"], "the 2000's (2)")
+        self.assertEqual(choices[2]["display"], "the 1990's (1)")
+        self.assertEqual(choices[3]["display"], "the 1980's (0)")
+        self.assertEqual(choices[4]["display"], "other decades (-)")
```
felixxm, line 1362: Can we add a hook for checking `choices`? For example:
```python
def assertChoicesDisplay(self, choices, expected_displays):
    for choice, expected_display in zip(choices[1:], expected_displays, strict=True):
        self.assertEqual(choice["display"], expected_display])
```
and use it as:
```python
    self.assertChoicesDisplay(
        choices, ["the 2000's (2)", "the 1990's (1)", "the 1980's (0)", "other decades (-)"]
    )
```

docs/releases/5.0.txt
```diff
@@ -37,6 +37,13 @@ compatible with Django 5.0.
 What's new in Django 5.0
 ========================
 
+Facet filters in the admin
+--------------------------
+
+Facet counts are now shown for applied filters in admin change-list UI. By
```
felixxm, line 43:
```suggestion
Facet counts are now shown for applied filters in the admin changelist. By
```

docs/releases/5.0.txt
```diff
@@ -37,6 +37,13 @@ compatible with Django 5.0.
...
+default, facet counts can be toggled via a ``_facets`` query parameter, but
+this behavior can be configured via the new :attr:`.ModelAdmin.show_facets`.
```
felixxm, line 45:
```suggestion
this behavior can be changed via the new :attr:`.ModelAdmin.show_facets` attribute.
```

docs/releases/5.0.txt
```diff
@@ -37,6 +37,13 @@ compatible with Django 5.0.
...
+
```
felixxm, line 46: I'd add a reference to the `Facets` section in the `docs/ref/contrib/admin/filters.txt`

docs/ref/contrib/admin/index.txt
```diff
@@ -965,6 +965,54 @@ subclass::
     editing, or deleting an object. You can have filters cleared by setting
     this attribute to ``False``.
 
+.. attribute:: ModelAdmin.show_facets
+
+    .. versionadded:: 5.0
+
+    Controls whether facet counts are displayed for filters in the change-list
+    UI. Default: :attr:`.ShowFacets.ALLOW`.
```
felixxm, line 973:
```suggestion
    Controls whether facet counts are displayed for filters in the admin changelist
    Defaults to :attr:`.ShowFacets.ALLOW`.
```

felixxm (commented):
@sarahboyce Thanks :+1: I left small suggestions.

Is `_facets` preserved as changelist filters?

docs/releases/5.0.txt
```diff
@@ -37,6 +37,13 @@ compatible with Django 5.0.
 What's new in Django 5.0
 ========================
 
+Facet filters in the admin
+--------------------------
+
+Facet counts are now shown for applied filters in admin change-list UI. By
+default, facet counts can be toggled via a ``_facets`` query parameter, but
```
carltongibson, line 44: Just one query: do we want to say, "can be toggled in the UI" now, rather than mentioning the query parameter? (Here and in docs.)

tests/admin_filters/tests.py
sarahboyce: Nice idea :+1:

tests/admin_filters/tests.py
sarahboyce: `self.assertSequenceEqual(queryset, list(Book.objects.order_by("-id")))`
I can remove the `list()` from queryset, but I need it wrapping the other one. Not sure why though, to have the query evaluated maybe? :thinking:

docs/releases/5.0.txt
sarahboyce: Yes true, it probably is less confusing also - I've tried to update


There is a picture at the top here which I would like to update, I don't suppose this is someones test app?
image

https://docs.djangoproject.com/en/4.1/ref/contrib/admin/filters/


I saw there is this ticket: https://code.djangoproject.com/ticket/34043 which would be generally useful and I could also use for updating documentation screenshots. Might be a good one for a Django sprint maybe

tests/admin_filters/tests.py
felixxm: Ah yes because we have two querysets :+1:

sarahboyce:
> Is `_facets` preserved as changelist filters?

It's preserved when clicking on other filter options (so the issue described earlier). I think it wouldn't preserve if you do things like click on another table/changelist view or click on the breadcrumbs.

alewish2196:
Your email has been received and will be addressed by our team shortly.

alewish2196:
I received your email and wanted to confirm that it has been received and
is being reviewed.

simonw:
> Toggle off image
> 
> Toggle on image
> 
> 🤷 maybe?

I think "show counts / hide counts" would be clearer for the button text here.

docs/releases/5.0.txt
simonw: I have a pretty elaborate mechanism for automated screenshots for my own project, I wonder if Django might benefit from something similar?

Definitely not as a blocker for this change though!

https://simonwillison.net/2022/Oct/14/automating-screenshots/

pauloxnet:
> I think "show counts / hide counts" would be clearer for the button text here.

I agree.

sarahboyce:
Thank you :+1: updated

Should I also localize the numbers? So `Test (1000)` looks like `Test (1,000)` / `Test (1.000)`
I think I need to use something like `formats.localize(count, use_l10n=True)` everywhere if we want it

felixxm:
> > Is `_facets` preserved as changelist filters?
> 
> It's preserved when clicking on other filter options (so the issue described earlier). I think it wouldn't preserve if you do things like click on another table/changelist view or click on the breadcrumbs.

It looks like it it preserved as a part of changelist filters (`_changelist_filters`) :+1:

felixxm:
`Hide/Show counts` should have the same font size as `Clear all filters`
image
We should also wrap both in `div` and move the bottom border from `#changelist-filter #changelist-filter-clear a` to the new `div`.

felixxm:
> > Unless you have a solution come to mind @sarahboyce (which you might! sweat_smile)
> 
> I have _something_ for the ordering [9c0f1f5](https://github.com/django/django/pull/16495/commits/9c0f1f5ba4361fad14c60c62ea11a9a7772176d9), I have put it in a seperate commit and we can still decide whether to include it or not. +1

I found ordering by facets quite disturbing as the list of choices is constantly changing. I'm also not sure what to do with facets when an option is selected because then all other options appear with `(0)`. It's confusing that changing the filter to an option with `(0)` shows a non-empty list. We should probably ignore the selected filter when calculating facets for the same option :thinking:

https://user-images.githubusercontent.com/2865885/221117042-45fe0c1f-daa6-4507-903a-e17993da994d.mp4

django/contrib/admin/templates/admin/change_list.html
```diff
@@ -75,9 +75,14 @@
         {% if cl.has_filters %}
           <div id="changelist-filter">
             <h2>{% translate 'Filter' %}</h2>
-            {% if cl.has_active_filters %}<h3 id="changelist-filter-clear">
-              <a href="{{ cl.clear_all_filters_qs }}">&#10006; {% translate "Clear all filters" %}</a>
-            </h3>{% endif %}
+            {% if cl.is_facets_optional or cl.has_active_filters %}<div id="changelist-filter-extra-actions">
+              {% if cl.is_facets_optional %}<h3 id="changelist-filter-facets">
+                {% if cl.add_facets %}<a href="{{ cl.remove_facet_link }}" class="hidelink">{% translate 'Hide counts' %}</a>{% else %}<a href="{{ cl.add_facet_link }}">{% translate 'Show counts' %}</a>{% endif %}
```
felixxm, line 80: `viewlink` disappeared:
```suggestion
                {% if cl.add_facets %}<a href="{{ cl.remove_facet_link }}" class="hidelink">{% translate 'Hide counts' %}</a>{% else %}<a href="{{ cl.add_facet_link }}" class="viewlink">{% translate 'Show counts' %}</a>{% endif %}
```

sarahboyce:
image
The size is updated :+1: nice suggestion

Also made it not include the current filter option when one of same group applied (screenshot hopefully illustrates it) but would appreciate feedback on that bit as feels a little hacky.

 On the reordering, I can remove it :shrug: I don't really mind, I added it based of @simonw 's suggestion here but also
> I don't have a particularly strong opinion about this. I feel that sorting by numerical count makes more sense that sorting alphabetically for most cases, but that's just an intuition based on how I've used these kinds of interfaces in the past - I'm not ready to confidently state that every Django project in existence would benefit more from numerical than from alphabetical sorting here.

I guess with the adjustment that it doesn't apply to the one you selected, it should move a little less (only the ones you weren't looking at move) maybe it's now less disturbing? Or is it still jarring?

felixxm:
@sarahboyce Can you drop reordering (at least for now)? This should simplify the current implementation, e.g. `OptionalFacetsMixin` would be unnecessary.

I'd like to merge it :star2:

sarahboyce:
It should be dropped :+1:

felixxm:
@sarahboyce I merged the first commit in 4cb55733527d39ebf58c1987f72a19b7c2e0bac6 and pushed small edits to a few files.

I noticed an issue while polishing this patch, i.e. `facets` is not preserved when using `X total` option:


https://user-images.githubusercontent.com/2865885/222430022-c0bffc24-31ff-4ebf-a08c-66590ab7ddf8.mp4



Can you take a look? Please push changes in a separate commit.

sarahboyce:
Well spotted :+1: thank you for testing :star:

felixxm:
@sarahboyce I tried to move some common logic to the `FacetsMixin`, what do you think?

django/contrib/admin/filters.py
```diff
@@ -269,7 +269,7 @@ def get_facet_counts(self, pk_attname, filtered_qs):
             for pk_val, _ in self.lookup_choices
         }
         if self.include_empty_choice:
-            counts["empty__c"] = models.Count(
+            counts["__c"] = models.Count(
```
felixxm, line 272: Using `empty__c` could potentially conflict with the _"empty"_ primary key.

sarahboyce:
> @sarahboyce I tried to move some common logic to the `FacetsMixin`, what do you think?

I like it :+1:

docs/releases/5.0.txt
```diff
@@ -37,6 +37,14 @@ compatible with Django 5.0.
 What's new in Django 5.0
 ========================
 
+Facet filters in the admin
+--------------------------
+
+Facet counts are now shown for applied filters in the admin changelist when
+toggled on via the UI. This behavior can be changed via the new
+:attr:`.ModelAdmin.show_facets` attribute. For more information see
```
smithdc1, line 45: Maybe the toggle in the UI should be referred to as headline feature rather than the `ModelAdmin` attribute?

felixxm:
I reorganized and simplified tests with `subTest()`, we're getting close... :crossed_fingers:

docs/releases/5.0.txt
felixxm: @smithdc1 Any wording suggestion?

smithdc1: I must have completely miss-read this earlier today 🤦.

The first sentence references the UI so I think it's fine as it is (for some reason I thought this section made no reference to the UI toggle at all). 

Sorry.

felixxm:
@sarahboyce Thanks for all your efforts :1st_place_medal: I updated screenshots and this is ready to merge! :tada:

felixxm:
buildbot, test on selenium.

simonw:
I deployed this on my blog and it's working great!

image

PR #16497, opened 2023-01-24
Fix UnicodeEncodeError when running unit tests in Windows

stephenatgithub:
When running unit test in Windows, got this error:
UnicodeDecodeError: 'cp950' codec can't decode byte 0xc3 in position 268: illegal multibyte sequence

Besides Linux or MacOS, this is to provide an extra resolution for Windows.
https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/unit-tests/#many-test-failures-with-unicodeencodeerror

### DIFF
--- a/docs/internals/contributing/writing-code/unit-tests.txt
+++ b/docs/internals/contributing/writing-code/unit-tests.txt
@@ -433,6 +433,26 @@ Run the ``locale`` command to confirm the change. Optionally, add those export
 commands to your shell's startup file (e.g. ``~/.bashrc`` for Bash) to avoid
 having to retype them.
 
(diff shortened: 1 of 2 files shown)


carltongibson (commented):
Can you set the [PYTHONUTF8](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONUTF8) environment variable, as per the [Common Pitfalls](https://docs.djangoproject.com/en/4.1/howto/windows/#common-pitfalls) section of the How to install Django on Windows
docs.

This should then pass. We don't support running with a non-utf-8 enconding. 

I'm going to close this, assuming that setting the env var resolves your issue. 

Thanks.

PR #16506, opened 2023-01-27
Fix max line length to match pep-8

ivnvxd:
Small fix to match PEP-8 max line length rule. Password validators' names in `settings.py` are longer than 79 characters.

### DIFF
--- a/django/conf/project_template/project_name/settings.py-tpl
+++ b/django/conf/project_template/project_name/settings.py-tpl
@@ -86,16 +86,20 @@ DATABASES = {
 
 AUTH_PASSWORD_VALIDATORS = [
     {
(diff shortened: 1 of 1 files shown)


ngnpope:
Thanks, but we format with `black` which is configured to have a maximum line length of 88 characters. Using an auto-formatter keeps a consistent style and avoids needing to be concerned with tweaks like this.

ivnvxd:
Thanks for the answer! I got it. But the length of the first line with `UserAttributeSimilarityValidator` is still more than 88 characters. Is that ok?

ngnpope:
Yes. Black won't aggressively wrap some long strings, but we generally don't make these changes to address that. As long as the auto-formatter is happy, that's fine 🙂

PR #16507, opened 2023-01-28
Fixed typo in the quertysets docs for alias()

deepak-s24:
Typo correction - "you" was missing in the example.


deepak-s24:
When I was reading the docs for alias() I found that "you" was missing in the example.

ngnpope:
Thanks, but this reads correctly without the additional "you".

PR #16543, opened 2023-02-11
Fixed #34331: Add support for prefetch_related to QuerySet.aiterator()

callerbear:
If possible, including this in the 4.2 production release would be very much appreciated...  we'll be on that LTS release until release 5.2 becomes available quite a long time from now.

This patch uses the "chunked" prefetch approach originally implemented in .iterator(), suitably modified for async.  For the sake of simplicity, It wraps the call to prefetch_related_items() within sync_to_async, instead of building a complete async version of that method.

### DIFF
--- a/django/db/models/query.py
+++ b/django/db/models/query.py
@@ -5,6 +5,7 @@ The main QuerySet implementation. This provides the public API for the ORM.
 import copy
 import operator
 import warnings
+from collections import deque
 from itertools import chain, islice
 
 from asgiref.sync import sync_to_async
@@ -544,19 +545,31 @@ class QuerySet(AltersData):
         An asynchronous iterator over the results from applying this QuerySet
         to the database.
         """
-        if self._prefetch_related_lookups:
-            raise NotSupportedError(
-                "Using QuerySet.aiterator() after prefetch_related() is not supported."
-            )
         if chunk_size <= 0:
             raise ValueError("Chunk size must be strictly positive.")
         use_chunked_fetch = not connections[self.db].settings_dict.get(
             "DISABLE_SERVER_SIDE_CURSORS"
         )
+
+        chunk = deque()
+        num_items = 0
         async for item in self._iterable_class(
             self, chunked_fetch=use_chunked_fetch, chunk_size=chunk_size
         ):
-            yield item
+            chunk.append(item)
+            num_items += 1
+            if num_items == chunk_size:
+                await sync_to_async(prefetch_related_objects)(
+                    chunk, *self._prefetch_related_lookups
+                )
+                for each_item in chunk:
+                    yield each_item
+                chunk.clear()
+                num_items = 0
+
+        if num_items:
+            for each_item in chunk:
+                yield each_item
 
     def aggregate(self, *args, **kwargs):
         """
--- a/django/db/models/query.py
+++ b/django/db/models/query.py
@@ -567,7 +567,10 @@ class QuerySet(AltersData):
                 chunk.clear()
                 num_items = 0
 
-        if num_items:
+        if chunk:
+           await sync_to_async(prefetch_related_objects)(
+                chunk, *self._prefetch_related_lookups
+            )
             for each_item in chunk:
                 yield each_item
 
--- a/django/db/models/query.py
+++ b/django/db/models/query.py
@@ -5,7 +5,6 @@ The main QuerySet implementation. This provides the public API for the ORM.
 import copy
 import operator
 import warnings
-from collections import deque
 from itertools import chain, islice
 
 from asgiref.sync import sync_to_async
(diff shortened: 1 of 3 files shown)


django/db/models/query.py
```diff
@@ -544,19 +545,31 @@ async def aiterator(self, chunk_size=2000):
...
         ):
-            yield item
+            chunk.append(item)
+            num_items += 1
+            if num_items == chunk_size:
+                await sync_to_async(prefetch_related_objects)(
+                    chunk, *self._prefetch_related_lookups
+                )
+                for each_item in chunk:
+                    yield each_item
```
charettes, line 566: I _think_ you can 

```suggestion
                yield from chunk
```

django/db/models/query.py
```diff
@@ -544,19 +545,31 @@ async def aiterator(self, chunk_size=2000):
...
+                chunk.clear()
+                num_items = 0
+
+        if num_items:
```
charettes, line 570: You'll want to prefetch related as well here if there are some times left in the buffer

```suggestion
        if chunk:
           await sync_to_async(prefetch_related_objects)(
                chunk, *self._prefetch_related_lookups
            )
```

django/db/models/query.py
```diff
@@ -544,19 +545,31 @@ async def aiterator(self, chunk_size=2000):
...
+            for each_item in chunk:
+                yield each_item
```
charettes, line 572:
```suggestion
            yield from chunk
```

django/db/models/query.py
```diff
@@ -544,19 +545,31 @@ async def aiterator(self, chunk_size=2000):
...
-                "Using QuerySet.aiterator() after prefetch_related() is not supported."
-            )
         if chunk_size <= 0:
             raise ValueError("Chunk size must be strictly positive.")
         use_chunked_fetch = not connections[self.db].settings_dict.get(
             "DISABLE_SERVER_SIDE_CURSORS"
         )
+
+        chunk = deque()
+        num_items = 0
```
charettes, line 555: What are the benefits of using a `deque` given we don't insert at `0` or bound the `deque` to `chunk_size`?

It seems we could use good ol' list here and possibly `len(chunk)` as well?

charettes (commented):
Thanks for taking the time to submit a PR @callerbear.

tests/async/test_async_queryset.py
```diff
@@ -62,6 +64,11 @@ async def test_aiterator_invalid_chunk_size(self):
                 async for m in qs:
                     pass
 
+    async def test_aiterator_prefetch_related(self):
+        async for s in SimpleModel.objects.prefetch_related("relatedmodel_set"):
+            count = await s.relatedmodel_set.acount()
+            self.assertEqual(count, 1)
```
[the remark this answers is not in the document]
callerbear: I thought about this point quite a bit while creating this test.  First, I wasn't able to use an "assertNumQueries", because the way it processes the connection object isn't valid in an async method.  Secondly, if the test does not execute a prefetch, then accessing the counts of the related manager would cause a lazy read, which would raise the "cannot execute a synchronous operation from an async context" error.  

I wasn't able to think of a test that was more precise, even after looking through other tests in the 'async' area.

Would you have any suggestions for improving this?

django/db/models/query.py
callerbear: I had tried that, but apparently the "yield from" syntax is not valid from an async method.  PEP 525 contains a discussion of why that syntax would be difficult to implement.

django/db/models/query.py
callerbear: Forehead smack -- of course we need a "prefetch_related items" in the last part!  Oops...

django/db/models/query.py
callerbear: When dealing with lists that have potentially thousands of entries (the default chunk size is 2000, and we commonly use a chunk size of 5000 for lengthy queries), I've been somewhat leery of the overhead involved in resizing lists.  I was also a little concerned about the cost of a len() function, although some quick googling a moment ago suggests that len() is an O(1) process.   Using len would certainly be cleaner, and using a list if we're not concerned about the overhead of reallocating list contents as the list grows.

django/db/models/query.py
```diff
@@ -544,19 +544,31 @@ async def aiterator(self, chunk_size=2000):
...
         )
+
+        chunk = []
         async for item in self._iterable_class(
             self, chunked_fetch=use_chunked_fetch, chunk_size=chunk_size
         ):
-            yield item
+            chunk.append(item)
+            if len(chunk) == chunk_size:
+                await sync_to_async(prefetch_related_objects)(
```
charettes, line 559: I guess a sense that we could reuse the `SyncToAsync(prefetch_related_objects)` to avoids creating multiple of them here by defining a `aprefetch_related_objects = sync_to_async(prefetch_related_objects)` outside of the loop and reusing it here and below

```suggestion
                await aprefetch_related_objects(chunk, *self._prefetch_related_lookups)
```

django/db/models/query.py
```diff
@@ -544,19 +544,31 @@ async def aiterator(self, chunk_size=2000):
...
+                    chunk, *self._prefetch_related_lookups
+                )
+                for each_item in chunk:
+                    yield each_item
+                chunk.clear()
+
+        if chunk:
+            await sync_to_async(prefetch_related_objects)(
+                chunk, *self._prefetch_related_lookups
+            )
```
charettes, line 569:
```suggestion
            await aprefetch_related_objects(chunk, *self._prefetch_related_lookups)
```

tests/async/test_async_queryset.py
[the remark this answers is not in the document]
charettes: > First, I wasn't able to use an "assertNumQueries", because the way it processes the connection object isn't valid in an async method.

Well that sucks :/

https://github.com/django/django/blob/9af0e9b3bade140fa3590b2f5831af932c35dc13/tests/async/test_async_queryset.py#L238-L240

I think that adding a similar TODO and making sure that `s` has the `_prefetched_objects_cache` set on it.

charettes:
👋 @callerbear, just passing by while reviewing PR and was wondering if you were still interested in working on this issue. Happy to provide more guidance if necessary.

callerbear:
Thanks, Simon.  I’m still interested, but life became very full for the
last many weeks. (Wife had shoulder replacement surgery, attended three
conferences, had some time-critical issues to solve for work.   I'm just
about caught up, and hope to be able to finish this one off soon.

By the way, our initial use of top-to-bottom async code (well, at least
down into Django's query sets) seems to be working very well so far.

baj

PR #16568, opened 2023-02-17
Fixed #18307 -- Fixed formset add_fields().

laurensverhoeven:
Fixes https://code.djangoproject.com/ticket/34349#ticket 

## Description
Formsets' `add_fields()` method fails in some circumstances if the argument `index` is `None`.

When a FormSet has the attributes `self.can_delete == True` and `self.can_delete_extra == False`, calling the `add_fields()` method on that FormSet fails if the argument `index` is `None`. This occurs for example when calling `FormSet.empty_form()`. The result is that the method raises the exception `TypeError: '<' not supported between instances of 'NoneType' and 'int'`. 

```python
MyFormSet = forms.formset_factory(
    form=MyForm,
    can_delete=True,
    can_delete_extra=False,
)
my_formset = MyFormSet(
    initial=None,
)
print(my_formset.empty_form)
```

### DIFF
--- a/django/forms/formsets.py
+++ b/django/forms/formsets.py
@@ -490,7 +490,12 @@ class BaseFormSet(RenderableFormMixin):
                     required=False,
                     widget=self.get_ordering_widget(),
                 )
-        if self.can_delete and (self.can_delete_extra or index < initial_form_count):
+        if (
+            self.can_delete and (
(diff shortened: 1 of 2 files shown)


laurensverhoeven:
Replaced by https://github.com/django/django/pull/16569

PR #16581, opened 2023-02-21
Stripped unused aggregate query values and aliases.

charettes:
This is still a draft, didn't meant to mark as ready for review.

### DIFF
--- a/django/db/models/sql/compiler.py
+++ b/django/db/models/sql/compiler.py
@@ -2074,7 +2074,7 @@ class SQLAggregateCompiler(SQLCompiler):
         inner_query_sql, inner_query_params = self.query.inner_query.get_compiler(
             self.using,
             elide_empty=self.elide_empty,
-        ).as_sql(with_col_aliases=True)
+        ).as_sql()
         sql = "SELECT %s FROM (%s) subquery" % (sql, inner_query_sql)
         params += inner_query_params
         return sql, params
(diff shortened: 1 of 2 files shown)


django/db/models/sql/query.py
```diff
@@ -484,10 +484,8 @@ def get_aggregation(self, using, aggregate_exprs):
             ):
                 # In case of Model.objects[0:3].count(), there would be no
                 # field selected in the inner query, yet we must use a subquery.
-                # So, make sure at least one field is selected.
-                inner_query.select = (
-                    self.model._meta.pk.get_col(inner_query.get_initial_alias()),
-                )
+                # So, make sure at least one value is selected.
+                inner_query.select = (Value(1),)
```
charettes, line 488: No need to actually select a field if it's discarded by the outer query anyway.

django/db/models/sql/compiler.py
```diff
@@ -2074,7 +2074,7 @@ def as_sql(self):
         inner_query_sql, inner_query_params = self.query.inner_query.get_compiler(
             self.using,
             elide_empty=self.elide_empty,
-        ).as_sql(with_col_aliases=True)
+        ).as_sql()
```
charettes, line 2077: These aliases are not used anymore as we systematically `GROUP BY` position.

charettes:
buildbot, test on oracle.

PR #16615, opened 2023-03-03
Make resolve_related_fields work with self referencing fields on an a…

zuzelvp:
…bstract model

We have an abstract base model `MyModel` which has a foreign key with `to='self'` . When calling `MyModel._meta.get_field('my_field').target_field` we get the exception **ValueError: Related model 'self' cannot be resolved** because `MyModel._meta.get_field('my_field').remote_field.model` is `'self'`. We also checked the values of `MyModel._meta.get_field('my_field').from_fields` which is `['self']` and `MyModel._meta.get_field('my_field').to_fields` which is `[None]`. This error was discovered because we had setup the mypy using django-stubs

### DIFF
--- a/django/db/models/fields/related.py
+++ b/django/db/models/fields/related.py
@@ -707,7 +707,7 @@ class ForeignObject(RelatedField):
             raise ValueError(
                 "Foreign Object from and to fields must be the same non-zero length"
             )
-        if isinstance(self.remote_field.model, str):
+        if isinstance(self.remote_field.model, str) and not self.remote_field.model == RECURSIVE_RELATIONSHIP_CONSTANT:
             raise ValueError(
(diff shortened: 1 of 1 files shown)


Aliquoted:
Your pull request fails checks. Don't worry, it's likely simple fixes.

Linter "flake8" is throwing an error. **Line 710 exceeds 88 characters**, which makes it fail. Separate the line.

Linter "black" is also throwing an error. I am not familiar with this.
`Oh no! 💥 💔 💥`
`1 file would be reformatted, 2744 files would be left unchanged.`
`Error: Process completed with exit code 1.`
Why it doesn't reformat the file and instead exits is beyond me. A more experienced contributor could help.

PR #16630, opened 2023-03-05
Fixed #25230, Fixed #30685 -- Optimized distinct().count().

charettes:
This still needs a battery of test as the `distinct().count()` coverage is relatively poor in the test suite but this will hopefully address ticket-25230 and ticket-30685.

Missing test coverage
- [ ] `Expression.deterministic`
- [ ] `distinct().count()` with `Random()` annotation
- [ ] `distinct().count()` with existing aggregation
- [ ] `distinct(*on).count()` coverage

### DIFF
--- a/django/db/models/functions/math.py
+++ b/django/db/models/functions/math.py
@@ -157,6 +157,7 @@ class Radians(NumericOutputFieldMixin, Transform):
 class Random(NumericOutputFieldMixin, Func):
     function = "RANDOM"
     arity = 0
+    deterministic = False
 
     def as_mysql(self, compiler, connection, **extra_context):
         return super().as_sql(compiler, connection, function="RAND", **extra_context)
--- a/django/db/models/sql/query.py
+++ b/django/db/models/sql/query.py
@@ -533,7 +533,30 @@ class Query(BaseExpression):
         Perform a COUNT() query using the current filter constraints.
         """
         obj = self.clone()
-        return obj.get_aggregation(using, {"__count": Count("*")})["__count"]
+        count = Count("*")
+        if self.distinct:
+            if self.distinct_fields:
+                # Only DISTINCT ON fields are relevant to provide an adequate
+                # count, every other fields can be discarded.
+                obj.set_values(self.distinct_fields)
+                obj.distinct_fields = ()
+                # A single non-null DISTINCT ON target can avoid a subquery
+                # pushdown by using COUNT(DISTINCT).
+                if len(obj.select) == 1 and not obj.select[0].target.null:
+                    obj.distinct = False
+                    count = Count(obj.select[0], distinct=True)
+            elif (
+                self.default_cols
+                and not self.extra_select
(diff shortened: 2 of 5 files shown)


django/db/models/sql/query.py
```diff
@@ -532,8 +532,24 @@ def get_count(self, using):
         """
         Perform a COUNT() query using the current filter constraints.
         """
+        count_distinct = None
+        if self.distinct and not self.is_sliced and not self.extra_select:
+            if len(self.distinct_fields) == 1:
+                count_distinct = self.distinct_fields[0]
+            elif len(self.values_select) == 1:
+                count_distinct = self.values_select[0]
```
charettes, line 540: I think we could use `self.select` here.

django/db/models/sql/query.py
```diff
@@ -532,8 +532,24 @@ def get_count(self, using):
...
+                count_distinct = "pk"
         obj = self.clone()
-        return obj.get_aggregation(using, {"__count": Count("*")})["__count"]
+        if count_distinct:
+            obj.distinct = False
+            obj.distinct_fields = ()
+            count = Count(count_distinct, distinct=True)
+        else:
+            count = Count("*")
+        return obj.get_aggregation(using, {"__count": count})["__count"]
```
charettes, line 552: It's possible that the logic above could be moved to `get_aggregation` based on `Aggregate.allow_distinct` but that's a larger can of worms.

django/db/models/functions/math.py
```diff
@@ -157,6 +157,7 @@ def as_oracle(self, compiler, connection, **extra_context):
 class Random(NumericOutputFieldMixin, Func):
     function = "RANDOM"
     arity = 0
+    volatile = True
```
ngnpope, line 160: Would `deterministic = False` be better for consistency with SQLite's `create_function()`? Also, PostgreSQL has [three classifications](https://www.postgresql.org/docs/15/xfunc-volatility.html): `VOLATILE`, `STABLE`, and `IMMUTABLE`. There is extra meaning attached to something being volatile, e.g. it can modify the database. But then again it mentions `random()`. So never mind, stick with `volatile`. Naming, huh!? 😅

charettes: yeah that's the doc I used t pick `volatile` after spending too much tinkering about a proper name 😅.

I think `deterministic` makes more sense in the use case we want to make of it in this particular context so I'll change it to that

By digging a bit more in the bug tracker today it seems the `DISTINCT` random problem also ties into other issues such as ticket-17990 which relates to ticket-24462 and the whole _extra ordering_ that gets added when ordering against columns that are not selected.

It seems there's a whole extra volume of fun associated with the notion of nondeterministic expressions :)

charettes:
Another case that must be covered here is that `COUNT(DISTINCT col)` cannot be used if `col` is nullable.

https://dbfiddle.uk/HBf0068t

charettes:
I guess I'll eventually come back to this but I'm not actively working on it.

PR #16636, opened 2023-03-08
Fixed #34391 -- Added async-compatible interface to auth functions and related methods test clients.

bigfootjon:
Ticket: https://code.djangoproject.com/ticket/34391
Forum discussion: https://forum.djangoproject.com/t/asyncifying-django-contrib-auth-and-signals-and-maybe-sessions/18770

This adds an async *interface* to `django.contrib.auth`. The actual underlying implementation remains synchronous by way of `sync_to_async` wrappers.

I also took the liberty of updating the implementation given in `AuthenticationMiddleware` to use the new common API, and also updated the test client helpers to use these new methods.

One open question: should we add an async version of `User.objects.create_user()`? After writing this I realized that they go hand in hand, but it wasn't part of the ticket I opened so I'm flagging here. I'm happy to add an async version of that function if reviewers agree.

### DIFF
--- a/django/contrib/auth/__init__.py
+++ b/django/contrib/auth/__init__.py
@@ -1,6 +1,8 @@
 import inspect
 import re
 
+from asgiref.sync import sync_to_async
+
 from django.apps import apps as django_apps
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured, PermissionDenied
@@ -91,6 +93,12 @@ def authenticate(request=None, **credentials):
     )
 
 
+@sensitive_variables("credentials")
+async def aauthenticate(request=None, **credentials):
+    """See authenticate()."""
+    return await sync_to_async(authenticate)(request, **credentials)
+
+
 def login(request, user, backend=None):
     """
     Persist a user id and a backend in the request. This way a user doesn't
@@ -144,6 +152,11 @@ def login(request, user, backend=None):
     user_logged_in.send(sender=user.__class__, request=request, user=user)
 
 
+async def alogin(request, user, backend=None):
+    """See login()."""
+    return await sync_to_async(login)(request, user, backend)
+
+
 def logout(request):
     """
     Remove the authenticated user's ID from the request and flush their session
@@ -160,6 +173,11 @@ def logout(request):
         from django.contrib.auth.models import AnonymousUser
 
         request.user = AnonymousUser()
+
+
+async def alogout(request):
+    """See logout()."""
+    return await sync_to_async(logout)(request)
 
 
 def get_user_model():
@@ -223,6 +241,11 @@ def get_user(request):
     return user or AnonymousUser()
 
 
+async def aget_user(request):
+    """See get_user()."""
+    return await sync_to_async(get_user)(request)
+
+
 def get_permission_codename(action, opts):
     """
     Return the codename of the permission for the specified action.
@@ -242,3 +265,8 @@ def update_session_auth_hash(request, user):
     request.session.cycle_key()
     if hasattr(user, "get_session_auth_hash") and request.user == user:
         request.session[HASH_SESSION_KEY] = user.get_session_auth_hash()
+
+
+async def aupdate_session_auth_hash(request, user):
+    """See update_session_auth_hash()."""
+    return await sync_to_async(update_session_auth_hash)(request, user)
--- a/django/test/client.py
+++ b/django/test/client.py
@@ -747,6 +747,9 @@ class ClientMixin:
         self.cookies[settings.SESSION_COOKIE_NAME] = session.session_key
         return session
 
+    async def asession(self):
+        return await sync_to_async(lambda: self.session)()
+
     def login(self, **credentials):
         """
         Set the Factory to appear as if it has successfully logged into a site.
@@ -762,19 +765,35 @@ class ClientMixin:
             return True
         return False
 
+    async def alogin(self, **credentials):
+        """See login()."""
+        from django.contrib.auth import aauthenticate
+
+        user = await aauthenticate(**credentials)
+        if user:
+            await self._alogin(user)
+            return True
+        return False
+
     def force_login(self, user, backend=None):
-        def get_backend():
-            from django.contrib.auth import load_backend
-
-            for backend_path in settings.AUTHENTICATION_BACKENDS:
-                backend = load_backend(backend_path)
-                if hasattr(backend, "get_user"):
-                    return backend_path
-
         if backend is None:
-            backend = get_backend()
+            backend = self._get_backend()
         user.backend = backend
         self._login(user, backend)
+
+    async def aforce_login(self, user, backend=None):
+        if backend is None:
+            backend = self._get_backend()
+        user.backend = backend
+        await self._alogin(user, backend)
+
+    def _get_backend(self):
+        from django.contrib.auth import load_backend
+
+        for backend_path in settings.AUTHENTICATION_BACKENDS:
+            backend = load_backend(backend_path)
+            if hasattr(backend, "get_user"):
+                return backend_path
 
     def _login(self, user, backend=None):
         from django.contrib.auth import login
@@ -789,6 +808,26 @@ class ClientMixin:
         login(request, user, backend)
         # Save the session values.
         request.session.save()
+        self._set_login_cookies(request)
+
+    async def _alogin(self, user, backend=None):
+        from django.contrib.auth import alogin
+
+        # Create a fake request to store login details.
+        request = HttpRequest()
+        session = await self.asession()
+        if session:
+            request.session = session
+        else:
+            engine = import_module(settings.SESSION_ENGINE)
+            request.session = engine.SessionStore()
+
+        await alogin(request, user, backend)
+        # Save the session values.
+        await sync_to_async(request.session.save)()
+        self._set_login_cookies(request)
+
+    def _set_login_cookies(self, request):
         # Set the cookie to represent the session.
         session_cookie = settings.SESSION_COOKIE_NAME
         self.cookies[session_cookie] = request.session.session_key
@@ -813,6 +852,21 @@ class ClientMixin:
             engine = import_module(settings.SESSION_ENGINE)
             request.session = engine.SessionStore()
         logout(request)
+        self.cookies = SimpleCookie()
+
+    async def alogout(self):
+        """See logout()."""
+        from django.contrib.auth import aget_user, alogout
+
+        request = HttpRequest()
+        session = await self.asession()
+        if session:
+            request.session = session
+            request.user = await aget_user(request)
+        else:
+            engine = import_module(settings.SESSION_ENGINE)
+            request.session = engine.SessionStore()
+        await alogout(request)
         self.cookies = SimpleCookie()
 
     def _parse_json(self, response, **extra):
--- a/docs/releases/5.0.txt
+++ b/docs/releases/5.0.txt
@@ -150,6 +150,12 @@ Minor features
 * The default iteration count for the PBKDF2 password hasher is increased from
   600,000 to 720,000.
 
+* The new asynchronous functions are now provided, using an
+  ``a`` prefix: :func:`django.contrib.auth.aauthenticate`,
+  :func:`~.django.contrib.auth.aget_user`,
+  :func:`~.django.contrib.auth.alogin`, :func:`~.django.contrib.auth.alogout`,
+  and :func:`~.django.contrib.auth.aupdate_session_auth_hash`.
+
 * ``AuthenticationMiddleware`` now adds an :meth:`.HttpRequest.auser`
   asynchronous method that returns the currently logged-in user.
 
@@ -366,7 +372,11 @@ Templates
 Tests
 ~~~~~
 
-* ...
+* :class:`~django.test.Client` and :class:`~django.test.AsyncClient` now
+  provide asynchronous methods, using an ``a`` prefix:
+  :meth:`~django.test.Client.asession`, :meth:`~django.test.Client.alogin`,
+  :meth:`~django.test.Client.aforce_login`, and
+  :meth:`~django.test.Client.alogout`.
 
 URLs
 ~~~~
--- a/docs/topics/testing/tools.txt
+++ b/docs/topics/testing/tools.txt
@@ -440,6 +440,9 @@ Use the ``django.test.Client`` class to make requests.
             The ``headers`` parameter was added.
 
     .. method:: Client.login(**credentials)
+    .. method:: Client.alogin(**credentials)
+
+        *Asynchronous version*: ``alogin()``
 
         If your site uses Django's :doc:`authentication system</topics/auth/index>`
         and you deal with logging in users, you can use the test client's
@@ -485,7 +488,14 @@ Use the ``django.test.Client`` class to make requests.
         :meth:`~django.contrib.auth.models.UserManager.create_user` helper
         method to create a new user with a correctly hashed password.
 
+        .. versionchanged:: 5.0
+
+            ``alogin()`` method was added.
+
     .. method:: Client.force_login(user, backend=None)
+    .. method:: Client.aforce_login(user, backend=None)
+
+        *Asynchronous version*: ``aforce_login()``
 
         If your site uses Django's :doc:`authentication
         system</topics/auth/index>`, you can use the ``force_login()`` method
@@ -509,7 +519,14 @@ Use the ``django.test.Client`` class to make requests.
         ``login()`` by :ref:`using a weaker hasher while testing
         <speeding-up-tests-auth-hashers>`.
 
+        .. versionchanged:: 5.0
+
+            ``aforce_login()`` method was added.
+
     .. method:: Client.logout()
+    .. method:: Client.alogout()
+
+        *Asynchronous version*: ``alogout()``
 
         If your site uses Django's :doc:`authentication system</topics/auth/index>`,
         the ``logout()`` method can be used to simulate the effect of a user
@@ -518,6 +535,10 @@ Use the ``django.test.Client`` class to make requests.
         After you call this method, the test client will have all the cookies
         and session data cleared to defaults. Subsequent requests will appear
         to come from an :class:`~django.contrib.auth.models.AnonymousUser`.
+
+        .. versionchanged:: 5.0
+
+            ``alogout()`` method was added.
 
 Testing responses
 -----------------
@@ -702,6 +723,13 @@ access these properties as part of a test condition.
             session = self.client.session
             session["somekey"] = "test"
             session.save()
+
+.. method:: Client.asession()
+
+    .. versionadded:: 5.0
+
+    This is similar to the :attr:`session` attribute but it works in async
+    contexts.
 
 Setting the language
 --------------------
--- a/django/contrib/auth/middleware.py
+++ b/django/contrib/auth/middleware.py
@@ -1,7 +1,5 @@
 from functools import partial
 
-from asgiref.sync import sync_to_async
-
 from django.contrib import auth
 from django.contrib.auth import load_backend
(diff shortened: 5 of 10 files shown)


docs/releases/5.0.txt
```diff
@@ -64,6 +64,10 @@ Minor features
 
 * The default iteration count for the PBKDF2 password hasher is increased from
   600,000 to 720,000.
+* Asynchronous versions of the following functions are now provided, using an
+  ``a`` prefix: :func:`~.contrib.auth.alogin`, :func:`~.contrib.auth.alogout`,
+  :func:`~.contrib.auth.aauthenticate`,
+  :func:`~.contrib.auth.aupdate_session_auth_hash`.
```
carltongibson, line 70: Blank line above. 

The following asynchronous functions... (they're not async versions of themselves)

docs/topics/async.txt
```diff
@@ -108,6 +108,18 @@ synchronous function and call it using :func:`sync_to_async`.
...
+
+Partial async support for some :doc:`contrib packages </ref/contrib/index>`
+are available:
+
+* :doc:`django.contrib.auth </topics/auth/index>` has an async interface
+
+.. versionchanged:: 5.0
+
+    Asynchronous interface for :mod:`django.contrib.auth` was added.
+
```
carltongibson, line 122: Not sure about this addition. 🤔

django/test/client.py
```diff
@@ -815,6 +859,20 @@ def logout(self):
         logout(request)
         self.cookies = SimpleCookie()
 
+    async def alogout(self):
+        """Log out the user by removing the cookies and session object."""
+        from django.contrib.auth import aget_user, alogout
+
+        request = HttpRequest()
+        if self.session:
```
carltongibson, line 867: How come not `session = await self.asession()`?

carltongibson (commented):
Hey @bigfootjon — Looking good. 

A few comments: 

* Can you review tests for the new functions, independently of the `Client` usage? e.g. `aauthenticate` should be covered other than via `ClientMixin.alogin`, and so on. 
* Test client API additions should be documented in `topics/testing.tools.txt`. 
* Can you re-group the methods in ClientMixin, so the `a` versions are next to their counterparts? e.g. I would expect `aforce_login` to be next to `force_login`, and so on. 

🚀

django/test/client.py
bigfootjon: oops!

bigfootjon:
Thanks for the review @carltongibson. Sorry for all the silly mistakes. I feel like I've turned in a rough draft 😓

docs/topics/async.txt
bigfootjon: Cool, I wasn't sure. Removed!

docs/topics/testing/tools.txt
```diff
@@ -703,6 +727,10 @@ access these properties as part of a test condition.
             session["somekey"] = "test"
             session.save()
 
+    .. versionchanged:: 5.0
+
+            ``asession()`` method was added.
```
felixxm, line 732:
```suggestion
        ``asession()`` method was added.
```

docs/releases/5.0.txt
```diff
@@ -64,6 +64,10 @@ Minor features
 
 * The default iteration count for the PBKDF2 password hasher is increased from
   600,000 to 720,000.
+* The following asynchronous functions are now provided, using an
+  ``a`` prefix: :func:`~.contrib.auth.alogin`, :func:`~.contrib.auth.alogout`,
+  :func:`~.contrib.auth.aauthenticate`,
+  :func:`~.contrib.auth.aupdate_session_auth_hash`.
```
felixxm, line 70: New `Client` methods should be listed in the release notes.

django/test/client.py
```diff
@@ -762,20 +765,41 @@ def login(self, **credentials):
...
+        return False
 
+    def force_login(self, user, backend=None):
         if backend is None:
-            backend = get_backend()
+            backend = self._get_backend()
         user.backend = backend
         self._login(user, backend)
 
+    async def aforce_login(self, user, backend=None):
```
felixxm, line 789: This method is untested.

django/test/client.py
```diff
@@ -762,20 +765,41 @@ def login(self, **credentials):
...
+        Return True if login is possible or False if the provided credentials
+        are incorrect.
+        """
+        from django.contrib.auth import aauthenticate
+
+        user = await aauthenticate(**credentials)
+        if user:
+            await self._alogin(user)
+            return True
+        return False
```
felixxm, line 781: This line is untested.

django/test/client.py
```diff
@@ -815,6 +859,21 @@ def logout(self):
         logout(request)
         self.cookies = SimpleCookie()
 
+    async def alogout(self):
```
felixxm, line 862: This method is untested.

felixxm (commented):
@bigfootjon Thanks :+1:

bigfootjon:
Thanks for the review @felixxm! I've fixed the things you flagged.

django/contrib/auth/__init__.py
```diff
@@ -91,6 +93,14 @@ def authenticate(request=None, **credentials):
     )
 
 
+@sensitive_variables("credentials")
+async def aauthenticate(request=None, **credentials):
+    """
+    If the given credentials are valid, return a User object.
+    """
+    return await sync_to_async(authenticate)(request, **credentials)
```
felixxm, line 101: It seems `@sensitive_variables` doesn't work with coroutines. For example, the following test crashes:
```diff
diff --git a/tests/auth_tests/test_auth_backends.py b/tests/auth_tests/test_auth_backends.py
index a535f329d2..81406f37e6 100644
--- a/tests/auth_tests/test_auth_backends.py
+++ b/tests/auth_tests/test_auth_backends.py
@@ -6,6 +6,7 @@ from django.contrib.auth import (
     BACKEND_SESSION_KEY,
     SESSION_KEY,
     _clean_credentials,
+    aauthenticate,
     authenticate,
     get_user,
     signals,
@@ -764,6 +765,28 @@ class AuthenticateTests(TestCase):
             status_code=500,
         )
 
+    @override_settings(
+        AUTHENTICATION_BACKENDS=["auth_tests.test_auth_backends.TypeErrorBackend"]
+    )
+    async def test_aauthenticate_sensitive_variables(self):
+        try:
+            await aauthenticate(
+                username="testusername", password=self.sensitive_password
+            )
+        except TypeError:
+            exc_info = sys.exc_info()
+        rf = RequestFactory()
+        response = technical_500_response(rf.get("/"), *exc_info)
+        self.assertNotContains(response, self.sensitive_password, status_code=500)
+        self.assertContains(response, "TypeErrorBackend", status_code=500)
+        self.assertContains(
+            response,
+            '<tr><td>credentials</td><td class="code">'
+            "<pre>&#39;********************&#39;</pre></td></tr>",
+            html=True,
+            status_code=500,
+        )
+
     def test_clean_credentials_sensitive_variables(self):
         try:
             # Passing in a list to cause an exception
```
We should start by fixing it in a separate PR :thinking:

bigfootjon: Oh wow! Nice catch. I've got a PR up that fixes this: https://github.com/django/django/pull/16668

bigfootjon: Hm but this still doesn't fix the test case you created. I'll see what's up there

bigfootjon: Alright I can't figure this out tonight, I'll take another look when I can

felixxm: Decorator is probably lost when we use `sync_to_async`:

image

bigfootjon: Oh then the problem isn't really the decorator handling coroutines (https://github.com/django/django/pull/16668) but more about how the implementation of `get_traceback_frame_variables` doesn't handle this case correctly? Am I on the right track here?

The reason I say this is I added some logging to `get_traceback_frame_variables` to print out the `current_frame` in the `while current_frame is not None` loop and it never actually hits the frames inside the asyncio executor (i.e. the one from your screenshot) which suggests to me that due to the way `asgiref` schedules the async function it might be missed by the traceback cleaning process. (I'm a little out of my depth here, I may be saying nonsense)

felixxm: As far as I'm aware, `sync_to_async` call should also be wrapped with `sensitive_variables` somehow :thinking: as I found this frame in the traceback:
```
<frame at 0x7f7e4e187840, file '/python3.10/site-packages/asgiref/sync.py', line 495, code thread_handler>

dict_items([
('self', <asgiref.sync.SyncToAsync object at 0x7f7e4e1a77f0>),
('loop', <_UnixSelectorEventLoop running=True closed=False debug=False>),
('source_task', <Task pending name='Task-1' coro=<AsyncToSync.main_wrap() running at /python3.10/site-packages/asgiref/sync.py:306> cb=[_run_until_complete_cb() at /python3.10/asyncio/base_events.py:184]>),
('exc_info', (None, None, None)),
('func', <built-in method run of _contextvars.Context object at 0x7f7e4e2a6a40>),
('args', (functools.partial(<function authenticate at 0x7f7e4f29e4d0>, None, username='testusername', password='mypassword'),)),
('kwargs', {}),
('current_thread', <_MainThread(MainThread, started 140180512552768)>),
('parent_set', False)])
```
and the previous frame is
```
<frame at 0x7fa4e89be3e0, file '/python3.10/site-packages/asgiref/current_thread_executor.py', line 26, code run>
```
instead of
```
<frame at 0x7fa4e899ba00, file '/django/django/views/decorators/debug.py', line 56, code sensitive_variables_wrapper>
```

carltongibson: Grrr. I just merged in #16668 having played with it this morning — the behaviour is OK there with the explicit `async def` declaration... — We may have to adjust that. 🤔

> Oh then the problem isn't really the decorator handling coroutines (https://github.com/django/django/pull/16668) but more about how the implementation of get_traceback_frame_variables doesn't handle this case correctly? Am I on the right track here?

Yes, there's something going on there. Musing aloud... once you hit the suspension at an `await` the stack is _"rebuilt"_ for the task to execute before the continuation puts you back where you were. It's not 100% clear to me that we can dig around in private parts of the traceback and expect to find exactly what we're looking for. (Need to dig more.)

carltongibson: Grrr. For four or five frames going through SyncToAsync the passed parameters are exposed, like... 

```
args | (functools.partial(<function authenticate at 0x106eecd60>, None, username='testusername', password='mypassword'),)
```

and similar. 

Once the other side they are handled again in the lower frames. 

There's no obvious way around this. A exception report filter that kept a track of sensitive ways, and knew to deal with SyncToAsync... — bah! More likely a fully async authentication backend.

bigfootjon: > Grrr. I just merged in https://github.com/django/django/pull/16668 having played with it this morning — the behaviour is OK there with the explicit async def declaration... — We may have to adjust that. thinking

I think it does need adjustment, but not

bigfootjon: Ok I think I have *a* solution. Not a great one, but it does actually work without requiring asyncification of all the `auth` internals at once!

I borrowed this technique from my day job: avoid assigning or logging a sensitive variable by just storing the sensitive variable in the return value of a lambda. (see below for this concept applied here).

This works just fine, but unfortunately the error reporter still isn't tuned correctly:

> Grrr. I just merged in https://github.com/django/django/pull/16668 having played with it this morning — the behaviour is OK there with the explicit async def declaration... — We may have to adjust that. thinking

Yeah unfortunately that looks like it might have been a partial solution, though the missing piece evades me. The full solution won't fix the `asgiref.sync` helpers in any event so I think  https://github.com/django/django/commit/b00046d2c25771bed2242680b08b524a44aa9798 is a good move, but there is a situation that my test case didn't cover that I can't quite pin down.

I took @felixxm's suggested test case and merged it to my current commit, then I added a new commit which just contains the "variable stored in a lambda" technique that I talked above above. I reran the test case Mariusz wrote and that commit reduced the rate of "mypassword" appearing from several (the `asgiref` stack) down to a single hit. This doesn't make sense to me as it should be cleansed by `@sensitive_variables` but it isn't.

When I was digging in to it I zeroed in on how `technical_500.html` tracebacks are actually generated and I noticed that, well first I have no idea how any of this code works, but after some very dumb debug `print`ing I noticed that `aauthenticate` doesn't have an `f_back` in the `tb_frame` object (...I started regretting opening this PR at this point). On a lark I passed in the previous `tb` and massaged it in to working. So I committed that fix to the third commit in the stack. I don't think this is the right solution (and not backwards compatible), but I'm not really sure why this "fix" is even necessary in the first place.

I'll have to take another look with fresh eyes sometime later (likely next week at the earliest), but I wanted to get this back up for another round of discussions, and maybe someone else has an idea on why commit #3 is necessary.

In summary, I have pushed the following commits:
1. The actual changes for this PR (including Mariusz's test case, which should stay no matter what)
2. The "variable stored in a lambda" magic, which I'm ok with but might be a bit too far of a reach
3. A maybe sorta hacky fix to the magical traceback machinery

IMO (2) should merge in to (1), if we agree that this is the right path, and (3) should be... rewritten.

bigfootjon: I've made the tiniest amount of progress. The following patch can be applied against prod and causes the `view_tests.tests.test_debug.ExceptionReporterFilterTests.test_async_sensitive_request` test I case that I wrote in https://github.com/django/django/pull/16668 to fail. I'm not sure how exactly to fix this though, but I'm staring at it very hard. if I can't figure this out we should probably revert that PR since it only protects the named sensitive variables if an exception is thrown *by the function wrapped by `@sensitive_variables`* which is basically no protection at all. 

```diff
diff --git a/tests/view_tests/views.py b/tests/view_tests/views.py
index 97febdaf83..1f01e31504 100644
--- a/tests/view_tests/views.py
+++ b/tests/view_tests/views.py
@@ -180,7 +180,7 @@ def sensitive_view(request):
 
 @sensitive_variables("sauce")
 @sensitive_post_parameters("bacon-key", "sausage-key")
-async def async_sensitive_view(request):
+async def async_nested_view(request):
     # Do not just use plain strings for the variables' values in the code
     # so that the tests don't return false positives when the function's source
     # is displayed in the exception report.
@@ -188,8 +188,13 @@ async def async_sensitive_view(request):
     sauce = "".join(  # NOQA
         ["w", "o", "r", "c", "e", "s", "t", "e", "r", "s", "h", "i", "r", "e"]
     )
+
+    raise Exception
+
+
+async def async_sensitive_view(request):
     try:
-        raise Exception
+        return await async_nested_view(request)
     except Exception:
         exc_info = sys.exc_info()
         send_log(request, exc_info)
```

bigfootjon:
Ok so I've managed to find a solution I'm reasonably happy with. We can use `__traceback_hide__ = True` to just skip the problematic frame:

https://github.com/django/django/blob/85b52d22fd2841c34e95b3a80d6f2b668ce2f160/django/views/debug.py#L523-L527

I replaced the previous implementation in commit 3 with the `__traceback_hide__` hack.

(sidenote: it might be a good idea to add this local variable to the `asgiref` machinery?)

I also added some comments to commits 2 and 3 to make it more clear why we're doing what we're doing.

IMO, revert https://github.com/django/django/pull/16668 as there's no way to handle it correctly right now due to the limitation in the cpython interpreter I flagged above. https://github.com/python/cpython/issues/91048

If reviewers are happy with this solution (enjoy your retirement Carlton! No rush) then I can merge the commits down to one commit, and I think this is otherwise in an acceptable state?

django/contrib/auth/__init__.py
felixxm: > If I can't figure this out we should probably revert that PR since it only protects the named sensitive variables if an exception is caught by the function wrapped by `@sensitive_variables` which is functionally not enough protection to warrant documentation that we DO support this.

Agreed, any error-prone solution in this area can be really damaging.

felixxm: See #16704.

carltongibson:
> ... it might be a good idea to add this local variable to the asgiref machinery?

It might be worth a separate targeted PR there, showing the change, and what's at stake etc. 
(That's going to be easier to review as a smaller independent piece too, I'd guess)

bigfootjon:
> It might be worth a separate targeted PR there, showing the change, and what's at stake etc.
> (That's going to be easier to review as a smaller independent piece too, I'd guess)
> (I'll need to sit down and think this through to say anything concrete...)

Filed https://github.com/django/asgiref/pull/382 with some initial discussion on both sides of the issue.

I also filed https://github.com/django/asgiref/pull/382, noticed this while I was working on this PR and was just waiting until I had a local clone of the `asgiref` repo to see if I could fix it :P

carltongibson:
Great. Thanks @bigfootjon — give me a cycle but, I will have a look 👍

django/contrib/auth/__init__.py
```diff
@@ -91,6 +93,28 @@ def authenticate(request=None, **credentials):
     )
 
 
+async def aauthenticate(request=None, **credentials):
+    """
+    If the given credentials are valid, return a User object.
+    """
```
nessita, line 99: I'm very new to this code but I was thinking whether duplicating the docstrings from the sync functions is desired or not. I did a quick check in the existing code and I see a (soft?) pattern of having the async version reference the sync ones:

```
django/core/cache/backends/base.py:207:    async def aget_many(self, keys, version=None):
django/core/cache/backends/base.py-208-        """See get_many()."""
--
django/core/cache/backends/base.py:235:    async def aget_or_set(self, key, default, timeout=DEFAULT_TIMEOUT, version=None):
django/core/cache/backends/base.py-236-        """See get_or_set()."""
--
django/core/cache/backends/base.py:273:    async def aincr(self, key, delta=1, version=None):
django/core/cache/backends/base.py-274-        """See incr()."""
--
django/core/cache/backends/base.py:341:    async def aclear(self):
django/core/cache/backends/base.py-342-        return await sync_to_async(self.clear, thread_sensitive=True)()
--
django/core/cache/backends/base.py:360:    async def aincr_version(self, key, delta=1, version=None):
django/core/cache/backends/base.py-361-        """See incr_version()."""
```

nessita (commented):
Added minor comment.

django/contrib/auth/__init__.py
bigfootjon: Yeah looks like you're right, I've updated the docstrings.

bigfootjon:
I just put up https://github.com/django/django/pull/16831. If that PR is accepted then we can just delete the second and third commits from this PR (I've verified locally that the tests from the first commit still pass).

felixxm:
@bigfootjon Please rebase after 38e391e95fe5258bc6d2467332dc9cd44ce6ba52.

django/contrib/auth/__init__.py
bigfootjon: Ok this issue has been resolved now due to the changes in https://github.com/django/django/commit/38e391e95fe5258bc6d2467332dc9cd44ce6ba52

bigfootjon:
@felixxm done!

felixxm:
See #16999 and #17000.

bigfootjon:
@felixxm: Compare the CI results from https://github.com/django/django/pull/16636/commits/007b5e4527b480279ab81abd615bbccb61dba1ec and https://github.com/django/django/pull/16636/commits/f557cc2b78be228acf5d4a4632a5fba308b8267c. Both in CI and locally the results of `auth_tests.test_auth_backends.AuthenticateTests.test_aauthenticate_sensitive_variables` differs between those 2 commits. I tried to narrow it down to a simple repro case and couldn't figure out what's going on.

felixxm (commented):
@bigfootjon Thanks :+1:

I pushed final edits, added docs for `aget_user()`, and reorganized tests.

PR #16645, opened 2023-03-12
Corrected example in CreateCollation() docs.

adamchainz:
`und-u-ks-level2` is a general purpose case-insensitive collation, so the name `german_phonebook` is misleading, and the `deterministic` flag should be used. I deconstructed what the parts of the name mean in my post: https://adamj.eu/tech/2023/02/23/migrate-django-postgresql-ci-fields-case-insensitive-collation/#what-to-do

I think there must have been some missed edits when writing the docs in #13469.

CC @knyghty @ngnpope


felixxm (approved):
@adamchainz Good catch :+1:

PR #16651, opened 2023-03-13
prefetching the earliest/latest related object

KommuSoft:
https://code.djangoproject.com/ticket/34413

The concept behind this is that we make a reverse filter that we use in a subquery to determine the *primary key* of the earliest/latest object, and then slightly alter the prefetch logic to fetch these objects, and do the JOINing at the Django layer.


charettes:
Thank you for your patch but an earlier conversation on the [Django developer list](https://groups.google.com/g/django-developers) or [the internals forum](https://forum.djangoproject.com/c/internals/5) might have pointed you at ticket-26780 / #15957 which fixed the _prefetch this slice of related objects_ problem in a single query in Django 4.2.

Closing per ticket.

PR #16679, opened 2023-03-23
Add space

rajsahu24:
I was going through django documents and I found a minor error regarding a space on this address. [https://docs.djangoproject.com/en/4.1/ref/contrib/messages/#displaying-messages ]

Check it out!


ngnpope:
Thanks, but these changes are incorrect. It's fine as it is.

PR #16703, opened 2023-03-30
Fixed #28900 -- Made SELECT respect the order specified by values()/values_list().

charettes:
ticket-28900

---

The TL;DR here is that previously when using `values` or `values_list` the resulting order of the `SELECT` members was not matching the one explicitly specified.

Instead the order was always `SELECT *extra_fields, *model_fields, *annotations` where each local group respected the explicitly specified order. This was leading to issues particularly when combining queries and trying to combine annotations with model fields as exemplified here

- https://code.djangoproject.com/ticket/28553#comment:3
- https://code.djangoproject.com/ticket/34502#comment:2

What this PR does is
1. Store a global order in `Query.selected`
2. Remove the unnecessary re-mapping of order in `ValuesListIterable`
3. Address ticket-28900 by having the combinator select inheritance logic rely on this new property

---

The only part that I think might be missing is a mention in the release notes. My thoughts are that it's possible that users tried to work around this bug in all sort of weird ways (mixing extras) or might not even realize that they have a buggy query because the swapped columns happen to share compatible types. They would benefit from being explained why this was solved and what to expect. Thoughts?

### DIFF
--- a/django/db/models/query.py
+++ b/django/db/models/query.py
@@ -200,12 +200,15 @@ class ValuesIterable(BaseIterable):
         query = queryset.query
         compiler = query.get_compiler(queryset.db)
 
-        # extra(select=...) cols are always at the start of the row.
-        names = [
-            *query.extra_select,
-            *query.values_select,
-            *query.annotation_select,
-        ]
+        if query.selected:
+            names = list(query.selected)
+        else:
+            # extra(select=...) cols are always at the start of the row.
+            names = [
+                *query.extra_select,
+                *query.values_select,
+                *query.annotation_select,
+            ]
         indexes = range(len(names))
         for row in compiler.results_iter(
             chunked_fetch=self.chunked_fetch, chunk_size=self.chunk_size
@@ -223,28 +226,6 @@ class ValuesListIterable(BaseIterable):
         queryset = self.queryset
         query = queryset.query
         compiler = query.get_compiler(queryset.db)
-
-        if queryset._fields:
-            # extra(select=...) cols are always at the start of the row.
-            names = [
-                *query.extra_select,
-                *query.values_select,
-                *query.annotation_select,
-            ]
-            fields = [
-                *queryset._fields,
-                *(f for f in query.annotation_select if f not in queryset._fields),
-            ]
-            if fields != names:
-                # Reorder according to fields.
-                index_map = {name: idx for idx, name in enumerate(names)}
-                rowfactory = operator.itemgetter(*[index_map[f] for f in fields])
-                return map(
-                    rowfactory,
-                    compiler.results_iter(
-                        chunked_fetch=self.chunked_fetch, chunk_size=self.chunk_size
-                    ),
-                )
         return compiler.results_iter(
             tuple_expected=True,
             chunked_fetch=self.chunked_fetch,
--- a/django/db/models/sql/compiler.py
+++ b/django/db/models/sql/compiler.py
@@ -247,11 +247,6 @@ class SQLCompiler:
         select = []
         klass_info = None
         annotations = {}
-        select_idx = 0
-        for alias, (sql, params) in self.query.extra_select.items():
-            annotations[alias] = select_idx
-            select.append((RawSQL(sql, params), alias))
-            select_idx += 1
         assert not (self.query.select and self.query.default_cols)
         select_mask = self.query.get_select_mask()
         if self.query.default_cols:
@@ -261,19 +256,39 @@ class SQLCompiler:
             # any model.
             cols = self.query.select
         if cols:
-            select_list = []
-            for col in cols:
-                select_list.append(select_idx)
-                select.append((col, None))
-                select_idx += 1
             klass_info = {
                 "model": self.query.model,
-                "select_fields": select_list,
+                "select_fields": list(
+                    range(
+                        len(self.query.extra_select),
+                        len(self.query.extra_select) + len(cols),
+                    )
+                ),
             }
-        for alias, annotation in self.query.annotation_select.items():
-            annotations[alias] = select_idx
-            select.append((annotation, alias))
-            select_idx += 1
+        selected = []
+        if self.query.selected is None:
+            selected = [
+                *(
+                    (alias, RawSQL(*args))
+                    for alias, args in self.query.extra_select.items()
+                ),
+                *((None, col) for col in cols),
+                *self.query.annotation_select.items(),
+            ]
+        else:
+            for alias, expression in self.query.selected.items():
+                # Reference to an annotation.
+                if isinstance(expression, str):
+                    expression = self.query.annotations[expression]
+                # Reference to a column.
+                elif isinstance(expression, int):
+                    expression = cols[expression]
+                selected.append((alias, expression))
+
+        for select_idx, (alias, expression) in enumerate(selected):
+            if alias:
+                annotations[alias] = select_idx
+            select.append((expression, alias))
 
         if self.query.select_related:
             related_klass_infos = self.get_related_selections(select, select_mask)
--- a/django/db/models/sql/compiler.py
+++ b/django/db/models/sql/compiler.py
@@ -591,20 +591,15 @@ class SQLCompiler:
                 # generate valid SQL.
                 compiler.elide_empty = False
         parts = ()
+        selected = self.query.selected
         for compiler in compilers:
             try:
                 # If the columns list is limited, then all combined queries
                 # must have the same columns list. Set the selects defined on
                 # the query on all combined queries, if not already set.
-                if not compiler.query.values_select and self.query.values_select:
+                if selected is not None and compiler.query.selected is None:
                     compiler.query = compiler.query.clone()
-                    compiler.query.set_values(
-                        (
-                            *self.query.extra_select,
-                            *self.query.values_select,
-                            *self.query.annotation_select,
-                        )
-                    )
+                    compiler.query.set_values(selected)
                 part_sql, part_args = compiler.as_sql(with_col_aliases=True)
                 if compiler.query.combinator:
                     # Wrap in a subquery if wrapping in parentheses isn't
--- a/django/db/models/sql/query.py
+++ b/django/db/models/sql/query.py
@@ -26,6 +26,7 @@ from django.db.models.expressions import (
     Exists,
     F,
     OuterRef,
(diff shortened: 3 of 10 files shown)


django/db/models/query.py
```diff
@@ -220,28 +223,6 @@ def __iter__(self):
...
-            if fields != names:
-                # Reorder according to fields.
-                index_map = {name: idx for idx, name in enumerate(names)}
-                rowfactory = operator.itemgetter(*[index_map[f] for f in fields])
-                return map(
-                    rowfactory,
-                    compiler.results_iter(
-                        chunked_fetch=self.chunked_fetch, chunk_size=self.chunk_size
-                    ),
-                )
```
charettes, line 244: `queryset._fields` is only truthy when fields are provided to `values_list` but when it's the case they are now properly selected in the right order so there is no need for re-positioning shenanigans.

django/db/models/sql/query.py
```diff
@@ -471,8 +473,7 @@ def get_aggregation(self, using, aggregate_exprs):
                         col_alias = f"__col{index}"
                         col_ref = Ref(col_alias, col)
                         col_refs[col] = col_ref
-                        inner_query.annotations[col_alias] = col
-                        inner_query.append_annotation_mask([col_alias])
+                        inner_query.add_annotation(col, col_alias)
```
charettes, line 476: This should be moved to a separate commit.

tests/queries/test_qs_combinators.py
```diff
@@ -256,7 +256,7 @@ def test_union_with_extra_and_values_list(self):
             )
             .values_list("num", "count")
         )
-        qs2 = Number.objects.filter(num=2).extra(select={"count": 1})
+        qs2 = Number.objects.filter(num=2).extra(select={"count": 1}).values_list("num", "count")
```
charettes, line 259: This test makes no sense and highlights the awkwardness of the extra API.

tests/postgres_tests/test_array.py
```diff
@@ -466,8 +466,8 @@ def test_group_by_order_by_select_index(self):
                 ],
             )
         sql = ctx[0]["sql"]
-        self.assertIn("GROUP BY 2", sql)
-        self.assertIn("ORDER BY 2", sql)
+        self.assertIn("GROUP BY 1", sql)
+        self.assertIn("ORDER BY 1", sql)
```
charettes, line 470: Expected as the usage of `annotate` after `values` now _append_ instead of _insert_ at the beginning.

tests/queries/test_qs_combinators.py
```diff
@@ -257,6 +257,23 @@ def test_union_with_two_annotated_values_list(self):
...
+            .values_list("num", "zero")
+        )
+        qs2 = (
+            Number.objects.filter(num=2)
+            .annotate(
+                zero=Value(0, IntegerField()),
+            )
+            .values_list("zero", "num")
+        )
+        self.assertCountEqual(qs1.union(qs2), [(1, 0), (0, 2)])
```
charettes, line 275: This use to fail because before it was doing

```sql
SELECT num, 0 AS zero
FROM number
UNION
SELECT num, 0 AS zero
FROM number
```

Completely disregarding the explicit order specified to `values_list`

tests/queries/tests.py
```diff
@@ -2200,7 +2210,7 @@ def test_col_alias_quoted(self):
                 {"tag_per_parent__max": 2},
             )
         sql = captured_queries[0]["sql"]
-        self.assertIn("AS %s" % connection.ops.quote_name("col1"), sql)
+        self.assertIn("AS %s" % connection.ops.quote_name("parent"), sql)
```
charettes, line 2213: The proper field alias is now used.

django/db/models/sql/compiler.py
```diff
@@ -261,19 +256,40 @@ def get_select(self, with_col_aliases=False):
...
+                *((None, col) for col in cols),
+                *self.query.annotation_select.items(),
+            ]
+        else:
+            selected = []
+            for alias, expression in self.query.selected.items():
+                # Reference to an annotation
+                if isinstance(expression, str):
+                    expression = self.query.annotations[expression]
+                # Reference to a column
```
felixxm, line 284:
```suggestion
                # Reference to an annotation.
                if isinstance(expression, str):
                    expression = self.query.annotations[expression]
                # Reference to a column.
```

felixxm (approved):
@charettes Thanks :+1: Fantastic!

django/db/models/sql/query.py
```diff
@@ -2415,13 +2417,28 @@ def set_annotation_mask(self, names):
         """Set the mask of annotations that will be returned by the SELECT."""
         if names is None:
             self.annotation_select_mask = None
+            if self.selected:
+                # Augment selected with all annotations as the mask was removed.
+                for name in self.annotations:
+                    self.selected[name] = name
```
sarahboyce, line 2423: I can also believe I can revert this without a test failure

django/db/models/sql/query.py
```diff
@@ -2415,13 +2417,28 @@ def set_annotation_mask(self, names):
...
+            self.annotation_select_mask = set(names)
+            if self.selected:
+                # Prune the masked annotations
+                self.selected = {
+                    key: value
+                    for key, value in self.selected.items()
+                    if not isinstance(value, str)
+                    or value in self.annotation_select_mask
+                }
+                # Append the unmasked annotations
```
sarahboyce, line 2434:
```suggestion
                # Prune the masked annotations.
                self.selected = {
                    key: value
                    for key, value in self.selected.items()
                    if not isinstance(value, str)
                    or value in self.annotation_select_mask
                }
                # Append the unmasked annotations.
```

sarahboyce (commented):
> The only part that I think might be missing is a mention in the release notes. My thoughts are that it's possible that users tried to work around this bug in all sort of weird ways (mixing extras) or might not even realize that they have a buggy query because the swapped columns happen to share compatible types. They would benefit from being explained why this was solved and what to expect. Thoughts?

I think it's a good idea
Maybe putting something in the [`values()` docs](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#django.db.models.query.QuerySet.values) explicitly mentioning the select ordering (with a versionchanged note explaining the change a little) might be a good start?

django/db/models/query.py
```diff
@@ -200,12 +200,15 @@ def __iter__(self):
...
         compiler = query.get_compiler(queryset.db)
 
-        # extra(select=...) cols are always at the start of the row.
-        names = [
-            *query.extra_select,
-            *query.values_select,
-            *query.annotation_select,
-        ]
+        if query.selected:
+            names = list(query.selected)
```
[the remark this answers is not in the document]
charettes: This one is interesting, we don't have tests for it but it ensures the insertion order in the `dict` returned by using `values` is preserved.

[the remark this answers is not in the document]
charettes: `annotations.tests.NonAggregateAnnotationTestCase.test_annotation_subquery_and_aggregate_values_chaining` was failing with this code commented out but I'll add a more explicit test for it.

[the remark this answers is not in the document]
charettes: See `test_values_fields_annotations_order`

django/db/models/sql/query.py
charettes: Interesting I'll try to see what this might be needed for by running the full suite.

charettes: There is effectively is no way to trigger it anymore (this PR is a few months old at this point 😅).

I checked all the `set_annotation_mask` call and the only one that can still pass `names = None` is the one in `get_aggregation` assuming `self.annotation_select_mask = None` prior

https://github.com/django/django/blob/e56a32b89bb7fadffdfaa2cdf12b4863ccd5af9b/django/db/models/sql/query.py#L456-L482

I'm happy to remove the code if you'd like as it's not activated by any code path but I wonder if a good preventive measure here might be drop support for passing `None` in the first place (and adjust `get_aggregation` accordingly) as if a future use case come up we'll definitely run into issues with `Query.selected` being out of sync.

The alternative would be write low level tests for `Query.set_annotation_mask` directly to cover for a potential future use case. Whichever you prefer works for me, just lmk @sarahboyce.

tests/annotations/tests.py
```diff
@@ -568,6 +568,16 @@ def test_values_annotation(self):
...
         self.assertEqual(book["other_isbn"], "155860191")
 
+    def test_values_fields_annotations_order(self):
+        qs = Book.objects.annotate(other_rating=F("rating") - 1).values(
+            "other_rating", "rating"
+        )
+        book = qs.get(pk=self.b1.pk)
+        self.assertEqual(
+            list(book.items()),
+            [("other_rating", self.b1.rating - 1), ("rating", self.b1.rating)],
```
charettes, line 578: `items()` is used to check the insertion order in the `dict` as well.

django/db/models/query.py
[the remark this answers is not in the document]
sarahboyce: Sorry and thank you 👍

sarahboyce (commented):
Thank you for this @charettes 🥳 looks really good

django/db/models/sql/query.py
charettes: Alright I figured I'd just drop the new logic after all as the gymnastic required in `get_aggregation` where getting out of hand.

docs/ref/models/querysets.txt
```diff
@@ -745,6 +745,11 @@ You can also refer to fields on related models with reverse relations through
     ``"true"``, ``"false"``, and ``"null"`` strings for
     :class:`~django.db.models.JSONField` key transforms.
 
+.. versionchanged:: 5.2
+
+    The ``SELECT`` clause generated when using ``values()`` was updated to
+    respect the order of the specified ``*fields`` and ``**expressions``.
+
```
charettes, line 752: @sarahboyce I applied your suggestion and I'll defer to you to determine if the release notes are enough.

IMO it might be as the previous behavior was borderline a bug and undocumented and based on the number of tickets I triaged where I had to _teach_ reporters about it I doubt many users even knew it. Still it might be the extra admonition to at least point out it behaves correctly.

sarahboyce (approved):
Thank you so much for this @charettes, looks fantastic ⭐

PR #16727, opened 2023-04-05
Fixed #34455 -- language_from_path returning None in LocaleMiddleware

sinsniwal:
Fixed: https://code.djangoproject.com/ticket/34455

When using LocaleMiddleware, the value of language_from_path was coming out to be None initially, and because of that, the redirect was not working.


felixxm:
Please create patch against the `main` branch.

sinsniwal:
Thanks Mariusz, on it, I am new to open source so it's taking a little bit
of time.

sinsniwal:
Can you please check this, should i send the pull request?

https://github.com/sinsniwal/django/tree/ticket_34455

felixxm:
Looks good as a draft patch (release notes `4.2.1.txt` and a regression test is missing). However, I don't believe it is the correct solution, see [comment](https://code.djangoproject.com/ticket/34455#comment:10).

felixxm:
Also, `get_language_from_request()` is intended to return a language from requests, not from settings.

sinsniwal:
But the thing is for the first time if someone enters a URL like admin,
there is no language to get from there, than in those cases we can redirect
too, that's what i got , like once i open the admin using en/admin, after
that i was able to access admin directly without using en, by implementing
this code instead of None, everything seems to work.
Even i will try to look more deeper.

sinsniwal:
Yes, Sir, You are correct, This issue is not even persistent, and now I am
unable to simulate the issue, the issue can be on edge cases, I was able to
simulate like an hour ago and now its working fine without any changes

PR #16730, opened 2023-04-05
Add Ruff Configuration

obi-jerome:
I use Ruff as my primary Linter for all my Python Projects.
https://github.com/charliermarsh/ruff

It's very fast and powerful, so I thought I could try it on Django Code.
With all options activated, Django is scanned in less than 1 second.
Here's the Configuration with all detected errors types ignored.

What do you think ?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,3 +5,213 @@ build-backend = 'setuptools.build_meta'
 [tool.black]
 target-version = ['py38']
 force-exclude = 'tests/test_runner_apps/tagged/tests_syntax_error.py'
+
+[tool.ruff]
+target-version = "py38"
+line-length = 98
+select = ["ALL"]
+ignore = [
+    "A001",    # Variable is shadowing a python builtin
+    "A002",    # Argument `format` is shadowing a python builtin
+    "A003",    # Class attribute is shadowing a python builtin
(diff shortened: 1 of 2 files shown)


obi-jerome:
Oh, shame. 
The goal was to have only one, that is way faster and detect way more anomalies / possible improvements.
Ruff is also able to fix many errors by himself too.
Please, feel free to check the list of what he detected in Django Code in just a fraction of the time the others take.
https://github.com/django/django/pull/16730/files
Hope you'll change your mind one day.
Best

aqeelat:
@felixxm I think it is worth it to revisit this PR. I disagree with the way the configurations were added to pyproject (I wouldn't select all rules then exclude them all) but I think we should replace flake8 with ruff.

Tronic:
Now with Ruff Format also released and other projects from Flask to Pydantic and FastAPI, even developers of previous linters having switched to it. All other linters are replaced, and Ruff also works better than Black & al, not only faster. Very good integration to VS Code as well, so my code actually gets formatted on Ctrl-S and I get immediate feedback on bugs as I am typing it (not with a delay like Pylance or others).

So, perhaps Django should also re-evaluate this?

https://docs.astral.sh/ruff/formatter/

PR #16790, opened 2023-04-21
Fixed #34462 Cleanup/Optimization Deletions in admin panel create N + 1 queries

Akash-Kumar-Sen:
- My first pull request in django, some changes might be required, please let me know after review.
- Ticket link : https://code.djangoproject.com/ticket/34462

### DIFF
--- a/django/contrib/admin/actions.py
+++ b/django/contrib/admin/actions.py
@@ -37,11 +37,10 @@ def delete_selected(modeladmin, request, queryset):
     if request.POST.get('post') and not protected:
         if perms_needed:
             raise PermissionDenied
-        n = queryset.count()
+        # Call log_deletions here
+        n = len(queryset)
         if n:
-            for obj in queryset:
-                obj_display = str(obj)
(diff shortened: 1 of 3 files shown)


nessita:
Hello @Akash-Kumar-Sen, thank you for your contribution. Please note that you have created the pull request against a branch that no longer accepts feature development, so we need to _close_ this PR and ideally you should propose a **new one** against the `main` development branch.

Besides the above, please note that every change that is proposed should include the corresponding set of tests. Particularly for this set of changes, I think you could add some new tests next to https://github.com/django/django/blob/main/tests/modeladmin/tests.py#L815 to ensure the new methods work as expected.

Then, ideally you should also add an entry to the release notes as mentioned in the ticket for the new methods, and perhaps consider a deprecation path for the existing per-item methods.

Akash-Kumar-Sen:
Thank you for the suggestion @nessita.  Will create a new PR including the test and the release notes and the depreciation path.

PR #16794, opened 2023-04-22
Shortcuts for add, change and delete permissions

ipakeev:
These shortcuts can be useful for quickly restricting user permissions.

### DIFF
--- a/django/contrib/admin/options.py
+++ b/django/contrib/admin/options.py
@@ -142,6 +142,10 @@ class BaseModelAdmin(metaclass=forms.MediaDefiningClass):
     show_full_result_count = True
     checks_class = BaseModelAdminChecks
 
+    can_add = True
+    can_change = True
+    can_delete = True
+
     def check(self, **kwargs):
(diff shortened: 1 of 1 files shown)


shangxiao:
Hi @ipakeev 

Thanks, but I'm not entirely sure what you're proposing here. We don't accept pull requests without a ticket; and all feature requests must first be discussed with Django developers on the Django forum: https://code.djangoproject.com/wiki/DevelopersMailingList

ipakeev:
1. Now ModelAdmin has a function
`def has_add_permission(self, request):`
but InlineModelAdmin has a slightly different function
`def has_add_permission(self, request, obj):`.
This difference makes it difficult to create common mixins.

2. Instead of writing in the ModelAdmin class:
`def has_add_permission(self, request): 
    return False`
we can define `can_add = False`

shangxiao:
Make a post on the Django forum or mailing list 😉 Only a select few people will read comments on this PR

PR #16795, opened 2023-04-23
corrected min and max value validators

mellertson:
The messages for MaxValueValidator reads, "Ensure this value is less than or equal to %(limit_value)s.", so the comparison should used `<=`, but it uses `<`.  And likewise for MinValueValidator, but it uses `>` instead of `>=`.  

Both of the bugs detailed above are fixed in this pull request.

Additionally, I looked at the unit tests for both validators, and they look correct.  So, not sure if the corresponding unit tests might have a failure in one or both, but just wanted to make a note of it here.

### DIFF
--- a/django/core/validators.py
+++ b/django/core/validators.py
@@ -380,7 +380,7 @@ class MaxValueValidator(BaseValidator):
     code = "max_value"
 
     def compare(self, a, b):
-        return a > b
+        return a >= b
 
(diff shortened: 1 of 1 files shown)


shangxiao:
Hi

The max/min validators are correct, please see https://code.djangoproject.com/ticket/34503

PR #16797, opened 2023-04-25
Fixed #34515 -- Updated to redirect to request language when from translated default url.

sarahboyce:
Ticket: https://code.djangoproject.com/ticket/34515

I'm not confident I found the issue, but this _might_ be what is happening.
- User has a cookie with a language code (not the default)
- We want to go to the path for the default language (no language prefix here)
- It doesn't let you because of the cookie

### DIFF
--- a/tests/i18n/urls_default_unprefixed.py
+++ b/tests/i18n/urls_default_unprefixed.py
@@ -7,5 +7,7 @@ urlpatterns = i18n_patterns(
     re_path(r"^(?P<arg>[\w-]+)-page", lambda request, **arg: HttpResponse(_("Yes"))),
     path("simple/", lambda r: HttpResponse(_("Yes"))),
     re_path(r"^(.+)/(.+)/$", lambda *args: HttpResponse()),
+    re_path(_(r"^users/$"), lambda *args: HttpResponse(), name="users"),
+    re_path(_(r"^account/"), lambda *args: HttpResponse()),
     prefix_default_language=False,
 )
--- a/django/middleware/locale.py
+++ b/django/middleware/locale.py
@@ -1,3 +1,5 @@
+from urllib.parse import quote
+
 from django.conf import settings
 from django.conf.urls.i18n import is_language_prefix_patterns_used
 from django.http import HttpResponseRedirect
@@ -5,6 +7,7 @@ from django.urls import get_script_prefix, is_valid_path
 from django.utils import translation
 from django.utils.cache import patch_vary_headers
 from django.utils.deprecation import MiddlewareMixin
+from django.utils.http import RFC3986_SUBDELIMS, escape_leading_slashes
 
 
 class LocaleMiddleware(MiddlewareMixin):
@@ -68,6 +71,8 @@ class LocaleMiddleware(MiddlewareMixin):
                 and not language_path.endswith("/")
                 and is_valid_path("%s/" % language_path, urlconf)
             )
+            with translation.override(settings.LANGUAGE_CODE):
+                valid_path_in_default = is_valid_path(request.path_info, urlconf)
 
             if path_valid or path_needs_slash:
                 script_prefix = get_script_prefix()
@@ -80,6 +85,18 @@ class LocaleMiddleware(MiddlewareMixin):
                 # get_language_from_request(). HTTP caches may cache this
                 # redirect, so add the Vary header.
                 redirect = self.response_redirect_class(language_url)
+                patch_vary_headers(redirect, ("Accept-Language", "Cookie"))
+                return redirect
+            elif (
+                not prefixed_default_language
+                and language_from_request != settings.LANGUAGE_CODE
+                and valid_path_in_default
+            ):
+                result, text_subs = valid_path_in_default.get_pattern_and_text_subs()
+                script_prefix = get_script_prefix()
+                path = f"{script_prefix}{language}/{result}"
+                url = quote(path % text_subs, safe=RFC3986_SUBDELIMS + "/~:@")
+                redirect = self.response_redirect_class(escape_leading_slashes(url))
                 patch_vary_headers(redirect, ("Accept-Language", "Cookie"))
                 return redirect
 
--- a/django/urls/resolvers.py
+++ b/django/urls/resolvers.py
@@ -101,6 +101,22 @@ class ResolverMatch:
     def __reduce_ex__(self, protocol):
         raise PicklingError(f"Cannot pickle {self.__class__.__qualname__}.")
 
+    def get_pattern_and_text_subs(self):
+        url_pattern = self.tried[-1][1]
+        result, params = normalize(url_pattern.pattern.regex.pattern)[0]
+        if self.args:
+            subs = dict(zip(params, self.args))
+        else:
+            subs = self.kwargs
+        # Convert the subs to text using Converter.to_url().
+        text_subs = {}
+        for k, v in subs.items():
+            if k in url_pattern.pattern.converters:
+                text_subs[k] = url_pattern.pattern.converters[k].to_url(v)
+            else:
+                text_subs[k] = str(v)
+        return result, text_subs
+
 
 def get_resolver(urlconf=None):
     if urlconf is None:
--- a/docs/releases/4.2.1.txt
+++ b/docs/releases/4.2.1.txt
@@ -30,6 +30,11 @@ Bugfixes
   ``prefix_default_language`` argument when a fallback language of the default
   language was used (:ticket:`34455`).
 
+* Fixed a regression in Django 4.2 where navigating to the translated url of
+  the default language was not possible when ``prefix_default_language`` was
+  False and a different language had been set in the cookie or the headers
+  (:ticket:`34515`).
+
 * Fixed a regression in Django 4.2 where creating copies and deep copies of
   ``HttpRequest``, ``HttpResponse``, and their subclasses didn't always work
   correctly (:ticket:`34482`, :ticket:`34484`).
--- a/tests/i18n/patterns/tests.py
+++ b/tests/i18n/patterns/tests.py
@@ -431,6 +431,35 @@ class URLResponseTests(URLTestCaseBase):
         self.assertEqual(response.context["LANGUAGE_CODE"], "nl")
 
 
+@override_settings(ROOT_URLCONF="i18n.urls_default_unprefixed", LANGUAGE_CODE="nl")
+class URLPrefixedFalseTranslatedTests(URLTestCaseBase):
+    def test_en_redirect_from_cookie_with_name(self):
+        self.client.cookies.load({settings.LANGUAGE_COOKIE_NAME: "en"})
+        response = self.client.get("/gebruikers/", follow=True)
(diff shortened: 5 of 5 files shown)


tests/i18n/urls_default_unprefixed.py
```diff
@@ -7,5 +7,6 @@
     re_path(r"^(?P<arg>[\w-]+)-page", lambda request, **arg: HttpResponse(_("Yes"))),
     path("simple/", lambda r: HttpResponse(_("Yes"))),
     re_path(r"^(.+)/(.+)/$", lambda *args: HttpResponse()),
+    re_path(_(r"^users/$"), lambda *args: HttpResponse(), name="users"),
```
shangxiao, line 10: path() seems more appropriate here ☺️

sarahboyce: I would need to add the translations of `users/` as here `^users/$` is being translated :cry:  
We don't have/I couldn't find any existing suitable translations of non-regex paths for the i18n tests to update this to something else.
Should I do this update and add translations?

aboutofpluto:
> @aboutofpluto Does it work for you?

Hello!

Nope, it does not seem to work. Because I do not have any language cookie installed. My browser send a HTTP header.

I am not sure this patch is going in the right direction. IMO, if `prefix_default_language` is `False` in `i18n_patterns`, an unprefixed url should use `LANGUAGE_CODE`, whatever the cookie or header is. I mean, getting the language should not fail in this specific case.

aboutofpluto:
A suggestion:

```python
    def process_request(self, request):
        urlconf = getattr(request, "urlconf", settings.ROOT_URLCONF)
        i18n_patterns_used, prefixed_default_language = is_language_prefix_patterns_used(urlconf)

        if i18n_patterns_used and prefixed_default_language is False:
            language = settings.LANGUAGE_CODE
            
        if not language:
            language = translation.get_language_from_request(
                request, check_path=i18n_patterns_used
            )
            
        if not language:
            language = self.get_fallback_language(request)

        translation.activate(language)
        request.LANGUAGE_CODE = translation.get_language()
```

tests/i18n/tests.py
```diff
@@ -1927,15 +1927,15 @@ def test_other_lang_with_prefix(self):
...
-        self.assertRedirects(response, "/fr/simple/")
+        self.assertEqual(response.status_code, 200)
 
     def test_unprefixed_language_with_cookie_language(self):
-        """A language set in the cookies is respected."""
+        """A language set in the cookies is ignored."""
         self.client.cookies.load({settings.LANGUAGE_COOKIE_NAME: "fr"})
         response = self.client.get("/simple/")
-        self.assertRedirects(response, "/fr/simple/")
+        self.assertEqual(response.status_code, 200)
```
sarahboyce, line 1938: @felixxm @shangxiao I needed to update these tests to comply with the new suggestion.
What is the expected behaviour in this case, am I right to update this to respect the path or should it redirect?

I think @aboutofpluto is suggesting it should respect the path in this case.

(On the new suggestion it ideally would need a depreciation note so it's not the most ideal "solution", but I'm still using this pr to think out loud)

sarahboyce: Right looking at the commit which added the behaviour, the redirect was kinda the whole point :sweat_smile: 

Which would mean that I think [the original attempt](https://github.com/django/django/commit/566fbe353ce9fab374373e9cf7b1404d2d57f39d) "fixed" the bug with translations and the cookie (and I could probably add a tweak for the header case).
So I am going to continue with that, otherwise the discussion is more whether the commit should be reverted.

sarahboyce:
@aboutofpluto I know this is not the behaviour you are advocating for but can you test again to confirm you are no longer experiencing 404s? :pray:

nessita:
Hello everyone, below an initial brain dump from me, it's a little long because I'm sharing my chain of thoughts but there is a conclusion at the end :-).

After some careful reading of the internationalization docs, and looking thru the related PRs, my personal (and initial) expectation for the reported case was to obtain a 200 OK response with the page in french: not because the default language is `fr`, but because the user explicitly requested a path that is the french translation of the original path (this would effectively mean ignoring the request information about lang preference).

Concretely, when using the [GH repo](https://github.com/aboutofpluto/django-prefix-default-language-bug-report/) that the reporter provided in the ticket, the resulting behavior from this branch is a redirect effectively fixing the original 404:

```
$ curl -I -H 'Accept-Language: en' http://127.0.0.1:8000/a-propos/
HTTP/1.1 302 Found
Date: Wed, 26 Apr 2023 20:49:49 GMT
Server: WSGIServer/0.2 CPython/3.11.3
Content-Type: text/html; charset=utf-8
Location: /en/about/
X-Content-Type-Options: nosniff
Referrer-Policy: same-origin
Cross-Origin-Opener-Policy: same-origin
```

But as mentioned I was expecting to get a 200 OK with:

```html
à propos | <a href='/en/about/'>En</a> | <a href='/a-propos/'>Fr</a> | <a href='/es/a-proposito/'>Es</a>
```

But one a second look, my expectation does not match what the docs says: since there is a request header indicating a lang preference, I think Django should do everything in their power to try to return a response in that lang. Thus, I think that the proposal from @sarahboyce is behaviorally correct. The diff itself looks decent though, to me (a first time reader of the `LocaleMiddleware` code), the logic in `process_request` and `process_response` feels extremely complicated and perhaps we could aim at simplifying that a bit in subsequent branches (do you think that's possible).

nessita:
> @aboutofpluto I know this is not the behaviour you are advocating for but can you test again to confirm you are no longer experiencing 404s? pray

I can confirm that your changes fix the 404 and the user gets a 302 instead to the `en` translation. As mentioned in my previous comment, after a lot of thought, I think this is correct.

nessita:
> > @aboutofpluto Does it work for you?
> 
> Hello!
> 
> Nope, it does not seem to work. Because I do not have any language cookie installed. My browser send a HTTP header.
> 
> I am not sure this patch is going in the right direction. IMO, if `prefix_default_language` is `False` in `i18n_patterns`, an unprefixed url should use `LANGUAGE_CODE`, whatever the cookie or header is. I mean, getting the language should not fail in this specific case.

Hey @aboutofpluto, I don't agree that the cookie or header should be ignored. The documentation clearly states that the middleware would do everything it can to determine (and honor, I would add) the user’s language preference. So IMHO that means that the request lang information should take precedence above all other settings.

aboutofpluto:
> > > @aboutofpluto Does it work for you?
> 
> Hey @aboutofpluto, I don't agree that the cookie or header should be ignored. The documentation clearly states that the middleware would do everything it can to determine (and honor, I would add) the user’s language preference. So IMHO that means that the request lang information should take precedence above all other settings.

Hello @nessita and thank you for your contribution.

Sorry, this is going to be a long message again.

As I already said, I do not agree in a specific case: `i18n_patterns`. The [doc says](https://docs.djangoproject.com/en/4.1/topics/i18n/translation/#django.conf.urls.i18n.i18n_patterns) (I highlight some parts):

_This function can be used in a root URLconf and Django will automatically **prepend the current active language code to all URL patterns defined within [i18n_patterns()](https://docs.djangoproject.com/en/4.1/topics/i18n/translation/#django.conf.urls.i18n.i18n_patterns)**._

_**Setting prefix_default_language to False removes the prefix from the default language** ([LANGUAGE_CODE](https://docs.djangoproject.com/en/4.1/ref/settings/#std-setting-LANGUAGE_CODE)). This can be useful when adding translations to existing site so that the current URLs won’t change._

In addition, the algorithm to decide what language to use is the following ([​django 4.2 documentation](https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#how-django-discovers-language-preference))

_LocaleMiddleware tries to determine the user’s language preference by following this algorithm:_
- _**First, it looks for the language prefix in the requested URL. This is only performed when you are using the i18n_patterns function in your root URLconf**. See Internationalization: in URL patterns for more information about the language prefix and how to internationalize URL patterns._
- _Failing that, it looks for a cookie. The name of the cookie used is set by the LANGUAGE_COOKIE_NAME setting. (The default name is django_language.)_
- _Failing that, it looks at the Accept-Language HTTP header. This header is sent by your browser and tells the server which language(s) you prefer, in order by priority. Django tries each language in the header until it finds one with available translations._
- _Failing that, it uses the global LANGUAGE_CODE setting._

First, it looks for the language prefix. But, IMO, `i18n_patterns` with `prefix_default_language = False` is meant to use the very same mechanism but specifying that the default language is not prefixed. So unprefixed URL within `i18n_patterns` should use `LANGUAGE_CODE`. By the way, that was the behavior before 4.2.
As a developer, when I use `i18n_patterns`, I want the URLs to take precedence over the user browser preferences.

To reinforce my point, it is not only a technical issue. My clients don't want their webpages to have a "fr" prefix for their main language. https://www.crousel.com/ is the home page, not https://www.crousel.com/fr/
Visitors can explicitly choose their language by clicking on the links in the header. The redirection to `/en/about` is not a solution, clicking on 'Fr' will never lead me to the french page because my browser sends a Accept-Language 'en' header.

aboutofpluto:
> Also @aboutofpluto thank you for your testing and suggestions, it's really useful star2

You're welcome, it's the least I can do :)

sarahboyce:
Thank you @nessita 
> logic in process_request and process_response feels extremely complicated

I completely agree, I've found this incredibly fiddly :sweat_smile: 
I think I have something that works (but needs more test cases) but it's complicated and adding to something already complicated. If anybody has any bright ideas I'm all ears

------

On the discussion, we are all in agreement that the 404s are wrong and I think the bug ticket captures that. I am hoping to resolve the specific 404 issue with this PR/ticket.

@aboutofpluto I understand you want the commit 94e7f471c4edef845a4fe5e3160132997b4cca81 to be reverted. This PR and the ticket is not the best place to host this discussion as I want to use this to resolve the 404s. Can you please post on the [django-developers](https://docs.djangoproject.com/en/4.2/internals/mailing-lists/#django-developers) mailing list? 
There you will have visibility from the wider community and steering committee. 
We have some differences of opinion here but if you can get an agreement the behaviour will be changed. :+1: 

So step by step, let's resolve the 404s first and see what the decision is :slightly_smiling_face:

aboutofpluto:
Dear Sarah,

You are right, this is probably not the best place for the discussion even if it seems obvious to me that redirecting "/a-propos/" to "/en/about" only solves the 404 issue but does not fix an inconsistency in django (`translate_url` and `process_request/process_response` do not agree).

I'll post to django-developers and add a comment to the issue 34515.

Thanks!

shangxiao:
@aboutofpluto just fyi the Django forum seems to be the preferred channel now and you may get more audience there

aboutofpluto:
> @aboutofpluto just fyi the Django forum seems to be the preferred channel now and you may get more audience there

Thanks @shangxiao. Here is the post (I hope it reports your opinion correctly @sarahboyce @nessita):
https://forum.djangoproject.com/t/django-behavior-on-translatable-url-patterns-i18n-patterns-when-prefix-default-language-false-is-used/20564

PR #16798, opened 2023-04-25
fix: correct 'decade' computation

sblondon:
The implementation of `Person.decade_born_in()` method in an example does not cut the year properly. It has no impact about how to use the django framework but it's better to have correct behavior.

Implementation demo:

```python-console
>>> 1950 // 10 * 10
1950
>>> '%d’s' % (1950 // 10 * 10)
'1950’s'
>>> 1950 % (10 * 10)
50
```

### DIFF
--- a/docs/ref/contrib/admin/index.txt
+++ b/docs/ref/contrib/admin/index.txt
@@ -607,7 +607,7 @@ subclass::
 
               @admin.display(description="Birth decade")
               def decade_born_in(self):
-                  decade = self.birthday.year // 10 * 10
+                  decade = self.birthday.year % (10 * 10)
                   return f"{decade}’s"
 
 


nessita:
Hello! Thank you for your contribution! While we appreciate your effort with this PR, we can't yet accept it since it's not bound to an Accepted ticket report. Could you please report a new issue in the Django tracker following the guidelines from https://docs.djangoproject.com/en/dev/internals/contributing/bugs-and-features/?

Once the ticket is accepted (we will triage it soon after you create it), we can link this PR with the ticket to continue with the process.

Having said the above, this PR also needs to provide a test that would fail without the proposed fix (and this test should ideally be the one provided in the ticket for us to properly triage).

I hope the above helps!

shangxiao:
Hi @sblondon,

Actually the original code in the example is correct:

```
>>> 2013 // 10 * 10
2010
```

2010 is the decade containing 2013 😜

Unless I'm missing something?

nessita:
> Hi @sblondon,
> 
> Actually the original code in the example is correct:
> 
> ```
> >>> 2013 // 10 * 10
> 2010
> ```
> 
> 2010 is the decade containing 2013 stuck_out_tongue_winking_eye
> 
> Unless I'm missing something?

@shangxiao I think that by "decade" the reporter is expecting a two digit value, á la "That 70s show", but IMHO the full year truncated to the decade is the correct implementation (from https://en.wikipedia.org/wiki/Decade)

sblondon:
By the way, my fix does not properly for what I wanted to do...

```python-console
>>> 2010 % 100 # ok
10
>>> 2011 % 100 # fail
11
```

PR #16805, opened 2023-04-27
Fixed #34521 -- Used __slots__ for more template classes.

adamchainz:
ticket-34521

Makes template rendering ~1% faster.

### DIFF
--- a/django/contrib/admin/templatetags/log.py
+++ b/django/contrib/admin/templatetags/log.py
@@ -5,7 +5,9 @@ register = template.Library()
 
 class AdminLogNode(template.Node):
     def __init__(self, limit, varname, user):
-        self.limit, self.varname, self.user = limit, varname, user
+        self.limit = limit
+        self.varname = varname
+        self.user = user
 
     def __repr__(self):
         return "<GetAdminLog Node>"
--- a/django/template/backends/django.py
+++ b/django/template/backends/django.py
@@ -45,6 +45,8 @@ class DjangoTemplates(BaseEngine):
 
 
 class Template:
+    __slots__ = ("template", "backend")
+
     def __init__(self, template, backend):
         self.template = template
         self.backend = backend
--- a/django/template/base.py
+++ b/django/template/base.py
@@ -99,6 +99,8 @@ class TokenType(Enum):
 
 
 class VariableDoesNotExist(Exception):
+    __slots__ = ("msg", "params")
+
     def __init__(self, msg, params=()):
         self.msg = msg
         self.params = params
@@ -108,6 +110,8 @@ class VariableDoesNotExist(Exception):
 
 
 class Origin:
+    __slots__ = ("name", "template_name", "loader")
+
     def __init__(self, name, template_name=None, loader=None):
         self.name = name
         self.template_name = template_name
@@ -136,6 +140,8 @@ class Origin:
 
 
 class Template:
+    __slots__ = ("name", "origin", "engine", "source", "nodelist")
+
     def __init__(self, template_string, origin=None, name=None, engine=None):
         # If Template is instantiated directly rather than from an Engine and
         # exactly one Django template engine is configured, use that engine.
@@ -288,6 +294,8 @@ def linebreak_iter(template_source):
 
 
 class Token:
+    __slots__ = ("token_type", "contents", "position", "lineno")
+
     def __init__(self, token_type, contents, position=None, lineno=None):
         """
         A token representing a string from the template.
@@ -335,6 +343,8 @@ class Token:
 
 
 class Lexer:
+    __slots__ = ("template_string", "verbatim")
+
     def __init__(self, template_string):
         self.template_string = template_string
         self.verbatim = False
@@ -396,6 +406,8 @@ class Lexer:
 
 
 class DebugLexer(Lexer):
+    __slots__ = ()
+
     def _tag_re_split_positions(self):
         last = 0
         for match in tag_re.finditer(self.template_string):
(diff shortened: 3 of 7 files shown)


felixxm:
@kezabelle Would you like to take a look?

django/contrib/admin/templatetags/log.py
```diff
@@ -5,7 +5,9 @@
 
 class AdminLogNode(template.Node):
     def __init__(self, limit, varname, user):
-        self.limit, self.varname, self.user = limit, varname, user
+        self.limit = limit
+        self.varname = varname
+        self.user = user
```
shangxiao, line 10: ~~Was this change intentional?~~ 🤔 

(I agree this is easier to grok but not sure what the standard is here)

Ah I see other places use this format so at least it's consistent.

shangxiao (approved):
slots ftw 🎉

adamchainz:
I just pushed an update adding to a couple more places.

`django.template.smartif` could use slots too, but I didn't do it since I think it could be optimized much more - hopefully will make a second ticket + PR for that.

django/contrib/admin/templatetags/log.py
adamchainz: Sorry I didn't hit the button to post the above comment - yes it's intentional, see the comment for details.

It could be put into a separate commit removing this pattern from everywhere.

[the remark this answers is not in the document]
shangxiao: Handy to know, cheers!

adamchainz: I'd put it the other way round: why leave in a slow pattern that can be copied in the future? IMO it doesn't clarify the code, it actually takes more mental load to figure out where variables are assigned.

felixxm: For me, it's just an extra noise in the history. We try to avoid trivial code reformatting unless it's a big problem.

kezabelle (commented):
👋 Thank you for stepping up and actually progressing this, Adam. 

From the POV of someone who wanted to achieve (at least some of) this but definitely hasn't had the time (or spare mental capacity 😆) to look at it, I can roughly say that it ought to be possible to take any of the `__slots__` changes where we're either just _adding_ a slots definition, or _adding_ a slots definition and _adjusting_ an existing `__init__` method to have a default value.

I'm _slightly_ wary of those changes where we add a slots definition and _introduce_ a new method call (either a `__new__` [e.g. `Node`] or an `__init__` [e.g. `NodeList`, `RenderContext`]) as we're essentially forcing another invisible function call that didn't previously exist, and I'm not _sure_ that the benchmark script in the ticket _necessarily_ exercises those changes well (though tbf, it might, because these are low-level details that won't necessarily be apparent in the test code itself). On balance I _imagine_ those changes are fine because they'll be init'd far less than they'll be called so the overall trend would be positive, but it's a bit of a murky 🤷 for me.[^1]

[^1]: As I've definitely gone down the rabbit hole in these kinds of micro-optimizations previously, I fully recognise that proving that the latter changes (`__new__` et al) are "fine" is potentially a bunch more work to potentially just reach the conclusion of "see, it was all OK and worthwhile from the beginning", and if that's where most of the gain lies (could be!), it'd be frankly annoying to do, for the amount of gain there is.

django/contrib/admin/templatetags/log.py
adamchainz: Okay, fair enough.

felixxm:
@adamchainz Can you take a look on Keryn's comments? I have a similar doubts about extra `__init__`/`__new__` calls, is it worth changing in these cases? :thinking:

felixxm:
Closing per ticket.

I moved the second cleanup to the #16869.

ngnpope:
> Can no longer create weak references to template objects. (This could be worked around but...)

Just for info, this would be resolved by adding `"__weakref__"` to `__slots__`.

> I appreciate that it's not part of the official contract for template object **not** to use slots, but I think it's a pretty major change.

On the other hand, arbitrary monkey patching of things has typically fallen into the "not supported" category.

Using `__slots__` could have a big advantage for reduced memory usage in cases with many nodes/includes I've encountered. So let's not rule this out entirely/forever, but wait and see what might work in the future after your template fragments work.

adamchainz:
Fair, I guess.

Perhaps we can start smaller. What about applying slots only to Node classes? That is the bulk of the savings and they’re much less likely to be interesting to monkey patch.

carltongibson:
Hey @ngnpope! 

> Just for info...

Yes. That was the point of my "can be worked around..." — but failure to do say (for lack of omniscience on our parts) leaves a pretty hard barrier.  

I'm not sure I really see attaching extra attributes as monkey patching, rather than just everyday Python. (See the DDT error...) But you use "monkey patching" like it's a bad thing, rather than a (to be used carefully) superpower of Python. 🤔

The usual approach to getting something experimental into Django would go:

1. Make it work any way you can to prove the concept. 
2. Work out the extra API Django needs to make that viable long-term
3. Maybe merge it into Django itself once stable. 

The trouble with the proposed change here is that it cuts that process off at step 1. You never get to make it work at all, unless you can say _a priori_ what would be needed at 2. (Which likely wouldn't be accepted as no-one wants to maintain purely speculative API)

@adamchainz Yes! That might work. Can we talk about it in Edinburgh? My worry would be making those classes difficult to customise, but I haven't dug into it. 

For my own use, I need `NodeList` and up, if that makes sense. 

Given the DDT breakage, I'd suggest testing against key ecosystem packages (especially those doing template things) before merging, to be safe. 

Less memory FTW, but we've survived forever. My concern is that we not nip an innovation bloom for the DTL in the bud, before it even gets going by making standard Python workarounds unavailable. 

🎁

PR #16817, opened 2023-05-01
Refs #24267 -- Replaced hardcoded null eliding by flags and methods.

charettes:


django/db/models/sql/query.py
```diff
@@ -2521,23 +2501,11 @@ def trim_start(self, names_with_path):
...
-        # to do currently.
-        return field.null or (
-            field.empty_strings_allowed
-            and connections[DEFAULT_DB_ALIAS].features.interprets_empty_strings_as_nulls
-        )
+    def is_nullable(self, nullable_aliases):
+        # A subquery can always return no rows and thus be nullable but the ORM
+        # currently has some expectations with regards to IN vs ANY lookups that
+        # must be revisited to allow this logic to work properly.
+        return False
```
charettes, line 2508: If we wanted to enable this we'd have to change our usage of `field IN (subquery)` to `field = ANY (subquery)`, see ticket-20024 which is related.

django/db/models/expressions.py
```diff
@@ -312,6 +312,16 @@ def _output_field_or_none(self):
             if not self._output_field_resolved_to_none:
                 raise
 
+    def is_nullable(self, nullable_aliases):
+        return any(
+            expr.is_nullable(nullable_aliases) for expr in self.get_source_expressions()
+        )
+
+    def exclude_nulls(self, nullable_aliases):
```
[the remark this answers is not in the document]
shangxiao: Devil's advocate: I'm a native english speaker and I had to look up the word "elide" when I first encountered it with migrations. I wonder if we should avoid using rarely used terms 🤔  I've had a few people ask what it means when I teach them about data migrations.

It's probably less of an issue if this is a private API… but still relevant for maintainers 🤷‍♂️

django/db/models/sql/query.py
```diff
@@ -1464,41 +1464,21 @@ def build_filter(
...
+        lookup = self.build_lookup(lookups, col, value)
+        conditions = [lookup]
 
-        require_outer = (
-            lookup_type == "isnull" and condition.rhs is True and not current_negated
-        )
-        if (
-            current_negated
-            and (lookup_type != "isnull" or condition.rhs is False)
-            and condition.rhs is not None
```
charettes, line 1477: The removal of this `JSONField` specialized case is no more thanks to the handling of `can_use_none_as_rhs` in `Lookup.exclude_nulls`.

tests/lookup/tests.py
```diff
@@ -1540,9 +1540,26 @@ def test_multivalued_join_reuse(self):
...
-                Exact(F("games__home"), "NY") & Exact(F("games__away"), "Boston")
+            Season.objects.get(IsNull(F("games"), True)),
+            self.s2,
+        )
+
+    def test_multivalued_coalesce(self):
+        product = Product.objects.create(qty_target=42)
+        self.assertEqual(
+            Product.objects.get(
+                Exact(Coalesce("stock__qty_available", "qty_target"), 42)
```
charettes, line 1562: The exact use case brought up by Anssi a few year ago https://gist.github.com/akaariai/a4c1acbfedac0f2cbf3e

django/db/models/sql/query.py
```diff
@@ -1992,7 +1982,12 @@ def split_exclude(self, filter_expr, can_reuse, names_with_path):
         lookup_class = select_field.get_lookup("exact")
         lookup = lookup_class(col, ResolvedOuterRef(trimmed_prefix))
         query.where.add(lookup, AND)
-        condition, needed_inner = self.build_filter(Exists(query))
+        condition, needed_inner = self.build_filter(
+            Exists(query),
+            current_negated=True,
+            branch_negated=True,
+            can_reuse=can_reuse,
```
charettes, line 1989: These [kwargs were originally passed along](https://github.com/django/django/commit/8593e162c9cb63a6c0b06daf045bc1c21eb4d7c1#diff-fd2300283d1546e36141373b0621f142ed871e3e8856e07efe5a22ecc38ad620L1813-R1811) but didn't carry any meaning since passing an expression to build_filter didn't make use of them until now.

Not doing so causes queries.tests.ManyToManyExcludeTest failures with the optimization enabled.

django/db/models/expressions.py
```diff
@@ -312,6 +312,22 @@ def _output_field_or_none(self):
             if not self._output_field_resolved_to_none:
                 raise
 
+    @cached_property
+    def constrains_nulls(self):
```
charettes, line 316: This property means _does this expression requires NULLs to be provided to behave correctly_. Only `IsNull(..., True)` and `Coalesce` have special null handling logic today but this flag allows other expressions to opt in instead of using hardcoded flags.

django/db/models/expressions.py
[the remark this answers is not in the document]
jacobtylerwalls: We have `elidable` already in the migrations framework, so although I hear what you're saying, it doesn't bother me here.

PR #16822, opened 2023-05-02
Update tutorial04.txt to make location of change explicit

fhightower:
This PR seeks to clarify where `question_id` was changed to `pk` to avoid any ambiguity or confusion.

### DIFF
--- a/docs/intro/tutorial04.txt
+++ b/docs/intro/tutorial04.txt
@@ -314,8 +314,8 @@ two views abstract the concepts of "display a list of objects" and
 
 * The :class:`~django.views.generic.detail.DetailView` generic view
   expects the primary key value captured from the URL to be called
-  ``"pk"``, so we've changed ``question_id`` to ``pk`` for the generic
-  views.
+  ``"pk"``, so we've changed ``question_id`` to ``pk`` in the URLconf
+  for the generic views.
 
 By default, the :class:`~django.views.generic.detail.DetailView` generic
 view uses a template called ``<app name>/<model name>_detail.html``.


nessita:
Hello @fhightower! Could you please clarify why do you think there is a confusion?

The tutorial Part 4 has a section that reads "amend URLconf" that explicitly talks about the param name:

image

IMHO the proposed changed does not add much, since "Amend views" comes right after "Amend URLconf" and the change should be clear to the reader if reading top-bottom.

kezabelle:
:wave: Hey. I know this got closed, but FWIW, I think @fhightower's proposed change (or something to that effect) does increase clarity. The text currently reads:

> The [...] generic view expects the primary key value captured from the URL to be called "pk", so we’ve changed question_id to pk for the generic views.

This would be all well and good, except the code snippet preceding this bullet point -- the changes being explained (`Amend views`) -- aren't relevant, and haven't actually been modified, because they've just been _introduced_; `pk` doesn't even appear in the GCBV definition, so we cannot have _changed_ anything, even in the process of deleting our previous implementations.
The line in question is actually just explaining _why_ the URLConf was amended (_in the previous section_, `Amend URLconf`) to use `<int:pk>` instead of `<int:question_id>`, which arguably could have been done under the
> Note that the name of the matched pattern in the path strings of the second and third patterns has changed

which doesn't currently explain _why_ I might need to note the change.

Explaining that the effective default for the CBV is `pk` is fine, and _good_, but if it doesn't correctly link the concept of the _urlconf_ parameter to the (hidden!) _CBV_ parameter (which is what this bullet point is notionally going for) then it ostensibly remains unexplained magic.

Or at least, that's my read of that part of the tutorial, naturally, YMMV.

nessita:
Hello @kezabelle! You have valid points. Would you consider opening a new PR addressing your more general concern? That is the lack of _why_s that you describe. We are always happy to review documentation improvements!

My suggestion for improving this part of the tutorial would be to swap instructions to modify views and urlconf: first have the reader modifying the views describing benefits of the generic views and their defaults, and only after then propose the change urlconf to comply with the `pk` default.

What do you think @fhightower @kezabelle?

nessita:
I created https://github.com/django/django/pull/16896 trying to sort what was discussed here.

PR #16828, opened 2023-05-04
Corrected code-block directives in docs.

jnns:
The combination of double colon before the block and `.. code-block: html` leads to the wrong formatting:

https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#resetcycle

image


felixxm (commented):
@jnns Thanks :+1: I found one more.

pauloxnet (approved):

PR #16849, opened 2023-05-11
Fixed #34555 allows Fields to be added to Models in __init_subclass__ 

hottwaj:


charettes:
Thanks for the PoC @hottwaj, this is way less invasive than I initially envisioned. The fact it piggy backs of the `contributable_attrs` pattern and is isolated to `__init_subclass__` usage makes it unlikely to cause regressions.

PR #9519, opened 2018-01-02
Refactor a test case

srinivasreddy:

PR #8661, opened 2017-06-20
Fixed #21961 -- added support for database level on delete cascade

NickStefan:
Ticket https://code.djangoproject.com/ticket/21961

General Approach:
- I've added a `models.DB_CASCADE` option for `on_delete`. Its essentially, `DO_NOTHING`, but the ForeignKey field sets a flag that affects the sql strings in `/backends/base/schemas`.

Caveats for `models.DB_CASCADE` :
- does not support on_delete signals
- will not cascade delete multiple inherited tables as expected
- will not trigger CASCADE on another model. E.g. Model A points to model B, via -
 DB_CASCADE. Model B points to model C, via CASCADE. A will cascade delete B, B will django delete C, but deleting A will not delete C!
- ManyToMany with db level cascade delete will require manually defining a through table and setting the ForeignKeys as DB_CASCADE

Existing Packages:
- I originally proof of concept'd this as a postgres only external package https://github.com/NickStefan/django-db-cascade. I think we can actually support Postgres, MySQL, and even SQLite. With that scope, I believe its better as part of django.

Other relevant tickets if we wish to support SQLite ON DELETE CASCADE
- we'd need to enable `PRAGMA foreign_keys = 1`
- ticket: https://code.djangoproject.com/ticket/14204
- thread: https://groups.google.com/forum/#!topic/django-developers/_NpZ9K4hpv0 . 
- I'd be happy to do something related, but I wasn't sure where we could parameterize the option of setting the SQLite foreign key functionality. `settings.DATABASES.OPTIONS["PRAGMA_foreign_keys"]` has been suggested, but is that an appropriate use of the that dictionary? We'd have to pop off that kwarg before passing it to the Database connection.

PR #16851, opened 2023-05-12
Refs  #21961 : Add support for database-level cascading options [Draft]

Akash-Kumar-Sen:
Ticket : https://code.djangoproject.com/ticket/21961
Following PR : #14550


felixxm (commented):
@Akash-Kumar-Sen Thanks for this patch :+1: However, I found this proposition very confusing, you introduced a second parameter `on_delete_db` and a constant `DB_CASCADE` which does nothing but allow `on_delete_db` to be set, so `DB_CASCADE` has nothing to do with cascade deletion. Also, having two arguments creates sync issues. I don't see much value in the two arguments, especially since they are mutually exclusive.

Personally, I prefer Nick's [proposition](https://github.com/django/django/pull/8661) to introduce `DB_*` options and support them in `on_delete`.

adamchainz (changes requested):
💪 Strong start! I have some suggestions for the docs right now. I am excited to see this get into Django, it is a sorely missed feature, especially at scale.

django/db/models/fields/related.py
```diff
@@ -1030,8 +1036,59 @@ def _check_on_delete(self):
...
+            on_delete in [DB_CASCADE, DB_SET_NULL]
+            and hasattr(self.model, "_meta")
+            and any(
+                not parent._meta.abstract
+                for parent in self.model._meta.get_parent_list()
+            )
+        ):
+            return [
+                checks.Error(
+                    "Field specifies unsupported on_delete=DB_CASCADE, on "
```
LilyFoote, line 1049: #This check will trigger for both `DB_CASCADE` and `DB_SET_NULL`, but only one of these is mentioned in the error message. We should mention both or parametrise the error message.

The same applies below.

tests/db_level_on_delete/models.py
```diff
@@ -0,0 +1,78 @@
+from django.db import models
+
+
+class Foo(models.Model):
+    """Initial model named Foo"""
+
+    pass
```
LilyFoote, line 7: `pass` is technically not required when the class has a docstring. I think it would be tidier to remove it.

tests/db_level_on_delete/tests.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class DatabaseLevelOnDeleteTests(TestCase):
+    def test_deletion_on_nested_cascades(self):
+        foo = Foo.objects.create()
+        bar = Bar.objects.create(foo=foo)
+        baz = Baz.objects.create(bar=bar)
+
+        self.assertEqual(bar, Bar.objects.get(pk=bar.pk))
+        self.assertEqual(baz, Baz.objects.get(pk=baz.pk))
```
LilyFoote, line 23: I think we can remove these asserts.

tests/db_level_on_delete/tests.py
```diff
@@ -0,0 +1,131 @@
...
+
+        foo.delete()
+
+        with self.assertRaises(Bar.DoesNotExist):
+            Bar.objects.get(pk=bar.pk)
```
LilyFoote, line 28:
```suggestion
            bar.refresh_from_db()
```

tests/db_level_on_delete/tests.py
```diff
@@ -0,0 +1,131 @@
...
+
+        with self.assertRaises(Baz.DoesNotExist):
+            Baz.objects.get(pk=baz.pk)
```
LilyFoote, line 31:
```suggestion
            baz.refresh_from_db()
```

tests/db_level_on_delete/tests.py
```diff
@@ -0,0 +1,131 @@
...
+            Bar.objects.create(foo=foo)
+
+        for bar in Bar.objects.all():
+            for i in range(3):
+                Baz.objects.create(bar=bar)
+
+        # one is the deletion
+        # three select queries for Bar, SetNullBar and RestrictBar
+        with self.assertNumQueries(4):
+            foo.delete()
```
LilyFoote, line 101: I'm curious why we have three select queries here. Is that part of the database's own cascade logic or is that something we're doing in Django? If the latter, do we need to do it or can we optimise down to just one (delete) query?

tests/db_level_on_delete_checks/models.py
```diff
@@ -0,0 +1,14 @@
+from django.db import models
+
+
+class Foo(models.Model):
+    """Initial model named Foo"""
+
+    pass
```
LilyFoote, line 7: This `pass` can also be tidied away.

django/db/models/fields/related.py
```diff
@@ -1051,6 +1108,35 @@ def _check_unique(self, **kwargs):
...
+                (
+                    isinstance(rel, ForeignKey)
+                    and hasattr(rel.remote_field, "on_delete")
+                    and rel.remote_field.on_delete in [DB_CASCADE, DB_SET_NULL]
+                    and on_delete in [CASCADE, SET_NULL]
+                )
+                for rel in model._meta.get_fields()
+            )
+            for model in [item for item in related_models if hasattr(item, "_meta")]
+        )
```
LilyFoote, line 1136: I think this would be more readable as nested for loops with explicit early returns instead of nested comprehensions/generator expressions.

django/db/models/fields/related.py
```diff
@@ -1051,6 +1108,35 @@ def _check_unique(self, **kwargs):
...
+            for rel in model._meta.get_fields()
+            if rel.related_model and not rel.auto_created
+        ]
+        has_related_cascade_db = any(
+            any(
+                (
+                    isinstance(rel, ForeignKey)
+                    and hasattr(rel.remote_field, "on_delete")
+                    and rel.remote_field.on_delete in [DB_CASCADE, DB_SET_NULL]
+                    and on_delete in [CASCADE, SET_NULL]
```
LilyFoote, line 1131: I think we can check `on_delete in [CASCADE, SET_NULL]` just after we create the `on_delete` variable (currently line 1119 above) instead of in this double-nested loop.

django/db/models/fields/related.py
```diff
@@ -1051,6 +1108,35 @@ def _check_unique(self, **kwargs):
...
+                )
+                for rel in model._meta.get_fields()
```
LilyFoote, line 1133: I wonder if the we can make the error message more useful by explicitly calling out which `rel` is incompatible.

django/db/models/fields/related.py
```diff
@@ -1030,8 +1036,59 @@ def _check_on_delete(self):
...
+                or any(  # generic foreign key
+                    hasattr(field, "get_content_type")
+                    for field in self.model._meta.private_fields
+                )
+            )
+        ):
+            return [
+                checks.Error(
+                    "Field specifies unsupported on_delete=DB_CASCADE on model "
+                    "declaring a GenericForeignKey.",
```
LilyFoote, line 1075: It might be useful to specify which `field` is the incompatible generic relation or generic foreign key.

django/db/models/fields/related.py
```diff
@@ -1030,8 +1036,59 @@ def _check_on_delete(self):
...
                 )
             ]
-        else:
-            return []
+        elif (
+            on_delete in [DB_CASCADE, DB_SET_NULL]
+            and hasattr(self.model, "_meta")
+            and any(
+                not parent._meta.abstract
+                for parent in self.model._meta.get_parent_list()
```
LilyFoote, line 1044: It might be helpful to specify which parent model(s) (or `parent_ptr` field(s)) are incompatible.

LilyFoote (changes requested):
Looking good! I'm excited to see this land!

I have left some minor comments and questions that I think deserve a bit more thought.

tests/db_level_on_delete/tests.py
Akash-Kumar-Sen: Initially I thought this was the Database that was making the queries, but while working on [#9519](https://code.djangoproject.com/ticket/9519) I realized that it can be optimized further as the delete() method is fetching the related objects. Will update the patch soon.

django/db/models/fields/related.py
Akash-Kumar-Sen: Sounds good, will update the patch accordingly.

django/db/models/fields/related.py
Akash-Kumar-Sen: Then we have to move to a for loop instead of the `any()`

django/db/models/fields/related.py
Akash-Kumar-Sen: The field name is being passed, so I think this will not be necessary
```
db_level_on_delete_checks.SomeAnotherModel.field_name: (fields.E345) Field specifies unsupported on_delete=DB_CASCADE on model declaring a GenericForeignKey.
        HINT: Change the on_delete rule.
```

Akash-Kumar-Sen:
Thank you for the suggestions @LilyFoote . Updated the patch with the follwing : 
- Reduced the number of queries to exactly 1.
- Optimized the logic inside the nested for loops.
- Make the error messages more understandable.

docs/ref/models/fields.txt
```diff
@@ -1723,6 +1724,41 @@ The possible values for :attr:`~ForeignKey.on_delete` are found in
...
 
+* .. attribute:: DB_CASCADE
+
+    .. versionadded:: 5.0
+
+    ``DB_CASCADE`` is a custom option introduced in Django that allows the
+    database to handle cascading deletes instead of relying on Django's
+    emulation of the SQL constraint ``ON DELETE CASCADE``. With ``DB_CASCADE``,
+    the database itself takes care of deleting related objects when a
+    referenced object is deleted.
```
smithdc1, line 1735: As reference documentation I think these can be short descriptions of what the option does. I don't have the precise words to describe this, so take this as inspiration rather than a suggestion. This also applies to the options below as well. 

```suggestion
    Enables the database handling of cascading deletes by adding the ``ON DELETE CASCADE``
    constraint.
```

django/db/models/fields/related.py
```diff
@@ -1030,8 +1049,44 @@ def _check_on_delete(self):
...
+            ]
+        elif related_model_status := self._has_related_models_with_db_cascading(
+            self.model, on_delete
+        ):
+            return [
+                checks.Error(
+                    "Using normal cascading with DB cascading referenced model is "
+                    "prohibited "
+                    f"Related model is {related_model_status.get('model')} "
+                    f"Related field is {related_model_status.get('field')}",
```
LilyFoote, line 1083: Can we do this for the `GenericForeignKey` check above?

LilyFoote (approved):
Looking good. I think the functionality is pretty much there now.

django/db/models/fields/related.py
Akash-Kumar-Sen: The field object is already being passed in the Error message so I don't think it would be necessary as the field name is already being shown to the user.

LilyFoote: I thought there was the field we're checking (`self`) and a different field was the GFK and only `self` was being shown in the error message. If both are seen, that's good.

Akash-Kumar-Sen: That would be a nice to have ig as we are not supporting db cascade to any kind of GFK.

felixxm (commented):
@Akash-Kumar-Sen Thanks :+1: I left initial comments (without checking tests and docs). We should add @NickStefan as a co-authored and add them to the `AUTHORS`.

Moreover, `DatabaseOnDelete` cannot be serialized:
```
ValueError: Cannot serialize: <django.db.models.deletion.DatabaseOnDelete object at 0x7f765c45f760>
There are some values Django cannot serialize into migration files.
```
it was already handled in #8661.

django/db/backends/base/operations.py
```diff
@@ -260,6 +260,17 @@ def limit_offset_sql(self, low_mark, high_mark):
             if sql
         )
 
+    def fk_on_delete_sql(self, operation):
+        """
+        Return the SQL to make an ON DELETE statement during a CREATE TABLE
+        statement.
+        """
+        if operation in ["CASCADE", "SET NULL", "RESTRICT"]:
+            return " ON DELETE %s " % operation
```
ngnpope, line 269:
```suggestion
            return f" ON DELETE {operation} "
```

django/db/backends/base/operations.py
```diff
@@ -260,6 +260,17 @@ def limit_offset_sql(self, low_mark, high_mark):
...
+        if operation == "":
+            return ""
+        raise NotImplementedError("ON DELETE %s is not supported." % operation)
```
ngnpope, line 272:
```suggestion
        raise NotImplementedError(f"ON DELETE {operation} is not supported.")
```

django/db/backends/base/operations.py
```diff
@@ -260,6 +260,17 @@ def limit_offset_sql(self, low_mark, high_mark):
             if sql
         )
 
+    def fk_on_delete_sql(self, operation):
+        """
+        Return the SQL to make an ON DELETE statement during a CREATE TABLE
+        statement.
+        """
+        if operation in ["CASCADE", "SET NULL", "RESTRICT"]:
```
ngnpope, line 268: What about `"NO ACTION"` and `"SET DEFAULT"`? See:

- https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK
- https://dev.mysql.com/doc/refman/8.0/en/create-table-foreign-keys.html
- https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/constraint.html#GUID-1055EA97-BA6F-4764-A15F-1024FD5B6DFE__GUID-8E388631-3D99-4445-A2CD-A8776E8CEFC1
- https://www.sqlite.org/foreignkeys.html#fk_actions

Looks like Oracle is the only one that wouldn't support these. (And notably, it also doesn't look like it supports `"RESTRICT"`.

For `"SET DEFAULT"` we'd also need to ensure that `db_default` is set on the field. (Recently added to Django 5.0)

django/db/models/fields/related.py
```diff
@@ -1051,6 +1106,50 @@ def _check_unique(self, **kwargs):
...
+                elif (
+                    isinstance(rel, ForeignKey) or isinstance(rel, OneToOneField)
+                ) and hasattr(rel.remote_field, "on_delete"):
+                    related_on_delete = rel.remote_field.on_delete
+                if related_on_delete in [DB_CASCADE, DB_SET_NULL, DB_RESTRICT]:
+                    non_db_related_models["model"] = rel_model
+                    non_db_related_models["field"] = related_remote_field
+                    return non_db_related_models
+
+        return non_db_related_models
```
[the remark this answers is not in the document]
Akash-Kumar-Sen: It might be helpful to specify which parent model(s) (or `parent_ptr` field(s)) are incompatible. As the current field does not refers to the exact field on the parent model, it will be helpful for the user to know which parent model field uses non_db on_delete options.

django/db/models/fields/related.py
```diff
@@ -1051,6 +1106,50 @@ def _check_unique(self, **kwargs):
...
+        db cascading return true
+        """
+        non_db_related_models = {}
+        # Optimization for the case when the model does not have non-db deletion
+        if not hasattr(model, "_meta") or on_delete in [
+            DB_CASCADE,
+            DB_SET_NULL,
+            DB_RESTRICT,
+        ]:
+            return non_db_related_models
```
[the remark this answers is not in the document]
Akash-Kumar-Sen: Some built in models does not have the `_meta` attribute, that caused some error in the tests. You can also refer to the following conversation on the past PR:
https://github.com/django/django/pull/14550/files#r719896252

django/db/models/fields/related.py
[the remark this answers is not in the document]
Akash-Kumar-Sen: Earlier I made it with just a true/false flag, got some community review regarding the specification : https://github.com/django/django/pull/16851#discussion_r1213358420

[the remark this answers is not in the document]
felixxm: You always return one model and one field, so why `non_db_related_models` is a dictionary? It's enough to return a field and get model from it, check out my [comment](https://github.com/django/django/pull/16851#discussion_r1251632988).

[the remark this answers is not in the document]
Akash-Kumar-Sen: Got it, will update the PR soon.

django/db/backends/base/operations.py
Akash-Kumar-Sen: Thank you for the review Nick. Will try to add the `DB_SET_DEFAULT` in the next commit and add a warning message for oracle database, in that case we may redirect to django based `RESTRICT`.

NickStefan: In the original PR, I think we left out NO_ACTION since it's the default FK behavior--even if you don't add an `ON DELETE...` statement. It's likely equivalent to django's DO_NOTHING. It wouldn't hurt to add it for completeness, but it might also add unnecessary options that do the same thing. Seems like a style preference. 

@ngnpope how sure are you about Oracle not supporting RESTRICT? Im not an oracle expert, but the language linked makes it seem like only _views_ dont support RESTRICT (but maybe tables would?).

> Restriction on ON DELETE
> You cannot specify this clause for a view constraint.

ngnpope: > In the original PR, I think we left out NO_ACTION since it's the default FK behavior--even if you don't add an ON DELETE... statement. It's likely equivalent to django's DO_NOTHING. It wouldn't hurt to add it for completeness, but it might also add unnecessary options that do the same thing. Seems like a style preference.

My concern was that `DO_NOTHING` makes no changes to the database. So if you've set, e.g. `DB_RESTRICT` and change to `DO_NOTHING` then the database would still be using `RESTRICT` because no migration would be generated? Because there is nothing to `ALTER` it to `NO ACTION`? I haven't had time to investigate this in detail - it's a hunch - but we should have tests for all transitions between database-managed and "traditional" `on_delete` options.

This is also why I've thought it might be cleaner to have a `db_on_delete` like we have a `db_default`... We need to get this right - not rush - as we'll be stuck with the decision.

NickStefan: Good point on migrations. It's been a while since Ive used django.

IIRC, the migrations aren't persisted as raw SQL though, right? So wouldnt changing the kwarg to DO_NOTHING possibly still *serialize* differently into the migration file than DB_CASCADE? And if it serializes differently, wont that then be able to generate new SQL for the table? I guess it's possible that wouldnt be backwards compatible though--if someone already has migrations manually adding CASCADE and they put DO_NOTHING in their python code...

Okay, thinking about it, I guess DB_NO_ACTION makes sense 👍 .

tests/db_level_on_delete_checks/tests.py
```diff
@@ -0,0 +1,183 @@
+from django.contrib.contenttypes.fields import GenericForeignKey
```
felixxm, line 1: Please move these tests to the `tests/invalid_models_tests/test_relative_fields.py` and reuse existing models.

tests/db_level_on_delete/tests.py
```diff
@@ -0,0 +1,171 @@
+from django.db import IntegrityError, transaction
```
felixxm, line 1: I'd move these tests to the `tests/delete/` and reuse existing models (when appropriate).

django/db/migrations/serializer.py
```diff
@@ -258,6 +258,11 @@ def serialize(self):
         return "pathlib.%s%r" % (prefix, self.value), {"import pathlib"}
 
 
+class DatabaseOnDeleteSerializer(BaseSerializer):
+    def serialize(self):
+        return f"models.{self.value.name}", {}
```
felixxm, line 263: Is this tested?

django/db/backends/base/operations.py
felixxm: > Best thing is to ensure that there are comprehensive tests migrating between all possible options.

Yes, we need migration tests.

tests/db_level_on_delete/tests.py
Akash-Kumar-Sen: Done.

django/db/backends/base/operations.py
Akash-Kumar-Sen: Added tests for migrations.

tests/db_level_on_delete_checks/tests.py
Akash-Kumar-Sen: Done

Akash-Kumar-Sen:
- Added tests for migrations.

**There are two issues I could not figure out after a lot of try, some help would be great : **
- MySQL specific error in the test ```delete.tests.DatabaseLevelOnDeleteTests.test_foreign_key_db_default```
- I am not sure how to test the serializer. (Some docs link or links of some similar tests would be helpful)

Akash-Kumar-Sen:
> Firstly, this is not where I think it is failing. To me it looks like it is struggling to create the table `DBDefaultsFK` (`delete_dbdefaultsfk`) which is in `tests/delete/models.py` (when the tests are being set up it first creates all tables so it isn't from a specific assert or test).

I have tried to print the queries. the delete query in ```foo.delete()``` is causing the error, the queries before that are the following, these seems to be executed succesfully:
```
query =  CREATE TABLE `delete_dbdefaultsfk` (`id` integer AUTO_INCREMENT NOT NULL PRIMARY KEY, `language_code_id` integer DEFAULT %s NOT NULL)
args =  [1]
query =  ALTER TABLE `delete_dbdefaultsfk` ADD CONSTRAINT `delete_dbdefaultsfk_language_code_id_04474ba9_fk_delete_foo_id` FOREIGN KEY (`language_code_id`) REFERENCES `delete_foo` (`id`) ON DELETE SET DEFAULT
args =  ()
```

> 
> I do not know what I am talking about here. It sounds like there might be some kinda [type mismatch](https://dba.stackexchange.com/questions/203509/mysql-foreign-key-constraint-is-incorrectly-formed) so I would first look at the generated SQL to create that table on MySQL.

This could be possible reason for the error, I'll take a look at this one.
> 
> (I would also be tempted to play around with the table such as changing `on_delete=models.DB_SET_DEFAULT` to `on_delete=models.RESTRICT` to check whether `db_default` itself is driving the error or the deletion behaviour - but that probably reflects that I don't understand how it all works so want to bump my way around)

I have looked into the tests for ```db_default``` and that seems to be working fine.
> 
> Have you looked into `tests/migrations/test_writer.py WriterTests` to me it looks like there are some serializer tests in there?

I missed that one, will take a look.
Thank you for the feedback Sarah

django/db/migrations/serializer.py
Akash-Kumar-Sen: Added tests for ```DatabaseOnDeleteSerializer``` - https://github.com/django/django/blob/a8789ab60cafc1576080bd5da200324b365659ac/tests/migrations/test_writer.py#L1005

sarahboyce:
Ah @Akash-Kumar-Sen I realised there are two errors, one for MariaDB (https://djangoci.com/job/pr-mariadb/database=mysql,label=mariadb,python=python3.11/21757/console) this is failing on the table creation and the other failures are (as you said) for the test `delete.tests.DatabaseLevelOnDeleteTests.test_foreign_key_db_default`

I am reading the MySQL documentation on foreign keys here: https://dev.mysql.com/doc/refman/8.0/en/create-table-foreign-keys.html
This sounds interesting:
> - SET DEFAULT: This action is recognized by the MySQL parser, but both [InnoDB](https://dev.mysql.com/doc/refman/8.0/en/innodb-storage-engine.html) and [NDB](https://dev.mysql.com/doc/refman/8.0/en/mysql-cluster.html) reject table definitions containing ON DELETE SET DEFAULT or ON UPDATE SET DEFAULT clauses.

> For storage engines that support foreign keys, MySQL rejects any [INSERT](https://dev.mysql.com/doc/refman/8.0/en/insert.html) or [UPDATE](https://dev.mysql.com/doc/refman/8.0/en/update.html) operation that attempts to create a foreign key value in a child table if there is no matching candidate key value in the parent table.

Apparently MariaDB uses the InnoDB storage engine by default - I think that explains why the table create is failing there. 

It makes me think _maybe_ Django should not allow `on_delete=DB_SET_DEFAULT` for MySQL :thinking:

Akash-Kumar-Sen (commented):
That seems to be the only way to go @sarahboyce . I will wait for a few more feedbacks and if nobody suggests otherwise I will disable ```ON DELETE SET DEFAULT``` for ```MySQL```

sarahboyce:
@Akash-Kumar-Sen I think go for the disabling MySQL support plan (also needs a docs adjustment). 👍  
The alpha feature freeze is coming up. If there is a solution which adds MySQL support, IMO we can always add it in a following release as to not block this work

Akash-Kumar-Sen:
> @Akash-Kumar-Sen I think go for the disabling MySQL support plan (also needs a docs adjustment). 

Done @sarahboyce .

django/db/models/fields/related.py
```diff
@@ -1051,6 +1104,42 @@ def _check_unique(self, **kwargs):
...
+                related_remote_field = rel
+                if isinstance(rel, OneToOneRel) and hasattr(rel, "on_delete"):
+                    related_on_delete = rel.on_delete
+                    related_remote_field = rel.remote_field
+                elif (
+                    isinstance(rel, ForeignKey) or isinstance(rel, OneToOneField)
+                ) and hasattr(rel.remote_field, "on_delete"):
+                    related_on_delete = rel.remote_field.on_delete
+                if isinstance(related_on_delete, DatabaseOnDelete):
+                    return related_remote_field
```
LilyFoote, line 1140: I think this might be a bit simpler to understand:

```suggestion
                if isinstance(rel, OneToOneRel) and isinstance(rel.on_delete, DatabaseOnDelete):
                    return rel.remote_field
                elif (
                    isinstance(rel, ForeignKey) or isinstance(rel, OneToOneField)
                ) and isinstance(getattr(rel.remote_field, "on_delete", None), DatabaseOnDelete):
                    return rel
```

I was able to verify that `rel` always has an `on_delete` attribute if it is an instance of `OneToOneRel`, so we can do `rel.on_delete` here. I think we need the `getattr` for `rel.remote_field`, but it would be nice if we didn't. Maybe it would be enough to set `on_delete = None` in `Field.__init__`?

ngnpope: You could also collapse `isinstance(rel, ForeignKey) or isinstance(rel, OneToOneField)` to `isinstance(rel, (ForeignKey, OneToOneField))` in the `elif` branch.

LilyFoote: > You could also collapse `isinstance(rel, ForeignKey) or isinstance(rel, OneToOneField)` to `isinstance(rel, (ForeignKey, OneToOneField))` in the `elif` branch.

Good point! I'll edit the suggestions.

carltongibson (commented):
@nessita You said to ping you on PRs we'd like in… — this one if possible, but it may be too much left to do. 

Can we add a test that migrates from a model from an `on_delete=CASCADE` to a `on_delete=DB_CASCADE`, and then performs an actual check on the DB that the migration is applied? 

I've just worked through applying this to an existing project on PostgreSQL 14.  

* The new checks are good, e.g. `(fields.E323) Using python based on_delete with database level on_delete referenced model is prohibited ...` Really nice. 
* Migration is correctly generated. 
* **BUT** applying the migration, the actual fields weren't updated. **Likely** I'm holding it wrong, but I have a `*_id` column the AFAICS should now have ON DELETE CASCADE, but in fact has nothing. 🤔

carltongibson:
@nessita I think that is the correct call 👍

Akash-Kumar-Sen (commented):
> Can we add a test that migrates from a model from an `on_delete=CASCADE` to a `on_delete=DB_CASCADE`, and then performs an actual check on the DB that the migration is applied?
>

@carltongibson I have already written the following test for this task it might be the case that I have missed something here if you can take a look that would be very helpful

https://github.com/django/django/blob/28699ebef224412fc80e3e7a73bd89ef7dc4845c/tests/migrations/test_operations.py#L244

django/db/models/fields/related.py
Akash-Kumar-Sen: This seems logical. I would try put these suggestions on test soon.

django/db/models/__init__.py
```diff
@@ -114,4 +119,9 @@
     "ManyToOneRel",
     "ManyToManyRel",
     "OneToOneRel",
+    "DB_CASCADE",
+    "DB_RESTRICT",
+    "DB_SET_NULL",
+    "DB_SET_DEFAULT",
+    "DatabaseOnDelete",
```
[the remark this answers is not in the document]
Akash-Kumar-Sen: I think we have a few imports later checking ```if isinstance(on_delete, DatabaseOnDelete)``` that's why I kept it there.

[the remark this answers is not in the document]
LilyFoote: Not in this file.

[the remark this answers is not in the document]
Akash-Kumar-Sen: Now I got it, will remove it

carltongibson:
Hi @Akash-Kumar-Sen — Yes (at quick read through) that looks like it's doing the right thing at the model-level. 

The issue I hit was that trying to update an actual project the actual field in the database wasn't changed. (`CASCADE` to `DB_CASCADE` on PostgreSQL 14) 

The change in the field was correctly detected, the migration was generated, and when applied it reported success, but the constraint wasn't applied to the field. (No `ON DELETE ...` in the DDL.) 

The test I'd like to see would hit the actual database. 

Something like: 

1. Create a pair of models with `CASCADE`. 
2. Create some instances. 
3. Delete the parent model. 
4. Assert the children were deleted (by Django)
5. Migrate the child model to use `DB_CASCADE`. 
6. Create some more instances. 
7. Delete the parent model again. 
8. Assert the children were deleted (by the database). 

It's step 8 that failed for me (because the field was never updated). 

As I say, I may have done something wrong. I didn't have time to look into this more fully yet. 

But all I did was update (a small group of) my models and migrate. It should have worked. (Or errored loudly.)

Akash-Kumar-Sen (commented):
> But all I did was update (a small group of) my models and migrate. It should have worked. (Or errored loudly.)

Thank you for the check @carltongibson I have also checked and this unexpected behavior is happening. I tried to print the queries and seen that out of the 16 combinations, (4 DB level on deletes X 4 non db on deletes) 5 are not making the query that updates the database.
Seems like it is not working alone, if we make another change with it, like  `null=True or db_default=1` then it is working fine. I am currently trying to figure out the reason, It might be the case I have not added the necessary code for migrations to be applied. @LilyFoote , if you can look into it that would be great, I have drawn a lot of testing insights from your `db_default` PR.

LilyFoote:
@Akash-Kumar-Sen Do you have some failing tests you can push?

Akash-Kumar-Sen:
> @Akash-Kumar-Sen Do you have some failing tests you can push?

Sure, pushing them in half an hour

tests/migrations/test_operations.py
```diff
@@ -241,6 +241,274 @@ def test_create_model_with_db_level_fk(self):
...
+
+        Rider = new_state.apps.get_model(app_label, fk_child_model)
+        Pony = new_state.apps.get_model(app_label, fk_parent_model)
+
+        pony = Pony.objects.create()
+        kwargs = {fk_fieldname: pony}
+        Rider.objects.create(**kwargs)
+
+        with self.assertNumQueries(1):
+            pony.delete()
```
LilyFoote, line 266: I see that we're testing that `pony.delete()` only causes one query.

I think we should also test that all the rows that we expect to be deleted *are* deleted.

tests/migrations/test_operations.py
```diff
@@ -241,6 +241,274 @@ def test_create_model_with_db_level_fk(self):
...
+
+        Rider.objects.all().delete()
+        Pony.objects.all().delete()
```
LilyFoote, line 269: What is the reason for these queries?

Also, this question applies to the other tests below.

tests/migrations/test_operations.py
```diff
@@ -241,6 +241,274 @@ def test_create_model_with_db_level_fk(self):
...
+        with connection.schema_editor() as editor:
+            operation.database_backwards(app_label, editor, new_state, project_state)
```
LilyFoote, line 271: We should also test that behaviour after `database_backwards` is what we expect. For example, I'd expect the equivalent `assertNumQueries` to above to assert more than one query.

tests/migrations/test_operations.py
```diff
@@ -241,6 +241,274 @@ def test_create_model_with_db_level_fk(self):
...
+
+        pony = Pony.objects.create()
+        kwargs = {fk_fieldname: pony}
+        rider = Rider.objects.create(**kwargs)
+
+        with self.assertNumQueries(1):
+            pony.delete()
+
+        rider.refresh_from_db()
+        self.assertEqual(getattr(rider, fk_fieldname), None)
```
LilyFoote, line 333:
```suggestion
        self.assertIsNone(getattr(rider, fk_fieldname))
```

LilyFoote (changes requested):
It's great to see these migrations tests!

I'd love to see more detailed asserts, to make sure we're really testing what we think we are.

I also think we might need to take a different approach to the `subTest`s, to allow us to cleanly test the backwards migrations also do the right thing.

Thanks for persisting with this PR, I think we're getting closer with every update!

tests/migrations/test_operations.py
Akash-Kumar-Sen: There are just two extra `BEGIN` and `COMMIT` queries, these extra queries are not there in `tests/delete/tests.py` that's why I didn't worried much about them. This might have something to do with applying migrations on the test itself

tests/migrations/test_operations.py
Akash-Kumar-Sen: Initially I thought of doing that, as it was already getting very long so didn't add them looking for a better loop based structure to reduce repetition. Will add after having the most optimal loop structure to write the tests.

tests/migrations/test_operations.py
```diff
@@ -241,6 +241,274 @@ def test_create_model_with_db_level_fk(self):
...
+
+        Rider = new_state.apps.get_model(app_label, fk_child_model)
+        Pony = new_state.apps.get_model(app_label, fk_parent_model)
+
+        pony = Pony.objects.create()
+        kwargs = {fk_fieldname: pony}
+        Rider.objects.create(**kwargs)
+
+        with self.assertRaises(IntegrityError):
+            pony.delete()
```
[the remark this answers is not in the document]
Akash-Kumar-Sen: I will add that.

tests/migrations/test_operations.py
```diff
@@ -241,6 +241,274 @@ def test_create_model_with_db_level_fk(self):
...
+        kwargs = {fk_fieldname: pony}
+        rider = Rider.objects.create(**kwargs)
+
+        # Extra begin and commit queries
+        with self.assertNumQueries(3):
+            pony.delete()
+
+        rider.refresh_from_db()
+        pony1.refresh_from_db()
+        self.assertEqual(getattr(rider, fk_fieldname), pony1)
```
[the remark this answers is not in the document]
Akash-Kumar-Sen: Sounds interesting, will add that.

tests/migrations/test_operations.py
```diff
@@ -241,6 +241,274 @@ def test_create_model_with_db_level_fk(self):
...
+            operation.database_forwards(app_label, editor, project_state, new_state)
+
+        Rider = new_state.apps.get_model(app_label, fk_child_model)
+        Pony = new_state.apps.get_model(app_label, fk_parent_model)
+
+        pony1 = Pony.objects.create(id=1)
+        pony = Pony.objects.create()
+        kwargs = {fk_fieldname: pony}
+        rider = Rider.objects.create(**kwargs)
+        pony.delete()
```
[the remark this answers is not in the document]
Akash-Kumar-Sen: This was a `non_db` test, so I haven't focused on the number of queries, Just the behavior was enough.

Akash-Kumar-Sen:
> I also think we might need to take a different approach to the `subTest`s, to allow us to cleanly test the backwards migrations also do the right thing.
>

Thank you for the review @LilyFoote , what's your suggestion on this?

LilyFoote:
> > I also think we might need to take a different approach to the `subTest`s, to allow us to cleanly test the backwards migrations also do the right thing.
> 
> Thank you for the review @LilyFoote , what's your suggestion on the `subTest`s framework?

I'm not really sure. Maybe there's a way to have a group of asserts associated with each `on_delete` option, and then call each after migrate and backwards migrate?

If you can't get that working, I think giving up on `subTest` might be reasonable - maybe create a new testcase or test file for these tests and just have all of them as distinct test methods? I know there's going to be loads of them, but maybe that's easier to write...

tests/migrations/test_operations.py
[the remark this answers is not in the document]
Akash-Kumar-Sen: Test : 
```
    @skipUnlessDBFeature("has_on_delete_db_default")
    def test_foreign_key_db_default(self):
        # Default parent exists
        default_parent = Foo.objects.create(pk=1)
        parent = Foo.objects.create(pk=2)
        child1 = DBDefaultsFK.objects.create(language_code=parent)
            parent.delete()
        child1.refresh_from_db()
        self.assertEqual(child1.language_code, default_parent)
        child1.delete()
        default_parent.delete()

        # default parent dne
        parent = Foo.objects.create(pk=2)
        child1 = DBDefaultsFK.objects.create(language_code=parent)
        parent.delete()
        DBDefaultsFK.objects.all()
```

```
ERROR: test_foreign_key_db_default (delete.tests.DatabaseLevelOnDeleteTests)
  File "C:\Users\akash\Desktop\Djangocon\django\django\db\backends\utils.py", line 103, in _execute
    return self.cursor.execute(sql)
psycopg2.errors.ForeignKeyViolation: insert or update on table "delete_dbdefaultsfk" violates foreign key constraint "delete_dbdefaultsfk_language_code_id_04474ba9_fk_delete_foo_id"
DETAIL:  Key (language_code_id)=(1) is not present in table "delete_foo".


  File "C:\Users\akash\Desktop\Djangocon\django\django\test\testcases.py", line 294, in _setup_and_call
    self._post_teardown()
  File "C:\Users\akash\Desktop\Djangocon\django\django\test\testcases.py", line 1112, in _post_teardown
    self._fixture_teardown()
  File "C:\Users\akash\Desktop\Djangocon\django\django\test\testcases.py", line 1334, in _fixture_teardown
    connections[db_name].check_constraints()
  File "C:\Users\akash\Desktop\Djangocon\django\django\db\backends\postgresql\base.py", line 383, in check_constraints
    cursor.execute("SET CONSTRAINTS ALL IMMEDIATE")
  File "C:\Users\akash\Desktop\Djangocon\django\django\db\backends\utils.py", line 79, in execute
  File "C:\Users\akash\Desktop\Djangocon\django\django\db\backends\utils.py", line 92, in _execute_with_wrappers
  File "C:\Users\akash\Desktop\Djangocon\django\django\db\backends\utils.py", line 100, in _execute
  File "C:\Users\akash\Desktop\Djangocon\django\django\db\utils.py", line 91, in __exit__
  File "C:\Users\akash\Desktop\Djangocon\django\django\db\backends\utils.py", line 103, in _execute
    return self.cursor.execute(sql)
django.db.utils.IntegrityError: insert or update on table "delete_dbdefaultsfk" violates foreign key constraint "delete_dbdefaultsfk_language_code_id_04474ba9_fk_delete_foo_id"
DETAIL:  Key (language_code_id)=(1) is not present in table "delete_foo".
```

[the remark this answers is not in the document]
Akash-Kumar-Sen: Let me know if that is expected.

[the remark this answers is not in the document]
LilyFoote: Yeah, looks correct to me.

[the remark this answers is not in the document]
Akash-Kumar-Sen: Getting some weird results here.
**The test : **
```
    @skipUnlessDBFeature("has_on_delete_db_default")
    def test_foreign_key_db_default(self):
        # Default parent exists
        default_parent = Foo.objects.create(pk=1)
        parent = Foo.objects.create(pk=2)
        child1 = DBDefaultsFK.objects.create(language_code=parent)
            parent.delete()
        child1.refresh_from_db()
        self.assertEqual(child1.language_code, default_parent)
        child1.delete()
        default_parent.delete()

        # default parent dne
            parent = Foo.objects.create(pk=2)
            child1 = DBDefaultsFK.objects.create(language_code=parent)
            parent.delete()
            DBDefaultsFK.objects.all()
```

**The error(or whatever)**
```
ERROR: test_foreign_key_db_default (delete.tests.DatabaseLevelOnDeleteTests)
  File "C:\Users\akash\Desktop\Djangocon\django\django\test\testcases.py", line 294, in _setup_and_call
    self._post_teardown()
  File "C:\Users\akash\Desktop\Djangocon\django\django\test\testcases.py", line 1112, in _post_teardown
    self._fixture_teardown()
  File "C:\Users\akash\Desktop\Djangocon\django\django\test\testcases.py", line 1334, in _fixture_teardown
    connections[db_name].check_constraints()
  File "C:\Users\akash\Desktop\Djangocon\django\django\db\backends\sqlite3\base.py", line 275, in check_constraints
    raise IntegrityError(
django.db.utils.IntegrityError: The row in table 'delete_dbdefaultsfk' with primary key '2' has an invalid foreign key: delete_dbdefaultsfk.language_code_id contains a value '1' that does not have a corresponding value in delete_foo.id.

FAIL: test_foreign_key_db_default (delete.tests.DatabaseLevelOnDeleteTests)
  File "C:\Users\akash\Desktop\Djangocon\django\django\test\testcases.py", line 1427, in skip_wrapper
  File "C:\Users\akash\Desktop\Djangocon\django\tests\delete\tests.py", line 899, in test_foreign_key_db_default
AssertionError: IntegrityError not raised
```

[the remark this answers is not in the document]
LilyFoote: Ah! I think the problem here is that we're using (a subclass of) `TestCase`, rather than `TransactionTestCase` - the `IntegrityError` doesn't get raised until the transaction ends, which is after the test method ends!

tests/migrations/test_operations.py
```diff
@@ -241,6 +241,274 @@ def test_create_model_with_db_level_fk(self):
...
+
+        Rider = new_state.apps.get_model(app_label, fk_child_model)
+        Pony = new_state.apps.get_model(app_label, fk_parent_model)
+
+        pony1 = Pony.objects.create(id=1)
+        pony = Pony.objects.create()
+        kwargs = {fk_fieldname: pony}
+        rider = Rider.objects.create(**kwargs)
+
+        # Extra begin and commit queries
```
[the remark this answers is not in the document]
Akash-Kumar-Sen: Thanks for responding fast! Now I also know the reason for this, it is `TransactionTestCase` that generates one extra `BEGIN` and `COMMIT` query

tests/migrations/test_operations.py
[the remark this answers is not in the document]
Akash-Kumar-Sen: Resolved now, added the tests here https://github.com/django/django/pull/16851/files#diff-a842e4474a8c70b274fabec1b44e377a498181dfe6c925801b7dc930917c3e03R904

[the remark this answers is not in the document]
Akash-Kumar-Sen: Added the test here : https://github.com/django/django/blob/d40bcc5b0b63fc1182247b8677a0d2f175e3c4da/tests/delete/tests.py#L904

tests/migrations/test_operations.py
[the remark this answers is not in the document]
Akash-Kumar-Sen: Added here : https://github.com/django/django/pull/16851/files#diff-7cc40f1cb85a8f19c80b2e901626d78e4729047dbbb0ed52861f5894ff688115R271

tests/migrations/test_operations.py
Akash-Kumar-Sen: Done over two roes here : https://github.com/django/django/pull/16851/files#diff-7cc40f1cb85a8f19c80b2e901626d78e4729047dbbb0ed52861f5894ff688115R258

django/db/backends/base/operations.py
```diff
@@ -269,6 +269,17 @@ def limit_offset_sql(self, low_mark, high_mark):
...
 
+    def fk_on_delete_sql(self, operation):
+        """
+        Return the SQL to make an ON DELETE statement during a CREATE TABLE
+        statement.
+        """
+        if operation in ["CASCADE", "SET NULL", "RESTRICT", "SET DEFAULT"]:
+            return f" ON DELETE {operation} "
+        if operation == "":
+            return ""
```
sobolevn, line 280: nit:

```suggestion
        if operation == "":
            return ""
        if operation in ("CASCADE", "SET NULL", "RESTRICT", "SET DEFAULT"):
            return f" ON DELETE {operation} "
```

This would be a bit faster.

django/db/models/deletion.py
```diff
@@ -216,7 +238,13 @@ def can_fast_delete(self, objs, from_field=None):
...
             # Foreign keys pointing to this model.
             all(
-                related.field.remote_field.on_delete is DO_NOTHING
+                related.field.remote_field.on_delete
+                in [
+                    DO_NOTHING,
+                    DB_CASCADE,
+                    DB_SET_NULL,
+                    DB_RESTRICT,
+                ]
```
sobolevn, line 247:
```suggestion
                in (
                    DO_NOTHING,
                    DB_CASCADE,
                    DB_SET_NULL,
                    DB_RESTRICT,
                )
```

sobolevn (commented):
(I am not really a part of Django team, but tuple creation is faster than list)

:)

django/db/backends/base/operations.py
adamchainz: Not really necessary... CPython’s compiler already converts lists used only for `in` into tuples:

```
In [1]: import dis

In [2]: def f():
   ...:     1 in (1, 2)
   ...:

In [3]: def g():
   ...:     1 in [1, 2]
   ...:

In [4]: dis.dis(f)
  1           0 RESUME                   0

  2           2 LOAD_CONST               1 (1)
              4 LOAD_CONST               2 ((1, 2))
              6 CONTAINS_OP              0
              8 POP_TOP
             10 LOAD_CONST               0 (None)
             12 RETURN_VALUE

In [5]: dis.dis(g)
  1           0 RESUME                   0

  2           2 LOAD_CONST               1 (1)
              4 LOAD_CONST               2 ((1, 2))
              6 CONTAINS_OP              0
              8 POP_TOP
             10 LOAD_CONST               0 (None)
             12 RETURN_VALUE

```

See how `g` also uses `LOAD_CONST` with constant 2, a tuple.

sobolevn: Oh, django now support 3.10+, then - yes, `list` can stay :)

adamchainz: This isn't a 3.8 thing, it goes back to Python 2.4 and improved in 3.1! https://github.com/python/cpython/commit/b7fbcd396fc4a366433cf6f26cae64fecb056099 / https://github.com/adamchainz/cpython/commit/2c31a058eb719da8908ea9d9c2f1b748ecc7a4ca

docs/releases/5.1.txt
```diff
@@ -26,6 +26,20 @@ only officially support the latest release of each series.
...
+--------------------------------------------------
+
+:attr:`ForeignKey.on_delete <django.db.models.ForeignKey.on_delete>` now
+supports four extra values to specify database-level actions:
+
+* ``DB_CASCADE`` - deletes the referring object.
+* ``DB_RESTRICT`` - prevents deletion of referred-to objects.
+* ``DB_SET_NULL`` - sets the referring foreign key to SQL ``NULL``.
+* ``DB_SET_DEFAULT`` - the value of the foreign key column in the referencing
+  table will be set to its ``db_default`` value.
```
adamchainz, line 40: 1. Use imperative.
2. Shorten `DB_SET_DEFAULT` description.

```suggestion
* ``DB_CASCADE`` - delete the referring object.
* ``DB_RESTRICT`` - prevent deletion of referred-to objects.
* ``DB_SET_NULL`` - set the referring foreign key to SQL ``NULL``.
* ``DB_SET_DEFAULT`` - set the referring column to its ``db_default`` value.
```

docs/releases/5.1.txt
```diff
@@ -26,6 +26,20 @@ only officially support the latest release of each series.
...
+
+
```
adamchainz, line 42: Only one blank line needed.
```suggestion

```

adamchainz (commented):
Some docs improvements.

django/db/backends/base/operations.py
sobolevn: Wow, time flies :)
I though it was 3.8 or something.

ColemanDunn:
Anything blocking this from getting approved?

docs/releases/5.1.txt
```diff
@@ -26,6 +26,17 @@ only officially support the latest release of each series.
 What's new in Django 5.1
 ========================
 
+Database-level values for ``ForeignKey.on_delete``
+--------------------------------------------------
+
+:attr:`ForeignKey.on_delete <django.db.models.ForeignKey.on_delete>` now
```
felixxm, line 32:
```suggestion
:attr:`.ForeignKey.on_delete` now ...
```

tests/delete/models.py
```diff
@@ -241,3 +241,115 @@ class GenericDeleteBottomParent(models.Model):
...
+
+
+class Baz(models.Model):
+    """Second level foreignkey child for Foo
+    Implemented using in DB cascading"""
+
+    bar = models.ForeignKey(
+        Bar,
+        on_delete=models.DB_CASCADE,
+    )
```
felixxm, line 267: This (and many others) can be single-lined.

tests/delete/models.py
```diff
@@ -241,3 +241,115 @@ class GenericDeleteBottomParent(models.Model):
     generic_delete_bottom = models.ForeignKey(
         GenericDeleteBottom, on_delete=models.CASCADE
     )
+
+
+class Foo(models.Model):
+    """Initial model named Foo"""
+
+
+class Bar(models.Model):
```
felixxm, line 250: Please try to avoid meaningless model names (e.g. `Foo`, `Bar`, `Baz`, etc.). Maybe `Child`, `Parent`, `Grandparent`, etc.

tests/delete/models.py
```diff
@@ -241,3 +241,115 @@ class GenericDeleteBottomParent(models.Model):
...
+    )
+
+
+class DBDefaultsFK(models.Model):
+    language_code = models.ForeignKey(
+        Foo, db_default=1, on_delete=models.DB_SET_DEFAULT
+    )
+
+    class Meta:
+        managed = getattr(connection.features, "has_on_delete_db_default", False)
```
felixxm, line 355: Use `Meta.required_db_features` instead.

tests/delete/tests.py
```diff
@@ -800,3 +818,187 @@ def test_fast_delete_full_match(self):
...
+        bar = Bar.objects.create(foo=foo)
+        baz = Baz.objects.create(bar=bar)
+
+        foo.delete()
+
+        with self.assertRaises(Bar.DoesNotExist):
+            bar.refresh_from_db()
+
+        with self.assertRaises(Baz.DoesNotExist):
+            baz.refresh_from_db()
```
felixxm, line 835: Use `count()` or `exists()` instead of catching errors:
```suggestion
        bar = Bar.objects.create(foo=foo)
        baz = Baz.objects.create(bar=bar)
        foo.delete()
        self.assertIs(Bar.objects.filter(pk=bar.pk).exists(), False)
        self.assertIs(Baz.objects.filter(pk=baz.pk).exists(), False)
```

docs/ref/checks.txt
```diff
@@ -293,9 +293,15 @@ Related fields
...
-* **fields.E320**: Field specifies ``on_delete=SET_NULL``, but cannot be null.
+* **fields.E320**: Field specifies ``on_delete=<set null>``, but cannot be null.
 * **fields.E321**: The field specifies ``on_delete=SET_DEFAULT``, but has no
   default value.
+* **fields.E322**: The field specifies unsupported database level ``on_delete``
+  on model declaring a ``GenericForeignKey``.
+* **fields.E323**: Using python based on_delete with database level
+  ``on_delete`` referenced parent model is prohibited.
+* **fields.E324**: The field specifies ``on_delete=DB_SET_DEFAULT``, but has no
+  ``db_default`` value.
```
felixxm, line 304: Documented error messages should match the error messages raised.

tests/migrations/test_operations.py
```diff
@@ -179,6 +179,835 @@ def test_create_model_with_duplicate_manager_name(self):
...
+                        project_state,
+                    )
+                    self.database_backward_assertion(
+                        app_label,
+                        project_state,
+                        on_delete_type,
+                        fk_parent_model,
+                        fk_child_model,
+                        fk_fieldname,
+                    )
```
felixxm, line 729: The number of nested hooks makes these tests/assertions extremely difficult to understand (at least for me).

django/db/backends/base/features.py
```diff
@@ -137,6 +137,9 @@ class BaseDatabaseFeatures:
     # which can't do it for MyISAM tables
     can_introspect_foreign_keys = True
 
+    # Features related to database level on delete
+    has_on_delete_db_default = True
```
felixxm, line 141: This comment means nothing, TBH.
```suggestion
    # Does the backend support ...
    supports_on_delete_db_default = True
```

django/db/backends/base/schema.py
```diff
@@ -1579,6 +1582,12 @@ def _rename_index_sql(self, model, old_name, new_name):
             new_name=self.quote_name(new_name),
         )
 
+    def _create_on_delete_sql(self, model, field):
+        on_delete = getattr(field.remote_field, "on_delete", None)
```
felixxm, line 1586: `on_delete` is required, do we need to use `getattr()`?

felixxm (commented):
@Akash-Kumar-Sen THhnks :+1: I left initial comments.

falexwolf:
I just arrived here from the original 10-year-old ticket and am blown away by how far this work has gotten now.

I'm looking forward to the feature and am grateful for the work in this PR. 🙏 

Good luck with what seems like minor requests from felixxm, @Akash-Kumar-Sen!

tests/migrations/test_operations.py
Akash-Kumar-Sen: This seems to be a persistent issue, will go forward with the suggestion made by Lily [here](https://github.com/django/django/pull/16851#issuecomment-1755491297). Let me know if you approve of creating a new file for this, as there is going to be a lot of tests. Will start after few days.
>If you can't get that working, I think giving up on subTest might be reasonable - maybe create a new testcase or test file for these tests and just have all of them as distinct test methods? I know there's going to be loads of them, but maybe that's easier to write...

docs/releases/5.1.txt
```diff
@@ -29,8 +29,8 @@ What's new in Django 5.1
 Database-level values for ``ForeignKey.on_delete``
 --------------------------------------------------
 
-:attr:`ForeignKey.on_delete <django.db.models.ForeignKey.on_delete>` now
-supports four extra values to specify database-level actions:
+:attr:`.ForeignKey.on_delete` now supports four extra values to specify
```
LilyFoote, line 32: Was the added `.` intentional?
```suggestion
:attr:`ForeignKey.on_delete` now supports four extra values to specify
```

Akash-Kumar-Sen: No, sorry for the little copy paste mistake.

django/db/backends/base/schema.py
Akash-Kumar-Sen: Sometimes `field.remote_field` can be `None`, getting the following error when tried without `getattr`
```
ERROR: test_remove_constraints_capital_letters (schema.tests.SchemaTests)
#23065 - Constraint names must be quoted if they contain capital letters.
  File "/usr/lib/python3.10/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/usr/lib/python3.10/unittest/case.py", line 591, in run
  File "/usr/lib/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/home/akash/Djangocon/django/tests/schema/tests.py", line 4474, in test_remove_constraints_capital_letters
    "on_delete_db": editor._create_on_delete_sql(model, field),
  File "/home/akash/Djangocon/django/django/db/backends/base/schema.py", line 1590, in _create_on_delete_sql
    on_delete = field.remote_field.on_delete
AttributeError: 'NoneType' object has no attribute 'on_delete'

```

tests/delete/models.py
```diff
@@ -241,3 +241,99 @@ class GenericDeleteBottomParent(models.Model):
     generic_delete_bottom = models.ForeignKey(
         GenericDeleteBottom, on_delete=models.CASCADE
     )
+
```
sarahboyce, line 244: Hi Akash 👋 thank you for being patient, I have started to test this PR.

In this file there is a model called `A` which has every on_delete possible except all of these new ones.
This is actually quite a similar approach to how I am testing in my local project.
It might make sense to review these tests with the model `A` and have each recreated with the DB delete counterparts when this is not already covered by existing tests.

 

For example, I think there might be an issue with the `DB_RESTICT` and `DB_CASCADE` relationship and we can add a test for very similar to `delete.tests.OnDeleteTests.test_restrict_path_cascade_direct`.

Here:

```diff
diff --git a/tests/delete/models.py b/tests/delete/models.py
index 5a3858d7ad..33f8094b6c 100644
--- a/tests/delete/models.py
+++ b/tests/delete/models.py
@@ -58,6 +58,7 @@ class A(models.Model):
         related_name="setnull_nullable_set",
     )
     cascade = models.ForeignKey(R, models.CASCADE, related_name="cascade_set")
+    db_cascade = models.ForeignKey(R, models.DB_CASCADE, related_name="db_cascade_set")
     cascade_nullable = models.ForeignKey(
         R, models.CASCADE, null=True, related_name="cascade_nullable_set"
     )
@@ -67,6 +68,9 @@ class A(models.Model):
     restrict = models.ForeignKey(
         R, models.RESTRICT, null=True, related_name="restrict_set"
     )
+    db_restrict = models.ForeignKey(
+        R, models.DB_RESTRICT, null=True, related_name="db_restrict_set"
+    )
     donothing = models.ForeignKey(
         R, models.DO_NOTHING, null=True, related_name="donothing_set"
     )
@@ -100,9 +104,11 @@ def create_a(name):
         "setdefault",
         "setdefault_none",
         "cascade",
+        "db_cascade",
         "cascade_nullable",
         "protect",
         "restrict",
+        "db_restrict",
         "donothing",
         "o2o_setnull",
     ):
diff --git a/tests/delete/tests.py b/tests/delete/tests.py
index e9a1c09240..bd6d671b2f 100644
--- a/tests/delete/tests.py
+++ b/tests/delete/tests.py
@@ -271,6 +271,26 @@ class OnDeleteTests(TestCase):
         self.assertFalse(A.objects.filter(name="restrict").exists())
         self.assertFalse(R.objects.filter(pk=a.restrict_id).exists())

+    def test_db_restrict_path_db_cascade_direct(self):
+        a = create_a("db_restrict")
+        a.db_restrict.p = P.objects.create()
+        a.db_restrict.save()
+        a.db_cascade_p = a.db_restrict.p
+        a.save()
+        a.db_restrict.p.delete()
+        self.assertFalse(A.objects.filter(name="db_restrict").exists())
+        self.assertFalse(R.objects.filter(pk=a.db_restrict_id).exists())
+
+    def test_db_restrict_path_cascade_direct(self):
+        a = create_a("db_restrict")
+        a.db_restrict.p = P.objects.create()
+        a.db_restrict.save()
+        a.cascade_p = a.db_restrict.p
+        a.save()
+        a.db_restrict.p.delete()
+        self.assertFalse(A.objects.filter(name="db_restrict").exists())
+        self.assertFalse(R.objects.filter(pk=a.db_restrict_id).exists())
+
     def test_restrict_path_cascade_indirect_diamond(self):
         delete_top = DeleteTop.objects.create()
         b1 = B1.objects.create(delete_top=delete_top)
```

The test `test_db_restrict_path_db_cascade_direct` fails with this error on SQLite:
```
ERROR: test_db_restrict_path_db_cascade_direct (delete.tests.OnDeleteTests.test_db_restrict_path_db_cascade_direct)
  File "C:\Users\boyce\OneDrive\Documents\GitHub\django\django\db\backends\utils.py", line 105, in _execute
  File "C:\Users\boyce\OneDrive\Documents\GitHub\django\django\db\backends\sqlite3\base.py", line 354, in execute
sqlite3.IntegrityError: FOREIGN KEY constraint failed

  File "<path_to_django>\tests\delete\tests.py", line 280, in test_db_restrict_path_cascade_direct
    a.db_restrict.p.delete()
  File "<path_to_django>\django\db\models\base.py", line 1249, in delete
    return collector.delete()
           ^^^^^^^^^^^^^^^^^^
  File "<path_to_django>\django\db\models\deletion.py", line 530, in delete
    count = query.delete_batch(pk_list, self.using)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<path_to_django>\django\db\models\sql\subqueries.py", line 42, in delete_batch
    num_deleted += self.do_query(
                   ^^^^^^^^^^^^^^
  File "<path_to_django>\django\db\models\sql\subqueries.py", line 20, in do_query
    cursor = self.get_compiler(using).execute_sql(CURSOR)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<path_to_django>\django\db\models\sql\compiler.py", line 1570, in execute_sql
    cursor.execute(sql, params)
  File "<path_to_django>\django\db\backends\utils.py", line 79, in execute
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<path_to_django>\django\db\backends\utils.py", line 92, in _execute_with_wrappers
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<path_to_django>\django\db\backends\utils.py", line 100, in _execute
  File "<path_to_django>\django\db\utils.py", line 91, in __exit__
  File "<path_to_django>\django\db\backends\utils.py", line 105, in _execute
  File "<path_to_django>\django\db\backends\sqlite3\base.py", line 354, in execute
django.db.utils.IntegrityError: FOREIGN KEY constraint failed

```
But `test_db_restrict_path_cascade_direct` passes. I think I would expect the behaviour to be the same and that they would both pass.

Can you confirm this?

Akash-Kumar-Sen: Thank you for the review @sarahboyce , I think you forgot to add the `db_cascade_p ` field in the model `A`, Updated the code with suggested tests - https://github.com/django/django/pull/16851/files#diff-a842e4474a8c70b274fabec1b44e377a498181dfe6c925801b7dc930917c3e03R324-R343

sarahboyce: You're completely right thank you ⭐

sarahboyce: Looked into it again. I still think there's something going on and these tests (might) show it:

```diff
diff --git a/tests/delete/tests.py b/tests/delete/tests.py
index f84831dea7..5b08357f4d 100644
--- a/tests/delete/tests.py
+++ b/tests/delete/tests.py
@@ -331,6 +331,22 @@ class OnDeleteTests(TestCase):
         self.assertFalse(A.objects.filter(name="db_restrict").exists())
         self.assertFalse(R.objects.filter(pk=a.db_restrict_id).exists())

+    def test_db_restrict_db_cascade_direct(self):
+        a = create_a("db_restrict")
+        a.db_cascade = a.db_restrict
+        a.save()
+        a.db_restrict.delete()
+        self.assertFalse(A.objects.filter(name="db_restrict").exists())
+        self.assertFalse(R.objects.filter(pk=a.db_restrict_id).exists())
+
+    def test_db_restrict_cascade_direct(self):
+        a = create_a("db_restrict")
+        a.cascade = a.db_restrict
+        a.save()
+        a.db_restrict.delete()
+        self.assertFalse(A.objects.filter(name="db_restrict").exists())
+        self.assertFalse(R.objects.filter(pk=a.db_restrict_id).exists())
+
     def test_db_restrict_path_cascade_direct(self):
         a = create_a("db_restrict")
         a.db_restrict.p = P.objects.create()
```
I have the FK error with the `test_db_restrict_db_cascade_direct` test but not `test_db_restrict_cascade_direct`

In my local project I have:
`models.py`
```python
from django.db import models


class Basic(models.Model):
    name = models.CharField(max_length=30)


class EveryOnDelete(models.Model):
    fk_cascade = models.ForeignKey(
        Basic,
        on_delete=models.CASCADE,
        null=True,
        related_name="fk_cascade",
    )
    fk_db_cascade = models.ForeignKey(
        Basic,
        on_delete=models.DB_CASCADE,
        null=True,
        related_name="fk_db_cascade",
    )
    fk_set_null = models.ForeignKey(
        Basic,
        on_delete=models.SET_NULL,
        null=True,
        related_name="fk_set_null",
    )
    fk_db_set_null = models.ForeignKey(
        Basic,
        on_delete=models.DB_SET_NULL,
        null=True,
        related_name="fk_db_set_null",
    )
    fk_set_default = models.ForeignKey(
        Basic,
        on_delete=models.SET_DEFAULT,
        default="1",
        null=True,
        related_name="fk_set_default",
    )
    fk_db_set_default = models.ForeignKey(
        Basic,
        on_delete=models.DB_SET_DEFAULT,
        null=True,
        db_default="1",
        related_name="fk_db_set_default",
    )
    fk_restrict = models.ForeignKey(
        Basic,
        on_delete=models.RESTRICT,
        null=True,
        related_name="fk_restrict",
    )
    fk_db_restrict = models.ForeignKey(
        Basic,
        on_delete=models.DB_RESTRICT,
        null=True,
        related_name="fk_db_restrict",
    )
    fk_protect = models.ForeignKey(
        Basic,
        on_delete=models.PROTECT,
        null=True,
        related_name="fk_protect",
    )
```
and some tests
```python
from django.test import TestCase
from .models import Basic, EveryOnDelete


class DBDeleteTests(TestCase):
    @classmethod
    def setUpTestData(cls):
        Basic.objects.create(name="id-1", id="1")

# Removed some to make this smaller

    def test_db_fk_restrict_basic_deleted_allowed_cascade(self):
        basic_1 = Basic.objects.create(name="not-protected")
        EveryOnDelete.objects.create(
            fk_db_restrict=basic_1,
            fk_cascade=basic_1,
        )
        basic_1.delete()
        self.assertEqual(Basic.objects.count(), 1)
        self.assertEqual(EveryOnDelete.objects.count(), 0)

    def test_db_fk_restrict_basic_deleted_allowed_db_cascade(self):
        basic_1 = Basic.objects.create(name="not-protected")
        EveryOnDelete.objects.create(
            fk_db_cascade=basic_1,
            fk_db_restrict=basic_1,
        )
        basic_1.delete()
        self.assertEqual(Basic.objects.count(), 1)
        self.assertEqual(EveryOnDelete.objects.count(), 0)
```
Here `test_db_fk_restrict_basic_deleted_allowed_db_cascade` is failing with the FK error, I think it should pass rather than raise an `IntegrityError`?

Akash-Kumar-Sen: Got some time this week after a long rush, will look into this. Thank you for finding it out @sarahboyce

Akash-Kumar-Sen: I have looked into this issue, the current implementation seems to be like this - 

There are four scenarios regarding this:
1. The ForeignKey parent has Non DB On Delete and the Child has Non DB On Delete - Allow
2. The ForeignKey parent has DB level On Delete and the Child has DB level On Delete - Allow 
3. The ForeignKey parent has DB level On Delete and the Child has Non DB On Delete - Not Allowed
(The check should throw an error, as if the Child is deleted by internal database operations, it might lead to
Integrity Error in those tables which references the child itself)
4. The ForeignKey parent has Non DB On Delete and the Child has DB level On Delete - Allow

Now the problem arises in only in the fourth scenario, as the 3rd scenario is not allowed and the first two comes up with a very consistent approach.

Let us assume there are three models
```python
class Parent(models.Model):
    pass

class Child(models.Model):
   parent = models.ForeignKey(Parent, models.CASCADE, null=True)

class GrandChild(models.Model):
   child = models.ForeignKey(Child, models.DB_RESTRICT, null=True)
```

As the django Non DB cascade will internally delete all the related objects, it will not care about the DB Restrict just like the internal RESTRICT, so I have to cover the scenario of `DB_RESTRICT` alone and not allow deletion if any of the related objects have `DB_RESTRICT` as FK

rickydua:
Looking forward for this go out, any update on this?

quiqueporta:
Any update about this feature?

CyrusNuevoDia:
Excited to see this merged!

We have a full-stack Next app using Supabase and Django for migrations/services and the fact that we can't enforce foreign key references at the database layer means that Django's migrations are lossy. 

All in all Django >>> every NodeJS migration system (so 100x thank yous to the team) but without things like this interop becomes really hard.

anupsingh3292:
Why is this not getting merged? This feature is really useful and required. The current cascade behaviour at ORM Level creates a lot of pain and performance Issues.  @carltongibson have look 👀

ngnpope:
> Why is this not getting merged? This feature is really useful and required. The current cascade behaviour at ORM Level creates a lot of pain and performance Issues. @carltongibson have look 👀

The [last comment](https://github.com/django/django/pull/16851#discussion_r1677153399) by the author indicates that there is still some work left to do and, after more than a year, this branch is also very stale.

Please desist from "Any updates?" comments as it doesn't progress things and also don't ping individual people if you have not already been engaged in the review and development work.

If you wish to see this feature progress, then you are welcome to re-review and come up with a summary of outstanding things that need to be completed. And even continue the work to finish it.

Remember that most contributors are not paid for the features they contribute, often in their personal time, and you have no right to make demands of them.

adamsol:
It's possible to have database cascade deletion in Django globally, using a custom "engine". For example, for PostgreSQL, we can put this into `custom_postgresql/base.py`:

```py
from django.db import models
from django.db.backends.postgresql import base


class DatabaseSchemaEditor(base.DatabaseSchemaEditor):
    sql_create_column_inline_fk = None

    def _create_fk_sql(self, model, field, suffix):
        statement = super()._create_fk_sql(model, field, suffix)
        if field.remote_field.on_delete is models.deletion.CASCADE:
            statement.template = statement.template.replace('%(deferrable)s', ' ON DELETE CASCADE' + '%(deferrable)s')
        return statement


class DatabaseWrapper(base.DatabaseWrapper):
    SchemaEditorClass = DatabaseSchemaEditor
```

and reference `custom_postgresql` as `ENGINE` in `DATABASES`. This can be easily extended to support other deletion actions (`RESTRICT`, `SET NULL`, `SET DEFAULT`).

Additionally, for performance, `can_fast_delete` should return `True` most of the time, to disable Django's default behaviour:

```py
models.deletion.Collector.can_fast_delete = lambda self, objs, *args, **kwargs: hasattr(objs, '_meta') or hasattr(objs, '_raw_delete')
```

This works for all models and foreign key fields (including M2M) automatically. If you want to use it in an existing application, you need to recreate the existing constraints first (by setting `db_constraint=False` everywhere, generating a migration, reverting the change, and generating a migration again).

---

Since this pull request seems to not be moving forward, is there a reason the above simpler solution couldn't be built into Django instead?

charettes:
> Since this pull request seems to not be moving forward, is there a reason the above simpler solution couldn't be built into Django instead?

There are few ones yes, I would invite you to read the full discussions here and on [the decade old ticket](https://code.djangoproject.com/ticket/21961).

One example is that your solution completely breaks signals and `delete`/`save` method overrides which might be an acceptable compromise for your project but not something we can silently enable given the impact it would have on third-party applications as well as first party one which have overridden these methods with business logic that would now be silently circumvented.

adamsol:
Of course no-one suggests changing the logic "silently". It could be an opt-in feature, as a global setting. Signals and method overrides have many documented gotchas already, like [`QuerySet.delete`](https://docs.djangoproject.com/en/5.2/ref/models/querysets/#delete) and [`QuerySet.update`](https://docs.djangoproject.com/en/5.2/ref/models/querysets/#update), so this would be no different. Anyway, I'll keep maintaining the workaround then.

tests/delete/models.py
felixxm: Actually, it is expected that `test_db_restrict_db_cascade_direct` raises `IntegrityError` because we no longer collect and check objects for them. If they try to remove objects where it should not be possible , according to defined constraints, database will raise an error.

tests/delete/tests.py
```diff
@@ -800,3 +838,189 @@ def test_fast_delete_full_match(self):
...
+
+    def test_restrict_on_inherited_model(self):
+        gp = ParentModel.objects.create()
+        child = DBLevelChild.objects.create(grandparent_ptr=gp)
+
+        with transaction.atomic():
+            with self.assertRaises(IntegrityError):
+                gp.delete()
+
+        self.assertIs(DBLevelChild.objects.filter(pk=child.pk).exists(), True)
```
felixxm, line 1026: IMO most of these tests are unnecessary, we don't do anything specific here so testing all different models configuration are overkill. We don't want to test database behavior.

felixxm:
Closing in favor of #19925.

PR #16859, opened 2023-05-15
Fixes #34566 - ModelAdmin get_field_queryset uses related admin ordering, but not related admin querysets.

Sodium-Hydrogen:
Fix issues if related_admin has annotations in get_queryset that are used in ordering.

See [#34566](https://code.djangoproject.com/ticket/34566) for a better explanation.

### DIFF
--- a/django/contrib/admin/options.py
+++ b/django/contrib/admin/options.py
@@ -250,9 +250,7 @@ class BaseModelAdmin(metaclass=forms.MediaDefiningClass):
         if related_admin is not None:
             ordering = related_admin.get_ordering(request)
             if ordering is not None and ordering != ():
-                return related_admin.get_queryset(request).order_by(
-                    *ordering
-                )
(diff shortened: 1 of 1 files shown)


django/contrib/admin/options.py
```diff
@@ -250,9 +250,7 @@ def get_field_queryset(self, db, db_field, request):
         if related_admin is not None:
             ordering = related_admin.get_ordering(request)
             if ordering is not None and ordering != ():
-                return db_field.remote_field.model._default_manager.using(db).order_by(
-                    *ordering
-                )
+                return related_admin.get_queryset(request).order_by(*ordering)
```
payamnj, line 253: This change seems to totally disregard the 'db' parameter. So, it looks like it'll always use the default data source, which might lead to some surprises in projects with multiple databases when users want to query sources other than the default one.

PR #16861, opened 2023-05-16
Fixed #23528 -- Doc'd django.db.migrations.swappable_dependency().

DevilsAutumn:
[Ticket #23528](https://code.djangoproject.com/ticket/23528)

### DIFF
--- a/docs/topics/migrations.txt
+++ b/docs/topics/migrations.txt
@@ -218,6 +218,26 @@ Apps without migrations must not have relations (``ForeignKey``,
 ``ManyToManyField``, etc.) to apps with migrations. Sometimes it may work, but
 it's not supported.
 
+Swappable dependencies
+----------------------
+
+.. function:: django.db.migrations.swappable_dependency(value)
+
+The ``swappable_dependency()`` function is used in migrations to declare
+"swappable" dependencies on migrations in the app of the swapped-in model.
+Currently, the first migration of this app and as a consequence the swapped-in
+model should be created in the initial migration. The argument ``value`` is a
+string ``"<app label>.<model>"`` describing an app label and a model name, e.g.
+``"myapp.MyModel"``.
+
+By using ``swappable_dependency()``, you inform the migration framework that
+the migration relies on a migration which sets up up a swappable model,
+allowing for the possibility of substituting the model with a different
+implementation in the future. This is typically used for referencing models
+that are subject to customization or replacement, such as the custom user
+model (``settings.AUTH_USER_MODEL``, which defaults to ``"auth.User"``) in
+Django's authentication system.
+
 .. _migration-files:
 
 Migration files


docs/topics/migrations.txt
```diff
@@ -218,6 +218,38 @@ Apps without migrations must not have relations (``ForeignKey``,
...
+- Create a custom user model by subclassing ``django.contrib.auth.models.AbstractBaseUser``
+  and ``django.contrib.auth.models.PermissionsMixin``.
+- Define an authentication backend by subclassing ``django.contrib.auth.backends.BaseBackend``
+  and implementing the necessary methods.
+- In your project's settings file, set the ``AUTH_USER_MODEL`` setting to point to your custom
+  user model and specify your custom authentication backend in the ``AUTHENTICATION_BACKENDS`` setting.
+
+Swappable dependencies in Django extend beyond just authentication and can be used in other areas,
+such as the database backend, email backends, cache backends, and more. It provides a powerful
+mechanism for customization and integration within the Django framework.
```
felixxm, line 251: There are some issue with rendering this proposition that make it hard to review. Please wrap at 79 chars, fix rendering issues and I will review it again

image

docs/topics/migrations.txt
```diff
@@ -218,6 +218,38 @@ Apps without migrations must not have relations (``ForeignKey``,
...
+it out with a custom authentication backend by defining your own user model and
+authentication backend class. By doing so, you can customize the authentication behavior
+and integrate with external systems or alternative authentication mechanisms.
+
+To make a component swappable, Django relies on the concept of "settings" where you
+can specify the classes or modules to be used for certain functionalities. These settings
+are typically defined in your project's ``settings.py`` file. By modifying
+the relevant settings, you can swap out the default implementations with your own.
+
+Here's an example of how swappable dependencies can be used for authentication::
```
felixxm, line 240:
```suggestion
Here's an example of how swappable dependencies can be used for authentication:
```

docs/topics/migrations.txt
```diff
@@ -218,6 +218,38 @@ Apps without migrations must not have relations (``ForeignKey``,
 ``ManyToManyField``, etc.) to apps with migrations. Sometimes it may work, but
 it's not supported.
 
+Swappable dependencies
+======================
```
felixxm, line 222: I'd put it under "Dependencies":
```suggestion
Swappable dependencies
----------------------
```

docs/topics/migrations.txt
```diff
@@ -218,6 +218,38 @@ Apps without migrations must not have relations (``ForeignKey``,
...
+
+A swappable dependency is a feature that allows you to replace or customize
```
felixxm, line 224: This entire paragraph document the term _"swappable dependency"_, but not the `swappable_dependency()` function itself, which was the main goal of the ticket.

felixxm, line 224: We should document a signature
```suggestion
.. function:: django.db.migrations.swappable_dependency(value)

A swappable dependency is a feature that allows you to replace or customize
```

felixxm (commented):
@DevilsAutumn Thanks :+1: I left comments.

DevilsAutumn:
@felixxm How does it look now?
Screenshot 2023-05-19 144613

felixxm:
> @felixxm How does it look now? Screenshot 2023-05-19 144613

Better, thanks :+1: I think it's too detailed, but after your edits, I will suggest some wording improvements.

docs/topics/migrations.txt
DevilsAutumn: Docs also need to be wrapped at 79 chars? 🤔

felixxm: I'd not ask otherwise, see [contributing guide](https://docs.djangoproject.com/en/4.2/internals/contributing/writing-code/coding-style/#python-style):
> _Documentation, comments, and docstrings should be wrapped at 79 characters,..._

DevilsAutumn: fixed 👍

shaib:
There seems to be a disconnect here. This documentation is supposed to be about swappable dependencies in migrations, and that has nothing to do with cache or email backends.

A swappable dependency, in this context, does not allow you to replace "certain components of the framework", but only helps dealing with the fact that specific **models** may be replaced; and the feature under consideration here is not even about this replacement per se, but only about one aspect of how it manifests in migrations.

DevilsAutumn:
I hope its more accurate now. 🤔

docs/topics/migrations.txt
```diff
@@ -218,6 +218,28 @@ Apps without migrations must not have relations (``ForeignKey``,
...
+the migration operation relies on a swappable model, allowing for the
+possibility of substituting the model with a different implementation in the
+future. This is typically used for referencing models that are subject to
+customization or replacement, such as the custom user model
+(``AUTH_USER_MODEL``) in Django's authentication system.
+
+In summary, ``swappable_dependency()`` is a function used in migrations to
+specify dependencies on swappable models, indicating that a migration
+operation relies on a model that can be swapped with alternative
+implementations while maintaining the same interface.
```
shangxiao, line 241:
```suggestion
```
Suggestion: don't summarise 2 paragraphs; it's unnecessary repetition 🙂  It kinda reminds me of when I was a kid in school trying to pad out a 1000 word essay with fillers. 😅

docs/topics/migrations.txt
```diff
@@ -218,6 +218,28 @@ Apps without migrations must not have relations (``ForeignKey``,
...
 
+Swappable dependencies
+----------------------
+
+.. function:: django.db.migrations.swappable_dependency(value)
+
+The ``swappable_dependency()`` function is used in migration files to specify
+a swappable dependency for a particular migration operation. It is not
+directly related to swapping models but rather influences how migrations
+handle the dependency resolution process.
```
shangxiao, line 229:
```suggestion
The ``swappable_dependency()`` function is used by migration operations to declare "swappable" dependencies such as swappable user models.
```
Suggestion: be direct and to-the-point:
 - used by migration operations
 - clarify what a "swappable dependency" is with the classic example: user models
 - avoid repetition "migration files … migration operation"; simply referring to migration operation is enough to indicate its usage

Additionally: I'm not sure the second sentence is useful. It appears contradictory as the very next paragraph talks about how it's commonly used with swappable models.

shangxiao (commented):
A couple of suggestions, hope you don't mind the criticism :)

I have a particular gripe with technical documentation that that strays down the path of elegant prose.

Technical documentation should convey accurate & complete information in the most efficient way possible; ie concise & clear.

docs/topics/migrations.txt
DevilsAutumn: ouch 😿 👍

DevilsAutumn:
>  hope you don't mind the criticism :)

Not at all! I appreciate constructive criticism and value your feedback. It helps me improve and grow.

docs/topics/migrations.txt
```diff
@@ -218,6 +218,21 @@ Apps without migrations must not have relations (``ForeignKey``,
...
+
+The ``swappable_dependency()`` function is used by migration operations to
+declare "swappable" dependencies such as swappable user models.
+
+By using ``swappable_dependency()``, you inform the migration framework that
+the migration operation relies on a swappable model, allowing for the
+possibility of substituting the model with a different implementation in the
+future. This is typically used for referencing models that are subject to
+customization or replacement, such as the custom user model
+(``AUTH_USER_MODEL``) in Django's authentication system.
```
felixxm, line 234:
```suggestion
(``settings.AUTH_USER_MODEL``) in Django's authentication system.
```

shaib (changes requested):
This is much, much better. Almost there, IMO. Thanks!

There's still an issue with the function's argument, see below.

docs/topics/migrations.txt
```diff
@@ -218,6 +218,21 @@ Apps without migrations must not have relations (``ForeignKey``,
 ``ManyToManyField``, etc.) to apps with migrations. Sometimes it may work, but
 it's not supported.
 
+Swappable dependencies
+----------------------
+
+.. function:: django.db.migrations.swappable_dependency(value)
```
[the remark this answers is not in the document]
DevilsAutumn:
```diff
 The ``swappable_dependency()`` function is used by migration operations to
-declare "swappable" dependencies such as swappable user models.
+declare "swappable" dependencies such as swappable user models. It just
+makes dependencies on the first migrations of the app instead of actual
+model name.
```
Does it look better here?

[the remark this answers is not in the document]
felixxm: @DevilsAutumn As far as I'm aware, Shai suggested to describe an expected `value`, e.g.

_"... declare "swappable" dependencies such as swappable user models. The argument ``value`` is a string
``"<app label>.<model>"`` describing an app label and a model name, e.g. ``` ``"myapp.MyModel"`` ```"_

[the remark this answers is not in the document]
felixxm: > (note also -- @felixxm I'm not sure if this belongs in this documentation, or in its own bug report -- the current code ignores the actual model name, and just makes the dependency on the first migration of the app. If, for some reason, anyone tries to use a swappable dependency with a model that is only created in the second migration of its app, things may break)

Tricky, but I would process it separately.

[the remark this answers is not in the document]
DevilsAutumn: @felixxm I changed that locally, but just for more clarification, i thought of adding this line too. 😄

docs/topics/migrations.txt
```diff
@@ -218,6 +218,23 @@ Apps without migrations must not have relations (``ForeignKey``,
...
+
+The ``swappable_dependency()`` function is used by migration operations to
+declare "swappable" dependencies such as swappable user models. The argument
+``value`` is a string "<app label>.<model>" describing an app label and a model
+name, e.g. ``"myapp.MyModel"``.
```
felixxm, line 229:
```suggestion
``value`` is a string ``"<app label>.<model>"`` describing an app label and a
model name, e.g. ``"myapp.MyModel"``.
```

docs/topics/migrations.txt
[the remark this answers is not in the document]
MarkusH: > > (note also -- @felixxm I'm not sure if this belongs in this documentation, or in its own bug report -- the current code ignores the actual model name, and just makes the dependency on the first migration of the app. If, for some reason, anyone tries to use a swappable dependency with a model that is only created in the second migration of its app, things may break)
> 
> Tricky, but I would process it separately.

I disagree. The currently proposed documentation is simply wrong. The key is, if the swapped-in model isn't created as part of the initial migration of an app, the dependencies can be wrong, causing migration failures.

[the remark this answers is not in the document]
felixxm: > The phrase `declare "swappable" dependencies such as swappable user models` isn't really correct. The `swappable_dependency()` function essentially creates a dependency on the first migration in the app of the swapped-in model. That means, it essentially adds `('auth', '__first__')` if `AUTH_USER_MODEL='auth.User'` and `('accounts', '__first__')` if `AUTH_USER_MODEL='accounts.MyUser'`.

Yeah, that was also pointed out by [Shai](https://github.com/django/django/pull/16861#discussion_r1212987567). Agreed, let's be more specific here.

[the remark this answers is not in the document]
felixxm: > > > (note also -- @felixxm I'm not sure if this belongs in this documentation, or in its own bug report -- the current code ignores the actual model name, and just makes the dependency on the first migration of the app. If, for some reason, anyone tries to use a swappable dependency with a model that is only created in the second migration of its app, things may break)
> > 
> > 
> > Tricky, but I would process it separately.
> 
> I disagree. The currently proposed documentation is simply wrong. The key is, if the swapped-in model isn't created as part of the initial migration of an app, the dependencies can be wrong, causing migration failures.

I took your opinion and _"my heart is still open"_ :wink: :green_heart: Let's correct this.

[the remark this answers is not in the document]
DevilsAutumn: Screenshot 2023-06-02 154102

docs/topics/migrations.txt
```diff
@@ -218,6 +218,23 @@ Apps without migrations must not have relations (``ForeignKey``,
 ``ManyToManyField``, etc.) to apps with migrations. Sometimes it may work, but
 it's not supported.
 
+Swappable dependencies
+----------------------
+
+.. function:: django.db.migrations.swappable_dependency(value)
+
+The ``swappable_dependency()`` function is used by migration operations to
```
shaib, line 226: Dependencies are at the level of the whole migration, not single operations.

```suggestion
The ``swappable_dependency()`` function is used in migrations to
```

docs/topics/migrations.txt
```diff
@@ -218,6 +218,23 @@ Apps without migrations must not have relations (``ForeignKey``,
...
+declare "swappable" dependencies such as swappable user models. The argument
+``value`` is a string ``"<app label>.<model>"`` describing an app label and a
+model name, e.g. ``"myapp.MyModel"``.
+
+By using ``swappable_dependency()``, you inform the migration framework that
+the migration operation relies on a swappable model, allowing for the
```
shaib, line 232:
```suggestion
the migration relies on a migration which sets up a swappable model, allowing for the
```

docs/topics/migrations.txt
```diff
@@ -218,6 +218,23 @@ Apps without migrations must not have relations (``ForeignKey``,
...
+possibility of substituting the model with a different implementation in the
+future. This is typically used for referencing models that are subject to
+customization or replacement, such as the custom user model
+(``settings.AUTH_USER_MODEL``) in Django's authentication system.
```
shaib, line 236: I think the distinction needs to be made -- I think `settings.AUTH_USER_MODEL` looks close enough to the `<app>.<Model>` pattern, that if we don't hammer this point in, some people will be confused.

```suggestion
(``settings.AUTH_USER_MODEL``, which defaults to ``"auth.User"``) in Django's authentication system.
```

docs/topics/migrations.txt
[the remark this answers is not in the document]
felixxm: There is no need to push the `value` description to a note. Also, please push edits instead of images, they are not really reviewable.

shaib (commented):
These are now details. The basics are in, up to the deeper discussion about the first-migration-of-app issue.

docs/topics/migrations.txt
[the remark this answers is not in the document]
shaib: > > (note also -- @felixxm I'm not sure if this belongs in this documentation, or in its own bug report -- the current code ignores the actual model name, and just makes the dependency on the first migration of the app. If, for some reason, anyone tries to use a swappable dependency with a model that is only created in the second migration of its app, things may break)
> 
> Tricky, but I would process it separately.

https://code.djangoproject.com/ticket/34623

docs/topics/migrations.txt
```diff
@@ -218,6 +218,26 @@ Apps without migrations must not have relations (``ForeignKey``,
...
+Swappable dependencies
+----------------------
+
+.. function:: django.db.migrations.swappable_dependency(value)
+
+The ``swappable_dependency()`` function is used in migrations to
+declare "swappable" dependencies on the first migration in the app of the
+swapped-in model. If the swapped-in model isn't created as part of the initial
+migration of an app, the dependencies can be wrong, causing migration
+failures. The argument ``value`` is a string ``"<app label>.<model>"``
```
felixxm, line 230: I'd remove _"If the swapped-in model isn't created as part of the initial migration of an app, the dependencies can be wrong, causing migration failures."_ or change it to e.g. _"Therefore, swapped-in model should be included in the initial migration."_.

We documented how it works, IMO, we don't need to scary users, and it may not be fixable, see ticket-23694.

docs/topics/migrations.txt
```diff
@@ -218,6 +218,26 @@ Apps without migrations must not have relations (``ForeignKey``,
...
+
+The ``swappable_dependency()`` function is used in migrations to
+declare "swappable" dependencies on the first migration in the app of the
+swapped-in model. Therefore, swapped-in model should be included in the
+initial migration. The argument ``value`` is a string
+``"<app label>.<model>"`` describing an app label and a model name,
+e.g. ``"myapp.MyModel"``.
+
+By using ``swappable_dependency()``, you inform the migration framework that
+the migration relies on a migration which setups up a swappable model,
```
shaib, line 234:
```suggestion
the migration relies on a migration which sets up a swappable model,
```

docs/topics/migrations.txt
```diff
@@ -218,6 +218,26 @@ Apps without migrations must not have relations (``ForeignKey``,
...
 
+Swappable dependencies
+----------------------
+
+.. function:: django.db.migrations.swappable_dependency(value)
+
+The ``swappable_dependency()`` function is used in migrations to
+declare "swappable" dependencies on the first migration in the app of the
+swapped-in model. Therefore, swapped-in model should be included in the
+initial migration. The argument ``value`` is a string
```
shaib, line 229: This suggestion includes some grammatical improvement, and some expression of my hope that ticket-23694 or ticket-34623 may yet be fixed.

```suggestion
declare "swappable" dependencies on migrations in the app of the
swapped-in model (currently, the first migration of this app; therefore, 
the swapped-in model should be created in the
initial migration). The argument ``value`` is a string
```

shaib (commented):
Minor phrasing edits

docs/topics/migrations.txt
felixxm: Agreed. I'd just try to avoid `()` and `;`, e.g.
```suggestion
declare "swappable" dependencies on migrations in the app of the swapped-in
model. Currently, the first migration of this app and as a consequence the
swapped-in model should be created in the initial migration). The argument
``value`` is a string ...
```

shaib (approved):
Yes, works for me. Thanks!

felixxm (approved):
@DevilsAutumn Thanks for updates :+1:
@shaib @MarkusH Thanks for reviews :star:

docs/topics/migrations.txt
```diff
@@ -218,6 +218,26 @@ Apps without migrations must not have relations (``ForeignKey``,
...
 
+Swappable dependencies
+----------------------
+
+.. function:: django.db.migrations.swappable_dependency(value)
+
+The ``swappable_dependency()`` function is used in migrations to declare
+"swappable" dependencies on migrations in the app of the swapped-in model.
+Currently, the first migration of this app and as a consequence the swapped-in
+model should be created in the initial migration. The argument ``value`` is a
```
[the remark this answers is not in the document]
felixxm: Ahh, right. That's my fault.

felixxm:
@smithdc1 I submitted follow up PR, see #16936.

PR #16864, opened 2023-05-17
Refs #15727 -- Added support for Content-Security-Policy (CSP) to SecurityMiddleware.

Anv3sh:
Added support for Content Security Policy(CSP) to SecurityMiddleware. It is still a WIP and I haven't added system checks yet. Please provide suggestions. 😄


nessita:
Hello @Anv3sh. Thank you for this work, below a couple of very high level questions:

1- Did you base your proposal on `django-csp`? If yes, what parts did you bring up here and why?
2- What kind of (manual?) testing you performed over the proposed code?
3- What's your plan regarding unit tests and documentation?

Thanks!

nessita:
@adamchainz @ubernostrum Since you were listed as possible GSoC mentors for this feature, I was wondering if you were interested in taking a look at this proposal. Thanks!

Anv3sh:
> 

Hey @nessita thanks for the quick reply.The proposed changes are based on my [GSOC'23 proposal](https://gist.github.com/Anv3sh/b0c130c1cd8a1996009954dd9f99586f). As for the manual testing I tested it on a sample project where it worked fine.

adamchainz:
I would prefer to see a separate CSP middleware. CSP is quite complicated and it would be best to separate the implementation from the other security features.

I think the two setting design, with `SECURE_CSP` and `SECURE_CSP_MULTIPLE`, is quite user unfriendly.

I would prefer an approach where we base the design on what's working in django-csp, but try to make it match Django's existing settings more closely. Something like the Clickjacking Protection ( https://docs.djangoproject.com/en/4.2/ref/clickjacking/ ) is what I imagined: a standalone feature with its own middleware, settings, and documentation page.

Another thing I'd like is the ability to specify separate `Content-Security-Policy` and `Content-Security-Policy-Report-Only` headers. The idea is that you can test out a new iteration of your security settings by configuring the report-only variant, whilst leaving the existing version deployed in the main header.

Anv3sh:
> I would prefer to see a separate CSP middleware. CSP is quite complicated and it would be best to separate the implementation from the other security features.
> 
> I think the two setting design, with `SECURE_CSP` and `SECURE_CSP_MULTIPLE`, is quite user unfriendly.
> 
> I would prefer an approach where we base the design on what's working in django-csp, but try to make it match Django's existing settings more closely. Something like the Clickjacking Protection ( https://docs.djangoproject.com/en/4.2/ref/clickjacking/ ) is what I imagined: a standalone feature with its own middleware, settings, and documentation page.
> 
> Another thing I'd like is the ability to specify separate `Content-Security-Policy` and `Content-Security-Policy-Report-Only` headers. The idea is that you can test out a new iteration of your security settings by configuring the report-only variant, whilst leaving the existing version deployed in the main header.

I agree with the provision to have a separate middleware for CSP because of how vastly complicated it is, I went with the current implementation as there were suggestions by [Tim Graham](https://github.com/django/django/pull/5776#issuecomment-163002994) and [Florian Apolloner](https://github.com/django/django/pull/5776#issuecomment-162289640).

jacobtylerwalls:
Thanks for advancing the conversation around CSP. Future work is happening in #18215

PR #16868, opened 2023-05-17
changed darkened-bg color

WELLWHADYAKNOW:
Fixed #ticket1313
Changed background color for warning admonition.


nessita:
Hello, it seems that this PR was created by mistake. The referenced ticket is already closed and the proposed diff has nothing to do with it.

PR #16874, opened 2023-05-19
Refs #31949 -- Made @no_append_slash decorator to work with async functions.

LomaxOnTheRun:

This PR makes the `no_append_slash` decorator able to handle both sync and async views.

### DIFF
--- a/django/views/decorators/common.py
+++ b/django/views/decorators/common.py
@@ -1,5 +1,7 @@
 from functools import wraps
 
+from asgiref.sync import iscoroutinefunction
+
 
 def no_append_slash(view_func):
     """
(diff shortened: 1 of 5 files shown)


django/views/decorators/common.py
```diff
@@ -9,9 +11,19 @@ def no_append_slash(view_func):
...
-    def wrapper_view(*args, **kwargs):
-        return view_func(*args, **kwargs)
 
-    wrapper_view.should_append_slash = False
-    return wrapper_view
+    if iscoroutinefunction(view_func):
+
+        async def _view_wrapper(request, *args, **kw):
+            response = await view_func(request, *args, **kw)
+            return response
```
LomaxOnTheRun, line 19: 🤔 I wasn't sure if this would be neater than the single line option:

```py
return await view_func(request, *args, **kw)
```

I think the more verbose version makes the code a little clearer, but I'm not sure if one less variable assignment is a bigger benefit.

LomaxOnTheRun: Made it into a single line to keep the styling in line with other decorator styles.

felixxm (approved):
@LomaxOnTheRun Thanks :+1:

PR #16880, opened 2023-05-20
Update overview.txt

sumitgirwal:
Correcting "it's" To "it has"
In the line "The data-model syntax offers many rich ways of representing your models – so far, it’s been solving many years’ worth of database-schema problems," "it's" should be "it has."


ngnpope:
Sorry, but "it's" is a valid contraction of "it has".

sumitgirwal:
Apologies for the confusion. You are correct, "it's" is a valid contraction of "it has." 
I appreciate you pointing that out. Thank you for your time!

PR #16894, opened 2023-05-24
Refs #34593 -- Commented that the extra count is necessary in ChangeList.get_results().

shangxiao:
Ref: https://github.com/django/django/pull/16891#discussion_r1202715868

### DIFF
--- a/django/contrib/admin/views/main.py
+++ b/django/contrib/admin/views/main.py
@@ -312,6 +312,9 @@ class ChangeList:
         result_count = paginator.count
 
         # Get the total number of objects, with no admin filters applied.
+        # Note this isn't necessarily the same as result_count in the case of
+        # no filtering. Filters defined in list_filters may still apply some
+        # default filtering which may be removed with query parameters.
         if self.model_admin.show_full_result_count:
             full_result_count = self.root_queryset.count()
         else:


django/contrib/admin/views/main.py
```diff
@@ -312,6 +312,9 @@ def get_results(self, request):
         result_count = paginator.count
 
         # Get the total number of objects, with no admin filters applied.
+        # Note this isn't necessarily the same as result_count in the case of
+        # no filtering. Filters defined in list_filters may still apply some
+        # default filtering which may be removed with query parameters.
```
nessita, line 317: Do you think future readers could benefit from an explicit: "The apparent query duplication when counting results can not be simplified since...."?

nessita (commented):
Thanks!

django/contrib/admin/views/main.py
shangxiao: Not really I feel this is clear & concise explanation that they're not (necessarily) equal without elaboration, which is how comments should be imho 😋

nessita: Fair enough, thanks!

nessita (approved):
LGTM

PR #16896, opened 2023-05-25
Reorganized tutorial's part 4 to better understand changes needed in URLConf.

nessita:
This work follows comments from PR #16822 and tries to ease the reader's flow and the understanding of the required changes across views and urls modules.

### DIFF
--- a/docs/intro/tutorial04.txt
+++ b/docs/intro/tutorial04.txt
@@ -215,8 +215,12 @@ the database according to a parameter passed in the URL, loading a template and
 returning the rendered template. Because this is so common, Django provides a
 shortcut, called the "generic views" system.
 
-Generic views abstract common patterns to the point where you don't even need
-to write Python code to write an app.
+Generic views abstract common patterns to the point where you don't even need to
+write Python code to write an app. For example, the
+:class:`~django.views.generic.list.ListView` and
+:class:`~django.views.generic.detail.DetailView` generic views
+abstract the concepts of "display a list of objects" and
+"display a detail page for a particular type of object" respectively.
 
 Let's convert our poll app to use the generic views system, so we can delete a
 bunch of our own code. We'll have to take a few steps to make the conversion.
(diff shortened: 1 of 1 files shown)


nessita:
@kezabelle Would you mind reviewing? Thanks!

shangxiao:
I remember that PR 🙂

When I read that you wanted to swap the order around I thought "great idea!" but assumed you meant swap the entire sections so that the updates to the URL conf were going come last. 🤔

docs/intro/tutorial04.txt
```diff
@@ -303,19 +312,8 @@ views and use Django's generic views instead. To do so, open the
     def vote(request, question_id):
         ...  # same as above, no changes needed.
 
-We're using two generic views here:
```
fhightower, line 306: I appreciate that the content on this line was moved earlier in this document as I think it helps readability. Do you think it would be worthwhile to remind the reader of what each template does here too?

I can see such a reminder being helpful for readers who jump back to this specific section without reading any previous material. I know some people (myself included) will jump back to a specific section of a tutorial when they need to be reminded of how to accomplish some task in a new framework - in these cases, I think it would be helpful to have explanations of the generic views we are using in the examples (`ListView` and `DetailView`) near the examples themselves.

fhightower (commented):
Thanks for creating this PR and I apologize for being so slow to follow up after creating my [original PR](https://github.com/django/django/pull/16822).

FWIW, I think this change is helpful - especially because it clarifies why we change `<question_id>` to `<pk>` without modifying any of the example code bocks.

I've shared one comment below for your consideration. Thanks!

docs/intro/tutorial04.txt
nessita: Thank you for the review and comment.

I have made some changes towards your request, though I haven't made explicit mention of the views' goals to avoid duplication, since the same section that holds this code has the views' descriptions (the section is "Use generic views: Less code is better").

I extended the paragraph right after the code with some extra details. Then, the following paragraph mentions the views again, *and* it provides links to their definition. So I would think that's a good compromise? After my latest push, this piece of docs looks like this:

image

What do you think?

fhightower: Thanks for the update!

That's very reasonable and LGTM.

nessita: Thanks! Merging then.

PR #16900, opened 2023-05-26
 Fixed #30355 -- Doc'd interaction between custom managers and prefetch_related().

Akash-Kumar-Sen:
[ticket #30355](https://code.djangoproject.com/ticket/30355)

### DIFF
--- a/docs/topics/db/queries.txt
+++ b/docs/topics/db/queries.txt
@@ -1790,6 +1790,18 @@ methods::
 
     b.entry_set(manager="entries").is_published()
 
+.. admonition:: Interaction with prefetching
+
+    When calling :meth:`~django.db.models.query.QuerySet.prefetch_related` with
+    a reverse relation, the default manager will be used. If you want to
+    prefetch related objects using a custom reverse manager, use
+    :class:`Prefetch() <django.db.models.Prefetch>`. For example::
+
+        from django.db.models import Prefetch
+
+        prefetch_manager = Prefetch("entry_set", queryset=Entry.entries.all())
+        Blog.objects.prefetch_related(prefetch_manager)
+
 Additional methods to handle related objects
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 


nessita:
Thank you @Akash-Kumar-Sen for this work. I have squashed commits, tweaked the text a little bit and changed the display to be an admonition:

image

I'll leave this PR here for a few days to see if other reviewers prefer other wording, otherwise I'll land before the week ends.

smithdc1:
Thank you @Akash-Kumar-Sen for picking this up and pushing it forward. 👍

I'm slightly nervous this sounds like we're just documenting a bug / limitation of Django, but I'm fairly sure this isn't the case. I'm left with more questions than answers. 

Could we add a little more to explain either why this is the correct behaviour or what the other means are to work around this are? (comments on ticket indicate there are, but it's not clear to me what that could be). 

Is it possible to write something like? 

> Using `prefetch_related` uses the default manager because ... therefore use ... instead.

felixxm:
> I'm slightly nervous this sounds like we're just documenting a bug / limitation of Django, but I'm fairly sure this isn't the case. I'm left with more questions than answers.

I have the same feelings. The current behavior is what it is. I wouldn't document it as unexpected, maybe:
```
``QuerySet.prefetch_related()`` on a reverse relation always uses the default manager not a custom one.  ...
```

Akash-Kumar-Sen:
@smithdc1  The possible explanation here is that `queryset` generated with prefetch does not modify the `queryset` or make a DB query when we are fetching the reviews, this is the sole purpose of prefetch, so if we try to use a custom manager with prefetch that will violate the purpose of using prefetch, hitting 1 extra query every time we try to fetch the reviews.
I will update the docs with this explanation for better understanding along with the alternative way using `from django.db.models import Prefetch`.

Attaching the testcase for better understanding: https://github.com/Akash-Kumar-Sen/django/blob/ticket_30355_3/tests/dummy_tests/tests.py

docs/topics/db/queries.txt
```diff
@@ -1790,6 +1790,14 @@ methods::
 
     b.entry_set(manager="entries").is_published()
 
+.. admonition:: Interaction with prefetching
+
+   :meth:`~django.db.models.query.QuerySet.prefetch_related` on a reverse
+   relation always uses the default manager. Therefore, using a custom manager
+   in conjunction with  
+   :meth:`~django.db.models.query.QuerySet.prefetch_related`
+   will not produce expected results.
```
felixxm, line 1799: We should document an alternative instead of documenting this as a bug, e.g.
```
Use :class:`Prefetch() <django.db.models.Prefetch>`, if you want to prefetch
related objects using a custom reverse manager. For example::

    from django.db.models import Prefetch

    prefetch_manager = Prefetch("entry_set", queryset=Entry.entries.all())
    Blog.objects.prefetch_related(prefetch_manager)
```

felixxm: ... so I'd remove _"Therefore, using a custom manager ... will not produce expected results."_ and add something similar to the :point_up:

felixxm:
@Akash-Kumar-Sen Thanks :+1: I pushed small formatting edits.

PR #16908, opened 2023-05-31
Create django.yml

n6wbb6:


nessita:
This seems to be a PR created by mistake, so I'll close it. If that's not the case, please create a ticket report in the Django trac instance with details about what you are trying to accomplish. Thanks!

PR #16927, opened 2023-06-02
Fixed #34622 -- Improved accessibility of related widget wrapper in admin.

allcaps:
Ref: https://code.djangoproject.com/ticket/34622

# After

- Images inside the links have empty alt.
- `aria-disabled` is toggled.

# Disabled state
Screenshot 2023-06-02 at 12 17 39

# Enabled state
Screenshot 2023-06-02 at 12 15 24

### DIFF
--- a/django/contrib/admin/static/admin/js/admin/RelatedObjectLookups.js
+++ b/django/contrib/admin/static/admin/js/admin/RelatedObjectLookups.js
@@ -79,9 +79,11 @@
             siblings.each(function() {
                 const elm = $(this);
                 elm.attr('href', elm.attr('data-href-template').replace('__fk__', value));
+                elm.removeAttr('aria-disabled');
             });
         } else {
             siblings.removeAttr('href');
+            siblings.attr('aria-disabled', true);
         }
     }
 
--- a/django/contrib/admin/templates/admin/widgets/related_widget_wrapper.html
+++ b/django/contrib/admin/templates/admin/widgets/related_widget_wrapper.html
@@ -9,7 +9,7 @@
             data-href-template="{{ change_related_template_url }}?{{ url_params }}"
             data-popup="yes"
             title="{% blocktranslate %}Change selected {{ model }}{% endblocktranslate %}">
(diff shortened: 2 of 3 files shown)


felixxm (commented):
@allcaps Thanks :+1: Tests are missing.

thibaudcolas (approved):
Thank you @allcaps, tested in desktop Safari VoiceOver, all working as expected (disabled links are skipped like disabled form controls would be).

There are no JS unit tests that I can see for RelatedObjectLookups.js. I’m not quite sure how you’d go about writing tests for empty `alt=""` – perhaps others will have suggestions.

allcaps:
What tests do we want? 

- Maybe a Python test, to assert the alt is empty?
- Maybe a JS test, to assert if the `aria-disabled="true"` is set? (I need some guidance here)
- Maybe both?

@felixxm What do you think?

thibaudcolas:
I can see three options for tests:

- Selenium. This seems the most solid if we can find an instance of `RelatedFieldWidgetWrapper`. We can use this to test both `aria-disabled`, and `alt=""`.
- QUnit JS tests. Though `RelatedObjectLookups.js` is loaded there, we don’t seem to have [any unit tests](https://github.com/django/django/blob/main/js_tests/tests.html#L127) for this component. So we could probably add some. This is quite a lot of work compared to the field involved there though – we’d need to write a JS-hardcoded version of [related_widget_wrapper.html](https://github.com/django/django/blob/main/django/contrib/admin/templates/admin/widgets/related_widget_wrapper.html) so we can test this with the correct HTML.
- Python tests, to check `alt=""`

allcaps:
I'll go with Selenium to test it all in one go. Gotta run, so'll come back
to this next week.

allcaps:
Tests are added, see `tests/admin_views/test_related_object_lookups.py`

felixxm:
buildbot, test on selenium.

felixxm:
buildbot, test on selenium.

felixxm:
@allcaps Thanks :+1:

PR #16929, opened 2023-06-02
Fixed #34034 -- Allow setting HTML attributes on Select widget options

khink:
https://code.djangoproject.com/ticket/34034

### DIFF
--- a/django/contrib/admin/widgets.py
+++ b/django/contrib/admin/widgets.py
@@ -490,6 +490,7 @@ class AutocompleteMixin:
         self.db = using
         self.choices = choices
         self.attrs = {} if attrs is None else attrs.copy()
+        self.option_attrs = {}
         self.i18n_name = get_select2_language()
 
     def get_url(self):
--- a/django/forms/widgets.py
+++ b/django/forms/widgets.py
@@ -618,12 +618,13 @@ class ChoiceWidget(Widget):
     checked_attribute = {"checked": True}
     option_inherits_attrs = True
 
-    def __init__(self, attrs=None, choices=()):
+    def __init__(self, attrs=None, choices=(), option_attrs=None):
         super().__init__(attrs)
         # choices can be any iterable, but we may need to render this widget
         # multiple times. Thus, collapse it into a list so it can be consumed
         # more than once.
         self.choices = list(choices)
+        self.option_attrs = {} if option_attrs is None else option_attrs.copy()
 
     def __deepcopy__(self, memo):
         obj = copy.copy(self)
@@ -689,9 +690,14 @@ class ChoiceWidget(Widget):
         self, name, value, label, selected, index, subindex=None, attrs=None
     ):
         index = str(index) if subindex is None else "%s_%s" % (index, subindex)
-        option_attrs = (
-            self.build_attrs(self.attrs, attrs) if self.option_inherits_attrs else {}
-        )
+
+        if self.option_attrs:
+            option_attrs = self.build_attrs(self.option_attrs)
+        elif self.option_inherits_attrs:
+            option_attrs = self.build_attrs(self.attrs, attrs)
+        else:
+            option_attrs = {}
+
         if selected:
             option_attrs.update(self.checked_attribute)
         if "id" in option_attrs:
--- a/tests/forms_tests/widget_tests/test_select.py
+++ b/tests/forms_tests/widget_tests/test_select.py
@@ -111,6 +111,28 @@ class SelectTest(WidgetTest):
             ),
         )
 
+    def test_constructor_option_attrs(self):
+        """
+        Select options shouldn't inherit the parent widget attrs.
+        """
+        widget = Select(
+            attrs={"class": "super", "id": "super"},
+            option_attrs={"data-test": "custom", "class": "other"},
+            choices=[(1, 1), (2, 2), (3, 3)],
+        )
+        self.check_html(
+            widget,
+            "num",
+            2,
+            html=(
+                """<select name="num" class="super" id="super">
+              <option value="1" data-test="custom" class="other">1</option>
+              <option value="2" data-test="custom" class="other" selected>2</option>
+              <option value="3" data-test="custom" class="other">3</option>
+            </select>"""
+            ),
+        )
+
     def test_compare_to_str(self):
         """
         The value is compared to its str().
@@ -332,13 +354,32 @@ class SelectTest(WidgetTest):
         self.assertEqual(options[0]["value"], "J")
         self.assertEqual(options[0]["label"], "John")
         self.assertEqual(options[0]["index"], "0")
+        self.assertEqual(options[0]["attrs"], {"selected": True})
         self.assertIs(options[0]["selected"], True)
         # Template-related attributes
         self.assertEqual(options[1]["name"], "name")
         self.assertEqual(options[1]["value"], "P")
         self.assertEqual(options[1]["label"], "Paul")
         self.assertEqual(options[1]["index"], "1")
+        self.assertEqual(options[1]["attrs"], {})
         self.assertIs(options[1]["selected"], False)
+
+    def test_options_with_option_attrs(self):
+        options = list(
+            self.widget(choices=self.beatles, option_attrs={"class": "other"}).options(
+                "name",
+                ["J"],
+                attrs={"class": "super"},
+            )
+        )
+        self.assertEqual(len(options), 4)
+        for option, (i, (value, label)) in zip(options, enumerate(self.beatles)):
+            self.assertEqual(option["name"], "name")
+            self.assertEqual(option["value"], value)
+            self.assertEqual(option["label"], label)
+            self.assertEqual(option["index"], str(i))
+            self.assertEqual(option["attrs"]["class"], "other")
+            self.assertIs(option["selected"], value == "J")
 
     def test_optgroups(self):
         choices = [
--- a/docs/ref/forms/widgets.txt
+++ b/docs/ref/forms/widgets.txt
@@ -195,6 +195,18 @@ Django will then include the extra attributes in the rendered output:
 You can also set the HTML ``id`` using :attr:`~Widget.attrs`. See
 :attr:`BoundField.id_for_label` for an example.
 
+For widgets that have options on their own, like those inheriting from :class:`ChoiceWidget`,
+you can also pass option-specific attributes using using `option_attrs`.
+For example, if you wish to set custom CSS classes for a RadioSelect's options, you could
+do as follows:
+
+.. code-block:: pycon
+
+    >>> from django.forms.widgets import RadioSelect
+    >>> widget = RadioSelect(choices=(("J", "John"),), option_attrs={"class": "special"})
+    >>> widget.render(name="beatle", value=["J"])
+    '<div><div>\n    <label><input type="radio" name="beatle" value="J" class="special" checked>\n John</label>\n\n</div>\n</div>'
+
 .. _styling-widget-classes:
 
 Styling widget classes
--- a/tests/forms_tests/widget_tests/test_radioselect.py
+++ b/tests/forms_tests/widget_tests/test_radioselect.py
@@ -200,6 +200,38 @@ class RadioSelectTest(WidgetTest):
         """,
         )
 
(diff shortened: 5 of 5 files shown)


nessita:
Hello @khink, thanks for this work! Could you fix the code style issues reported by the failed "Linters / black" test?

Alternatively, you could automate the code formatting in your repo by configuring the [pre-commit checks](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#pre-commit-checks).

django/forms/widgets.py
```diff
@@ -618,12 +618,13 @@ class ChoiceWidget(Widget):
...
     option_inherits_attrs = True
 
-    def __init__(self, attrs=None, choices=()):
+    def __init__(self, attrs=None, choices=(), option_attrs=None):
         super().__init__(attrs)
         # choices can be any iterable, but we may need to render this widget
         # multiple times. Thus, collapse it into a list so it can be consumed
         # more than once.
         self.choices = list(choices)
+        self.option_attrs = option_attrs
```
nessita, line 627: I think this should mimic what `Widget` does in `__init__` for `attrs`:

```suggestion
        self.option_attrs = {} if option_attrs is None else option_attrs.copy()
```

nessita (changes requested):
Great start!

I think this is on the right path but I noticed that we need to be more generous with the tests that are added. Particularly, I don't think this proposal is completely testing all the branches `if self.option_attrs: ... `elif self.option_inherits_attrs: ...`. More so, we should have tests for this change on every widget that inherits from `ChoiceWidget` (those are `Select` and `RadioSelect`).

The `Select` widget particularly sets `option_inherits_attrs = False` and there are no new tests for this class. As an example, I've extended some checks in `test_options` and added a new `test_options_with_option_attrs` to make the tests more robust (you can check how these fail if some of the new code is removed):

```diff
--- a/tests/forms_tests/widget_tests/test_select.py
+++ b/tests/forms_tests/widget_tests/test_select.py
@@ -111,6 +111,28 @@ class SelectTest(WidgetTest):
             ),
         )

+    def test_constructor_option_attrs(self):
+        """
+        Select options shouldn't inherit the parent widget attrs.
+        """
+        widget = Select(
+            attrs={"class": "super", "id": "super"},
+            option_attrs={"data-test": "custom", "class": "other"},
+            choices=[(1, 1), (2, 2), (3, 3)],
+        )
+        self.check_html(
+            widget,
+            "num",
+            2,
+            html=(
+                """<select name="num" class="super" id="super">
+              <option value="1" data-test="custom" class="other">1</option>
+              <option value="2" data-test="custom" class="other" selected>2</option>
+              <option value="3" data-test="custom" class="other">3</option>
+            </select>"""
+            ),
+        )
+
     def test_compare_to_str(self):
         """
         The value is compared to its str().
@@ -332,14 +354,33 @@ class SelectTest(WidgetTest):
         self.assertEqual(options[0]["value"], "J")
         self.assertEqual(options[0]["label"], "John")
         self.assertEqual(options[0]["index"], "0")
+        self.assertEqual(options[0]["attrs"], {"selected": True})
         self.assertIs(options[0]["selected"], True)
         # Template-related attributes
         self.assertEqual(options[1]["name"], "name")
         self.assertEqual(options[1]["value"], "P")
         self.assertEqual(options[1]["label"], "Paul")
         self.assertEqual(options[1]["index"], "1")
+        self.assertEqual(options[1]["attrs"], {})
         self.assertIs(options[1]["selected"], False)

+    def test_options_with_option_attrs(self):
+        options = list(
+            self.widget(choices=self.beatles, option_attrs={"class": "other"}).options(
+                "name",
+                ["J"],
+                attrs={"class": "super"},
+            )
+        )
+        self.assertEqual(len(options), 4)
+        for option, (i, (value, label)) in zip(options, enumerate(self.beatles)):
+            self.assertEqual(option["name"], "name")
+            self.assertEqual(option["value"], value)
+            self.assertEqual(option["label"], label)
+            self.assertEqual(option["index"], str(i))
+            self.assertEqual(option["attrs"]["class"], "other")
+            self.assertIs(option["selected"], value == "J")
+
     def test_optgroups(self):
         choices = [
             (
```

khink:
Thanks for the suggestions, I'll take a look next week!

django/contrib/admin/widgets.py
```diff
@@ -580,8 +580,8 @@ def media(self):
 
 
 class AutocompleteSelect(AutocompleteMixin, forms.Select):
-    pass
+    option_attrs = None
```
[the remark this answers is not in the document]
khink: The test `admin_views.tests.AdminViewBasicTest.test_trailing_slash_required` broke because `AutocompleteSelect` had no attribute `option_attrs`. 

But after some thinking, i think it's better to solve this in `ChoiceWidget.create_option() ` as i have done now:
```python
        if not hasattr(self, "option_attrs"):
            # AutocompleteSelect subclasses AutocompleteMixin first, and then Select
            # (which subclasses this class).
            # So the ChoiceWidget's __init__ method (which is how other Select classes
            # get their option_attrs set) is not called.
            # But AutocompleteSelect does call create_option, so we need to allow for
            # a missing option_attrs attribute.
            option_attrs = {}
```

khink:
@nessita Thanks for your remarks. I have copied your patch as-is (see last commit) for easy reviewing, thanks for this. I agree all widgets need to be tested. Do you think the commits should be squashed at some point? 

And is there any place where we can see which lines in a PR are covered by tests, and which are not?

Looking forward to your reply!

django/forms/widgets.py
```diff
@@ -689,9 +690,22 @@ def create_option(
...
-            self.build_attrs(self.attrs, attrs) if self.option_inherits_attrs else {}
-        )
+
+        if not hasattr(self, "option_attrs"):
+            # AutocompleteSelect subclasses AutocompleteMixin first, and then Select
+            # (which subclasses this class).
+            # So the ChoiceWidget's __init__ method (which is how other Select classes
+            # get their option_attrs set) is not called.
+            # But AutocompleteSelect does call create_option, so we need to allow for
+            # a missing option_attrs attribute.
```
shangxiao, line 700: All this can be simplified to just say that subclasses have mixins which do not call `super().__init__()` but do call `create_option()` … or something to that effect… :)

django/forms/widgets.py
```diff
@@ -689,9 +690,22 @@ def create_option(
...
+            option_attrs = {}
+        elif self.option_attrs:
+            option_attrs = self.build_attrs(self.option_attrs)
+        elif self.option_inherits_attrs:
+            option_attrs = self.build_attrs(self.attrs, attrs)
+        else:
+            option_attrs = {}
```
shangxiao, line 707: There are 3 different places that `option_attrs` is defaulted to `{}`… which may indicate a different approach is needed.

django/forms/widgets.py
shangxiao: To nessita, felix or carlton: can we not make `AutocompleteMixin` just honour the contract of `__init__()`? It's rare to have a mixin break the contract like this and I feel like if subclasses & mixins want to do this then they must behave properly.

shangxiao (changes requested):
Thanks 🏆 😊

Personally I don't think this class should bend over backwards to accommodate subclasses. The subclasses must fall into line, but let's see what others think.

nessita:
> Do you think the commits should be squashed at some point?

Yes, commits should be eventually squashed into one or perhaps two commits, depending on logical units (it's usually only one commit).

> And is there any place where we can see which lines in a PR are covered by tests, and which are not?

I'm not sure, what I do is comment out the code block that look suspicious, and I run the tests. If no test fail, then that's a clear indication that the commented code is either unnecessary or is untested.

I'll re-review before this week ends!

shangxiao:
> > And is there any place where we can see which lines in a PR are covered by tests, and which are not?
> 
> I'm not sure, what I do is comment out the code block that look suspicious, and I run the tests. If no test fail, then that's a clear indication that the commented code is either unnecessary or is untested.

You can use coverage.py for this, here are the instructions: https://docs.djangoproject.com/en/4.2/internals/contributing/writing-code/unit-tests/#code-coverage

khink:
I've shortened the comment as suggested by @shangxiao.

khink:
Hi @nessita and @shangxiao, is there anything else i can to to further the progress on this?

shangxiao:
> Hi @nessita and @shangxiao, is there anything else i can to to further the progress on this?

For me, I think [this question](https://github.com/django/django/pull/16929#discussion_r1218919305) would need checking

khink:
@shangxiao Thanks, that would be (please correct me if i'm wrong):
> To nessita, felix or carlton: can we not make AutocompleteMixin just honour the contract of __init__()? It's rare to have a mixin break the contract like this and I feel like if subclasses & mixins want to do this then they must behave properly.

I think @nessita @felixxm or @carltongibson would be the ones to comment on that.

If there's anything i myself can do to further progress on this issue, please let me know.

shangxiao:
> There's no **just** here... -- Folks are already using (and extending) this in the wild, so we'd need to consider any migration path.
> 
> I'm afraid I haven't looked in depth at this specific question, but there have been a few related tickets in this area, that reviewing might help move forward. (Like **maybe** it would be OK, but without measuring twice I wouldn't like to just say yes.)

I mean… there are grades of modifications from **just** to **not-just** 😛

To be clear what I'm referring to is making `AutocompleteMixin` behave according to what I consider its responsibilities; ie making sure it initialises a _complete_ instance that its supers expect. If a subclass purposefully avoids the "normal" initialisation flow then it should make sure it manually initialises all expected attributes properly:

```
--- a/django/contrib/admin/widgets.py
+++ b/django/contrib/admin/widgets.py
@@ -490,6 +490,7 @@ class AutocompleteMixin:
         self.db = using
         self.choices = choices
         self.attrs = {} if attrs is None else attrs.copy()
+        self.option_attrs = {}
         self.i18n_name = get_select2_language()

     def get_url(self):
```

carltongibson:
@shangxiao Yes, makes sense. (My worry was only breaking existing usage, but squinting, it's likely OK...)

khink:
@shangxiao @carltongibson Thanks for the input! I changed this in 09741910448a9c9d64ed9c2b7a8e95ebdfb9eecb. I think this is what you meant, right?

tests/forms_tests/widget_tests/test_select.py
```diff
@@ -111,6 +111,28 @@ def test_constructor_attrs(self):
             ),
         )
 
+    def test_constructor_option_attrs(self):
```
nessita, line 114: I believe we need an analogous test for radioselect, and even more so that radioselect has the `option_inherits_attrs` set to True, so we'd need tests for `attrs` and `option_attrs` being passed to ensure that `option_attrs` takes precedence.

nessita (commented):
Thanks for the further work. If you could work on the latest comments, that would be great, otherwise I'm happy to push this to completion next week.

django/forms/widgets.py
khink: I see only one now, so i think this comment is resolved now.

django/forms/widgets.py
```diff
@@ -618,12 +618,13 @@ class ChoiceWidget(Widget):
...
     option_inherits_attrs = True
 
-    def __init__(self, attrs=None, choices=()):
+    def __init__(self, attrs=None, choices=(), option_attrs=None):
         super().__init__(attrs)
         # choices can be any iterable, but we may need to render this widget
         # multiple times. Thus, collapse it into a list so it can be consumed
         # more than once.
         self.choices = list(choices)
+        self.option_attrs = {} if option_attrs is None else option_attrs.copy()
```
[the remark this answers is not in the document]
khink: I don't know about that code, but i'll dive into it, thanks for the pointer.

khink:
I've pushed an update for the docs, i'll look into the other comments later.

felixxm:
@khink Do you have time to keep working on this?

khink:
@felixxm Good of you to check, i had forgotten. Unfortunately the answer is no, i have no more time to spend on this.

felixxm:
> @felixxm Good of you to check, i had forgotten. Unfortunately the answer is no, i have no more time to spend on this.

Thanks for letting us know.

PR #16932, opened 2023-06-02
Fixed #34627 -- Highlighted active row in admin UI when forced-colors mode is enabled.

nmenezes0:
Fixes https://code.djangoproject.com/ticket/34627

This PR adds highlighting on selected models in the admin sidebar and changelist when using "forced-colors".

For selected models in the admin and changelist:
image

This is what you currently see with forced-colors: 
image

This PR changes this to:
image

### DIFF
--- a/django/contrib/admin/static/admin/css/changelists.css
+++ b/django/contrib/admin/static/admin/css/changelists.css
@@ -267,10 +267,10 @@
 
 @media (forced-colors: active) {
     #changelist table tbody tr.selected {
-        border: 2px SelectedItem solid;
+        background-color: SelectedItem;
     }
     #changelist table tbody tr:has(input[type=checkbox]:checked) {
-        border: 2px SelectedItem solid;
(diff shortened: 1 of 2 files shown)


nessita:
Thank you for this work! For those (like me) that had no idea how to test this in Firefox, I followed https://hacks.mozilla.org/2020/07/adding-prefers-contrast-to-firefox/ and set the colors to "Override the colors specified by the page with your selections above" -> "Always".

I used the default colors from Firefox, and this is how it looks:
image

I was wondering if the row highlight could be... ligther? different? so the text on the highlighted row is easier to read. Right now the bold violet on top of dark cyan (?) seems challenging.

What do you think?

nessita (approved):
Approving since this is working as expected, will wait a few days before merging to ensure other reviewers get the chance to review as well.

thibaudcolas (approved):
Thank you for giving this a go @nmenezes0. Result is looking great.

PR #16934, opened 2023-06-02
Fixed #34628 -- Improved icons accessibility of related widget wrapper in admin.

nmenezes0:
Fixes https://code.djangoproject.com/ticket/34628.

Makes icons larger, greys them out when disabled.

### DIFF
--- a/django/contrib/admin/static/admin/css/widgets.css
+++ b/django/contrib/admin/static/admin/css/widgets.css
@@ -579,16 +579,13 @@ ul.timelist, .timelist li {
 }
 
 .related-widget-wrapper-link {
-    opacity: 0.3;
+    opacity: .6;
+    filter: grayscale(1);
 }
 
 .related-widget-wrapper-link:link {
-    opacity: .8;
-}
-
-.related-widget-wrapper-link:link:focus,
-.related-widget-wrapper-link:link:hover {
     opacity: 1;
+    filter: grayscale(0);
 }
 
 select + .related-widget-wrapper-link,
--- a/django/contrib/admin/templates/admin/widgets/related_widget_wrapper.html
+++ b/django/contrib/admin/templates/admin/widgets/related_widget_wrapper.html
@@ -9,7 +9,7 @@
             data-href-template="{{ change_related_template_url }}?{{ url_params }}"
             data-popup="yes"
             title="{% blocktranslate %}Change selected {{ model }}{% endblocktranslate %}">
-            <img src="{% static 'admin/img/icon-changelink.svg' %}" alt="">
+            <img src="{% static 'admin/img/icon-changelink.svg' %}" alt="" width="20" height="20">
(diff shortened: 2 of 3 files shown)


django/contrib/admin/static/admin/css/widgets.css
```diff
@@ -579,11 +579,13 @@ ul.timelist, .timelist li {
...
+    filter: grayscale(1);
 }
 
 .related-widget-wrapper-link:link {
-    opacity: .8;
+    opacity: 1;
+    filter: grayscale(0);
 }
 
 .related-widget-wrapper-link:link:focus,
```
thibaudcolas, line 591: Those hover / focus styles should be removed if we always have full opacity for links.

django/contrib/admin/templates/admin/widgets/related_widget_wrapper.html
```diff
@@ -32,7 +32,7 @@
           <a class="related-widget-wrapper-link view-related" id="view_id_{{ name }}"
```
thibaudcolas, line 32: ^ Above – we also need to add the width and height for the "Delete" action link.

thibaudcolas (changes requested):
Thank you for giving this a go @nmenezes0! The visual changes look good to me. Having bigger, less transparent icons will be great for everyone. In the future, please make sure to add a more descriptive title for your pull request and ideally include a screenshot (or before/after) in the description so it’s easier to understand the changes you made.

There is a further refactoring to implement, and an extra icon to update.

nmenezes0:
> Thank you for giving this a go @nmenezes0! The visual changes look good to me. Having bigger, less transparent icons will be great for everyone. In the future, please make sure to add a more descriptive title for your pull request and ideally include a screenshot (or before/after) in the description so it’s easier to understand the changes you made.
> 
> There is a further refactoring to implement, and an extra icon to update.

Thanks - will have a go at those changes, and will add more detail next time I create a PR.

django/contrib/admin/templates/admin/widgets/related_widget_wrapper.html
nmenezes0: I have added the width and height, though I got a bit stuck trying to test (trying to find out when the delete should appear).

By doing something hacky, this is the before:
image

And this is the after:
image

django/contrib/admin/static/admin/css/widgets.css
nmenezes0: Now have been removed.

felixxm (approved):
@nmenezes0 Thanks :+1: I squashed commits and added a small test for `width`/`height`.

felixxm:
buildbot, test on selenium.

nmenezes0:
> 

Thanks @felixxm and thanks @thibaudcolas for all your help with this!

PR #16935, opened 2023-06-02
Minor improvement in auth.logout

AliRn76:
Replace `getattr` with `hasattr`


nessita:
Hello! Thank you for your effort in contributing but the proposed change is not logically-equivalent to the existing code. It's actually incorrect since your proposal will set the `user` to None if the `authenticated` attribute is not defined and the code does something different.

AliRn76:
> Hello! Thank you for your effort in contributing but the proposed change is not logically-equivalent to the existing code. It's actually incorrect since your proposal will set the `user` to None if the `authenticated` attribute is not defined and the code does something different.

Thank you for your explanation 🙌🏽

PR #16942, opened 2023-06-05
Fixed #34429 -- Allowed setting unusable passwords for users in the auth forms.

fsbraun:
https://code.djangoproject.com/ticket/34429

## Feature added
This PR introduces a new functionality to the admin change password form: the checkbox (on by default) "Allow login"

image

If unchecked, the password field disappears and the form sets an unusable password upon submission. The user is warned before.

image

## Tests

Existing tests have been changed to contain the new form field.

A new test was added to test the functionality.

## Other effects
This PR adds three new strings which will have to be translated.

### DIFF
--- a/django/contrib/admin/static/admin/js/unusable_password_field.js
+++ b/django/contrib/admin/static/admin/js/unusable_password_field.js
@@ -1,0 +1,29 @@
+"use strict";
+// Fallback JS for browsers which do not support :has selector used in
+// admin/css/unusable_password_fields.css
+// Remove file once all supported browsers support :has selector
+try {
+    // If browser does not support :has selector this will raise an error
+    document.querySelector("form:has(input)");
+} catch (error) {
+    console.log("Defaulting to javascript for usable password form management: " + error);
+    // JS replacement for unsupported :has selector
+    document.querySelectorAll('input[name="usable_password"]').forEach(option => {
+        option.addEventListener('change', function() {
+            const usablePassword = (this.value === "true" ? this.checked : !this.checked);
+            const submit1 = document.querySelector('input[type="submit"].set-password');
+            const submit2 = document.querySelector('input[type="submit"].unset-password');
+            const messages = document.querySelector('#id_unusable_warning');
+            document.getElementById('id_password1').closest('.form-row').hidden = !usablePassword;
+            document.getElementById('id_password2').closest('.form-row').hidden = !usablePassword;
+            if (messages) {
+                messages.hidden = usablePassword;
+            }
+            if (submit1 && submit2) {
+                submit1.hidden = !usablePassword;
+                submit2.hidden = usablePassword;
+            }
+        });
+        option.dispatchEvent(new Event('change'));
+    });
+}
--- a/django/contrib/admin/templates/admin/auth/user/add_form.html
+++ b/django/contrib/admin/templates/admin/auth/user/add_form.html
@@ -1,5 +1,5 @@
 {% extends "admin/change_form.html" %}
-{% load i18n %}
+{% load i18n static %}
 
 {% block form_top %}
   {% if not is_popup %}
@@ -7,4 +7,12 @@
   {% else %}
     <p>{% translate "Enter a username and password." %}</p>
   {% endif %}
+{% endblock %}
+{% block extrahead %}
+  {{ block.super }}
+  <link rel="stylesheet" href="{% static 'admin/css/unusable_password_field.css' %}">
+{% endblock %}
+{% block admin_change_form_document_ready %}
+  {{ block.super }}
+  <script src="{% static 'admin/js/unusable_password_field.js' %}" defer></script>
 {% endblock %}
--- a/django/contrib/admin/templates/admin/auth/user/change_password.html
+++ b/django/contrib/admin/templates/admin/auth/user/change_password.html
@@ -2,7 +2,11 @@
 {% load i18n static %}
 {% load admin_urls %}
 
-{% block extrastyle %}{{ block.super }}<link rel="stylesheet" href="{% static "admin/css/forms.css" %}">{% endblock %}
+{% block extrastyle %}
+  {{ block.super }}
+  <link rel="stylesheet" href="{% static "admin/css/forms.css" %}">
+  <link rel="stylesheet" href="{% static 'admin/css/unusable_password_field.css' %}">
+{% endblock %}
 {% block bodyclass %}{{ block.super }} {{ opts.app_label }}-{{ opts.model_name }} change-form{% endblock %}
 {% if not is_popup %}
 {% block breadcrumbs %}
@@ -11,7 +15,7 @@
 &rsaquo; <a href="{% url 'admin:app_list' app_label=opts.app_label %}">{{ opts.app_config.verbose_name }}</a>
 &rsaquo; <a href="{% url opts|admin_urlname:'changelist' %}">{{ opts.verbose_name_plural|capfirst }}</a>
 &rsaquo; <a href="{% url opts|admin_urlname:'change' original.pk|admin_urlquote %}">{{ original|truncatewords:"18" }}</a>
-&rsaquo; {% translate 'Change password' %}
+&rsaquo; {% if form.user.has_usable_password %}{% translate 'Change password' %}{% else %}{% translate 'Set password' %}{% endif %}
 </div>
 {% endblock %}
 {% endif %}
@@ -27,10 +31,23 @@
 {% endif %}
 
 <p>{% blocktranslate with username=original %}Enter a new password for the user <strong>{{ username }}</strong>.{% endblocktranslate %}</p>
+{% if not form.user.has_usable_password %}
+  <p>{% blocktranslate %}This action will <strong>enable</strong> password-based authentication for this user.{% endblocktranslate %}</p>
+{% endif %}
 
 <fieldset class="module aligned">
 
 <div class="form-row">
+  {{ form.usable_password.errors }}
+  <div class="flex-container">{{ form.usable_password.label_tag }} {{ form.usable_password }}</div>
+  {% if form.usable_password.help_text %}
+  <div class="help"{% if form.usable_password.id_for_label %} id="{{ form.usable_password.id_for_label }}_helptext"{% endif %}>
+    <p>{{ form.usable_password.help_text|safe }}</p>
+  </div>
+  {% endif %}
+</div>
+
+<div class="form-row field-password1">
   {{ form.password1.errors }}
   <div class="flex-container">{{ form.password1.label_tag }} {{ form.password1 }}</div>
   {% if form.password1.help_text %}
@@ -38,7 +55,7 @@
   {% endif %}
 </div>
 
-<div class="form-row">
+<div class="form-row field-password2">
   {{ form.password2.errors }}
   <div class="flex-container">{{ form.password2.label_tag }} {{ form.password2 }}</div>
   {% if form.password2.help_text %}
@@ -49,9 +66,15 @@
 </fieldset>
 
 <div class="submit-row">
-<input type="submit" value="{% translate 'Change password' %}" class="default">
+  {% if form.user.has_usable_password %}
+  <input type="submit" name="set-password" value="{% translate 'Change password' %}" class="default set-password">
+  <input type="submit" name="unset-password" value="{% translate 'Disable password-based authentication' %}" class="unset-password">
+  {% else %}
+  <input type="submit" name="set-password" value="{% translate 'Enable password-based authentication' %}" class="default set-password">
+  {% endif %}
 </div>
 
 </div>
 </form></div>
+<script src="{% static 'admin/js/unusable_password_field.js' %}" defer></script>
 {% endblock %}
--- a/django/contrib/auth/forms.py
+++ b/django/contrib/auth/forms.py
@@ -89,27 +89,65 @@ class UsernameField(forms.CharField):
         }
 
 
-class BaseUserCreationForm(forms.ModelForm):
+class SetPasswordMixin:
     """
-    A form that creates a user, with no privileges, from the given username and
-    password.
+    Form mixin that validates and sets a password for a user.
     """
 
     error_messages = {
         "password_mismatch": _("The two password fields didn’t match."),
     }
-    password1 = forms.CharField(
-        label=_("Password"),
-        strip=False,
-        widget=forms.PasswordInput(attrs={"autocomplete": "new-password"}),
-        help_text=password_validation.password_validators_help_text_html(),
-    )
-    password2 = forms.CharField(
-        label=_("Password confirmation"),
-        widget=forms.PasswordInput(attrs={"autocomplete": "new-password"}),
-        strip=False,
-        help_text=_("Enter the same password as before, for verification."),
-    )
+
+    @staticmethod
+    def create_password_fields(label1=_("Password"), label2=_("Password confirmation")):
+        password1 = forms.CharField(
+            label=label1,
+            strip=False,
+            widget=forms.PasswordInput(attrs={"autocomplete": "new-password"}),
+            help_text=password_validation.password_validators_help_text_html(),
+        )
+        password2 = forms.CharField(
+            label=label2,
+            widget=forms.PasswordInput(attrs={"autocomplete": "new-password"}),
+            strip=False,
+            help_text=_("Enter the same password as before, for verification."),
+        )
+        return password1, password2
+
+    def validate_passwords(
+        self, password1_field_name="password1", password2_field_name="password2"
+    ):
+        password1 = self.cleaned_data.get(password1_field_name)
+        password2 = self.cleaned_data.get(password2_field_name)
+        if password1 and password2 and password1 != password2:
+            error = ValidationError(
+                self.error_messages["password_mismatch"],
+                code="password_mismatch",
+            )
+            self.add_error(password2_field_name, error)
+
+    def validate_password_for_user(self, user, password_field_name="password2"):
+        password = self.cleaned_data.get(password_field_name)
+        if password:
+            try:
+                password_validation.validate_password(password, user)
+            except ValidationError as error:
+                self.add_error(password_field_name, error)
+
+    def set_password_and_save(self, user, password_field_name="password1", commit=True):
+        user.set_password(self.cleaned_data[password_field_name])
+        if commit:
+            user.save()
+        return user
+
+
+class BaseUserCreationForm(SetPasswordMixin, forms.ModelForm):
+    """
+    A form that creates a user, with no privileges, from the given username and
+    password.
+    """
+
+    password1, password2 = SetPasswordMixin.create_password_fields()
 
     class Meta:
         model = User
@@ -123,34 +161,21 @@ class BaseUserCreationForm(forms.ModelForm):
                 "autofocus"
             ] = True
 
-    def clean_password2(self):
-        password1 = self.cleaned_data.get("password1")
-        password2 = self.cleaned_data.get("password2")
-        if password1 and password2 and password1 != password2:
-            raise ValidationError(
-                self.error_messages["password_mismatch"],
-                code="password_mismatch",
-            )
-        return password2
+    def clean(self):
+        self.validate_passwords()
+        return super().clean()
 
     def _post_clean(self):
         super()._post_clean()
         # Validate the password after self.instance is updated with form data
         # by super().
-        password = self.cleaned_data.get("password2")
-        if password:
-            try:
-                password_validation.validate_password(password, self.instance)
-            except ValidationError as error:
-                self.add_error("password2", error)
+        self.validate_password_for_user(self.instance)
 
     def save(self, commit=True):
         user = super().save(commit=False)
-        user.set_password(self.cleaned_data["password1"])
-        if commit:
-            user.save()
-            if hasattr(self, "save_m2m"):
-                self.save_m2m()
+        user = self.set_password_and_save(user, commit=commit)
+        if commit and hasattr(self, "save_m2m"):
+            self.save_m2m()
         return user
 
 
@@ -383,48 +408,27 @@ class PasswordResetForm(forms.Form):
             )
 
 
-class SetPasswordForm(forms.Form):
+class SetPasswordForm(SetPasswordMixin, forms.Form):
     """
     A form that lets a user set their password without entering the old
     password
     """
 
-    error_messages = {
-        "password_mismatch": _("The two password fields didn’t match."),
-    }
-    new_password1 = forms.CharField(
-        label=_("New password"),
-        widget=forms.PasswordInput(attrs={"autocomplete": "new-password"}),
-        strip=False,
-        help_text=password_validation.password_validators_help_text_html(),
-    )
-    new_password2 = forms.CharField(
-        label=_("New password confirmation"),
-        strip=False,
-        widget=forms.PasswordInput(attrs={"autocomplete": "new-password"}),
+    new_password1, new_password2 = SetPasswordMixin.create_password_fields(
+        label1=_("New password"), label2=_("New password confirmation")
     )
 
     def __init__(self, user, *args, **kwargs):
         self.user = user
         super().__init__(*args, **kwargs)
 
-    def clean_new_password2(self):
-        password1 = self.cleaned_data.get("new_password1")
-        password2 = self.cleaned_data.get("new_password2")
-        if password1 and password2 and password1 != password2:
-            raise ValidationError(
-                self.error_messages["password_mismatch"],
-                code="password_mismatch",
-            )
-        password_validation.validate_password(password2, self.user)
-        return password2
+    def clean(self):
+        self.validate_passwords("new_password1", "new_password2")
+        self.validate_password_for_user(self.user, "new_password2")
+        return super().clean()
 
     def save(self, commit=True):
-        password = self.cleaned_data["new_password1"]
-        self.user.set_password(password)
-        if commit:
-            self.user.save()
-        return self.user
+        return self.set_password_and_save(self.user, "new_password1", commit=commit)
 
 
 class PasswordChangeForm(SetPasswordForm):
@@ -462,52 +466,27 @@ class PasswordChangeForm(SetPasswordForm):
         return old_password
 
 
-class AdminPasswordChangeForm(forms.Form):
+class AdminPasswordChangeForm(SetPasswordMixin, forms.Form):
     """
     A form used to change the password of a user in the admin interface.
     """
 
-    error_messages = {
-        "password_mismatch": _("The two password fields didn’t match."),
-    }
     required_css_class = "required"
-    password1 = forms.CharField(
-        label=_("Password"),
-        widget=forms.PasswordInput(
-            attrs={"autocomplete": "new-password", "autofocus": True}
-        ),
-        strip=False,
-        help_text=password_validation.password_validators_help_text_html(),
-    )
-    password2 = forms.CharField(
-        label=_("Password (again)"),
-        widget=forms.PasswordInput(attrs={"autocomplete": "new-password"}),
-        strip=False,
-        help_text=_("Enter the same password as before, for verification."),
-    )
+    password1, password2 = SetPasswordMixin.create_password_fields()
 
     def __init__(self, user, *args, **kwargs):
         self.user = user
         super().__init__(*args, **kwargs)
+        self.fields["password1"].widget.attrs["autofocus"] = True
 
-    def clean_password2(self):
-        password1 = self.cleaned_data.get("password1")
-        password2 = self.cleaned_data.get("password2")
-        if password1 and password2 and password1 != password2:
-            raise ValidationError(
-                self.error_messages["password_mismatch"],
-                code="password_mismatch",
-            )
-        password_validation.validate_password(password2, self.user)
-        return password2
+    def clean(self):
+        self.validate_passwords()
+        self.validate_password_for_user(self.user)
+        return super().clean()
 
     def save(self, commit=True):
         """Save the new password."""
-        password = self.cleaned_data["password1"]
-        self.user.set_password(password)
-        if commit:
-            self.user.save()
-        return self.user
+        return self.set_password_and_save(self.user, commit=commit)
 
     @property
     def changed_data(self):
--- a/django/contrib/auth/forms.py
+++ b/django/contrib/auth/forms.py
@@ -92,33 +92,78 @@ class UsernameField(forms.CharField):
 class SetPasswordMixin:
     """
     Form mixin that validates and sets a password for a user.
+
+    This mixin also support setting an unusable password for a user.
     """
 
     error_messages = {
         "password_mismatch": _("The two password fields didn’t match."),
     }
+    usable_password_help_text = _(
+        "Whether the user will be able to authenticate using a password or not. "
+        "If disabled, they may still be able to authenticate using other backends, "
+        "such as Single Sign-On or LDAP."
+    )
 
     @staticmethod
     def create_password_fields(label1=_("Password"), label2=_("Password confirmation")):
         password1 = forms.CharField(
             label=label1,
+            required=False,
             strip=False,
             widget=forms.PasswordInput(attrs={"autocomplete": "new-password"}),
             help_text=password_validation.password_validators_help_text_html(),
         )
         password2 = forms.CharField(
             label=label2,
+            required=False,
             widget=forms.PasswordInput(attrs={"autocomplete": "new-password"}),
             strip=False,
             help_text=_("Enter the same password as before, for verification."),
         )
         return password1, password2
 
+    @staticmethod
+    def create_usable_password_field(help_text=usable_password_help_text):
+        return forms.ChoiceField(
+            label=_("Password-based authentication"),
+            required=False,
+            initial="true",
+            choices={"true": _("Enabled"), "false": _("Disabled")},
+            widget=forms.RadioSelect(attrs={"class": "radiolist inline"}),
+            help_text=help_text,
+        )
+
     def validate_passwords(
-        self, password1_field_name="password1", password2_field_name="password2"
+        self,
+        password1_field_name="password1",
+        password2_field_name="password2",
+        usable_password_field_name="usable_password",
     ):
+        usable_password = (
+            self.cleaned_data.pop(usable_password_field_name, None) != "false"
+        )
+        self.cleaned_data["set_usable_password"] = usable_password
         password1 = self.cleaned_data.get(password1_field_name)
         password2 = self.cleaned_data.get(password2_field_name)
+
+        if not usable_password:
+            return self.cleaned_data
+
+        if not password1:
+            error = ValidationError(
+                self.fields[password1_field_name].error_messages["required"],
+                code="required",
+            )
+            self.add_error(password1_field_name, error)
+
+        if not password2:
+            error = ValidationError(
+                self.fields[password2_field_name].error_messages["required"],
+                code="required",
+            )
+            self.add_error(password2_field_name, error)
+
         if password1 and password2 and password1 != password2:
             error = ValidationError(
                 self.error_messages["password_mismatch"],
@@ -128,14 +173,17 @@ class SetPasswordMixin:
 
     def validate_password_for_user(self, user, password_field_name="password2"):
         password = self.cleaned_data.get(password_field_name)
-        if password:
+        if password and self.cleaned_data["set_usable_password"]:
             try:
                 password_validation.validate_password(password, user)
             except ValidationError as error:
                 self.add_error(password_field_name, error)
 
     def set_password_and_save(self, user, password_field_name="password1", commit=True):
-        user.set_password(self.cleaned_data[password_field_name])
+        if self.cleaned_data["set_usable_password"]:
+            user.set_password(self.cleaned_data[password_field_name])
+        else:
+            user.set_unusable_password()
         if commit:
             user.save()
         return user
@@ -148,6 +196,7 @@ class BaseUserCreationForm(SetPasswordMixin, forms.ModelForm):
     """
 
     password1, password2 = SetPasswordMixin.create_password_fields()
+    usable_password = SetPasswordMixin.create_usable_password_field()
 
     class Meta:
         model = User
@@ -205,7 +254,7 @@ class UserChangeForm(forms.ModelForm):
         label=_("Password"),
         help_text=_(
             "Raw passwords are not stored, so there is no way to see this "
-            "user’s password, but you can change the password using "
+            "user’s password, but you can change or unset the password using "
             '<a href="{}">this form</a>.'
         ),
     )
@@ -219,6 +268,11 @@ class UserChangeForm(forms.ModelForm):
         super().__init__(*args, **kwargs)
         password = self.fields.get("password")
         if password:
+            if self.instance and not self.instance.has_usable_password():
+                password.help_text = _(
+                    "Enable password-based authentication for this user by setting a "
+                    'password using <a href="{}">this form</a>.'
+                )
             password.help_text = password.help_text.format(
                 f"../../{self.instance.pk}/password/"
             )
@@ -472,12 +526,22 @@ class AdminPasswordChangeForm(SetPasswordMixin, forms.Form):
     """
 
     required_css_class = "required"
+    usable_password_help_text = SetPasswordMixin.usable_password_help_text + (
+        '<ul id="id_unusable_warning" class="messagelist"><li class="warning">'
+        "If disabled, the current password for this user will be lost.</li></ul>"
+    )
     password1, password2 = SetPasswordMixin.create_password_fields()
 
     def __init__(self, user, *args, **kwargs):
         self.user = user
         super().__init__(*args, **kwargs)
         self.fields["password1"].widget.attrs["autofocus"] = True
+        if self.user.has_usable_password():
+            self.fields["usable_password"] = (
+                SetPasswordMixin.create_usable_password_field(
+                    self.usable_password_help_text
+                )
+            )
 
     def clean(self):
         self.validate_passwords()
@@ -491,7 +555,6 @@ class AdminPasswordChangeForm(SetPasswordMixin, forms.Form):
     @property
     def changed_data(self):
         data = super().changed_data
-        for name in self.fields:
-            if name not in data:
-                return []
-        return ["password"]
+        if "set_usable_password" in data or "password1" in data and "password2" in data:
+            return ["password"]
+        return []
--- a/docs/releases/5.1.txt
+++ b/docs/releases/5.1.txt
@@ -46,6 +46,12 @@ Minor features
 * The default iteration count for the PBKDF2 password hasher is increased from
   720,000 to 870,000.
 
+* :class:`~django.contrib.auth.forms.BaseUserCreationForm` and
+  :class:`~django.contrib.auth.forms.AdminPasswordChangeForm` now support
+  disabling password-based authentication by setting an unusable password on
+  form save. This is now available in the admin when visiting the user creation
+  and password change pages.
+
 :mod:`django.contrib.contenttypes`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/topics/auth/default.txt
+++ b/docs/topics/auth/default.txt
@@ -1623,9 +1623,17 @@ provides several built-in forms located in :mod:`django.contrib.auth.forms`:
 
 .. class:: AdminPasswordChangeForm
 
-    A form used in the admin interface to change a user's password.
+    A form used in the admin interface to change a user's password, including
+    the ability to set an :meth:`unusable password
+    <django.contrib.auth.models.User.set_unusable_password>`, which blocks the
+    user from logging in with password-based authentication.
 
     Takes the ``user`` as the first positional argument.
+
+    .. versionchanged:: 5.1
+
+        Option to disable (or reenable) password-based authentication was
+        added.
 
 .. class:: AuthenticationForm
 
@@ -1717,12 +1725,21 @@ provides several built-in forms located in :mod:`django.contrib.auth.forms`:
     A :class:`~django.forms.ModelForm` for creating a new user. This is the
     recommended base class if you need to customize the user creation form.
 
-    It has three fields: ``username`` (from the user model), ``password1``,
-    and ``password2``. It verifies that ``password1`` and ``password2`` match,
-    validates the password using
+    It has four fields: ``username`` (from the user model), ``password1``,
+    ``password2``, and ``usable_password`` (the latter is enabled by default).
+    If ``usable_password`` is enabled, it verifies that ``password1`` and
+    ``password2`` are non empty and match, validates the password using
     :func:`~django.contrib.auth.password_validation.validate_password`, and
     sets the user's password using
     :meth:`~django.contrib.auth.models.User.set_password()`.
+    If ``usable_password`` is disabled, no password validation is done, and
+    password-based authentication is disabled for the user by calling
+    :meth:`~django.contrib.auth.models.User.set_unusable_password()`.
+
+    .. versionchanged:: 5.1
+
+        Option to create users with disabled password-based authentication was
+        added.
 
 .. class:: UserCreationForm
 
@@ -1837,6 +1854,8 @@ You should see a link to "Users" in the "Auth"
 section of the main admin index page. The "Add user" admin page is different
 than standard admin pages in that it requires you to choose a username and
 password before allowing you to edit the rest of the user's fields.
+Alternatively, on this page, you can choose a username and disable
+password-based authentication for the user.
 
 Also note: if you want a user account to be able to create users using the
 Django admin site, you'll need to give them permission to add users *and*
@@ -1858,4 +1877,4 @@ Changing passwords
 User passwords are not displayed in the admin (nor stored in the database), but
 the :doc:`password storage details </topics/auth/passwords>` are displayed.
 Included in the display of this information is a link to
-a password change form that allows admins to change user passwords.
+a password change form that allows admins to change or unset user passwords.
--- a/tests/admin_views/test_password_form.py
+++ b/tests/admin_views/test_password_form.py
@@ -1,0 +1,144 @@
+from django.contrib.admin.tests import AdminSeleniumTestCase
+from django.contrib.auth.models import User
+from django.test import override_settings
+from django.urls import reverse
+
+
+@override_settings(ROOT_URLCONF="auth_tests.urls_admin")
+class SeleniumAuthTests(AdminSeleniumTestCase):
+    available_apps = AdminSeleniumTestCase.available_apps
+
+    def setUp(self):
+        self.superuser = User.objects.create_superuser(
+            username="super",
+            password="secret",
+            email="super@example.com",
+        )
+
+    def test_add_new_user(self):
+        """A user with no password can be added.
+
+        Enabling/disabling the usable password field shows/hides the password
+        fields when adding a user.
+        """
+        from selenium.common import NoSuchElementException
+        from selenium.webdriver.common.by import By
+
+        user_add_url = reverse("auth_test_admin:auth_user_add")
+        self.admin_login(username="super", password="secret")
+        self.selenium.get(self.live_server_url + user_add_url)
+
+        pw_switch_on = self.selenium.find_element(
+            By.CSS_SELECTOR, 'input[name="usable_password"][value="true"]'
+        )
+        pw_switch_off = self.selenium.find_element(
+            By.CSS_SELECTOR, 'input[name="usable_password"][value="false"]'
+        )
+        password1 = self.selenium.find_element(
+            By.CSS_SELECTOR, 'input[name="password1"]'
+        )
+        password2 = self.selenium.find_element(
+            By.CSS_SELECTOR, 'input[name="password2"]'
+        )
+
+        # Default is to set a password on user creation.
+        self.assertIs(pw_switch_on.is_selected(), True)
+        self.assertIs(pw_switch_off.is_selected(), False)
+
+        # The password fields are visible.
+        self.assertIs(password1.is_displayed(), True)
+        self.assertIs(password2.is_displayed(), True)
+
+        # Click to disable password-based authentication.
+        pw_switch_off.click()
+
+        # Radio buttons are updated accordingly.
+        self.assertIs(pw_switch_on.is_selected(), False)
+        self.assertIs(pw_switch_off.is_selected(), True)
+
+        # The password fields are hidden.
+        self.assertIs(password1.is_displayed(), False)
+        self.assertIs(password2.is_displayed(), False)
+
+        # The warning message should not be shown.
+        with self.assertRaises(NoSuchElementException):
+            self.selenium.find_element(By.ID, "id_unusable_warning")
+
+    def test_change_password_for_existing_user(self):
+        """A user can have their password changed or unset.
+
+        Enabling/disabling the usable password field shows/hides the password
+        fields and the warning about password lost.
+        """
+        from selenium.webdriver.common.by import By
+
+        user = User.objects.create_user(
+            username="ada", password="charles", email="ada@example.com"
+        )
+        user_url = reverse("auth_test_admin:auth_user_password_change", args=(user.pk,))
+        self.admin_login(username="super", password="secret")
+        self.selenium.get(self.live_server_url + user_url)
+
+        pw_switch_on = self.selenium.find_element(
+            By.CSS_SELECTOR, 'input[name="usable_password"][value="true"]'
+        )
+        pw_switch_off = self.selenium.find_element(
+            By.CSS_SELECTOR, 'input[name="usable_password"][value="false"]'
+        )
+        password1 = self.selenium.find_element(
+            By.CSS_SELECTOR, 'input[name="password1"]'
+        )
+        password2 = self.selenium.find_element(
+            By.CSS_SELECTOR, 'input[name="password2"]'
+        )
+        submit_set = self.selenium.find_element(
+            By.CSS_SELECTOR, 'input[type="submit"].set-password'
+        )
+        submit_unset = self.selenium.find_element(
+            By.CSS_SELECTOR, 'input[type="submit"].unset-password'
+        )
+
+        # By default password-based authentication is enabled.
+        self.assertIs(pw_switch_on.is_selected(), True)
+        self.assertIs(pw_switch_off.is_selected(), False)
+
+        # The password fields are visible.
+        self.assertIs(password1.is_displayed(), True)
+        self.assertIs(password2.is_displayed(), True)
+
+        # Only the set password submit button is visible.
+        self.assertIs(submit_set.is_displayed(), True)
+        self.assertIs(submit_unset.is_displayed(), False)
+
+        # Click to disable password-based authentication.
+        pw_switch_off.click()
+
+        # Radio buttons are updated accordingly.
+        self.assertIs(pw_switch_on.is_selected(), False)
+        self.assertIs(pw_switch_off.is_selected(), True)
+
+        # The password fields are hidden.
+        self.assertIs(password1.is_displayed(), False)
+        self.assertIs(password2.is_displayed(), False)
+
+        # Only the unset password submit button is visible.
+        self.assertIs(submit_unset.is_displayed(), True)
+        self.assertIs(submit_set.is_displayed(), False)
+
+        # The warning about password being lost is shown.
+        warning = self.selenium.find_element(By.ID, "id_unusable_warning")
+        self.assertIs(warning.is_displayed(), True)
+
+        # Click to enable password-based authentication.
+        pw_switch_on.click()
+
+        # The warning disappears.
+        self.assertIs(warning.is_displayed(), False)
+
+        # The password fields are shown.
+        self.assertIs(password1.is_displayed(), True)
+        self.assertIs(password2.is_displayed(), True)
+
+        # Only the set password submit button is visible.
+        self.assertIs(submit_set.is_displayed(), True)
+        self.assertIs(submit_unset.is_displayed(), False)
--- a/tests/auth_tests/test_forms.py
+++ b/tests/auth_tests/test_forms.py
@@ -221,6 +221,16 @@ class BaseUserCreationFormTest(TestDataMixin, TestCase):
             form["password2"].errors,
         )
 
+        # passwords are not validated if `usable_password` is unset
+        data = {
+            "username": "othertestclient",
+            "password1": "othertestclient",
+            "password2": "othertestclient",
+            "usable_password": "false",
+        }
+        form = BaseUserCreationForm(data)
+        self.assertIs(form.is_valid(), True, form.errors)
+
     def test_custom_form(self):
         class CustomUserCreationForm(BaseUserCreationForm):
             class Meta(BaseUserCreationForm.Meta):
@@ -349,6 +359,19 @@ class BaseUserCreationFormTest(TestDataMixin, TestCase):
             ["The password is too similar to the first name."],
         )
 
+        # passwords are not validated if `usable_password` is unset
+        form = CustomUserCreationForm(
+            {
+                "username": "testuser",
+                "password1": "testpassword",
+                "password2": "testpassword",
+                "first_name": "testpassword",
+                "last_name": "lastname",
+                "usable_password": "false",
+            }
+        )
+        self.assertIs(form.is_valid(), True, form.errors)
+
     def test_username_field_autocapitalize_none(self):
         form = BaseUserCreationForm()
         self.assertEqual(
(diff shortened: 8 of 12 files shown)


django/contrib/admin/templates/admin/auth/user/change_password.html
```diff
@@ -54,4 +70,5 @@
 
 </div>
 </form></div>
+<script src="{% static 'admin/js/change_password_form.js' %}" async></script>
```
ngnpope, line 73: Stick this in the `extrahead` block:

```html
{% block extrahead %}
    {{ block.super }}
    <script src="{% static 'admin/js/change_password_form.js' %}" async></script>
{% endblock %}
```

django/contrib/admin/templates/admin/auth/user/change_password.html
```diff
@@ -28,8 +28,24 @@
 
 <p>{% blocktranslate with username=original %}Enter a new password for the user <strong>{{ username }}</strong>.{% endblocktranslate %}</p>
 
+<ul id="id_unusable_warning" class="messagelist">
+  <li class="warning">
+    {% translate "Upon submission this will set a unusable password and prevent the user from logging in." %}
```
ngnpope, line 33:
```suggestion
    {% translate "Upon submission an unusable password will be set which prevents the user from logging in." %}
```

ngnpope (changes requested):
Thanks for this, I've made some comments.

Please can you update the PR title to match the format in the contributing guidelines:

```
Fixed #34429 -- Allowed setting unusable password via admin UI.
```

This will also need documentation changes and release notes adding.

When you've done this you can uncheck `Needs documentation` and `Patch needs improvement` on ticket-34429.

tests/auth_tests/test_forms.py
```diff
@@ -1371,3 +1402,21 @@ def test_html_autocomplete_attributes(self):
...
+                "password1": "",
+                "password2": "test",
+            },
+        )
+
+        with mock.patch.object(User, "set_unusable_password") as set_unusable_password:
+            valid = form.is_valid()
+            form.save(commit=True)
+
+        self.assertTrue(valid)  # Valid despite password empty/mismatch
```
ngnpope, line 1421:
```suggestion
        self.assertIs(valid, True)  # Valid despite password empty/mismatch.
```

tests/auth_tests/test_forms.py
```diff
@@ -1371,3 +1402,21 @@ def test_html_autocomplete_attributes(self):
...
+        set_unusable_password.assert_called_once()
```
ngnpope, line 1422: Instead of mocking, can we just use the following:

```suggestion
        self.assertIs(user.has_usable_password(), False)
```

fsbraun:
@ngnpope Thank you for the suggestions! Let me briefly comment and ask one more question:

* Probably we do not need to wait for `DOMContentLoaded` since the script is loaded as `async`. It would probably miss the event (which might be fired before it is loaded). I'll remove the event in total.
* Using the `hidden` property might clash with custom css which might set something like `.form-row { display: block }` effectively disabling the `hidden` property. OK, if I stay with `style.display`?

Will, of course, also make the other changes.

ngnpope (changes requested):

fsbraun:
One more thing to add: I did use `function () { ... }` instead of `() => { ... }` since otherwise `this` is not defined properly.

django/contrib/auth/forms.py
```diff
@@ -469,42 +469,67 @@ class AdminPasswordChangeForm(forms.Form):
...
     )
     password2 = forms.CharField(
         label=_("Password (again)"),
         widget=forms.PasswordInput(attrs={"autocomplete": "new-password"}),
         strip=False,
+        required=False,
         help_text=_("Enter the same password as before, for verification."),
     )
+    usable_password = forms.BooleanField(
+        label=_("Allow login?"),
```
xi, line 484: I am not sure I like this label. Setting an unusable login is also an option if you use a different authentication backends.

fsbraun: @xi What label would you like? Would a different authentication backend allow logins with an unusable login? If so, this would need explanation.

Would "Allow login with password?" be better? Help text then could be "If unchecked, the user will not be able to log in using their password."

shangxiao:
I have a remark about the wording 

> Allow login?
> If unchecked, the user will not be able to log in.

I think people are going to be confused about what the difference is between this and the active flag on the edit screen:

> Active.
> Designates whether this user should be treated as active. Unselect this instead of deleting accounts.

Nick I know you requested the addition of ? and I'm not saying it's good or bad but it is now inconsistent with other checkboxes (I'd just leave it off personally) 😛

shangxiao:
Can anything be done to stop the jumping checkbox? 🤔

https://github.com/django/django/assets/1845938/fe192289-704e-4db0-9f6f-9667d913c698

shangxiao:
Anyhoo nice work so far fsbraun!

fsbraun:
@shangxiao Thanks very much for the comments:
* The jumping can be stopped by putting the warning below the checkbox and above the submit button
* Thanks for pointing out the "active" alternative. Really, an unusable password only prevents login using username and password (as @xi probably also pointed out). I tried to better name those.

login

django/contrib/auth/forms.py
xi: I like your suggestions :+1:

django/contrib/admin/templates/admin/auth/user/change_password.html
```diff
@@ -30,6 +34,25 @@
...
+  {% if form.usable_password.help_text %}
+  <div class="help"{% if form.usable_password.id_for_label %} id="{{ form.usable_password.id_for_label }}_helptext">{% endif %}{{ form.usable_password.help_text|safe }}</div>
+  {% endif %}
+</div>
+
+<ul id="id_unusable_warning" class="messagelist">
+  <li class="warning">
+    {% blocktranslate %}
+      Upon submission an unusable password will be set which prevents the user from logging in
+      using his username and password. Potential other means of authentication remain unaffected.
```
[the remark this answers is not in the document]
fsbraun: Right! I try to pay attention to those things but they still slip through 🤦‍♂️

fsbraun:
Thanks, everybody, for the constructive feedback. I have updated the PR:

* @knyghty The help text now reads
  
   _Upon submission an unusable password will be set which prevents the user from logging in
   using their username and password. The user's current password will be lost. Potential other
   means of authentication remain unaffected._ 
 
* @xi This is conveniently gender-neural and clarifies that the current password will be lost

* @knyghty I have moved the warning into the help_text div. This implies that once the form field sets the `aria-describedby`  property, it will include the help text **and** the warning. No need for two id to be provided to `aria-describedby` (which would be possible as far as I understand the docs).

fsbraun:
Is there anything I can do to move this forward?

django/contrib/admin/templates/admin/auth/user/change_password.html
```diff
@@ -30,6 +34,28 @@
 
 <fieldset class="module aligned">
 
+<div class="form-row">
+  {{ form.usable_password.errors }}
+  <div class="flex-container checkbox-row">{{ form.usable_password }}
+    <label class="vCheckboxLabel" for="id_usable_password">{{ form.usable_password.label }}</label>
+  </div>
+  {% if form.usable_password.help_text %}
+  <div class="help"{% if form.usable_password.id_for_label %} id="{{ form.usable_password.id_for_label }}_helptext">{% endif %}
```
felixxm, line 43: ">" should be outside of `{% if ... %}...{% endif %}`/

tests/auth_tests/test_forms.py
```diff
@@ -1371,3 +1402,20 @@ def test_html_autocomplete_attributes(self):
...
+                "password1": "",
+                "password2": "test",
+            },
+        )
+
+        valid = form.is_valid()
+        form.save(commit=True)
+
+        self.assertIs(valid, True)  # Valid despite password empty/mismatch.
+        self.assertIs(user.has_usable_password(), False)
```
felixxm, line 1421: Can we also add assertions for `password`.

felixxm (commented):
@fsbraun Thanks :+1: I left initial comments.

According to the ticket description we should have the same feature when adding a new user:
image

django/contrib/auth/forms.py
```diff
@@ -469,42 +469,70 @@ class AdminPasswordChangeForm(forms.Form):
             attrs={"autocomplete": "new-password", "autofocus": True}
         ),
         strip=False,
+        required=False,  # Can be empty if usable_password is False
+        validators=[],  # Password validation happens in the check_password1() method
```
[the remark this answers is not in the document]
fsbraun: Yes, it's both needed and related. The password validation can only happen in the form's clean method, since a password is only required or needs to be validated if `has_usable_password` is `True`.

django/contrib/admin/templates/admin/auth/user/change_password.html
fsbraun: Indeed! 🤦‍♂️

tests/auth_tests/test_forms.py
fsbraun: The only assertion I can think of is that `user.password` starts with `UNUSABLE_PASSWORD_PREFIX`. I feel this is somewhat redundant to `user.has_usable_password()` being `False´, but certainly, I can add this. Or were you thinking of something else?

felixxm: Right, probably not needed.

fsbraun:
@felixxm Thank you for your detailed feedback. Indeed, it overlooked that this should also apply to the user add form. This is fixed now:
* Since the change password and user add form are quite different (form with template and an admin form) I have refactored the HTML to avoid repetition and make it usable for both forms.
* Indentation has been fixed
* The initial value of `usable_password` now is `has_usable_password()`
* I annotated the docs with `versionchanged`
* Updated the docstrings for the tests
* Added a Selenium test
* Fixed the HTML layout issue (wrongly placed `>`)
Please let me know if there is anything else I can do!

fsbraun:
@felixxm Is there anything I can do to get this PR over the finish line? Would you like to take a final look?

nessita:
> I think people are going to be confused about what the difference is between this and the active flag on the edit screen:

I started reviewing this PR and the first think I thought was exactly this. Was there a consideration of being more "honest" with the admin user operating the form, something like:

Label: "Unset the password"
Help text: "Unsetting the password will render the login credentials unusable. The User will no longer be able to log in until a new password is set."

fsbraun:
@nessita Thanks for taking the time to look through this. 

Indeed, the wording has been discussed [here](https://github.com/django/django/pull/16942#discussion_r1218526204) (and I also tried to get some input [here](https://code.djangoproject.com/ticket/34429#comment:6)).

I did consider "Set unusable password" but this would require the user to understand the implications, namely, that the user cannot log in using a username and password. That's the root of the wording.

IMHO, "unsetting" a password could be confused with "set no password". Also, even with an unusable password, you can potentially still log in with a different authentication backend.

Current text: 
**Allow login with username and password**

Help text: 
**If unchecked, the user will have an unusable password and not be able to log in using their username and any password.** (I just realized this had a spelling error)

Warning: 
**Upon submission, an unusable password will be set which prevents the user from logging in
           using their username and password. The user's current password will be lost. Potential other
           means of authentication remain unaffected.**

I am not hung up on this, but also would like it to be as clear as possible. What's your call?

nessita:
> @nessita Thanks for taking the time to look through this.
> 
> Indeed, the wording has been discussed [here](https://github.com/django/django/pull/16942#discussion_r1218526204) (and I also tried to get some input [here](https://code.djangoproject.com/ticket/34429#comment:6)).
> 
> I did consider "Set unusable password" but this would require the user to understand the implications, namely, that the user cannot log in using a username and password. That's the root of the wording.
> 
> IMHO, "unsetting" a password could be confused with "set no password". Also, even with an unusable password, you can potentially still log in with a different authentication backend.
> 
> Current text: **Allow login with username and password**

My main concern with this sentence is that, given that Django allows the User model to be customized so that the `USERNAME_FIELD` could be something other than `username`, like email or something system specific, saying `username` here would not be a accurate for those cases.

Also, considering that this flow occurs within the context of the admin, I would suggest to be more direct with the admin user and explicitly saying that the password will be set and it will be unususable. Otherwise I can see admin user scratching their head wondering what's the difference with marking a user as inactive.

> Help text: **If unchecked, the user will have an unusable password and not be able to log in using their username and any password.** (I just realized this had a spelling error)

Same concern about saying `username`,  we could say "credentials" as I suggested?
 
> Warning: **Upon submission, an unusable password will be set which prevents the user from logging in using their username and password. The user's current password will be lost. Potential other means of authentication remain unaffected.**

I would omit "upon submission", I think this is implied from every help text. With that in mind, I would suggest: 

An unusable password will be set which prevents the user from logging in until a new password is set. The user's current password will be lost, though other means of authentication (if configured) remain unaffected.

> I am not hung up on this, but also would like it to be as clear as possible. What's your call?

I proposed a few alternatives above. The summary would be:
* not saying "username and password" explicitly, and
* being clear about how this action is different from making a user inactive

nessita:
There is also a potentially awkward user flow when:
1. A user `tesuser1` has an unusable password
2. Later on, the admin user goes to this user's change form to set a valid password:
image
4. The admin user is presented with a "change password" form where no password fields are available:
image

I wonder if we should enable the password fields when going from "change the password using <this form>"?

fsbraun:
@nessita Indeed, I initially had the password fields enabled by default. @felixxm requested it differently here: https://github.com/django/django/pull/16942#discussion_r1261060471 

I can revert the change. A third option might be to change the help text if the user does not have a usable password to something like: "Enable this to set a password for this user". 

An update for wording is coming up.

nessita:
> @nessita Indeed, I initially had the password fields enabled by default. @felixxm requested it differently here: [#16942 (comment)](https://github.com/django/django/pull/16942#discussion_r1261060471)

Oh, I see. I can understand the goal of having the form properly initialized, but I guess is weird when coming from "no password set". I guess we can proceed as is and then evaluate potential/future ticket reports.

> I can revert the change.

No need for now.

> A third option might be to change the help text if the user does not have a usable password to something like: "Currently this user does not have a usable password. Enable this to set a password for this user".

I like this proposal. May I suggest:

The current password is not usable, check this option to set a new password for user authentication.

fsbraun:
@nessita **Thanks once more for helping to clarify all this!** Here's my update. Messages are:

Label:
**Set usable password**

Help text (1/2) for "add user" and "change password" if the current password is usable: 
**An unusable password will render the login credentials unusable.  The user will no longer be able authenticate using a password until a new password is set.** 

Help text (2/2) for "change password" if the current password is unusable: 
**The current password is not usable, check this option to set a new password for user authentication.**

Warning: 
**An unusable password prevents the user from authenticating using a password.
          A user's current password will be lost. Potential other means of authentication
          remain unaffected.**

tests/auth_tests/test_forms.py
```diff
@@ -372,6 +375,15 @@ def test_case_insensitive_username(self):
             ["A user with that username already exists."],
         )
 
+    def test_unusable_password(self):
+        """Create user with unusable password"""
```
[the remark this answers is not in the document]
fsbraun: Removed.

tests/auth_tests/test_forms.py
```diff
@@ -372,6 +375,15 @@ def test_case_insensitive_username(self):
...
+        data = {
+            "username": "new-user-which-does-not-exist",
+            "usable_password": False,
+        }
+        form = UserCreationForm(data)
+        self.assertTrue(form.is_valid())
```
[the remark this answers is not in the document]
fsbraun: Test now saves user and double checks username and that the password is not usable.

django/contrib/admin/templates/admin/auth/user/add_form.html
```diff
@@ -8,3 +8,11 @@
     <p>{% translate "Enter a username and password." %}</p>
   {% endif %}
 {% endblock %}
+{% block extrahead %}
+     {{ block.super }}
+     <script src="{% static 'admin/js/unusable_password_field.js' %}" async></script>
+{% endblock %}
```
nessita, line 14:
```suggestion
{% block extrahead %}
  {{ block.super }}
  <script src="{% static 'admin/js/unusable_password_field.js' %}" async></script>
{% endblock %}
```

django/contrib/auth/forms.py
```diff
@@ -138,7 +162,11 @@ def _post_clean(self):
 
     def save(self, commit=True):
         user = super().save(commit=False)
-        user.set_password(self.cleaned_data["password1"])
+        if self.cleaned_data["usable_password"]:
+            password = self.cleaned_data["password1"]
+            user.set_password(password)
```
nessita, line 167:
```suggestion
            user.set_password(self.cleaned_data["password1"])
```

docs/topics/auth/default.txt
```diff
@@ -1831,7 +1838,12 @@ Creating users
 You should see a link to "Users" in the "Auth"
 section of the main admin index page. The "Add user" admin page is different
 than standard admin pages in that it requires you to choose a username and
-password before allowing you to edit the rest of the user's fields.
+password (or setting the password as unusable) before allowing you to edit the rest of
```
nessita, line 1841: I forgot to mention that docs should always be wrapped at 79cols, so this line may need an adjustment.

nessita (commented):
I did another pass at the code. I would like to find a way to avoid the logic duplication between the base user creation form and the admin password change form. I know the duplication was there already, but with the adding of "usable_password", it feels like we should revisit the code and consider a refactoring.

Perhaps we could consider a "set password mixin" to be used in both form? I can give that a try but is unlikely I can do it before next week.

What do you think?

fsbraun:
I'll give it a try. There's an issue with postgres selenium tests which I cannot reproduce.

sarahboyce:
> I'll give it a try. There's an issue with postgres selenium tests which I cannot reproduce.

It appears to be an issue with flakiness rather than to do with your change set 👍

nessita:
> I'll give it a try. There's an issue with postgres selenium tests which I cannot reproduce.

Thanks! re: selenium, see Sarah's comment (test is flaky).

I'll do another UI review now. If possible, would you also, please, rebase onto main?

nessita:
Hey Fabian, I did a UI review as promised. This is looking good! Thank you for being so responsive and open to our feedback.

Before going any deeper in a potential "password change mixin", I'd like to explore a different path for the user details/edit page. I was thinking that we could consider leaving the change password form as is, and instead have a new, simpler, form for just setting the password as unusable.

So, when viewing user details, the admin user is presented with a legend similar to this:

If the user has a usable password: "Raw passwords are not stored, so there is no way to see this user’s password, but you can change the password using <this form>, or disable authentication via <this form>".

If the user has a unusable password: "Re-enable authentication for this user by setting a password using <this form>"

The user creation form will remain as proposed in this PR.

I know we have been going back and forth with this PR, and I'm deeply thankful for your patience, hard work and willingness to adapt. At the same time, I'm evaluating these user experience adjustments as these pages are very high profile and it's important we get this right for both new users and advanced users.

What do you think?

nessita:
Hey @fsbraun, I see that your pushed the changes for the password mixin, thanks! Perhaps let's hold on my last UI change proposal until I review this latest revision, I would not want you to keep investing in this without more concrete feedback from me.

Are you in the Discord Django channel? I was wondering if a sync chat would help the (my?) thinking process.

fsbraun:
@nessita Just pushed the commit including the Mixin before I saw your thoughts on different forms. Here's my 5 cents on the option to offer two different password change forms:

* On the plus side: I find the text in the user change form explanation texts better. They are also closer to the "competing" active switch for the user. I'd slightly change them to:
  * Raw passwords are not stored, so there is no way to see this user’s password, but you can change the password using <this form>, or disable **password-based** authentication via <this form>"
  * Re-enable **password-based** authentication for this user by setting a password using <this form>
* A potential minus: The click-flow will be different for creating a user and changing a user's password

To me, this is not a clear call. Given the repeated discussion about active users and users with an unusable password, I'd probably nudge to separating the forms.

fsbraun:
@nessita Can you share a link for the discord channel?

nessita:
> @nessita Can you share a link for the discord channel?

I think so, I'm new to discord so I think I got the right link :-D:

https://discord.gg/WHW9YkFg

I'll be relocating for about an hour, but I'll be online afterwards. I would suggest joining the `#contributor-discussion` channel, where we specifically discuss Django contributions. I'm nessita there as well!

fsbraun:
@nessita Thinking a bit more, I come to the conclusion that I would keep the single password change form **and** change the message on the user change form. Something like this:

  * Raw passwords are not stored, so there is no way to see this user’s password, but you can change the password or disable password-based authentication via `<this form>`"
  * Re-enable password-based authentication for this user by setting a password using `<this form>`

fsbraun:
Hi @nessita ! I will give it a stab next week.

thibaudcolas:
@nessita @fsbraun based on your request, we’ve reviewed this together with Wagtail UI designers & UI developers @benenright, @laymonage, @nicklee and myself. 

Our specific goal was to provide:

> Review IA of the fields + help text as-is in Fabian’s PR
>  Some thoughts on "one form vs. two forms"

---

So their guidance is:

1. Use the same "one form with conditional fields" pattern for both "add" and "edit" for consistency and clarity. With separate "set password" and "remove password" forms, the individual forms would be clearer, but having to choose between forms (or understand there are two forms) would add confusion.
2. On the forms, switch from a "Set password" checkbox to two radio buttons, so both alternatives can be more clearly labelled.
3. As far as field order, makes sense for the "Set password" field to be where it is now.
4. The "set password: true" should always be selected by default, as that’s the most likely intent for both a new user, and editing an existing user regardless of that user’s current status.

And additionally, change the wording and placement of this "this form" link so it’s more obvious what the link takes you to. Possibly use a button rather than a link in text, with that button label being clear about the form’s purpose.

I didn’t ask them too much input on exact wording but they did say:

- The help text needs work
- Consider terms like "Password-less" or "Other authentication mechanism"
- Even better if we could add "most common use cases" scenarios as examples in the help text, like "Use this for SSO" or similar

I’ll try to get back to the wording on my own based on their input and propose specific tweaks.

---

Congratulations @benenright & @nicklee on a first contribution to Django 😉

fsbraun:
Thank you so much, @thibaudcolas, @benenright, and @nicklee!

django/contrib/auth/forms.py
```diff
@@ -194,6 +238,11 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         password = self.fields.get("password")
         if password:
+            if self.instance and not self.instance.has_usable_password():
+                password.help_text = _(
+                    "Enable password-based authentication for this user by setting "
+                    'a password using <a href="{}">this form</a>.'
```
smithdc1, line 244: https://design-system.service.gov.uk/components/text-input/#hint-text

What do you think about avoiding links in help text?

> Do not include links within hint text. While screen readers will read out the link text when describing the field, they will not tell users that the text is a link.

fsbraun: @smithdc1 I personally think this should be a **button** but did not want to interfere with #17419.

nessita: @smithdc1 I kind of agree with you but if we remove the link from the help text, how/where can we expose the change password functionality? Any ideas or suggestions? (in any case, I would do that in a separated ticket/PR)

docs/releases/5.1.txt
```diff
@@ -32,7 +32,9 @@ Minor features
 :mod:`django.contrib.admin`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-* ...
+* When adding a new user or changing a password in the admin UI, the
+  administrator can chose to set an unusable password effectively disabling
+  a user from logging in with a password.
```
nessita, line 37:
```suggestion
* Enhanced User's password management now allows administrators to enable or
   disable password-based authentication for any user by setting a usable or
   unusable password. This applies to both to user creation and user change
   forms.
```

docs/topics/auth/default.txt
```diff
@@ -1632,7 +1632,14 @@ provides several built-in forms located in :mod:`django.contrib.auth.forms`:
 
 .. class:: AdminPasswordChangeForm
 
-    A form used in the admin interface to change a user's password.
+    A form used in the admin interface to change a user's password. This form
+    also allows the password to be set as an :meth:`unusable password
+    <django.contrib.auth.models.User.set_unusable_password>` effectively
+    barring the user from logging in.
```
nessita, line 1638:
```suggestion
    A form used in the admin interface to change a user's password, including
    the ability to set an :meth:`unusable password
    <django.contrib.auth.models.User.set_unusable_password>`, effectively
    barring the user from logging in using a password-based authentication.
```

docs/topics/auth/default.txt
```diff
@@ -1845,7 +1852,12 @@ Creating users
 You should see a link to "Users" in the "Auth"
 section of the main admin index page. The "Add user" admin page is different
 than standard admin pages in that it requires you to choose a username and
-password before allowing you to edit the rest of the user's fields.
+password (or setting the password as unusable) before allowing you to edit the
+rest of the user's fields.
+
+    .. versionchanged:: 5.1
+
+        Checkbox option to set an unusable password was added.
```
nessita, line 1860:
```suggestion
    .. versionchanged:: 5.1

        Option to set an unusable password was added.
```

nessita (commented):
I've done yet another UX/string review. Thank you for the changes! :balloon: 

In all honesty I don't like radio buttons for when there is a True/False choice, but I'll abide with the advice of the design experts. With that in mind, I most certainly would change:

* "_Password-based login_" to be consistently "_Password-based authentication_" (and ensure that we use this wording everywhere, when referencing the same thing)
* The radio buttons labels for this option should be "_Enabled_" and "_Disabled_", so:
```
Password-based Authentication: [X] Enabled [ ] Disabled
```
* The sentence "_Potential other authentication mechanisms such as Single Sign-On remain unaffected._ " should be rephrased to:
"_Other authentication mechanisms, such as Single Sign-On or LDAP, remain unaffected._"
* I would suggest that we remove both help texts for the new radio button field, because I think they pollute the UI without adding much clarity (see next point).
* I would slightly change the phrase "_Not setting a password will render the login credentials unusable. The user will no longer be able authenticate using a password until a new password is set._"  to:
"_Not setting a password will render the login credentials unusable. The user will not be able to authenticate using a password until a new password is set._" and use this as the yellow warning main text when "Password-base Authentication" is **disabled**.
* I would join all warning sentences into one paragraph except for "_The user's current password will be lost._", which I think needs to be a separated, red-like-warning message, because is a destructive action.

@fsbraun if you could work on the changes above, that would be great, otherwise no worries, I can do those myself next week.

docs/releases/5.1.txt
nessita, line 37: I think this would be best placed under the `django.contrib.auth` subsection. Also, I'd suggest a rewording as per:
```suggestion
* Improved password management in the User model now allows admins to enable or
  disable password-based authentication by setting a usable or unusable        
  password, applicable in user creation and change forms.                      
```

django/contrib/auth/forms.py
fsbraun: @nessita I'd make it a button, just blow the help text. It'll be a simple string change that can be easily added to this PR. It would look like this: 
image
I can add this here or create a separate mini-PR. What's your preference?

thibaudcolas: I think this would be an amazing change. Whether it’s a separate PR or not – not my call.

django/contrib/auth/forms.py
```diff
@@ -180,7 +219,7 @@ class UserChangeForm(forms.ModelForm):
         label=_("Password"),
         help_text=_(
             "Raw passwords are not stored, so there is no way to see this "
-            "user’s password, but you can change the password using "
+            "user’s password, but you can change or unset the password via "
```
thibaudcolas, line 222: 👍 seems like a reasonable adjustment to me.

django/contrib/admin/templates/admin/auth/user/unusable_password_warning.html
```diff
@@ -0,0 +1,19 @@
+{% load i18n %}
+<template id="id_unusable_warning_template">
+  <ul id="id_unusable_warning" class="messagelist">
+    <li class="warning">
+      {% blocktranslate %}
+        Not setting a password will render the login credentials unusable.
+        The user will not be able to authenticate using a password until a
+        new password is set. Other authentication mechanisms, such as
+        Single Sign-On or LDAP, remain unaffected.
```
thibaudcolas, line 9: I think the first sentence just adds confusion so should definitely be removed. Starting with "not" makes it tough to read, and "login credentials" will be lost on people. So at least, do this:

```suggestion
        The user will not be able to authenticate using a password until a
        new password is set. Other authentication mechanisms, such as
        Single Sign-On or LDAP, remain unaffected.
```

I’d go further and use even simpler language. Don’t care too much about the exact wording but the specific changes:

- "authenticate" / "authentication mechanism" -> "log in"
- "remain unaffected" -> needs to use simpler language too.

Or even simpler:

```suggestion
        The user will not be able to log in with a password.
        They may still be able to log in through other options, such as Single Sign-On or LDAP.
```

thibaudcolas (changes requested):
Just reviewing the copy and nothing else – seems like a clear improvement to me over previous iterations. I’d suggest going further. I’d really like to see changes to `unusable_password_warning.html`. For everything else here – I’d suggest going further but I don’t feel strongly about it.

## Field label

Currently this is:

- Password-based authentication
    - 🔘 Enabled
    - ⭘ Disabled

I think this is an improvement over the previous "Set password" checkbox, but we could go further and make more use of having descriptive labels.

- Authentication type
    - 🔘 Password
    - ⭘ Other (no password set)

## Unusable password

I don’t feel strongly about this – but we did discuss that this was just an implementation detail, so I question whether it should even be in the release notes / docs, aside from the method name.

## "Change password" form title

Not a big deal for me but I don’t get why we would want different form titles for those two options, as the form technically allows both types of changes. Why not always a generic "Set password", or an open-ended "Change or set password"?

One exception would be the form’s submit button, where I think it’s great we’re swapping depending on what the form would do in its current state.

django/contrib/auth/forms.py
thibaudcolas: I think this would be an amazing change. Whether as a separate PR or not – not my call.

fsbraun:
@thibaudcolas Thanks so much! I pushed an improved `unusable_password_warning.html` based on your second suggestion (keeping the term "authenticate" instead of "log in" as @nessita earlier pointed out the need for consistent language). It now reads:

```
The user will not be able to authenticate using a password. 
They may still be able to authenticate using other options, 
such as Single Sign-On or LDAP.
```

Also, I relabelled the radio buttons to
```
Authentication type:   [X] Password   [ ] Only other (no password)
```
I was struggling a bit to avoid the impression that you'd have to select "Other" for other authentication backends to work, while in fact they are independent.

django/contrib/auth/forms.py
fsbraun: @smithdc1 @thibaudcolas I created this PR #17489 to introduce the set/change password button.

django/contrib/admin/templates/admin/auth/user/unusable_password_warning.html
nessita: @thibaudcolas the thing is that "log in" is not (strictly) [the same](https://english.stackexchange.com/questions/80862/difference-between-log-in-authenticate-and-sign-in) as "authenticate". One could authenticate via other means (think about [Basic Authentication](https://en.wikipedia.org/wiki/Basic_access_authentication)) and still not do a "log in" flow.

This new UI affects how authentication works, not necessarily the login process.

nessita:
> ```
> Authentication type:   [X] Password   [ ] Only other (no password)
> ```
> I was struggling a bit to avoid the impression that you'd have to select "Other" for other authentication backends to work, while in fact they are independent.

@thibaudcolas @fsbraun Precisely for this reason,  I find `Other` confusing. OTOH, `Only other`, while more accurate, feels cryptic and that it conveys an effect with unclear side-effects.

In my view, we should use as context the default authentication backend, avoiding mixed references with other backends that may or may not exist. This prevents confusion, especially when users may not be aware of other backends. Even more, an admin user should be able to disable password-based authentication altogether, even if no other auth backend is configured, so saying `Other`  could hint that you can't do that *unless* you define other auth backend.

I stand by my original suggestion which limits the messaging to the authentication backend that makes use of the password:

```
Password-based authentication:  [x] Enabled  [ ] Disabled
```

thibaudcolas:
👍 fair enough :)

fsbraun:
@thibaudcolas Great! I've reverted the radio buttons as @nessita suggested. 🚀. From my side, this is read for take-off 😎.

nessita:
> @thibaudcolas Great! I've reverted the radio buttons as @nessita suggested. 🚀. From my side, this is read for take-off 😎.

Understood, will review again soon! Thanks for your constant work and amazing predisposition.

django/contrib/auth/forms.py
```diff
@@ -89,28 +89,86 @@ def widget_attrs(self, widget):
...
+            user = super().save(commit=False)
+        if self.cleaned_data["usable_password"] == "True":
+            user.set_password(self.cleaned_data["password1"])
+        else:
+            user.set_unusable_password()
+        if commit:
+            user.save()
+            if hasattr(self, "save_m2m"):
+                self.save_m2m()
+        return user
```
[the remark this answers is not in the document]
fsbraun: Great idea!

django/contrib/admin/templates/admin/auth/user/change_password.html
```diff
@@ -11,10 +11,14 @@
 &rsaquo; <a href="{% url 'admin:app_list' app_label=opts.app_label %}">{{ opts.app_config.verbose_name }}</a>
 &rsaquo; <a href="{% url opts|admin_urlname:'changelist' %}">{{ opts.verbose_name_plural|capfirst }}</a>
 &rsaquo; <a href="{% url opts|admin_urlname:'change' original.pk|admin_urlquote %}">{{ original|truncatewords:"18" }}</a>
-&rsaquo; {% translate 'Change password' %}
+&rsaquo; {% if form.has_usable_password %}{% translate 'Change password' %}{% else %}{% translate 'Set password' %}{% endif %}
```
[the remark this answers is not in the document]
fsbraun: Added a regression test.

fsbraun:
@nessita I made the changes, but kept the JS as discussed.

fsbraun:
@nessita Any news on this?

django/contrib/admin/templates/admin/auth/user/unusable_password_warning.html
```diff
@@ -0,0 +1,18 @@
...
+        They may still be able to authenticate using other options,
+        such as Single Sign-On or LDAP.
+      {% endblocktranslate %}
+    </li>
+    {% if form.user.has_usable_password %}
+      <li class="error">
+        {% translate "The user's current password will be lost. " %}
+      </li>
+    {% endif %}
+
```
nessita, line 16: Could we remove this empty line?
```suggestion
```

django/contrib/admin/templates/admin/auth/user/unusable_password_warning.html
```diff
@@ -0,0 +1,18 @@
+{% load i18n %}
+<template id="id_unusable_warning_template">
+  <ul id="id_unusable_warning" class="messagelist">
+    <li class="warning">
+      {% blocktranslate %}
+        The user will not be able to authenticate using a password.
+        They may still be able to authenticate using other options,
```
nessita, line 7:
```suggestion
        They may still be able to authenticate using other backends,
```

django/contrib/auth/forms.py
```diff
@@ -180,7 +220,7 @@ class UserChangeForm(forms.ModelForm):
         label=_("Password"),
         help_text=_(
             "Raw passwords are not stored, so there is no way to see this "
-            "user’s password, but you can change the password using "
+            "user’s password, but you can change or unset the password via "
```
nessita, line 223:
```suggestion
            "user’s password, but you can change or unset the password using "
```

docs/releases/5.1.txt
```diff
@@ -44,6 +43,10 @@ Minor features
 
 * The default iteration count for the PBKDF2 password hasher is increased from
   720,000 to 870,000.
+* Improved password management in the User model now allows admins to enable or
```
nessita, line 46:
```suggestion

* Improved password management in the User model now allows admins to enable or
```

docs/releases/5.1.txt
```diff
@@ -32,7 +32,6 @@ Minor features
 :mod:`django.contrib.admin`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-* ...
```
nessita, line 35: I think we need to restore this 3 dots.

docs/releases/5.1.txt
```diff
@@ -44,6 +43,10 @@ Minor features
 
 * The default iteration count for the PBKDF2 password hasher is increased from
   720,000 to 870,000.
+* Improved password management in the User model now allows admins to enable or
+  disable password-based authentication by setting a usable or unusable
+  password, applicable in user creation and change forms.
+
```
nessita, line 49: Only one empty line afterwards
```suggestion
```

docs/topics/auth/default.txt
```diff
@@ -1632,7 +1632,14 @@ provides several built-in forms located in :mod:`django.contrib.auth.forms`:
...
 
-    A form used in the admin interface to change a user's password.
+    A form used in the admin interface to change a user's password, including
+    the ability to set an :meth:`unusable password
+    <django.contrib.auth.models.User.set_unusable_password>`, effectively
+    barring the user from logging in using password-based authentication.
+
+    .. versionchanged:: 5.1
+
+        Option to disable password-based authentication was added.
```
nessita, line 1642:
```suggestion
        Option to disable (or re-enable) password-based authentication was added.
```

docs/topics/auth/default.txt
```diff
@@ -1845,7 +1852,12 @@ Creating users
 You should see a link to "Users" in the "Auth"
 section of the main admin index page. The "Add user" admin page is different
 than standard admin pages in that it requires you to choose a username and
-password before allowing you to edit the rest of the user's fields.
+password (or setting the password as unusable) before allowing you to edit the
+rest of the user's fields.
+
+    .. versionchanged:: 5.1
+
+        Option to disable password-based authentication was added.
```
nessita, line 1860:
```suggestion
        Option to disable (or re-enable) password-based authentication was added.
```

tests/admin_views/test_password_form.py
```diff
@@ -0,0 +1,73 @@
...
+        # Finally check that the warning is displayed
+        warning = self.selenium.find_element(By.ID, "id_unusable_warning")
+        self.assertTrue(warning.is_displayed())
+
+        # Switch usable_password selector once more
+        pw_switch_on.click()
+
+        # Warning disappears
+        with self.assertRaises(StaleElementReferenceException):
+            self.assertFalse(warning.is_displayed())
```
nessita, line 73: This test is great, we'd need a couple more for:
* existing user starting with usable password
* existing user starting with non usable password

tests/admin_views/test_password_form.py
```diff
@@ -0,0 +1,73 @@
...
+        )
+        password1 = self.selenium.find_element(
+            By.CSS_SELECTOR, 'input[name="password1"]'
+        )
+        password2 = self.selenium.find_element(
+            By.CSS_SELECTOR, 'input[name="password2"]'
+        )
+
+        # For add user the default is a usable password
+        self.assertTrue(pw_switch_on.is_selected())
```
nessita, line 46: From [the contributing guide](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/), in general is preferred to use:
```suggestion
        self.assertIs(pw_switch_on.is_selected(), True)
```

tests/admin_views/test_password_form.py
```diff
@@ -0,0 +1,73 @@
...
+
+        # Check the click did switch the checkbox
+        self.assertFalse(pw_switch_on.is_selected())
+        self.assertTrue(pw_switch_off.is_selected())
+
+        # Hence the password fields should be hidden now
+        self.assertFalse(password1.is_displayed())
+        self.assertFalse(password2.is_displayed())
+
+        # Finally check that the warning is displayed
```
nessita, line 64: I think we should also be testing for the expected button legend, since we are changing it via javascript?

tests/auth_tests/test_views.py
```diff
@@ -1440,14 +1440,35 @@ def test_user_change_password(self):
...
         self.assertEqual(urljoin(user_change_url, rel_link), password_change_url)
 
+        response = self.client.get(password_change_url)
+        # Test the form title with original (usable) password
+        self.assertContains(response, "<h1>Change password")  # Title
+        self.assertContains(response, "&rsaquo; Change password")  # Breadcrumb
+        self.assertNotContains(response, "<h1>Set password")
+        self.assertNotContains(response, "&rsaquo; Set password")
+
+        # Test for title with unusable password with a test user
```
nessita, line 1455: Could we split this in two tests? one starting with a user with a usable password, and the other for a user without.

django/contrib/admin/templates/admin/auth/user/unusable_password_warning.html
fsbraun: Whole file is gone now... 😉. It is not needed any more if CSS manages what is viewable.

django/contrib/auth/forms.py
fsbraun: Changed.

docs/releases/5.1.txt
fsbraun: Done.

tests/admin_views/test_password_form.py
fsbraun: Since JS is gone now, this can be checked by the Python tests.

tests/auth_tests/test_views.py
fsbraun: Done.

tests/admin_views/test_password_form.py
fsbraun: Done.

tests/admin_views/test_password_form.py
fsbraun: … and just pushed.

django/contrib/admin/templates/admin/auth/user/unusable_password_warning.html
```diff
@@ -0,0 +1,18 @@
...
+        such as Single Sign-On or LDAP.
+      {% endblocktranslate %}
+    </li>
+    {% if form.user.has_usable_password %}
+      <li class="error">
+        {% translate "The user's current password will be lost. " %}
```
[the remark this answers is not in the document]
fsbraun: File is gone. Changed here: 
https://github.com/fsbraun/django/blob/7c1e0b7c16b079b6f3311eed94d48784af697d5e/django/contrib/auth/forms.py#L101-L106

tests/admin_views/tests.py
```diff
@@ -7192,6 +7192,7 @@ def test_save_button(self):
             reverse("admin:auth_user_add"),
             {
                 "username": "newuser",
+                "usable_password": True,
```
ngnpope, line 7195: Following on from my comment earlier, this is relying on `str(True)` rather than an explicit value. As these are values of a query string, I'd expect them to be in serialized form, i.e. `"True"`, that matches the actual value used in the form choices as defined in `SetPasswordMixin`. _(Although maybe `"1"` as for `_continue` below would be more conventional/consistent?)_

ngnpope (changes requested):
Thanks for your continued efforts on this one. It's been a while since I dropped in - this is shaping up to be a very nice improvement 👍🏻

django/contrib/admin/static/admin/js/unusable_password_field.js
```diff
@@ -0,0 +1,25 @@
+"use strict";
+// Fallback JS for browsers which do not support :has selector used in
+// admin/css/unusable_password_fields.css
+// Remove file once all supported browsers support :has selector
+try {
+    // If browser does not support :has selector this will raise an error
+    document.querySelector("form:has(input)");
+} catch (error) {
```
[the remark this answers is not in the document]
ngnpope: But I'll defer to @nessita's decision on this 😄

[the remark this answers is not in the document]
felixxm: I disagree, what you're proposing is a really aggressive policy :fire: . We cannot drop support for 1 months old Firefox releases just because the latest one supports `has()`. I (for example) have 119 locally :older_adult:. We need to remember that admin is for app users and not all of them bump Firefox daily.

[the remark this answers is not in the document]
ngnpope: I was looking at this with a lens of what the state of things will look in 8 months time rather than right now. In that regard we'll [be expecting](https://whattrainisitnow.com/calendar/) v129 to be the latest version[^1]. _(Also v128 will be the next ESR and, yes, there'll be one more release of v115 ESR to come in September before it hits end-of-life in October.)_

Our documentation states "recent modern versions" and I get that this is quite nebulous, but I would hardly call what I proposed "really aggressive" in light of that. I will concede though that, in this case, it's likely to cause more than "minor stylistic differences" to the feature if the browser doesn't have the required support.

Regardless, I was just making a case that including this seemed pointless and, if that's rejected, it's fine ❤️

[^1]: I certainly hope you've upgraded to something more recent than v119 by then for your own security 😛

django/contrib/auth/forms.py
```diff
@@ -89,50 +89,112 @@ def widget_attrs(self, widget):
...
     def __init__(self, *args, **kwargs):
+        extra_warning = kwargs.pop("extra_warning", None)
         super().__init__(*args, **kwargs)
-        if self._meta.model.USERNAME_FIELD in self.fields:
-            self.fields[self._meta.model.USERNAME_FIELD].widget.attrs[
-                "autofocus"
-            ] = True
+        warnings = f'<li class="warning">{self.unusable_warning}</li>'
+        if extra_warning:
+            warnings += f'<li class="error">{extra_warning}</li>'
```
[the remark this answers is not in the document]
fsbraun: The semantically correct levels would be "info" and "warning". But info has a checkbox as an icon, which I find misleading. But that's representation. It's probably best to stay with the semantics. @nessita What do you think?

django/contrib/auth/forms.py
```diff
@@ -89,50 +89,112 @@ def widget_attrs(self, widget):
...
+    )
+
+    usable_password = forms.ChoiceField(
+        label=_("Password-based authentication"),
+        required=False,
+        initial="True",
+        choices=(
+            ("True", _("Enabled")),
+            ("False", _("Disabled")),
+        ),
```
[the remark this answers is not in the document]
fsbraun: `"true"`/ `"false"` make sense to me and I would prefer it to 1 and 0.

tests/admin_views/tests.py
fsbraun: I'd go for `"true"`; otherwise we have `[value="0"]` and `[value="1"]` as CSS selectors which are not very explicit.

nessita:
Hello @fsbraun! I'm back to this review :muscle: Details below for UI review, code review coming next.

As discussed in Discord, I agree that the need for JS is (at this point) justified and we'll have to wait to 5.2 to remove this JS and then rely solely on the `:has()` CSS pseudo-class as you suggested.

I can make the changes requested below if you are busy, let me know! Thanks again :trophy: 

UI review
-------------
* The separation between the radio buttons feels weird, the options are too apart from eachother. Is there something that can be done?  
image

* In the message shown in the screenshot, I see two issues with how it is displayed right now:
  1. the left margin/padding makes it looks a bit unpolished, and
  2. the fact that is green (usually meaning success) is a bit unexpected for the tone of the message. Could we instead make it a "traditional" help text with its usual formatting?
image

What about something like:
image

With diff:
```diff
diff --git a/django/contrib/auth/forms.py b/django/contrib/auth/forms.py
index a5d7f8017b..f2978bf992 100644
--- a/django/contrib/auth/forms.py
+++ b/django/contrib/auth/forms.py
@@ -98,9 +98,9 @@ class SetPasswordMixin(forms.Form):
     error_messages = {
         "password_mismatch": _("The two password fields didn’t match."),
     }
-    unusable_warning = _(
-        "The user will not be able to authenticate using a password. "
-        "They may still be able to authenticate using other backends, "
+    usable_help_text = _(
+        "Whether the user will be able to authenticate using a password or not. "
+        "If disabled, they may still be able to authenticate using other backends, "
         "such as Single Sign-On or LDAP."
     )
 
@@ -112,6 +112,7 @@ class SetPasswordMixin(forms.Form):
         widget=forms.RadioSelect(
             attrs={"class": "radiolist inline"},
         ),
+        help_text=usable_help_text,
     )
     password1 = forms.CharField(
         label=_("Password"),
@@ -128,17 +129,14 @@ class SetPasswordMixin(forms.Form):
         help_text=_("Enter the same password as before, for verification."),
     )
 
-    def __init__(self, *args, **kwargs):
-        extra_warning = kwargs.pop("extra_warning", None)
+    def __init__(self, *args, extra_warning=None, **kwargs):
         super().__init__(*args, **kwargs)
-        warnings = f'<li class="info">{self.unusable_warning}</li>'
         if extra_warning:
-            warnings += f'<li class="warning">{extra_warning}</li>'
-        self.fields[
-            "usable_password"
-        ].help_text = (
-            f'<ul id="id_unusable_warning" class="messagelist">{warnings}</ul>'
-        )
+            self.fields[
+                "usable_password"
+            ].help_text += (
+                f'<ul id="id_unusable_warning" class="messagelist">'
+                f'<li class="warning">{extra_warning}</li></ul>')
 
     def clean_password1(self):
         password1 = self.cleaned_data.get("password1")
@@ -528,7 +526,7 @@ class AdminPasswordChangeForm(SetPasswordMixin, forms.Form):
     """
 
     required_css_class = "required"
-    password_lost_warning = _("The current password for this user will be lost.")
+    password_lost_warning = _("If disabled, the current password for this user will be lost.")
 
     def __init__(self, user, *args, **kwargs):
         self.user = user
```

User workflow review
------------------------------
1. When javascript is disabled, on user add form, if password-based authentication is disabled, the values from the password fields should be cleaned in a way where no validation is done but form errors are returned indicating that fields should be empty in this case. Right now, when submitting the form with password auth disabled **and** some values in the password forms, the user get form errors regarding the password fields.
2. Password-based authentication value on forms for existing users should default to the correct/set value (currently, visiting "change password" form with a user with unusable password shows a form with "Enabled" chosen by default).
4. Even with javascript enabled, I see these two buttons for an existing user with non usable password:
image

fsbraun:
@nessita Thank you very much! A few remarks/questions:
* The spacing between radio buttons is defined for all forms by the admin style set. I would not change it for this specific form. Maybe that's another thing to change in a separate PR for all of django admin's radio buttons?
* I'm fine with the "success" message going into the help_text. ☑️ **Can you commit your patch?**
* Good catch that the hidden password fields can cause a **form validation failure. I will update the PR accordingly**.
* If a user has an unset password, clicking "set password" should IMHO not prompt them to unset it again. We discussed this here: https://github.com/django/django/pull/16942#issuecomment-1717755980 and I already changed it back and forth. I suggest not changing this.
* How did you create the scenario #3 in user workflow (the two submit buttons visible)? I tried and cannot recreate it even with Firefox 118.

I will check the form validation.

nessita:
> @nessita Thank you very much! A few remarks/questions:
> 
>     * The spacing between radio buttons is defined for all forms by the admin style set. I would not change it for this specific form. Maybe that's another thing to change in a separate PR for all of django admin's radio buttons?

That's fair, and I agree, let's leave this as is, using the shared CSS style.

>     * I'm fine with the "success" message going into the help_text. ☑️ **Can you commit your patch?**

Yes, will do that next, along with a `main` rebase.
 
>     * Good catch that the hidden password fields can cause a **form validation failure. I will update the PR accordingly**.

Yey!

>     * If a user has an unset password, clicking "set password" should IMHO not prompt them to unset it again. We discussed this here: [Fixed #34429 -- Allowed setting unusable password via admin UI. #16942 (comment)](https://github.com/django/django/pull/16942#issuecomment-1717755980) and I already changed it back and forth. I suggest not changing this.

You are absolutely right, sorry for the back and forth. I guess we could (eventually) ask for a second opinion to Thibaud, but we may do this when we are closer to landing.
 
>     * How did you create the scenario no. 3 in user workflow (the two submit buttons visible)? I tried and cannot recreate it even with Firefox 118.

All I did was:
1. Add a user with unusable password, submit
2. Visit this user's details page
3. Click on the "set the password using this link"
4. Two buttons are shown (this is NOT the case for a user with a previously usable password)

Does this help?

> I will check the form validation.

Excellent! Thank you.

fsbraun:
@nessita I just pushed an update. Now, the test for password1 and password2 matching is skipped if an unusable password is to be set.

I cannot, however, reproduce the situation with the two submit buttons. Their visibility should be managed by the JS polyfill. Which browser are you seeing this? Is there any message on the JS console? Do both buttons stay if you flip the radio buttons for a usable password? Does this persist if you empty the browser's cache?

nessita:
Hello @fsbraun! I pushed my changes though the selenium tests need fixing one we decide on the final flow.

Regarding the two buttons for changing the password for a user with unusable password, this is working on Chrome Version 114.0.5735.198 but is not working on Firefox 120.0.1. I tested this on a private window, there errors in the developer console as shown below:
```
Uncaught TypeError: document.getElementById(...) is null
    <anonymous> http://localhost:9000/static/admin/js/unusable_password_field.js:17
    <anonymous> http://localhost:9000/static/admin/js/unusable_password_field.js:23
    <anonymous> http://localhost:9000/static/admin/js/unusable_password_field.js:10
```
See screenshot:
image

nessita:
Double buttons on Firefox is still present even if browser cache is clear.

fsbraun:
@nessita Ah, I see now: The double buttons are a result of your replacing the warnings with the help text: 
* When the polyfill tries to find the warning to set its visibility, it fails. 
* Probably the same for the selenium tests: They must not test that the help text is hidden if it is supposed to be visible all the time.

I can quickly fix both, but am unsure what decision has to be made on the "final flow". Will push the fix and wait for further feedback or instructions.

nessita:
Hey @fsbraun, I'm doing another review now.

In order to help with the review (for myself, and for others), I started isolating the creation of the `SetPasswordMixin` into a logical unit (separated commit). I've found a potential blocker, this line I'm highlighting is removed from `AdminPasswordChangeForm` and not added anywhere:

```
password_validation.validate_password(password2, self.user)
```

Also,  I've noticed that a rebase onto main causes conflicts because older commits cause conflicts which are later fixed by newer commits. Given that there are 71 commits, would you agree to squash commits into isolated logical units? This way, it's easier to review newer commits, and also we'll comply with [making the commit history as usable as possible](https://docs.djangoproject.com/en/5.0/internals/contributing/committing-code/#handling-pull-requests).

Following these two items, would you have time/interest in creating a separated commit (with commit message suffix `Refs #34429`) that refactors the common logic for setting a password into `SetPasswordMixin`?
We would need to evaluate how to add this lost line back in and add a few tests for it (since no test fails when removed). The tests adding would ideally be done in its own isolated PR, to be landed as soon as ready and separated from this work.

Let me know! Thank you.

fsbraun:
@nessita I added back the missing line and regression tests for the future for both creating users and changing a user's password. This should remove the blocker. The set of valid form inputs stays unchanged, but the validation errors shown to the user now are consistent across all the forms again. The relevant edge case is an empty password1 and a non-empty non-validating password2.

I'm strongly supporting to squash all commits of this PR upon merging. With the amount of back and forth we have had in this PR, I doubt looking at single commits will be in any way illuminating. 

I just merged the changes from main back into the PR. So there should not be any conflicts (though I cannot see this in GitHub).

Do we really have to split the PR into two commits? I'd be fine with a single squashed commit. Certainly for me, the procedure would cost considerable time and probably also be error-prone. The benefits to me seem not so great in comparison, as the result is a bit artificial: The mixin solely is a tool to not repeat oneself.

nessita:
> @nessita I added back the missing line and regression tests for the future for both creating users and changing a user's password. This should remove the blocker. The set of valid form inputs stays unchanged, but the validation errors shown to the user now are consistent across all the forms again. The relevant edge case is an empty password1 and a non-empty non-validating password2.

Great, thank you!

> I'm strongly supporting to squash all commits of this PR upon merging. With the amount of back and forth we have had in this PR, I doubt looking at single commits will be in any way illuminating.

I beg to differ, though I understand your view. I can imagine you may find the ongoing round of reviews and sometimes mildly contradictory requests a bit repetitive.
I want to say again that we appreciate your patience, and that PR review is an iterative process. Django source code is of extremely high quality both in terms of correctness and git history, and as reviewers, we need to ensure that it continues that way.

> I just merged the changes from main back into the PR. So there should not be any conflicts (though I cannot see this in GitHub).

Thanks! For future reference, it'd be preferred to do a rebase instead of a merge (`git rebase main`), that way the PR commits are replayed on top of latest `main` and is a bit cleaner/easier to review and later to merge.

> Do we really have to split the PR into two commits?

Yes, I think so.

> I'd be fine with a single squashed commit. Certainly for me, the procedure would cost considerable time and probably also be error-prone. The benefits to me seem not so great in comparison, as the result is a bit artificial: The mixin solely is a tool to not repeat oneself.

I understand if you'd prefer not to tackle this yourself. However, to maintain commits as isolated logical units and anticipate a potential future need for a revert, I believe the mixin refactor (without any logic change) should serve as a sort of prerequisite change. As mentioned in my previous comment, (eventually) consolidating all the changes into a single commit can make the review process more convoluted, cumbersome, and error-prone—it's easier to overlook something.

Because of the above, my plan is to produce and push these 3 commits as time permits:

1. isolate the newly added (missing in main) tests for the password management form
2. do the mixin
3. the actual ticket changes 

Reorganizing these commits will significantly facilitate the review process. It's more straightforward (at least in my experience) to complete reviews for each logical unit, potentially landing those, and then concentrate on the remaining changes with a more focused approach (the classic divide and conquer).

nessita:
> @nessita OK! I just feel not sure how to do that exactly. It might be my limited mastery of git that scares me. (I just clicked update branch on my fork in GitHub, for example. Next time I will run `git rebase main`.)

Not being sure is completely fine! Also do note that you need to update your `main` before doing the rebase.

The big advantage for using rebasing (IMHO) is that your commits always end up at the "top of the stack of commits". For example, `git rebase <ref>` means (I'll handwave a bit, there are tons of articles out there explaining this with more technically correct terms):

* all your commits since `<ref>` are put aside
* repo is updated with latest changes for `<ref>` (this is why you need to update `main` before rebasing)
* the commits are re-played on top of the latest changes 

Doing this consistently and often, ensures that your branch history looks like this:
```
main-hash-1
main-hash-2
...
main-hash-N
your-branch-hash-1
your-branch-hash-2
...
your-branch-hash-M
```
When newer revnos are landed in main, say revnos `main-hash-N+1` and `main-hash-N+2`, after a rebase your branch looks like this:
```
main-hash-1
main-hash-2
...
main-hash-N
main-hash-N+1
main-hash-N+2
your-branch-hash-1
your-branch-hash-2
...
your-branch-hash-M
```

To the contrary, when doing merges from main into your branch, your history starts looking like this:
```
main-hash-1
main-hash-2
...
main-hash-N
your-branch-hash-1
your-branch-hash-2
...
your-branch-hash-M
your-branch-hash-M+1
```

The difference is that `your-branch-hash-M+1` is a single merge commit of both revnos from main ( `main-hash-N+1` and `main-hash-N+2` merged into a single commit). So your branch history incorporates newer changes from `main` but starts diverging quite considerably. 

I can go on and on about this, but I figured you could continue the investigation on your own with these pointers. 

> Isolating the changes to the tests should be easy. There's probably some thinking involved when separating out the mixin.

I agree. I have separated and simplified the test in PR #17729. I removed the testing on both values of password1 because as far as I understood, the fact that password1 would be empty or not was irrelevant for the call:

`password_validation.validate_password(password2, self.user)`

If I'm missing something here, please let me know.

> Out of curiosity, how do you do this? You would need to "undo" all commits while keeping the changes and then redoing the changes in a new order.

What I did was doing multiple *interactive* rebases. An interactive rebase is a rebase as described above but before replaying the commits from your branch, you get a screen to "interact" with git and instruct it to do things, perhaps differently, with the commits that is about to replay.

Specifically for this PR, which had 76 commits before my work, the rebasing was very painful and slow. It took me almost 2 hours to resolve all the conflicts that were arising after every commit from the 76 pending ones was applied.

I think I got the squash and rebase correct. Will push the result in a few.

nessita:
One thing that I noticed while rebasing and fixing conflicts is that the adding in various places (tests) of `"usable_password": "true"` to form data could be removed, since that's the default and I don't think is necessary to set it explicitly in tests that are not relevant to this feature. Let me know if there is a reason to add it!

nessita:
If you wanna give it a shot to splitting the Mixin into its own commit, [this article](https://dev.to/timmouskhelichvili/how-to-split-a-git-commit-into-multiple-ones-3g6f) may help when it comes to finding the necessary git magic incantations.

django/contrib/auth/forms.py
```diff
@@ -89,50 +89,108 @@ def widget_attrs(self, widget):
...
+        else:
+            user.set_unusable_password()
+        if commit:
+            user.save()
+            if hasattr(self, "save_m2m"):
+                self.save_m2m()
+        return user
+
+
+class BaseUserCreationForm(SetPasswordMixin, forms.ModelForm):
```
felixxm, line 176: This class is [documented](https://docs.djangoproject.com/en/dev/topics/auth/default/#django.contrib.auth.forms.BaseUserCreationForm), and its documentation no longer matches the implementation.

docs/topics/auth/default.txt
```diff
@@ -1632,7 +1632,15 @@ provides several built-in forms located in :mod:`django.contrib.auth.forms`:
...
 
-    A form used in the admin interface to change a user's password.
+    A form used in the admin interface to change a user's password, including
+    the ability to set an :meth:`unusable password
+    <django.contrib.auth.models.User.set_unusable_password>`, effectively
+    barring the user from logging in using password-based authentication.
+
+    .. versionchanged:: 5.1
+
+        Option to disable (or re-enable) password-based authentication was
```
felixxm, line 1642: `reenable` is even in the `spelling_wordlist`:
```suggestion
        Option to disable (or reenable) password-based authentication was
```

tests/admin_views/test_password_form.py
```diff
@@ -0,0 +1,145 @@
...
+            password="secret",
+            email="super@example.com",
+        )
+        self.user = User.objects.create_user(
+            username="ada", password="charles", email="ada@example.com"
+        )
+
+    def test_add_user_unusable_password_css(self):
+        """Selecting and deselecting the unusable password field shows/hides the
+        fields and the warning when adding a user"""
```
felixxm, line 25: We don't wrap docstrings this way.

tests/admin_views/test_password_form.py
```diff
@@ -0,0 +1,145 @@
...
+        self.assertIs(password2.is_displayed(), True)
+
+        # and only the set password submit button is visible
+        self.assertIs(submit_set.is_displayed(), True)
+        self.assertIs(submit_unset.is_displayed(), False)
+
+        # Switch usable_password selector
+        pw_switch_off.click()
+
+        # Check the click did switch the checkbox
```
felixxm, line 125: To add more explanation, docstrings should state the expected behavior and omit prefixes like "Check" since all tests check things. This guideline is from [Python coding style](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#python-style).

tests/auth_tests/test_views.py
```diff
@@ -1440,14 +1440,24 @@ def test_user_change_password(self):
...
         # Test the link inside password field help_text.
         rel_link = re.search(
-            r'you can change the password using <a href="([^"]*)">this form</a>',
+            r'change or unset the password using <a href="([^"]*)">this form</a>',
             response.content.decode(),
         )[1]
         self.assertEqual(urljoin(user_change_url, rel_link), password_change_url)
 
+        response = self.client.get(password_change_url)
+        # Test the form title with original (usable) password
```
felixxm, line 1449: To add more explanation, comments should state the expected behavior and omit prefixes like "Test the" since all tests test things. This guideline is from [Python coding style](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#python-style).

django/contrib/auth/forms.py
fsbraun: @felixxm Thanks for pointing this out. I've updated the documentation.

fsbraun:
@nessita Did I do this OK?

nessita:
> @nessita Is this what you expected?

Yes, quite close! Could you please adjust the Mixin docstring to not mention the unusable password bit in the first commit? That piece should only be added in the second commit.

fsbraun:
@nessita There you go!! 😎😃

Thanks to @thibaudcolas for his patient introduction to interactive rebase.

django/contrib/admin/static/admin/js/unusable_password_field.js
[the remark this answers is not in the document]
nessita: I have thoroughly tested the three paths:
1. Browser with `has` support
2. No `has` support and JS enabled
3. No `has` support and JS disabled

User experience is quite decent in all cases (after I made some tweaks to the forms and the view flows). I think we can safely land as is and cleanup in a future iteration. Thank you!

nessita:
Hey @fsbraun, thanks for your patience. Last week and today I've been working on final tweaks for this PR. I have reviewed logic and tests in detail, making some adjustments to the UI and UX, and adding more tests to ensure proper coverage of changes.

Tomorrow I'll give the docs another pass and hopefully I will approve and merge. :crossed_fingers:

nessita:
I believe this is ready for merge, will do so later today after the CI runs complete.

PR #16951, opened 2023-06-06
Fixed #34638 -- Fixed admin change list selected row highlight on editable boolean fields.

sainipray:


shangxiao (approved):
Thanks! 🏆

shangxiao:
@sainipray hey could you do us a favour and update the commit message? 😊

Should be in the format: `Fixed #<ticket-number> -- <Message in past tense>` - very similar to what you have in the PR title except just say what was fixed in past tense 😉

I've marked it as a release blocker because it's a regression in 4.2 which means we'll also need an entry in the 4.2.3 release notes.

shangxiao:
Also we tend to thank the reporter in the commit message (see commit log for examples) if you felt like doing that… otherwise a merger will update the message.

shangxiao:
Ideally we should also have a selenium test for this. Doesn't seem like there were any added for this particular CSS-only featurette but now might be a good time to think about adding some? 🤔

felixxm:
This is not the right solution as it highlight only action checkboxes:
image

django/contrib/admin/static/admin/css/changelists.css
```diff
@@ -261,7 +261,7 @@
     background-color: var(--selected-row);
 }
 
-#changelist table tbody tr:has(input[type=checkbox]:checked) {
+#changelist table tbody tr .action-checkbox:has(input[type=checkbox]:checked) {
```
felixxm, line 264: As far as I'm aware, we should use
```suggestion
#changelist table tbody tr:has(input.action-select[type=checkbox]:checked) {
```
Also the same selector must be updated for `forced-colors`:
https://github.com/django/django/blob/7cc138a58f73c17f07cfaf459ef8e7677ac41ac0/django/contrib/admin/static/admin/css/changelists.css#L272

sainipray: Let me check and update this.

shangxiao: Ah I see now… you can get the same behaviour by deleting the CSS declaration above with the comment "Once the :has() is supported… the tr.selected can be removed".

The correct behaviour I observed was due to that ^ polyfill for FF selecting the rest of the tr.

sainipray: @felixxm you're right, I tried and it was looking same as your screenshot,  So updated code that you suggested and remove table tag from CSS for unnecessarily nesting selectors

felixxm:
@sainipray Thanks :+1: I added release notes.

> Ideally we should also have a selenium test for this. Doesn't seem like there were any added for this particular CSS-only featurette but now might be a good time to think about adding some?

@shangxiao I have no idea how we could test it, or in general how we could test any CSS logic :exploding_head: . Suggestions are welcome :gift: . As for me, we can merge it without a regression test :shrug: .

felixxm:
buildbot, test on selenium.

PR #16963, opened 2023-06-09
Refs #34043 -- Added GitHub action to capture screenshots in Selenium tests.

sarahboyce:
This is a proposal to help us confidently make UI changes. :muscle: 
Testing UI is always hard and many of us would not describe ourselves as frontend developers. Adding a layer of visual testing to our automated test suite will not only make our lives easier but can open up possibilities to us such as tests to cover visual regressions.

In [ticket 34043](https://code.djangoproject.com/ticket/34043), there is a suggestion that we should have a test suite/tooling to keep track of UI changes automatically. 

This PR adds the screenshot tests to our GitHub actions. This is currently triggered when a PR has the `selenium` label.
You can see the screenshots (in a zip file at the bottom called screenshots) here: https://github.com/django/django/actions/runs/5386413722?pr=16963

The action "fails" because I do not have write permissions to the repo. please see this comment https://github.com/django/django/pull/16963#discussion_r1243194672. I have tested this against my fork where I do have write permissions. This should be resolved on merge.

### DIFF
--- a/.github/workflows/screenshots.yml
+++ b/.github/workflows/screenshots.yml
@@ -1,0 +1,77 @@
+name: Visual Regression Tests
+
+on:
+  pull_request:
+    types: [labeled, synchronize, opened, reopened]
+    paths-ignore:
+      - 'docs/**'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  selenium-screenshots:
+    if: contains(join(github.event.pull_request.labels.*.name, '|'), 'screenshots')
+    runs-on: ubuntu-latest
+    name: Screenshots
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+          cache: 'pip'
+          cache-dependency-path: 'tests/requirements/py3.txt'
+      - name: Install and upgrade packaging tools
+        run: python -m pip install --upgrade pip setuptools wheel
+      - run: python -m pip install -r tests/requirements/py3.txt -e .
+      - name: Run Selenium tests with screenshots
+        id: generate-screenshots
+        working-directory: ./tests/
+        run: |
+          python -Wall runtests.py --verbosity 2 --noinput --selenium=chrome --headless --screenshots --settings=test_sqlite --parallel 2
+          echo "date=$(date)" >> $GITHUB_OUTPUT
+          echo "🖼️ **Screenshots created**" >> $GITHUB_STEP_SUMMARY
+          echo "Generated screenshots for ${{ github.event.pull_request.head.sha }} at $(date)" >> $GITHUB_STEP_SUMMARY
+
+      - name: Upload screenshots
+        uses: actions/upload-artifact@v3
+        with:
+          name: screenshots-${{ github.event.pull_request.head.sha }}
+          path: tests/screenshots/
+
+      - name: Find comment to update
+        uses: peter-evans/find-comment@v2
+        id: find-comment
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          comment-author: 'github-actions[bot]'
+          body-includes: You can download the generated screenshots from the workflow artifacts.
+
+      - name: Create comment
+        if: steps.find-comment.outputs.comment-id == ''
+        uses: peter-evans/create-or-update-comment@v3
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          body: |
+            🖼️ **Screenshots created**
+
+            You can download the generated screenshots from the workflow artifacts.
+
+            _Please note that artifacts are only available for download for ${{ github.retention_days }} days._
+
+            - Generated screenshots for ${{ github.event.pull_request.head.sha }} at ${{ steps.generate-screenshots.outputs.date }} ([download](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})).
+
+      - name: Update comment
+        if: steps.find-comment.outputs.comment-id != ''
+        uses: peter-evans/create-or-update-comment@v3
+        with:
+          comment-id: ${{ steps.find-comment.outputs.comment-id }}
+          body: |
+            - Generated screenshots for ${{ github.event.pull_request.head.sha }} at ${{ steps.generate-screenshots.outputs.date }} ([download](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})).


smithdc1:
I'm not sure if this is helpful? Maybe you've already seen it!

https://simonwillison.net/2022/Mar/10/shot-scraper/

sarahboyce:
> I'm not sure if this is helpful? Maybe you've already seen it!
> 
> https://simonwillison.net/2022/Mar/10/shot-scraper/

I haven't but it's quite similar (I am also using playwright here but using the python version because Python = :heart:), will take a look!
I am trying also take inspiration from Adam's blog (https://adamj.eu/tech/2023/03/17/django-parameterized-tests-model-admin-classes/) around testing the admin and what I want in the end is something that will give you just changed screenshots with before and after so hopefully the css changes are easier when seeing unexpected changes (visual regressions are just difficult)

As an example, you can download and see screenshots here in the Artifacts section at the bottom: https://github.com/django/django/actions/runs/5225589580, it's not quite what I want yet but you get the idea. Need to check with people if we think this adds value.

.github/workflows/selenium_tests.yml
```diff
@@ -0,0 +1,34 @@
...
+          python-version: 3.11
+          cache: 'pip'
+          cache-dependency-path: 'tests/requirements/selenium.txt'
+      - name: Install and upgrade packaging tools
+        run: python -m pip install --upgrade pip setuptools wheel
+      - run: python -m pip install -r tests/requirements/selenium.txt .
+
+      - name: Run selenium tests for current PR
+        working-directory: ./tests/
+        run: python runtests.py --selenium=chrome --headless --parallel 2 admin_changelist admin_inlines admin_widgets forms_tests
```
sarahboyce, line 28: This is stating which tests to run because without it I was getting database threading errors
```
django.db.utils.DatabaseError: DatabaseWrapper objects created in a thread can only be used in that same thread. The object with alias 'default' was created in thread id 139881739424192 and this is thread id 139880491378240.
```

I had no idea where this was coming from but this [blog article](https://www.dimagi.com/blog/debugging-a-thread-related-test-failure/) implied that something is being imported in the test discovery and causing this issue. By only stating the test folders which have selenium tests, this went away but it is not ideal as it's a people are unlikely to know they must update this when adding selenium tests to new test folders.
I can try to find a better solution but if anyone knows how to fix this that would be lovely.

smithdc1:
I expect that there is value to move the selenium tests to GHA irrespective of the new screenshotting feature. 

Maybe it's worth splitting that change out from here to get that merged first? 

We can then review that the new action reflects what's in Jenkins before possibly adding a selenium job here in addition to the two on Jenkins.

@felixxm -- is my understanding that it is desired to move these tests from Jenkins correct? 🤔

ngnpope (commented):
Thank for looking into this @sarahboyce. I was actually thinking about this very recently in relation to a change I was thinking of making in the admin and how we'd be able to tell that it didn't break anything. So this is most timely 😄 

I see you've shown some examples of the deltas using `pixelmatch`. It'd also be nice if we can come up with a way to easily review the changes, whether that's posting a comment with the diff (although that might get very noisy) or adding an HTML page that can be published to GitHub pages with a link added via a comment to enable easy review?

In addition, how confident are we that all styling stuff is captured in the selenium tests that exist? Do we want an example project somewhere that exercises everything which we can use to run this visual testing against?

django/contrib/admin/tests.py
```diff
@@ -231,3 +232,18 @@ def has_css_class(self, selector, klass):
...
+
+    def take_screenshot(self):
+        current_directory = os.getcwd()
+        screenshot_directory = os.path.join(current_directory, "screenshots")
+        if not os.path.exists(screenshot_directory):
+            os.makedirs(screenshot_directory)
+        screenshot_name = os.path.join(
+            screenshot_directory, f"{self._testMethodName}.png"
+        )
+        self.selenium.set_window_size(1366, 784)
```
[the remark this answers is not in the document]
sarahboyce: The default screen size of selenium was strange, so expanded it to a "standard" desktop size. This actually [might not be the most common](https://www.hobo-web.co.uk/best-screen-size/) one. Need to pick one and put an appropriate comment :+1: 
We can take multiple screenshots at different sizes. I would be tempted to either start with desktop or desktop and mobile rather than do too many too early. :thinking:

django/contrib/admin/tests.py
```diff
@@ -231,3 +232,18 @@ def has_css_class(self, selector, klass):
...
+        self.selenium.save_screenshot(screenshot_name)
```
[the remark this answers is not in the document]
sarahboyce: Thank you, pushed an update! That's much nicer

django/contrib/admin/tests.py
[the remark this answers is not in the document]
sarahboyce: Another idea, similar to how `runtests.py` has the cli argument `--selenium chrome,firefox` (list of browsers) maybe I should do something similar like `--screenshots 1366x784,1920x1200` (list of resolutions) and then you can run this locally and check different screen-sizes and the ci just does 1 or 2. :thinking:

sarahboyce:
As the end of a test might not give us a screenshot at the right point, made the screenshot taking explicitly part of the test.
@ngnpope also added a mobile screenshot :+1: 

I added screenshots for one of the test files to give us an idea. You can see the screenshots here: https://github.com/django/django/actions/runs/5280983768?pr=16963
How do we feel about this approach? Am I getting warmer or colder do we think? (@smithdc1)

felixxm:
I moved the last commit to the separate PR, see #16989.

django/contrib/admin/tests.py
```diff
@@ -231,3 +233,24 @@ def has_css_class(self, selector, klass):
...
+    def take_screenshot(self, name):
+        if not self.screenshots:
+            return
+        window_size = self.selenium.get_window_size()
+        screenshot_specs = [
+            {"name": "desktop", "width": 1920, "height": 1200},
+            {"name": "mobile", "width": 360, "height": 800},
+        ]
+        for spec in screenshot_specs:
+            self.selenium.set_window_size(spec["width"], spec["height"])
```
sabderemane, line 246: If I'm not mistaken, we don't have screenshot with rtl direction in HTML. I think it will be interesting if we can have that.
I'm not completely familiar with selenium, do you think it's possible at this stage ?

sarahboyce: It should be possible for sure and it's a good idea, do you think it's similar to mobile where we would want it for every page?

sabderemane: I think so, but if it's too many screenshots to have/check on, maybe we could have screenshots on list page and detail page, IMO these are the most significant pages for rtl

knyghty: IMO it's good to have everywhere because while I usually remember to take screens of RTL, I always forget to test it when making further commits.

sarahboyce: Ok I added logic that you can take the screenshots also in RTL. You can view screenshots here: https://github.com/django/django/actions/runs/5359753461/
It's an attachment named screenshots at the bottom :+1:

django/test/selenium.py
```diff
@@ -133,3 +137,86 @@ def disable_implicit_wait(self):
...
+
+    return wrapper
+
+
+@tag("screenshot")
+class ScreenshotTestCase(SeleniumTestCase):
+    screenshots = False
+
+    @classmethod
+    def __init_subclass__(cls, **kwargs):
```
sarahboyce, line 158: @adamchainz I took a lot of inspiration from your package [unittest-parametrize](https://github.com/adamchainz/unittest-parametrize) here :heart: 
Feel free to add your thoughts on how best to structure this :+1:

sarahboyce: Also @ngnpope if you want to look at the PR again, I approach has changed quite a bit to accommodate different requirements and would be good to get some feedback :star:

django/test/selenium.py
```diff
@@ -133,3 +137,86 @@ def disable_implicit_wait(self):
...
+        super().__init_subclass__(**kwargs)
+
+        for name, func in list(cls.__dict__.items()):
+            if not cls.screenshots:
+                continue
+            if not isinstance(func, FunctionType):
+                continue
```
adamchainz, line 165: I suggest deleting this since the attribute can come only from the decorator
```suggestion
```

I know you copied it from unittest-parametrize but that needs to be a bit more defensive because of the chance of interaction with other code. Whilst I'd say don't add code here that's not necessary.

django/test/selenium.py
adamchainz: Glad to inspire!

django/test/selenium.py
```diff
@@ -133,3 +137,86 @@ def disable_implicit_wait(self):
...
+                setattr(cls, test.__name__, test)
+
+    def desktop(self):
+        self.window_size = self.selenium.get_window_size()
+        self.selenium.set_window_size(1280, 720)
+
+    def disable_desktop(self):
+        self.selenium.set_window_size(
+            self.window_size["width"], self.window_size["height"]
+        )
```
adamchainz, line 192: turn these into context managers? Otherwise failures can pollute other tests.

adamchainz (changes requested):

django/test/selenium.py
sarahboyce: Yes! I was doing this going, why is this so ugly?! :joy: thank you :+1:

tests/runtests.py
```diff
@@ -747,6 +754,9 @@ def paired_tests(paired_test, options, test_labels, start_at, start_after):
             SeleniumTestCaseBase.external_host = options.external_host
         SeleniumTestCaseBase.headless = options.headless
         SeleniumTestCaseBase.browsers = options.selenium
+        if options.screenshots:
+            options.tags = ["screenshot"]
```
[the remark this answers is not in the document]
sarahboyce: Not sure what you mean here, when I have `--screenshots` it only makes sense to run the tests with screenshots so I don't think it's worth running other selenium tests without screenshots for example.
This quite specific to how we run our tests, do you think I need to handle some case I'm missing?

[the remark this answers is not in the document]
adamchainz: I mean if `--tags` has been specified along with `--screenshots`, there should be an error, similar to the above for checking that `--selenium` has been passed.

django/contrib/admin/tests.py
sabderemane: LGTM 🚀

.github/workflows/screenshots.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Find comment to replace
+        uses: peter-evans/find-comment@v2
+        id: find-comment
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          comment-author: 'github-actions[bot]'
+          body-includes: Screenshots created
+
+      - name: Create or update comment
```
sarahboyce, line 53: The GitHub action fails at this point because this PR comes from a fork and does not have write permissions for the repository. This should be resolved if the PR is accepted.
I opened the same PR on my forked repository so that we can test the comment, you can see that here: https://github.com/sarahboyce/django/pull/1
Looks to work :tada: (@ngnpope I think that's what you wanted)

.github/workflows/screenshots.yml
```diff
@@ -0,0 +1,69 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  selenium-sqlite-screenshots:
+    if: contains(github.event.pull_request.labels.*.name, 'selenium')
```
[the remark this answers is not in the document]
sarahboyce: Yes I think we should. Can you create a label for me? 🙏

docs/internals/contributing/writing-code/unit-tests.txt
```diff
@@ -269,6 +271,10 @@ faster and more stable. Add the ``--headless`` option to enable this mode.
 
 .. _selenium.webdriver: https://github.com/SeleniumHQ/selenium/tree/trunk/py/selenium/webdriver
 
+For testing changes to the admin UI, the selenium tests can be run with the
+``--screenshots`` option enabled. Screenshots will be saved to the
+``tests/screenshots/`` directory.
+
```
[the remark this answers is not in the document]
sarahboyce: Yeah I agree. Added an example in here 79648a4 👍

django/test/selenium.py
```diff
@@ -134,3 +137,82 @@ def disable_implicit_wait(self):
...
+        get_language_bidi_mock = mock.patch(
+            "django.utils.translation.get_language_bidi", return_value=True
+        )
+        self.selenium.set_window_size(1280, 720)
+        get_language_bidi_mock.start()
+        try:
+            yield
+        finally:
+            self.selenium.set_window_size(window_size["width"], window_size["height"])
+            get_language_bidi_mock.stop()
```
felixxm, line 211: I wonder if these 3 context managers could be moved to `SeleniumTestCase` and used as decorators directly on test methods (we have some tests that use `set_window_size()`), and reused here. This could standardize tests.

Also, we should use the same boundary widths as defined in admin CSS files, i.e.
- <= 767 (mobile)
- <= 1024 (small screens)
- standard

so maybe:

- `(767, 800)` -> `@mobile`
- `(1024, 768)` -> `@small_screen`
- `(1280, 720)` -> `@desktop`

We can do this in a separate PR.

django/test/selenium.py
```diff
@@ -134,3 +137,82 @@ def disable_implicit_wait(self):
...
+        finally:
+            self.selenium.set_window_size(window_size["width"], window_size["height"])
+
+    @contextmanager
+    def rtl(self):
+        window_size = self.selenium.get_window_size()
+        get_language_bidi_mock = mock.patch(
+            "django.utils.translation.get_language_bidi", return_value=True
+        )
+        self.selenium.set_window_size(1280, 720)
```
felixxm, line 205: Can we reuse `desktop` here?

.github/workflows/screenshots.yml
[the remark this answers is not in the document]
felixxm: Added `screenshots 🖼️` with emoji :wink: .

pauloxnet:
Do you think it's possible to take one image for each style mode (dark and light) for each screenshot so that we can test the admin UI in both style mode?

.github/workflows/screenshots.yml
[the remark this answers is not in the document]
sarahboyce: Ahh amazing! Makes me very happy 🖼️❤️

django/test/selenium.py
felixxm: So I would start with adding four context processors that can be used as test decorators, `@mobile`, `@small_screen`, `@desktop`, and `@rtl`.

sarahboyce:
> Do you think it's possible to take one image for each style mode (dark and light) for each screenshot so that we can test the admin UI in both style modes?

Yes, everything is possible 😉 

Question though, do we need a "matrix" approach where every combination has a screenshot (so mobile vs desktop, ltr vs rtl, light vs dark - 2 x 2 x 2 = test is ran 8 times) or is it enough to test against a default which is what I am doing here (so rlt is only tested in desktop) which ends up with less test runs (including the dark mode, test would be ran 4 times).
Does that make sense?

felixxm:
Personally, I don't think it's worth doing.

pauloxnet:
> Personally, I don't think it's worth doing.

The risk is not noticing bugs introduced in the admin's dark theme mode.

felixxm:
> > Personally, I don't think it's worth doing.
> 
> The risk is not noticing bugs introduced in the admin's dark theme mode.

But those will be bugs in colors, not in a structure, am I right?

pauloxnet:
> > > Personally, I don't think it's worth doing.
> > 
> > The risk is not noticing bugs introduced in the admin's dark theme mode.
> 
> But those will be bugs in colors, not in a structure, am I right?

Not exactly, I would say that there would be bugs in the User eXperience and User Interface.

Potentially some sections of the interface could "disappear" simply by mistakenly setting the same color for a text and its background, or similar.

And this may not be visible by analyzing only the clear theme.

sarahboyce:
This might be my fault I answered @pauloxnet and then asked a seperate question.
@felixxm when you said you don't see the worth of doing it, did you just mean my 2 x 2 x 2 matrix question in which every combination should be tested or the dark mode screenshot or both?

felixxm:
> @felixxm when you said you don't see the worth of doing it, did you just mean my 2 x 2 x 2 matrix question in which every combination should be tested or the dark mode screenshot or both?

Yes, that's definitely not worth doing, maybe a separate `dark` marker to generate desktop screenshots in the dark mode.

django/test/selenium.py
sarahboyce: Interesting suggestion. Originally I assumed these are only used/wanted when doing screenshots, if you agree with that then do you think the naming should be more exclicit around that it's related to taking a screenshot if this was updated? (like `@screenshot_mobile`)
If you think these are useful just in general, then I guess when doing the standard selenium tests we would want to re-run the test for each case regardless of whether it is taking a screenshot 🤔

felixxm: > If you think these are useful just in general, then I guess when doing the standard selenium tests we would want to re-run the test for each case regardless of whether it is taking a screenshot 🤔

Yes, that's why I left this suggestion :smile: I imagine something like:
```python
class ChangeWindowSize:
    def __init__(self, width, height, selenium):
        self.new_size = (width, height)
        self.old_size = tuple(self.selenium.get_window_size().values())

    def __enter__(self):
        self.old_size = self.selenium.get_window_size()
        self.selenium.set_window_size(*self.new_size)
        return self

    def __exit__(self, exc_type, exc_value, traceback):
        self.selenium.set_window_size(*self.old_size)

class SeleniumTestCase(...)

    @contextmanager
    def desktop(self):
        with ChangeWindowSize(1280, 720, self.selenium):
            yield

    @contextmanager
    def small_screen(self):
        with ChangeWindowSize(1024, 768, self.selenium):
            yield

    ...
```
and in tests, e.g.
```diff
diff --git a/tests/admin_widgets/tests.py b/tests/admin_widgets/tests.py
index 80bb6233ab..5ef69ce049 100644
--- a/tests/admin_widgets/tests.py
+++ b/tests/admin_widgets/tests.py
@@ -1076,6 +1076,7 @@ class DateTimePickerSeleniumTests(AdminWidgetSeleniumTestCase):
 
         self.assertEqual(len(selected), 0)
 
+    @self.small_screen
     def test_calendar_show_date_from_input(self):
         """
         The calendar shows the date from the input field for every locale
@@ -1083,7 +1084,6 @@ class DateTimePickerSeleniumTests(AdminWidgetSeleniumTestCase):
         """
         from selenium.webdriver.common.by import By
 
-        self.selenium.set_window_size(1024, 768)
         self.admin_login(username="super", password="secret", login_url="/")
 
         # Enter test data
```

The same context processors would be reused later in `_screenshot_cases`.

sarahboyce: Ah thank you, I see what you're going for 👍

sarahboyce:
Dark mode added 🕶️ (you can download the screenshots at the bottom here: https://github.com/django/django/actions/runs/6494619719/)

pauloxnet:
> Dark mode added 🕶️ (you can download the screenshots at the bottom here: https://github.com/django/django/actions/runs/6494619719/)

Great work @sarahboyce!

I see strange image height values (678px or 598px), different from what I read in your PR.

django/test/selenium.py
```diff
@@ -118,6 +161,51 @@ def setUpClass(cls):
...
+            try:
+                yield [desktop]
+            finally:
+                get_language_bidi_mock.stop()
+
+    @contextmanager
+    def dark(self):
+        # Need to navigate to a page before executing a script.
+        self.selenium.get(self.live_server_url)
+        self.selenium.execute_script("localStorage.setItem('theme', 'dark');")
```
ngnpope, line 195: I forget with selenium... But do we need to wait for anything to load before executing the script here? Or does it work fine?

docs/internals/contributing/writing-code/unit-tests.txt
```diff
@@ -271,6 +271,32 @@ faster and more stable. Add the ``--headless`` option to enable this mode.
...
+To define when screenshots should be taken during a selenium test, the test
+class must use the ``django.test.selenium.screenshot_cases`` decorator with a
+list of supported screenshot types (currently ``"desktop"``, ``"mobile"``,
+``"rtl"`` and ``"dark"``). It can then call
```
ngnpope, line 281: You also have `small_screen` above? Or was the intent to get rid of that?

ngnpope: Not sure whether it's worth stating that `rtl` and `dark` only operate with `desktop` size?

docs/internals/contributing/writing-code/unit-tests.txt
```diff
@@ -271,6 +271,32 @@ faster and more stable. Add the ``--headless`` option to enable this mode.
...
+    class SeleniumTests(SeleniumTestCase):
+        @screenshot_cases(["desktop", "mobile", "rtl", "dark"])
+        def test_login_button_centered(self):
+            self.selenium.get(self.live_server_url + reverse("admin:login"))
+            self.take_screenshot("login")
+            ...
+
+
+This generates three screenshots of the login page - one for a desktop screen,
+one for a mobile screen and one for right to left languages.
```
ngnpope, line 298: With the addition of `dark`, I think it's now **four**... 😂 

Maybe we can avoid specifying a number and just list the screenshots generated?

ngnpope (changes requested):
Hey @sarahboyce 👋🏻 

Finally got around to having a look. Nice to see this coming together.

ngnpope:
> I see strange image height values (678px or 598px), different from what I read in your PR.

@pauloxnet I guess this is because it is setting window size, not viewport size?

.github/workflows/screenshots.yml
```diff
@@ -0,0 +1,69 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  selenium-sqlite-screenshots:
+    if: contains(join(github.event.pull_request.labels.*.name, '|'), 'screenshots')
```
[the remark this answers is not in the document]
sarahboyce: It is because of the emoji (though I do love it) and I wasn't able to get it working by just putting the emoji into the string for some reason (I will commit this suggestion shortly to triple check)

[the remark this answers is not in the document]
ngnpope: Might be worth checking how other projects that use emojis in labels handle this?

.github/workflows/screenshots.yml
```diff
@@ -0,0 +1,69 @@
...
+
+            You can find the attached screenshots to the below linked Workflow Run URL (Logs).
+
+            Please note that files only stay for around 90 days!
+
+            | Name      | Link                                                                                |
+            |-----------|-------------------------------------------------------------------------------------|
+            | Commit    | ${{ github.event.pull_request.head.sha }}                                           |
+            | Logs      | ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }} |
+          edit-mode: replace
```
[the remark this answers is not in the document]
sarahboyce: Updated, you can see the example on my test PR: https://github.com/sarahboyce/django/pull/1

django/test/selenium.py
sarahboyce: It appears to be working but maybe this is something I can do instead of navigating to a url 🤔

.github/workflows/screenshots.yml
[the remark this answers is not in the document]
sarahboyce: I think there might be a hidden character at the end of the label (if you hover over the different labels with emojis I think there's something at the end of the screenshot label: https://github.com/django/django/labels). 
@felixxm can you check the label and maybe delete the extra character? 🙏

django/test/selenium.py
```diff
@@ -118,6 +161,51 @@ def setUpClass(cls):
...
+        with self.desktop() as desktop:
+            try:
+                yield [desktop]
+            finally:
+                self.selenium.execute_script("localStorage.removeItem('theme');")
```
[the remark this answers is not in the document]
sarahboyce: Any suggestions? I can add `_mode` to the end of each of them `dark_mode` `rtl_mode` etc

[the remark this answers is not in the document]
ngnpope: Perhaps… 🤔

Could also go with `desktop_size` and `mobile_size`? (Although not sure `small_screen_size` would be better…)

Thoughts, @felixxm?

.github/workflows/screenshots.yml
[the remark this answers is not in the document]
ngnpope: Nearly there… But I think you forgot to add `id: create-screenshots` to the "Run Selenium tests with screenshots" step as the steps output stuff isn't working?

Would also be good to see how the extra list items get added to the comment when the commits are updated. We should see that if you update the branch to fix the above.

.github/workflows/screenshots.yml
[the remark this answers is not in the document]
felixxm: > I think there might be a hidden character at the end of the label (if you hover over the different labels with emojis I think there's something at the end of the screenshot label: https://github.com/django/django/labels). @felixxm can you check the label and maybe delete the extra character? 🙏

There is nothing there (unfortunately).

django/test/selenium.py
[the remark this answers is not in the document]
felixxm: `_mode` and `_size` suffixes sound good to me.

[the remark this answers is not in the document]
felixxm: Also, please move "size" helpers to a separate PR, they can be added and used without screenshots so I'd like to merge them first.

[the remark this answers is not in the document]
sarahboyce: https://github.com/django/django/pull/17360 👍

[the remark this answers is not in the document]
felixxm: Merged in f6629ee2c986d3bf59b4c1b3058f370a00bdc573. Please rebase.

felixxm:
@sarahboyce Thanks for all your efforts :+1: I would like to merge/discuss the remaining code in two steps:

1. new `screenshots` option to the `runtests.py`,
2. new GitHub action.

django/test/selenium.py
```diff
@@ -147,6 +174,36 @@ def mobile_size(self):
         with ChangeWindowSize(360, 800, self.selenium):
             yield
 
+    @contextmanager
+    def rtl(self):
+        get_language_bidi_mock = mock.patch(
+            "django.utils.translation.get_language_bidi", return_value=True
+        )
+        get_language_bidi_mock.start()
```
felixxm, line 182: Can we use one of bidirectional languages instead of this mocking?
```python
with override_settings(LANGUAGE_CODE="he"):
    ...
```
or
```python
with override_settings(LANGUAGE_CODE=settings.LANGUAGES_BIDI[-1]):
    ...
```

docs/internals/contributing/writing-code/unit-tests.txt
```diff
@@ -271,6 +271,33 @@ faster and more stable. Add the ``--headless`` option to enable this mode.
...
+
+
+    class SeleniumTests(SeleniumTestCase):
+        @screenshot_cases(["desktop_size", "mobile_size", "rtl", "dark"])
+        def test_login_button_centered(self):
+            self.selenium.get(self.live_server_url + reverse("admin:login"))
+            self.take_screenshot("login")
+            ...
+
+
```
felixxm, line 296: Remove an extra blank line.

docs/internals/contributing/writing-code/unit-tests.txt
```diff
@@ -271,6 +271,33 @@ faster and more stable. Add the ``--headless`` option to enable this mode.
...
+This generates multiple screenshots of the login page - one for a desktop
+screen, one for a mobile screen, one for right to left languages on desktop
+and one for dark mode on desktop.
```
felixxm, line 299: `versionchanged` annotation is missing, for example:

```
.. versionchanged:: 5.1

     The ``--screenshots`` option and `@screenshot_cases` decorator were added.
```

django/test/selenium.py
sarahboyce: This works 🎉 
It does mean that in tests when there is any selection based off link text `find_element(By.LINK_TEXT, "Example")` - it doesn't work out of the box 🤔 
What do you think? Is it better because the language is realistic? Or is it nice to keep it in english and have it go the other side via the mocking?

felixxm:
I moved the first commit with small edits (release notes, comments) to the #17374.

felixxm:
> I moved the first commit with small edits (release notes, comments) to the #17374.

Merged in be56c982c0805f62a2948d9d7a3e21215c352174.

felixxm:
@sarahboyce Many thanks! rebased.

felixxm:
I simplified the generated comment a bit to:

image

ngnpope:
> I simplified the generated comment a bit to:
> 
> image

This reverts a bunch of what was discussed in [this comment](https://github.com/django/django/pull/16963#discussion_r1356743588), in particular reducing the visibility of screenshots taken on previous runs. Being able to easily access multiple downloads is useful for comparison. 

Maybe an alternative is not to edit existing comments, but just create new ones every time?

felixxm:
> Maybe an alternative is not to edit existing comments, but just create new ones every time?

Yes, I was thinking about the same :+1:

felixxm:
@ngnpope You can check the result in https://github.com/felixxm/django/pull/6.

felixxm:
This also make it easier to compare results when pushing fixes.

sarahboyce:
> > Maybe an alternative is not to edit existing comments, but just create new ones every time?
> 
> Yes, I was thinking about the same 👍

I'm not a fan of a new message with each commit. When you get enough comments, like this PR, conversations become hidden even when they're not resolved. I missed that Nick's suggestion earlier was to append to an existing message. Can try that? Or that a message is created in a thread? Worried a new message might pollute a review process 🤔

felixxm:
All good, but `date` is not there :thinking:

felixxm:
Finally, it works! :tada: https://github.com/felixxm/django/pull/9

sarahboyce:
> Finally, it works! 🎉 [felixxm#9](https://github.com/felixxm/django/pull/9)

Ah I love it!

PR #16968, opened 2023-06-12
Added TIME_ZONE override test

smithdc1:
Locally, this doesn't work for me on Windows (Python 3.11.3), but it does on Ubuntu WSL (22.04 | Python 3.10.6). On Windows even with the settings overridden I get my local timezone. 

Opening to see if I see the same failure in CI.

### DIFF
--- a/tests/settings_tests/tests.py
+++ b/tests/settings_tests/tests.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import unittest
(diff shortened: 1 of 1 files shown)


smithdc1:
I guess time travelling is only allowed in Unix. 

See docs.

https://docs.python.org/3/library/time.html#time.tzset

PR #16969, opened 2023-06-12
Swap user_can_authenticate and check_password in if statement

ebram96:
Make `user_can_authenticate` the left operand and `check_password` the right operand in if statement because `user_can_authenticate` does no complex logic but `check_password` does (like hashing the raw password). This can save computation in case of `user_can_authenticate` just returns `False`.

### DIFF
--- a/django/contrib/auth/backends.py
+++ b/django/contrib/auth/backends.py
@@ -45,7 +45,7 @@ class ModelBackend(BaseBackend):
             # difference between an existing and a nonexistent user (#20760).
             UserModel().set_password(password)
         else:
-            if user.check_password(password) and self.user_can_authenticate(user):
+            if self.user_can_authenticate(user) and user.check_password(password):
                 return user
 
     def user_can_authenticate(self, user):


django/contrib/auth/backends.py
```diff
@@ -45,7 +45,7 @@ def authenticate(self, request, username=None, password=None, **kwargs):
             # difference between an existing and a nonexistent user (#20760).
             UserModel().set_password(password)
         else:
-            if user.check_password(password) and self.user_can_authenticate(user):
+            if self.user_can_authenticate(user) and user.check_password(password):
```
charettes, line 48: Looking at the line above I think the original order was intended to blur the lines in terms of timing difference and prevent attacks.

Basically no matter what input is provided hashing takes place, which is a slow operation, to make it hard to differentiate between
1. A user that doesn't exist
2. A user that can't authenticate
3. A user with the wrong password
4. A user with the right password

ebram96: I see.. thank you for explaining this!

PR #16972, opened 2023-06-13
Add action for deactivating selected users to User admin panel

pktiuk:
I think it is handy to have default action for disabling users from the list view.

image


pktiuk:
I created a followup therad on django forum: https://forum.djangoproject.com/t/proposal-add-more-default-actions-to-default-user-admin-page/21584

PR #16975, opened 2023-06-14
Fixed #34643 -- Move admin form labels to a more accessible place

hrushikeshrv:
Fixes #34643 by moving admin form labels above the inputs instead of left aligned.

Before - 
image

After -
image


hrushikeshrv:
I've updated the failing tests. @felixxm is the approach of the fix okay?

shangxiao:
I could be wrong but I thought @thibaudcolas' recommendation was to look at just making the small-width responsive CSS be used for all widths. I didn't think any HTML changes were required with this? 🤔

knyghty (commented):
I have a vague feeling from my previous work trying to get this stuff aligned after changing things that actually a lot of CSS can be removed that supports aligning stuff in different cases. But if I'm right on that I'm not sure it really needs to be done here.

I didn't do any review (yet?)

knyghty (changes requested):
Hi @hrushikeshrv, apologies for the delay here. I made a quick review locally, and found two issues with help text alignment:

One when using the [`wide` fieldset class](https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.fieldsets) used by default in the user add admin:

Screenshot 2023-11-21 at 19 14 21

And another using RTL:

Screenshot 2023-11-21 at 19 13 05

You can test this by setting your `LANGUAGE_CODE` to e.g. `he`.

hrushikeshrv:
Thanks for the review, @knyghty 👍🏻 
I've removed the extra padding given by the `.wide` class and the extra margin for the `.wide` class in RTL.

I also found [these lines in `forms.css`](https://github.com/django/django/blob/101a85a5a06585ba16ecb25860146d034a8a55ec/django/contrib/admin/static/admin/css/forms.css#L147) that add the same margin to paragraphs and divs with the `.help` class, but I couldn't find a page or form state on the admin site where these styles were applied. I've preemptively removed those lines as well.

However, now that form labels are placed above their respective inputs, validation errors on fields are displayed above the labels, like this -

image

IMO, we should move these errors to be below the inputs and above the help text (if any).

knyghty:
@hrushikeshrv thanks for bringing this up, it's indeed in a bad place. I talked with @thibaudcolas a bit and we suggest following gov.uk and putting it between the help text and the input: https://design-system.service.gov.uk/components/error-message/

I also wonder if the `aligned` and `wide` classes are meaningless now. `wide` in particular is an issue because it's documented: https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.fieldsets

hrushikeshrv:
> I also wonder if the aligned and wide classes are meaningless now.

I agree. `.wide` mainly just increases the with of `label`s from `160px` to `200px` and decreases the width of `.vLargeTextField` and `.vXMLLargeTextField` - this generally won't make a visual difference unless the label is very long. 

Without the `wide` class - 

image

With the `wide` class -

image

IMO we should increase the width of the `label`s to `200px` anyway, now that they are placed above the fields instead of beside them.
  
There are only 3 selectors for `.wide` in `forms.css` which largely have no noticeable effect now (from the cases I have seen).

I'm not so sure about `.aligned` being meaningless now, since it defines very specific margin and padding numbers for most form elements.

knyghty:
@hrushikeshrv my feeling is we should just remove the label widths. They might be excessively long in some cases I suppose, but I feel like intuitively they should take up the full width of the content. If it feels too long for very verbose labels, maybe that's a great impetus to increase the font size (in another PR).

For .aligned, maybe, I didn't look into it too much. Removing widths is probably enough for here and we can do any clean up, again, in another PR.

knyghty (commented):
@hrushikeshrv, again apologies for the delay here.

I noticed another problem we'll face here. It's possible that you can define multiple fields on the same line. The upshots of this:

- We need to handle it, by making sure it works (e.g. we'd have a column per field, but the label / field / help text / etc. is still in a single column). I didn't test this yet myself.
- The wide class is probably still of value here.

hrushikeshrv:
@knyghty I tested defining multiple fields on the same line, and it was indeed not working. I've handled that case so it now works as expected.

image

Regarding the `wide` class, it used to work by setting a width of 200px for field labels, which pushed the field inputs to the right (or left in case of rtl), leading to more horizontal space. Now that the labels are on top of the inputs, just setting the label widths won't lead to more horizontal space, so I've modified the effect of the `wide` class to instead set an increased margin on the `fieldBox` class, which leads to more horizontal space. I think this is the right approach, and I've checked to make sure that there aren't any overflow issues on different screen sizes, but please do confirm.

knyghty:
I will do some testing myself but I think otherwise this is ready for a review from someone other than me. I've updated the flags on the ticket. Release notes are probably also required.

hrushikeshrv:
Thanks @knyghty. I've fixed the merge conflicts as well.

This is my first time writing release notes so I had a few questions. Should I write the release notes in `docs/releases/5.1.txt`? If so, would writing them under the [django.contrib.admin heading](https://github.com/django/django/blob/7a05b8a2fac57e32a61726893d4601352c1d1c8d/docs/releases/5.1.txt#L32) be okay? 😅

nessita:
> Thanks @knyghty. I've fixed the merge conflicts as well.
> 
> This is my first time writing release notes so I had a few questions. Should I write the release notes in `docs/releases/5.1.txt`? If so, would writing them under the [django.contrib.admin heading](https://github.com/django/django/blob/7a05b8a2fac57e32a61726893d4601352c1d1c8d/docs/releases/5.1.txt#L32) be okay? 😅

Yes, release notes should be in the `5.1.txt` file following the existing style (verb tense, etc). The `django.contrib.admin` is correct as well.

Also, could you please rebase your PR onto main? This mean first of all, updating your `main` branch with the latest revisions and then running a command like this: `git rebase main` in your branch. More details [in this blogpost](https://adamj.eu/tech/2022/03/25/how-to-squash-and-rebase-a-git-branch/).

When all of the above is complete, please make sure that the title references the corresponding ticket and to adjust the ticket flags in the Trac system so this PR gets added to the "branches needing review" section of the [Django Developer Dahsboard](https://dashboard.djangoproject.com/). More information in [these docs](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/submitting-patches/#patch-style).

If you need any help, please ask! Thank you, Natalia.

nessita:
After an initial review, I see these issues in the admin UI. Let me know your thoughts!

* Different widths for groups and permissions for User:
image

* Help text on top of (read only) field is very confusing about which phrase is help and which phrase is the field value:
image

* Since we are gaining horizontal space for the fields, it seems like it would be a good idea to let them grow horizontally as wide as they can, so we avoid long texts to the chopped off like this:
image

nessita:
Hello @hrushikeshrv, would you have time to keep working on this? Thank you for your efforts so far!

hrushikeshrv:
Hi @nessita, sorry for the late response! I was a little swamped for the last few weeks, but I will be able to work on this patch now. I'll rebase this PR and address the width and horizontal space concerns asap!

hrushikeshrv:
> Since we are gaining horizontal space for the fields, it seems like it would be a good idea to let them grow horizontally as wide as they can, so we avoid long texts to the chopped off 

I agree, I've implemented this and it fixes the issue of different widths for groups and permissions as well. The permissions widget was appearing wider because it was growing to try to fit its content. 

However, now that both the widgets grow to fill the entire width, the link to add a related object (the ➕ button) seems a little out of place IMO -

image


Moreover, on mobile layouts, it appears even more disconnected after this patch -

image


This may be out of scope for this ticket, but what if we changed that link to appear alongside the `Available groups` heading, like so -

image

Please let me know if that seems reasonable or if I should try another approach.

nessita:
Hey @hrushikeshrv, thank you for restarting work on this.

Specifically about the "Add" (plus sign) button: I love the new proposed location. Do you think that moving it to that place could be done in isolation/independently of this branch? I'm wondering we could do a separated commit/PRs with that.

Of course we should also confirm with the accessibility team (cc @django/accessibility @knyghty) that this "Add" button change is acceptable. Did you happen to check screen readers and tab keyboard navigation?

hrushikeshrv:
@nessita Yes, I think we could do this independently in a separate PR. I haven't checked with screen readers or tab keyboard navigation yet. I can do that and create a ticket if the accessibility team agrees.

knyghty:
@hrushikeshrv @nessita, I don't see an immediate problem with the new position - I do think it is actually easier to see from a UX perspective and good to have a clear text label as well. However, it's easier to review the accessibility in PR form, so I'd certainly encourage creating one (with a ticket).

nessita:
> Hi @nessita, I tried to rebase this branch onto main, but it looks like I messed up somehow. I followed the steps you mentioned and the steps in Adam's article, but it looks like I committed every change that has been made since I created my branch. Sorry for my mixup, I'm not sure how to fix this 😅

No worries, let's see: what command did you run, and how did you resolve conflicts, if there were any?

In general, you can undo changes to your git repo by (make a copy of your work before proceeding, just in case). See more, for example, in [this stackoverflow link](https://stackoverflow.com/questions/134882/undoing-a-git-rebase):

* Run `git reflog` and **read carefully** the list of changes shown there. There will be a line for the latest known "good state" of your branch, something like:
```
<a hash> (origin/ticket_34643) HEAD{@<number>}: <action description>
```
* Write down that hash (for example `a123b456c7`, likely belonging to the last good commit you made.
* Reset your git tree to that revno. **Careful** that this reset is a "hard" reset, changes newer than `a123b456c7` will be lost:
```
git reset --hard a123b456c7
```
With that, your git tree will look like it was at `a123b456c7`. Then, you can try the rebase again (`git rebase main`).

hrushikeshrv:
Thanks for your help @nessita, I think I've fixed it.
I've also added release notes.

nessita:
> Thanks for your help @nessita, I think I've fixed it. I've also added release notes.

Amazing! Please remember to update the ticket flags ("has patch") so the branch appears in the review queue.

hrushikeshrv:
Has patch was already set. I've also unset needs documentation.

nessita:
@knyghty is this something that you would like to see progressed?

Also @hrushikeshrv could you please rebase and bring latest changes from `main`? The release notes need to be moved from 5.1 to 5.2. Thanks!

hrushikeshrv:
Moved the release notes to `5.2.txt` 👍🏻

knyghty:
> @knyghty is this something that you would like to see progressed?

Yes. I've done some basic testing only. There are probably quite some pitfalls though and worth a closer look. I will try to look at it more closely during the Vigo sprints, if I'm not too busy with other stuff. Another pair of eyes would also be great.

docs/releases/5.1.txt
```diff
@@ -112,7 +112,7 @@ Minor features
 * The default iteration count for the PBKDF2 password hasher is increased from
   720,000 to 870,000.
 
-* The default ``parallelism`` of the ``ScryptPasswordHasher`` is 
+* The default ``parallelism`` of the ``ScryptPasswordHasher`` is
```
sarahboyce, line 115: Please revert this change

docs/releases/5.2.txt
```diff
@@ -37,6 +37,18 @@ Minor features
...
+* Help text in the change form is now shown after the field label and before
+  the field input.
+
+* Validation errors in the change form are now shown after the help text and
+  before the field input.
+
+* The ``.related-widget-wrapper`` class now stretches to occupy 100% of the
+  available width.
+
 * ...
```
sarahboyce, line 52:
```suggestion
```

docs/releases/5.2.txt
```diff
@@ -37,6 +37,18 @@ Minor features
...
+  instead of next to them.
+
+* Help text in the change form is now shown after the field label and before
+  the field input.
+
+* Validation errors in the change form are now shown after the help text and
+  before the field input.
+
+* The ``.related-widget-wrapper`` class now stretches to occupy 100% of the
+  available width.
```
sarahboyce, line 50:
```suggestion
```
I'm not sure this one is necessary

docs/releases/5.2.txt
```diff
@@ -37,6 +37,18 @@ Minor features
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* Form fields in the change form are now shown below their respective labels
+  instead of next to them.
+
+* Help text in the change form is now shown after the field label and before
+  the field input.
+
+* Validation errors in the change form are now shown after the help text and
+  before the field input.
```
sarahboyce, line 47: I think this should have some justification as to why we're doing this change, maybe "For accessibility purposes..." or something more specific

sarahboyce (commented):
Thank you for your work on this @hrushikeshrv ⭐ 

Personally, I would increase the padding between the label and the input (desktop ltr mode)
test_first_field_focus--focus-single-widget--desktop_size

I would like us to add a screenshot test cases for errors to check where they are being rendered 

A few images in the docs will need updating after this change such as:
- `docs/ref/contrib/admin/_images/fieldsets.png`
- `docs/ref/contrib/admin/_images/raw_id_fields.png`
- `docs/intro/_images/admin05t.png`
- `docs/intro/_images/admin05t.png`
- `docs/intro/_images/admin08t.png`
- `docs/intro/_images/admin09.png`
- `docs/intro/_images/admin10t.png`
- `docs/intro/_images/admin14t.png`
Do this when we're not updating the html and css (otherwise you'll have to generate these again)

hrushikeshrv:
Thanks for the review @sarahboyce, and sorry for the late response.  
I've been a little busy for the last few days, but I'll make the changes you requested asap!

knyghty:
@hrushikeshrv I just wanted to check if you still had time to work on this as I'd like to land it in 5.2 if possible. If you don't, I'm happy to take your work and get it over the line, and you'll still have the co-authorship of course as you've done all the work so far.

Ideally I'd prefer not to take it over as I already don't have enough time for all my other projects :)

docs/releases/5.1.txt
hrushikeshrv: Sorry about that. PyCharm auto-formatted the file and I didn't notice. I'll revert it.

hrushikeshrv:
@knyghty Sorry for the inactivity. I will be able to work on this, but I might be delayed, since I have just moved, and I'm still in the process of setting up my new house.

hrushikeshrv:
Hi @sarahboyce, I've added some extra margin between labels and inputs and also updated the release notes. If this looks good, I can generate the screenshots. Let me know!

sarahboyce:
Thank you for the updates! 

To me it looks like the margin between the fields and the label still needs some work. Looking at the [screenshots generated from the visual regression tests](https://github.com/django/django/actions/runs/10635244416/) you can see this margin appears to be missing in some cases

test_ForeignKey--raw_id_widget--desktop_size

The "Start date" and "Start time" fields looks great but "Main band" and "Description" look to be quite close to their input - I think the spacing should be consistent and I would like to see similar spacing as to the "Start date" and "Start time" fields

Can you also add a screenshot test cases for errors? This will help ease the review (see the [screenshots section](https://docs.djangoproject.com/en/5.1/internals/contributing/writing-code/unit-tests/#running-the-selenium-tests) of running the selenium tests docs for some context)

hrushikeshrv:
Thanks for the review, I'll see what's going wrong. I'll add screenshot test cases for errors as well

hrushikeshrv:
Hi @sarahboyce, is it okay to add the new tests in `tests/admin_views/tests.py` in the `SeleniumTests` class?

sarahboyce:
> Hi @sarahboyce, is it okay to add the new tests in `tests/admin_views/tests.py` in the `SeleniumTests` class?

Yes :+1:

hrushikeshrv:
I've added the screenshot test to check for error message positions. I've checked the screenshots locally as well and they seem to be coming out right. Can you confirm @sarahboyce?

sarahboyce:
Thank you @hrushikeshrv :star: 
This looks really strong and getting the screenshots both locally and in the CI
I have squashed the commits and done a minor tweak to a test

I have two questions

I see that our multi widgets (see "Date:" and "Time:" below) will have the labels next to the field rather than above
How do we feel about this? Should those also change? Is there a reason why this shouldn't change? (cc @django/accessibility)

test_collapsible_fieldset--collapsed--desktop_size

A non-blocking thought is "why do the labels end with a colon : ?" If the reason was due to the layout where the label was on the left, then we perhaps could/should remove it.

@thibaudcolas is at the sprint I have have asked if he could also do some testing but this is very positive and I think we could generate the screenshots for the docs

hrushikeshrv:
I'll generate the screenshots for the docs 👍🏻 

Comparing side by side, IMO having the labels for date and time inputs next to the fields feels better and more "continuous". Also, the multi widgets will have a label above them for the actual field name -

image

MHLut:
About the date and time fields, those two may need to be wrapped in a `<fieldset>` with the overlapping label as a `<legend>`. (Looking at the other accessibility members for a second opinion.)

Such a change is out of scope for this PR, though.

knyghty:
FWIW I would also prefer the labels to not have a colon but also keeping the date and time fields roughly the way they are, which might be better with a colon. Which is probably the most annoying thing I could ask for :)

But I don't think this should be blocking, I'd be happy with something like this in a follow-up

tests/admin_views/tests.py
```diff
@@ -6206,6 +6206,22 @@ def test_first_field_focus(self):
         )
         self.take_screenshot("focus-multi-widget")
 
+    @screenshot_cases(["desktop_size", "mobile_size", "rtl", "dark", "high_contrast"])
+    def test_error_message_position(self):
+        """Error messages are shown above the field and below the label"""
```
sarahboyce, line 6211:
```suggestion
    def test_error_message(self):
```
As we're not asserting based off the position (it won't fail if we move it) - I would call this something generic on errors

tests/admin_views/tests.py
```diff
@@ -6206,6 +6206,22 @@ def test_first_field_focus(self):
...
+
+        url = reverse("admin:admin_views_article_add")
+        full_url = self.live_server_url + url
+        self.admin_login(
+            username="super", password="secret", login_url=reverse("admin:index")
+        )
+        self.selenium.get(full_url)
+        # Submit an empty form to trigger validation errors
+        self.selenium.find_element(By.CSS_SELECTOR, 'input[type="submit"]').click()
+        self.wait_page_ready()
```
sarahboyce, line 6222:
```suggestion
        self.admin_login(
            username="super", password="secret", login_url=reverse("admin:index")
        )
        with self.wait_page_loaded():
            self.selenium.get(
                self.live_server_url + reverse("admin:admin_views_picture_add")
            )

        # Submit an empty form to trigger validation errors.
        with self.wait_page_loaded():
            self.selenium.find_element(By.CSS_SELECTOR, 'input[type="submit"]').click()
        error = self.selenium.find_elements(By.CSS_SELECTOR, ".errornote")
        self.assertEqual(len(error), 1)
```

hrushikeshrv:
@sarahboyce For the docs screenshots, should I use the same models as the current screenshots? If so, where can I find those models?

sarahboyce:
> I'll generate the screenshots for the docs 👍🏻
> 
> Comparing side by side, IMO having the labels for date and time inputs next to the fields feels better and more "continuous". Also, the multi widgets will have a label above them for the actual field name -
> 
> image

We also have a `AdminDateWidget` and a `AdminTimeWidget`, I think for consistency the help text for your time should be under the label/legend but we can keep that the labels are to the side when it is a fieldset
AdminDateWidget with help text below input

docs/releases/5.2.txt
```diff
@@ -41,6 +41,17 @@ Minor features
...
+* In order to improve accessibility:
+
+  * Form fields in the change form are now shown below their respective labels
+    instead of next to them.
+
+  * Help text in the change form is now shown after the field label and before
+    the field input.
+
+  * Validation errors in the change form are now shown after the help text and
+    before the field input.
```
sarahboyce, line 53:
```suggestion
* In order to improve accessibility of the admin change forms:

  * Form fields are now shown below their respective labels instead of next to
    them.

  * Help text is now shown after the field label and before the field input.

  * Validation errors are now shown after the help text and before the field
    input.
```
I wonder if this should be in the `Miscellaneous Backwards incompatible changes in 5.2` section as it isn't so much a feature but rather a change we wish to alert folks of :thinking:

sarahboyce (commented):
I discussed this with @nessita and we are happy to merge this before screenshots are updated, screenshots can be updated in a PR prior to the 5.2 release(s)

sarahboyce:
I also want an approval from the accessibility team so re-requested a review here :+1:

knyghty:
Happy to have another look, but will wait until the final widgets are done.

hrushikeshrv:
Hi @sarahboyce, I ran into a few issues with merge conflicts and rebasing, and I mistakenly erased some history by force pushing. I've tried for a while to try to fix things, but I'm not able to. Is it okay if I close this PR and open a new one from a new branch from current `main`? I have a backup of all changes made in this PR until now.

Sorry for the inconvenience 😅

sarahboyce:
Yeah that's fine - do what you need to do :+1:

YounesDjelloul:
Hi @hrushikeshrv, curious if you are still working on this? I would love to continue if not?
CC @sarahboyce

hrushikeshrv:
Hi @YounesDjelloul, I'm unfortunately not able to work on this PR at the moment, please do feel free to take over 👍🏻

YounesDjelloul:
Hi, new here and exploring the codebase, would love to get a little help! I have forked the main branch and tried to run the selenium tests locally! Though some of them are failing, is this normal or am I missing something?

Antoliny0919:
Hello @YounesDjelloul , if you’re not planning to work on this, would it be okay if I give it a try?

YounesDjelloul:
Yes sir, please go ahead

Antoliny0919:
Superseded by https://github.com/django/django/pull/19713

PR #16979, opened 2023-06-15
Fixed #22125 -- Stopped unnecessary creation of index for ManyToManyField

Akash-Kumar-Sen:
- [ticket 22125](https://code.djangoproject.com/ticket/22125)
- Following PR #10435

### DIFF
--- a/django/db/backends/base/features.py
+++ b/django/db/backends/base/features.py
@@ -360,6 +360,9 @@ class BaseDatabaseFeatures:
     # Does the backend support unlimited character columns?
     supports_unlimited_charfield = False
 
+    # Does this backend use composite index to search by any prefix of its key?
+    composite_index_supports_prefix_search = True
+
     # Collation names for use by the Django test suite.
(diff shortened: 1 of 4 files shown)


rodbv:
Hi @Akash-Kumar-Sen!

I am looking into the [Trac issue related](https://github.com/django/django/pull/10435), and this PR seems quite advanced on the fix. 

Based on @timgraham's comment on the [original PR](https://github.com/django/django/pull/10435#pullrequestreview-169924339) and the fact that all relational DBs supported by Django (including MariaDB - I tested [here](https://sqlfiddle.com/mariadb/online-compiler?id=6af2699e-0a91-403e-a758-0a331294916a)) optimizes the usage of unique composite indexes when searching by the first column, I wonder if the fix could even drop the need of a feature `composite_index_supports_prefix_search`? So maybe it could be as simple as setting `db_index=False` for the first column in the M2M table and adjusting/creating related tests for all the backends.

PR #16998, opened 2023-06-22
Signing session keys #34672 [WIP]

fatih-erikli:
This is a work in progress pull request.

Currently the session keys in `django_auth` is used to authenticate users (or any other session data).

If we sign the session keys when setting them in the cookie, and unsign them when reading, a BadSignature will be thrown if the session key is not signed or signed with a different secret key.

### DIFF
--- a/django/contrib/sessions/middleware.py
+++ b/django/contrib/sessions/middleware.py
@@ -7,6 +7,7 @@ from django.contrib.sessions.exceptions import SessionInterrupted
 from django.utils.cache import patch_vary_headers
 from django.utils.deprecation import MiddlewareMixin
 from django.utils.http import http_date
+from django.core.signing import Signer
 
 
 class SessionMiddleware(MiddlewareMixin):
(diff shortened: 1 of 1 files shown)


charettes:
The accepted consensus was that we wanted to hash session keys instead. See ticket-21076 / #12814

fatih-erikli:
@charettes Hashing sounds good, however I a hashed session key is stored on the database, it still could be used to authenticate a user on a data breach.

charettes:
> still could be used to authenticate a user on a data breach.

Could you elaborate on that? The hashing function is not reversible by definition and a database breach would only contain the hash of the session ID not the session ID itself.

fatih-erikli:
> The hashing function is not reversible 

This is for sure but what will be used for setting the cookies? What I meant was it will not fix the problem if the same hash is used for logging the users in with cookie.

charettes:
> This is for sure but what will be used for setting the cookies? What I meant was it will not fix the problem if the same hash is used for logging the users in with cookie.

On session creation a `session_id` is generating, assigned to `request.cookies[session_cookie_name] = session_id`, and a corresponding `hash(session_id)` record is created

On authentication attempts `hash(request.cookies[session_cookie_name])` is used to lookup corresponding session.

If a database leak occurs the attacker is only in possession of a collection on unreversible hashed session IDs so they cannot reuse them for anything. In other words, there doesn't exist a way to take a leaked session id hash and get the session id out it.

fatih-erikli:
Ok, it's clear now. Thanks for explaining.

PR #17004, opened 2023-06-24
Fixed #34642 -- File.open to support different encodings

SecondPort:
fixed problem to support different encodings

### DIFF
--- a/django/core/files/base.py
+++ b/django/core/files/base.py
@@ -105,11 +105,12 @@ class File(FileProxyMixin):
     def __exit__(self, exc_type, exc_value, tb):
         self.close()
 
-    def open(self, mode=None):
+    def open(self, mode=None, encoding=None):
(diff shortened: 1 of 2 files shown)


shangxiao:
Hi.

Please make sure you read through all the discussion before picking up a ticket because iirc the agreed fix was to pass args and kwargs through without gatekeeping. The PR also has some unrelated changes that need to be removed.

Thanks 😊

SecondPort:
> Hi.
> 
> Please make sure you read through all the discussion before picking up a ticket because iirc the agreed fix was to pass args and kwargs through without gatekeeping. The PR also has some unrelated changes that need to be removed.
> 
> Thanks 😊

Hello, a thousand apologies for the error caused, enter the channel and I will contact you for more guidance when implementing a change, thank you very much for your response, have a good day. 👍

PR #17006, opened 2023-06-24
Fixed #34676 -- Normalized Distance()/Area() exceptions for nonexistent units.

anorthall:
Fixes [34676](https://code.djangoproject.com/ticket/34676#ticket).


felixxm (commented):
@anorthall Thanks :+1: I pushed small edits.

PR #17014, opened 2023-06-26
Fixed #34654 -- Post-normalization performed on the Username field leading to the bypass of the whitespace stripping

SecondPort:
I think the correct solution to this security error was to add a validation step that rejects user names with certain characters, I hope this is the correct solution and if something else needs to be changed, please let me know, thank you very much.

### DIFF
--- a/django/contrib/auth/forms.py
+++ b/django/contrib/auth/forms.py
@@ -147,9 +147,16 @@ class BaseUserCreationForm(forms.ModelForm):
 
 
 class UserCreationForm(BaseUserCreationForm):
+    def contains_unwanted_characters(self, username):
+        """Check if the username contains characters that normalize to whitespace."""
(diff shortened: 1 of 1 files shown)


django/contrib/auth/forms.py
```diff
@@ -147,9 +147,16 @@ def save(self, commit=True):
 
 
 class UserCreationForm(BaseUserCreationForm):
+    def contains_unwanted_characters(self, username):
+        """Check if the username contains characters that normalize to whitespace."""
+        normalized_username = unicodedata.normalize("NFKC", username)
+        return username != normalized_username.strip()
+
     def clean_username(self):
```
lhmzhou, line 155: To separate concerns, consider breaking down `clean_username()` to handle checking for unwanted characters and case for insensitive duplicates. Another request is to add unit tests for this form, covering the various scenarios.

PR #17017, opened 2023-06-26
Added dedicated section for output_field in query expressions docs.

nessita:


nessita:
Current proposal looks like this:
image

LilyFoote (approved):

docs/ref/models/expressions.txt
```diff
@@ -18,6 +18,25 @@ Django supports negation, addition, subtraction, multiplication, division,
...
+``IntegerField()`` or ``BooleanField()``. Usually no arguments are needed when
+instantiating the model field as any arguments relating to data validation
+(``max_length``, ``max_digits``, etc.) will not be enforced on the expression's
+output value.
+
+Note that ``output_field`` is only required when Django is unable to determine
+what field type the result should be. Complex expressions that mix field types
+should define the desired ``output_field``. For example, adding an
+``IntegerField()`` and a ``FloatField()`` together should probably have
+``output_field=FloatField()`` defined.
```
adamchainz, line 38: Wordsmithing...

```suggestion
Output field
============

Many of the expressions documented in this section support an optional
``output_field`` parameter. If given, Django will Django will load the
value into that field after retrieving it from the database.

``output_field`` takes a model field instance, like ``IntegerField()`` or
``BooleanField()``. Usually, the field doesn't need any arguments, like
``max_length``, since field arguments relate to data validation which will not
be performed on the expression's output value.

``output_field`` is only required when Django is unable to automatically
determine the result's field type, such as with complex expressions that mix
field types. For example, adding a ``DecimalField()`` and a ``FloatField()``
requires an output field, like ``output_field=FloatField()``.
```

It didn't sound like you were sure `IntegerField()` + `FloatField()` actually requires `output_field`. I checked the tests and found that `DecimalField()` + `FloatField()` definitely requires it: https://github.com/django/django/blob/88cb8c638cca5c2cdb5a2dd86a9c7f49ed684646/tests/aggregation/tests.py#L1130-L1139

docs/ref/models/expressions.txt
```diff
@@ -389,7 +408,8 @@ driver would escape them.
 
 The ``function``, ``template``, and ``arg_joiner`` keywords can be used to
 replace the attributes of the same name without having to define your own
-class. ``output_field`` can be used to define the expected return type.
+class. ``output_field`` can be used to define the expected return type as
+previously explained.
```
adamchainz, line 412: Link to the section?

docs/ref/models/expressions.txt
```diff
@@ -515,15 +522,10 @@ The ``value`` argument describes the value to be included in the expression,
...
-instantiating the model field as any arguments relating to data validation
-(``max_length``, ``max_digits``, etc.) will not be enforced on the expression's
-output value. If no ``output_field`` is specified it will be tentatively
-inferred from the :py:class:`type` of the provided ``value``, if possible. For
-example, passing an instance of :py:class:`datetime.datetime` as ``value``
-would default ``output_field`` to :class:`~django.db.models.DateTimeField`.
+If no ``output_field`` is specified it will be tentatively inferred from the
+:py:class:`type` of the provided ``value``, if possible. For example, passing
+an instance of :py:class:`datetime.datetime` as ``value`` would default
+``output_field`` to :class:`~django.db.models.DateTimeField`.
```
adamchainz, line 528:
```suggestion
If no ``output_field`` is specified, it will be inferred from the type of the
provided ``value`` for many common types. For example, passing an instance of
:py:class:`datetime.datetime` as ``value`` defaults ``output_field`` to
:class:`~django.db.models.DateTimeField`.
```

adamchainz (commented):
Nice one. Using a dedicated section looks good.

I couldn't resist tweaking some wording.

docs/ref/models/expressions.txt
nessita: Thanks! Initially I grouped existing paragraphs into this new section. A review of those with fresh eyes is definitely appreciated! I applied your suggestions.

docs/ref/models/expressions.txt
nessita: Applied!

carltongibson (approved):
Lovely. Thanks @nessita 🎁

nessita:
Thank you everyone for the suggestions and reviews!

I have applied all fixes suggested by @adamchainz, and considering that Lilly and Carlton have approved this PR, I'll be landing this now since I won't be available starting tomorrow (for 2 weeks).

adamchainz:
Have a good holiday! ☀️

PR #17036, opened 2023-07-01
rewritten the page_range function

awsl1110:
In the process of learning django, I found that the page_range of Paginator does not display well in the case of a large number of pages, so I optimized it.

### DIFF
--- a/django/core/paginator.py
+++ b/django/core/paginator.py
@@ -118,13 +118,23 @@ class Paginator:
         hits = max(1, self.count - self.orphans)
         return ceil(hits / self.per_page)
 
-    @property
-    def page_range(self):
+    def page_range(self, page_current, page_group=11):
         """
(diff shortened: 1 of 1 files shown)


DevilsAutumn:
@awsl1110 Thanks 👍 .It is always better to first [open a ticket](https://code.djangoproject.com/newticket) for these types of issues (for discussion) or search for similar a issue on the [Django's issue tracker](https://code.djangoproject.com/query) (might be a ticket already available).

awsl1110:
> @awsl1110 Thanks 👍 .It is always better to first [open a ticket](https://code.djangoproject.com/newticket) for these types of issues (for discussion) or search for similar a issue on the [Django's issue tracker](https://code.djangoproject.com/query) (might be a ticket already available).

Thanks, but as you can see, it tested badly, so I will close this pull request

PR #17052, opened 2023-07-07
Fixed #34696 -- Updated selection counter in admin changelist on Chrome.

cauethenorio:
https://code.djangoproject.com/ticket/34696

### DIFF
--- a/django/contrib/admin/static/admin/js/actions.js
+++ b/django/contrib/admin/static/admin/js/actions.js
@@ -179,6 +179,9 @@
                 }
             });
         }
+        // Sync counter when navigating to the page, such as through the back
+        // button.
+        window.addEventListener('pageshow', (event) => updateCounter(actionCheckboxes, options));
     };
 
     // Call function fn when the DOM is loaded and ready. If it is already
--- a/tests/admin_changelist/tests.py
+++ b/tests/admin_changelist/tests.py
@@ -1741,6 +1741,41 @@ class SeleniumTests(AdminSeleniumTestCase):
             self.assertIs(c.get_property("checked"), True)
         self.assertIs(checkboxes[-1].get_property("checked"), False)
 
+    def test_selection_counter_is_synced_when_page_is_shown(self):
+        from selenium.webdriver.common.by import By
+
+        self.admin_login(username="super", password="secret")
+        self.selenium.get(self.live_server_url + reverse("admin:auth_user_changelist"))
+
+        form_id = "#changelist-form"
+        first_row_checkbox_selector = (
+            f"{form_id} #result_list tbody tr:first-child .action-select"
+        )
+        selection_indicator_selector = f"{form_id} .action-counter"
+        selection_indicator = self.selenium.find_element(
+            By.CSS_SELECTOR, selection_indicator_selector
+        )
+        row_checkbox = self.selenium.find_element(
+            By.CSS_SELECTOR, first_row_checkbox_selector
+        )
+        # Select a row.
+        row_checkbox.click()
+        self.assertEqual(selection_indicator.text, "1 of 1 selected")
+        # Go to another page and get back.
+        self.selenium.get(
+            self.live_server_url + reverse("admin:admin_changelist_parent_changelist")
+        )
+        self.selenium.back()
+        # The selection indicator is synced with the selected checkboxes.
+        selection_indicator = self.selenium.find_element(
+            By.CSS_SELECTOR, selection_indicator_selector
+        )
+        row_checkbox = self.selenium.find_element(
+            By.CSS_SELECTOR, first_row_checkbox_selector
+        )
+        selected_rows = 1 if row_checkbox.is_selected() else 0
+        self.assertEqual(selection_indicator.text, f"{selected_rows} of 1 selected")
+
     def test_select_all_across_pages(self):
         from selenium.webdriver.common.by import By
 


mgaligniana:
Perfect! 🙌 

I think add a Selenium test would be great!

cauethenorio:
@mgaligniana done, Selenium test added.

tests/admin_changelist/tests.py
```diff
@@ -1741,6 +1741,46 @@ def test_modifier_allows_multiple_section(self):
...
+        # Ensure that the selection indicator is in sync with the selected checkboxes
+        selection_indicator = self.selenium.find_element(
+            By.CSS_SELECTOR, "%s .action-counter" % form_id
+        )
+        row_checkbox = self.selenium.find_element(
+            By.CSS_SELECTOR,
+            "%s .action-select" % first_row_selector,
+        )
+        selected_rows = 1 if row_checkbox.is_selected() else 0
+        self.assertEqual(selection_indicator.text, "%s of 1 selected" % selected_rows)
```
sarahboyce, line 1782:
```suggestion
        selected_rows = self.selenium.find_elements(
            By.CSS_SELECTOR,
            "% input:checked .action-select" % first_row_selector,
        )
        self.assertEqual(selection_indicator.text, "%s of 1 selected" % len(selected_rows))
```
I feel like we can do something like this?

tests/admin_changelist/tests.py
```diff
@@ -1741,6 +1741,46 @@ def test_modifier_allows_multiple_section(self):
...
+        # Select a row
+        row_checkbox.click()
+        self.assertEqual(selection_indicator.text, "1 of 1 selected")
+
+        # Go to another page and get back
+        detail_page_link = self.selenium.find_element(
+            By.CSS_SELECTOR,
+            "%s a" % first_row_selector,
+        )
+        detail_page_link.click()
```
sarahboyce, line 1770:
```suggestion
        self.selenium.get(self.live_server_url + reverse("admin:auth_group_changelist"))
```
Optional/Idea - might be enough just to navigate to a different page without finding a link and clicking

tests/admin_changelist/tests.py
```diff
@@ -1741,6 +1741,46 @@ def test_modifier_allows_multiple_section(self):
...
+
+        selection_indicator = self.selenium.find_element(
+            By.CSS_SELECTOR, "%s .action-counter" % form_id
+        )
+        row_checkbox = self.selenium.find_element(
+            By.CSS_SELECTOR,
+            "%s .action-select" % first_row_selector,
+        )
+
+        # Select a row
```
sarahboyce, line 1761: Minor - can you add full stops/periods to the end of each comment? (I think this is in the coding style but I actually can't find it documented :thinking:)

sarahboyce (commented):
Thank you for the patch @cauethenorio! This looks great!
I have super minor suggestions/ideas - let me know what you think :+1:

django/contrib/admin/static/admin/js/actions.js
```diff
@@ -179,6 +179,9 @@
                 }
             });
         }
+
+        // sync counter when page loads
```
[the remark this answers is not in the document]
cauethenorio: I think keeping the comment will help out future people browsing the code to understand why we are listening to the `pageshow` event here and not `DOMContentLoaded` or another more common event.

tests/admin_changelist/tests.py
cauethenorio: That was my first attempt, but on Firefox the selection will be cleared, therefore `selenium.find_elements` will return no elements and Selenium will keep waiting for them until it times out after 10 seconds.

My solution then was to find all checkboxes (it always returns one element) and check manually if it's checked out or not.

tests/admin_changelist/tests.py
cauethenorio: Done :-)

cauethenorio:
Thanks for the awesome suggestions, @sarahboyc.
I implemented most of them, just one can't be implemented as it will make Selenium timeout when using Firefox.
Please give it another look when you have a chance.

tests/admin_changelist/tests.py
sarahboyce: I think `selenium.find_elements` returns an empty list when there are no elements, so this still works for me on firefox

tests/admin_changelist/tests.py
```diff
@@ -1741,6 +1741,44 @@ def test_modifier_allows_multiple_section(self):
             self.assertIs(c.get_property("checked"), True)
         self.assertIs(checkboxes[-1].get_property("checked"), False)
 
+    def test_selection_counter_is_synced_when_page_is_show(self):
```
sarahboyce, line 1744:
```suggestion
    def test_selection_counter_is_synced_when_page_is_shown(self):
```

tests/admin_changelist/tests.py
cauethenorio: I just tested your suggestion again to make sure I'm not talking nonsense.
It works on Firefox, but it waits 10 seconds until it times out and then returns an empty list.
It doesn't happen when you run the test?

I'm running the test with:
``` 
./runtests.py --selenium=firefox --parallel=1 admin_changelist.tests.SeleniumTests.test_selection_counter_is_synced_when_page_is_show
```

I mean, the test passes with the suggested code but it adds 10s to the test runtime.

sarahboyce (approved):
Thank you @cauethenorio - tested and looks good to me!
Sorry for the wrong view name suggestion! :sweat_smile:

tests/admin_changelist/tests.py
cauethenorio: Just adding a little bit more context I found while researching the reason for the 10s wait:

Django's `SeleniumTestCase` class enables Selenium's implicit wait by default with a timeout of 10 seconds:
https://github.com/django/django/blob/4.2.3/django/test/selenium.py#L100-L101

So Seleniun's `find_elements` function will block the text execution until it times out or it finds any element matching the specified selector:
> An implicit wait is to tell WebDriver to poll the DOM for a certain amount of time when trying to find an element or elements if they are not immediately available.

https://www.selenium.dev/documentation/webdriver/waits/

When running on Firefox, the checkboxes won't restore the previous selection state, therefore the `input:checked.action-select` selector will return no elements, blocking the execution until it times out.

sarahboyce: You’re completely right 👍 I agree your approach is much better

This is interesting though! On a side note I’m tempted to have a look into reducing the implicit wait time (10 seconds feels excessive) 🤔

tests/admin_changelist/tests.py
```diff
@@ -1741,6 +1741,44 @@ def test_modifier_allows_multiple_section(self):
...
         self.assertIs(checkboxes[-1].get_property("checked"), False)
 
+    def test_selection_counter_is_synced_when_page_is_shown(self):
+        from selenium.webdriver.common.by import By
+
+        self.admin_login(username="super", password="secret")
+        self.selenium.get(self.live_server_url + reverse("admin:auth_user_changelist"))
+
+        form_id = "#changelist-form"
+        first_row_selector = "%s #result_list tbody tr:first-child" % form_id
```
mgaligniana, line 1751: I would use [f-strings](https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals), I think is more clear

tests/admin_changelist/tests.py
```diff
@@ -1741,6 +1741,44 @@ def test_modifier_allows_multiple_section(self):
...
+
+        # Go to another page and get back.
+        self.selenium.get(
+            self.live_server_url + reverse("admin:admin_changelist_parent_changelist")
+        )
+        self.selenium.back()
+
+        # Ensure that the selection indicator is in sync with the selected checkboxes.
+        selection_indicator = self.selenium.find_element(
+            By.CSS_SELECTOR, "%s .action-counter" % form_id
```
mgaligniana, line 1773: I would use f-strings as well here and as it's used twice, I would use a constant for this selector

mgaligniana (commented):
Great! Minor suggestions!

PR #17055, opened 2023-07-08
Refs #31949 -- Made @csrf_exempt decorator work with async functions.

LomaxOnTheRun:

This PR makes the `csrf_exempt` decorator able to handle both sync and async views.

### DIFF
--- a/django/views/decorators/csrf.py
+++ b/django/views/decorators/csrf.py
@@ -1,5 +1,7 @@
 from functools import wraps
 
+from asgiref.sync import iscoroutinefunction
+
 from django.middleware.csrf import CsrfViewMiddleware, get_token
 from django.utils.decorators import decorator_from_middleware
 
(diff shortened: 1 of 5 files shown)


django/views/decorators/csrf.py
```diff
@@ -51,9 +53,17 @@ def csrf_exempt(view_func):
...
-    return wrapper_view
+    if iscoroutinefunction(view_func):
+
+        async def _view_wrapper(request, *args, **kwargs):
+            return await view_func(request, *args, **kwargs)
+
+    else:
+
+        def _view_wrapper(request, *args, **kwargs):
+            return view_func(request, *args, **kwargs)
```
LomaxOnTheRun, line 65: This commit mirrors the logic of the [commit](https://github.com/django/django/pull/16874/commits/23abec9192c88f9372e5784ae69c9aa5be2b21be) which makes the `@no_append_slash` decorator able to handle asynchronous views. That commit also added the initial `request` argument (rather than just having `*args` and `**kwargs`, so I assume it's fine for this commit to do the same?

felixxm (approved):
@LomaxOnTheRun Thanks :+1:

PR #16268, opened 2022-11-07
Fixed #34036 -- Corrected low link color contrast in admin light-theme. 

nimra200:
This PR is for [improving color contrast in the admin light theme](https://code.djangoproject.com/ticket/34036). Google Chrome has a useful tool for website readability called [CSS Overview](https://developer.chrome.com/docs/devtools/accessibility/contrast/#overview-contrast). The tool indicates the color contrast ratios now meet AA guidelines of [WebAIM ](https://webaim.org/standards/wcag/#current).
Screen Shot 2022-11-07 at 3 43 34 PM
@carltongibson - Would appreciate if you could look at this!

PR #17073, opened 2023-07-14
Fixed #34036 -- Improved color contrast in admin light theme.

marianaasilva:
Fix for [ticket 34036](https://code.djangoproject.com/ticket/34036) to improve text contrast over light blue backgrounds in admin light-theme

Based on @nimra200's  patch [#16268](https://github.com/django/django/pull/16268)

Screenshot 2023-07-14 at 23 13 24

### DIFF
--- a/django/contrib/admin/static/admin/css/base.css
+++ b/django/contrib/admin/static/admin/css/base.css
@@ -22,11 +22,11 @@ html[data-theme="light"],
 
     --breadcrumbs-fg: #c4dce8;
     --breadcrumbs-link-fg: var(--body-bg);
-    --breadcrumbs-bg: var(--primary);
+    --breadcrumbs-bg: #264b5d;
 
     --link-fg: #417893;
     --link-hover-color: #036;
-    --link-selected-fg: #5b80b2;
+    --link-selected-fg: var(--secondary);
 
     --hairline-color: #e8e8e8;
     --border-color: #ccc;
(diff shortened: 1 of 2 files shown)


sarahboyce:
image
The colour contrasts look much better! @thibaudcolas can you take another look?

django/contrib/admin/static/admin/css/base.css
```diff
@@ -515,7 +516,7 @@ select[multiple] {
     padding: 10px 15px;
     border: none;
     border-radius: 4px;
-    color: var(--button-fg);
+    color: var(--body-fg);
```
felixxm, line 519: Check out [comment](https://github.com/django/django/pull/16268/files#discussion_r1217213838).

django/contrib/admin/static/admin/css/base.css
```diff
@@ -538,6 +539,7 @@ a.button {
     border: none;
     font-weight: 400;
     background: var(--default-button-bg);
+    color: var(--default-button-fg);
```
felixxm, line 542: Do we need to change it? It's not in the ticket description and it already has AA contrast: 
image

django/contrib/admin/static/admin/css/base.css
felixxm: Shouldn't we instead change a `button-bg`? This will not fix contrast issues in the other places where the `button-bg` is used, e.g. on pagination, and it's not enough for hover:
image

felixxm (commented):
@marianaasilva Thanks for this patch :+1: I left comments.

felixxm (commented):
@marianaasilva Thanks :+1: Welcome aboard :boat:

I found and fixed two more issues:
- selected date:
  image
- and _"All"_:
  image

PR #17077, opened 2023-07-16
Fixed #34657 -- Testing assertions are made more understandable

abdulaev-sh-m:
When `assertContains` and `assertInHTML` tests failed, it was not clear why the tests failed, because general information was displayed and it took a long time to find what caused the tests to fail.

In this regard, a [ticket](https://code.djangoproject.com/ticket/34657)was created and in this PR I fixed this drawback and made the tests more understandable.

### DIFF
--- a/django/test/testcases.py
+++ b/django/test/testcases.py
@@ -534,7 +534,8 @@ class SimpleTestCase(unittest.TestCase):
             )
         else:
             self.assertTrue(
-                real_count != 0, msg_prefix + "%s not found in %s" % (safe_repr(text_repr), safe_repr(response.content))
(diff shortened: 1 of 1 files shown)


sarahboyce:
Hi @abdulaev-sh-m :wave: thank you for working on this!

These are the next steps:
- You need to get the test suite passing and you would need to update some existing tests for the new behaviour
- This change needs a documentation update for the two methods
- This change needs a release note 

I've assigned the ticket to you on Trac and marked that it needs tests, docs and that the patch needs improvement. When you think you have resolved each of these things, you can update the ticket to mark all of those fields to "No" and we will take another look :+1:

PR #17082, opened 2023-07-17
Display cursor as pointer on admin filter summaries

christophehenry:
On the django admin site, it is not clear that the filters can be collapsed. Styling the cursor as pointer should solve that.


felixxm (approved):
@christophehenry Thanks for this patch :+1:

PR #17095, opened 2023-07-19
Refs #30052 -- Clarified that defer() and only() do not work with aggregated fields.

jestenough:
I think it would be useful to mention in the documentation about the behavior of **QuerySet.only()** with annotated fields (this has already been discussed in [ticket #30052](https://code.djangoproject.com/ticket/30052), but for some reason, it hasn't been updated in the documentation). Additionally, I believe it would be great to specify that by default, besides the fields provided to only(), '**id**' and consequently '**annotated fields**' are also included.

### DIFF
--- a/docs/ref/models/querysets.txt
+++ b/docs/ref/models/querysets.txt
@@ -1709,6 +1709,11 @@ never defer the loading of the primary key. If you are using
 loading of the field that connects from the primary model to the related
 one, doing so will result in an error.
 
+Similarly, calling ``defer()`` (or its counterpart :meth:`only()`) including an
+argument from an aggregation (e.g. using the result of :meth:`annotate()`)
+doesn't make sense: doing so will raise an exception. The aggregated values
+will always be fetched into the resulting queryset.
+
 .. note::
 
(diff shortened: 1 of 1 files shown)


sulaiman-coder (approved):

nessita:
Hello, thank you for your contribution!

I rebased this branch onto main and adjusted the text (and its location) a bit. Current changes look like:

* In the `defer` section:
image

* in the `only` section:
image

jestenough:
Hey,

It looks nice. But are you sure that the fields that were declared using **alias()** will also be loaded immediately?
I can vouch for the **annotated fields**, they are really included in the empty **only()**, but I'm not sure about **alias**, because **_explain_** doesn't show it to me

nessita:
> Hey,
> 
> It looks nice. But are you sure that the fields that were declared using **alias()** will also be loaded immediately? I can vouch for the **annotated fields**, they are really included in the empty **only()**, but I'm not sure about **alias**, because **_explain_** doesn't show it to me

Good point, thank you for raising it! I have made a small edition. If you are happy, once the tests pass I can land this.

nessita (approved):

PR #17110, opened 2023-07-25
Fixed #34657 -- Testing assertions are made more understandable

abdulaev-sh-m:
When assertContains and assertInHTML tests failed, it was not clear why the tests failed, because general information was displayed and it took a long time to find what caused the tests to fail.

In this regard, a [ticket](https://code.djangoproject.com/ticket/34657) was created and in this PR I fixed this drawback and made the tests more understandable.


.gitignore
```diff
@@ -6,6 +6,7 @@
 *.pot
 *.py[co]
 .tox/
+.idea/
```
sarahboyce, line 9: Can you remove this change? I know you probably wanted to ignore your pycharm config files or something like this but it's important to keep the PRs to only fix 1 thing at a time :+1:

package-lock.json
sarahboyce, line 1: Can you delete this file? I don't think we have decided to commit the `package-lock.json` yet (and similar idea about solve 1 thing per PR :+1: )

sarahboyce (commented):
Thank you @abdulaev-sh-m! I have some minor cleanup comments before we put this in the review queue :+1:

package-lock.json
abdulaev-sh-m: Thank you for helping. Deleted local files and changes in `.gitignore`

abdulaev-sh-m:
> Looks good!
> 
> Next steps:
> 
> * There are tests in `tests/test_utils/tests.py` such as `HTMLEqualTests.test_unequal_html` `HTMLEqualTests.test_contains_html`, I think we might need test cases to cover each of the asserts that have changed. Can you add these in?
> * Can you add a release note for this change?
> * Can you also add a documentation note like `.. versionchanged:: 5.0` for the documentation around these assert methods?

Tests to check for updates have already been written and I have updated them.
At the expense of documentation, I think that the update will not be needed, because there was no update as such

sarahboyce:
> Tests to check for updates have already been written and I have updated them. 

`assertInHTML` is not included in any updated tests

> At the expense of documentation, I think that the update will not be needed, because there was no update as such

I'm not sure I understand what you mean. 
As this would change the behaviour of `assertContains` `assertNotContains` and `assertInHTML`, we should document the behaviour change both in the release notes and the documentation. Otherwise our users might find this change surprising

tests/test_client_regress/tests.py
```diff
@@ -141,11 +141,15 @@ def test_contains(self):
...
-            self.assertIn("Couldn't find 'thrice' in response", str(e))
+            self.assertIn(
+                "False is not true : \"'thrice'\" not found in 'thrice'", str(e)
+            )
         try:
             self.assertContains(response, "thrice", msg_prefix="abc")
         except AssertionError as e:
-            self.assertIn("abc: Couldn't find 'thrice' in response", str(e))
+            self.assertIn(
+                "False is not true : abc: \"'thrice'\" not found in 'thrice'", str(e)
```
sarahboyce, line 151: This doesn't seem right, I would expect to see the response which doesn't contain the word `thrice`. It looks like the wrong thing has been added in here

PR #17116, opened 2023-07-27
Simplified django.http.request.split_domain_port().

ngnpope:
Use the capture groups from the regular expression that has already been matched to avoid re-splitting and the need to special case for IPv6.

### DIFF
--- a/tests/requests_tests/tests.py
+++ b/tests/requests_tests/tests.py
@@ -1013,10 +1013,37 @@ class HostValidationTests(SimpleTestCase):
         ):
             request.get_host()
 
-    def test_split_domain_port_removes_trailing_dot(self):
-        domain, port = split_domain_port("example.com.:8080")
-        self.assertEqual(domain, "example.com")
-        self.assertEqual(port, "8080")
+    def test_split_domain_port(self):
+        for host, expected in [
+            ("<invalid>", ("", "")),
+            ("<invalid>:8080", ("", "")),
+            ("example.com 8080", ("", "")),
+            ("example.com:invalid", ("", "")),
+            ("[::1]", ("[::1]", "")),
+            ("[::1]:8080", ("[::1]", "8080")),
+            ("[::ffff:127.0.0.1]", ("[::ffff:127.0.0.1]", "")),
+            ("[::ffff:127.0.0.1]:8080", ("[::ffff:127.0.0.1]", "8080")),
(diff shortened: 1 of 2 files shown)


tests/requests_tests/tests.py
```diff
@@ -1013,10 +1013,27 @@ def test_get_host_suggestion_of_allowed_host(self):
...
-        self.assertEqual(domain, "example.com")
-        self.assertEqual(port, "8080")
+    def test_split_domain_port(self):
+        for host, expected in [
+            ("<invalid>", ("", "")),
+            ("<invalid>:8080", ("", "")),
+            ("example.com 8080", ("", "")),
+            ("example.com:invalid", ("", "")),
+            ("[::1]", ("[::1]", "")),
+            ("[::1]:8080", ("[::1]", "8080")),
```
nessita, line 1023: Given that the regex supports IPv6 addresses, shall we add a test for one, such as:

            "[1851:0000:3238:DEF1:0177:0000:0000:0125]:8080"

And I guess we should add a few cases with domain with uppercase?

nessita (commented):
LGTM! Thanks

tests/requests_tests/tests.py
ngnpope: But `[::1]` is an IPv6 address!? Specifically it matches this part of the regex: `\[[a-f0-9]*:[a-f0-9.:]+\]`.

ngnpope: I guess we could add in other types of IPv6 address, e.g. an IPv4-mapped IPv6 address, e.g. `[::ffff:127.0.0.1]` which is why there is the `.` in that part of the regex pattern... 🤷🏻

ngnpope: Ok. I added both of the above - to ensure that IPv4-mapped IPv6 addresses work as expected. And also your IPv6 address above has uppercase characters in so it tests the existing behavior that lowercases the value.

nessita: Excellent, thank you! Looks great.

nessita (approved):

nessita:
I'll wait a bit in case @felixxm wants to review, otherwise I'll merge later today or tomorrow morning.

felixxm (approved):
@ngnpope Thanks :+1:

PR #17125, opened 2023-07-29
minor grammar fix in comment

therealrinku:
small grammar fix in comment . nothing crazy 😅


smithdc1:
Thanks for the patch. 

However, I think the current words are ok. I'm not sure it's worth changing.

shangxiao:
Yep just FYI "do something _if_ something else" is grammatically correct.

PR #17139, opened 2023-08-03
Fixed call read before call body

runtangr:
1. request obj call read function, self._read_started = True
2. request obj call body function, request obj don't have `_body`, but self._read_started = True.
    the `You cannot access body after reading from request's data stream` exception will raise

so, I add `_body` in read function.


runtangr:
I get it 
https://github.com/django/django/blob/9b9c805cedb08621bd5dc58a01a6478eb7cc49a9/tests/requests_tests/tests.py#L391-L393

PR #17158, opened 2023-08-11
Fixed #34624 - Removed edit and view options in RelatedFieldWidgetWrapper for radioselects.

allcaps:
Ref: https://code.djangoproject.com/ticket/34624

# Before

Current situation: There is only one pencil and eye, and they are disabled.
Expected: A pencil and eye for each item, and enabled OR no pencils and eyes.

In the code I found the comment: `# XXX: The UX does not support multiple selected values.`. But radio select was not excluded. I went for the solution to remove the pencils and eyes to make radio select behave the same as multi-select.

Screenshot 2023-08-11 at 14 25 12

# After

Screenshot 2023-08-11 at 14 25 31

I'd like to know if this approach is acceptable. If so, I'll update the PR with tests.

### DIFF
--- a/django/contrib/admin/tests.py
+++ b/django/contrib/admin/tests.py
@@ -23,6 +23,7 @@ class AdminSeleniumTestCase(SeleniumTestCase, StaticLiveServerTestCase):
         "django.contrib.contenttypes",
         "django.contrib.sessions",
         "django.contrib.sites",
+        "django.contrib.redirects",
     ]
 
     def wait_until(self, callback, timeout=10):
--- a/django/contrib/admin/widgets.py
+++ b/django/contrib/admin/widgets.py
@@ -270,13 +270,15 @@ class RelatedFieldWidgetWrapper(forms.Widget):
         if can_add_related is None:
             can_add_related = admin_site.is_registered(rel.model)
         self.can_add_related = can_add_related
-        # XXX: The UX does not support multiple selected values.
-        multiple = getattr(widget, "allow_multiple_selected", False)
-        self.can_change_related = not multiple and can_change_related
+        # Widgets that display multiple values are not supported.
+        supported = not getattr(
+            widget, "allow_multiple_selected", False
+        ) and not isinstance(widget, AdminRadioSelect)
+        self.can_change_related = supported and can_change_related
         # XXX: The deletion UX can be confusing when dealing with cascading deletion.
         cascade = getattr(rel, "on_delete", None) is CASCADE
-        self.can_delete_related = not multiple and not cascade and can_delete_related
-        self.can_view_related = not multiple and can_view_related
+        self.can_delete_related = supported and not cascade and can_delete_related
+        self.can_view_related = supported and can_view_related
         # so we can check if the related object is registered with this AdminSite
         self.admin_site = admin_site
 
--- a/tests/admin_views/admin.py
+++ b/tests/admin_views/admin.py
@@ -8,6 +8,10 @@ from django.contrib.admin import BooleanFieldListFilter
 from django.contrib.admin.views.main import ChangeList
 from django.contrib.auth.admin import GroupAdmin, UserAdmin
 from django.contrib.auth.models import Group, User
+from django.contrib.redirects.admin import RedirectAdmin
+from django.contrib.redirects.models import Redirect
+from django.contrib.sites.admin import SiteAdmin
+from django.contrib.sites.models import Site
 from django.core.exceptions import ValidationError
 from django.core.mail import EmailMessage
 from django.db import models
@@ -1299,9 +1303,11 @@ site.register(Box)
 site.register(Country, CountryAdmin)
 site.register(Traveler, TravelerAdmin)
 
-# Register core models we need in our tests
+# Register core and contrib models we need in our tests
 site.register(User, UserAdmin)
 site.register(Group, GroupAdmin)
+site.register(Redirect, RedirectAdmin)
+site.register(Site, SiteAdmin)
 
 # Used to test URL namespaces
 site2 = admin.AdminSite(name="namespaced_admin")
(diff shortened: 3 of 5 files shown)


nessita:
> I'd like to know if this approach is acceptable. If so, I'll update the PR with tests.

Hello, thank you for your contribution! I took a quick look, and I'm not sure the proposal fixes the ticket. The goal for proposal 2 is to make the `RelatedFieldWidgetWrapper` to actually allow for editing/inspecting the selected item, not to remove the disabled buttons.

Did you consider/try to work towards that solution? Did you find any blocker?

allcaps:
In the [Django tracker ticket](https://code.djangoproject.com/ticket/34624) the suggestion is made to remove the RadioSelect on the SitesAdmin. To me, that hides the issue, not solve it. So I initially set out to only enable the _single_ pencil and eye. I've discovered some things:

1. The pencil/eye should probably be on each option, not only the last one.
2. The WidgetWrapper wraps the widget, not individual options.
3. The WidgetWrapper is never intended to work with multi-select. See the comment `# XXX: The UX does not support multiple selected values.`.

I think, the original author(s) never intended to support a flat list of options. Maybe they just forgot to handle the radio select scenario? Maybe they just where not aware the JS also applies to radio selects?

With the knowledge I have now, I believe that enabling the single pencil and eye is _not_ a solution. It does not make sense to be able to edit one option.

I think adding pencils and eyes to both multi-select and radio select is an valid option. But this is a big change. We can't reuse any of the current wrapper widget, JS, and tests. As we need to target individual options. This is way more work then I'd like to do.

Removing a broken pencil/eye seems a valid solution. If the multi-select was never supported, why would we implement this for radio selects?

nessita:
> Yes, I suggested to go for option 2.
> 
> In the [Django tracker ticket](https://code.djangoproject.com/ticket/34624) the suggestion is made to remove the RadioSelect on the SitesAdmin. To me, that hides the issue, not solve it.

I agree!

> So I initially set out to only enable the _single_ pencil and eye. I've discovered some things:
> 
>     1. The pencil/eye should probably be on each option, not only the last one.

I sort of agree. Arguably the edit/view option could be only on the selected row, right? That would be similar to the behavior for the select widget.

>     2. The WidgetWrapper wraps the widget, not individual options.
> 
>     3. The WidgetWrapper is never intended to work with multi-select. See the comment `# XXX: The UX does not support multiple selected values.`.

Right, though please note that the radio select does not allow for multiple selected values.

> I think, the original author(s) never intended to support a flat list of options. Maybe they just forgot to handle the radio select scenario? Maybe they just where not aware the JS also applies to radio selects?
> 
> With the knowledge I have now, I believe that enabling the single pencil and eye is _not_ a solution. It does not make sense to be able to edit one option.
> 
> I think adding pencils and eyes to both multi-select options and radio select option is an valid solutions. But this is a lot of work. We can't reuse any of the current wrapper widget, JS, and tests. As we need to target individual options. This is way more work then I'd like to do.
> 
> Removing a broken pencil/eye seems a valid solution. If the multi-select was never supported, why would we implement this for radio selects?

Well, radio selects do not allow for a multi-selection of values, right? What do you think about enabling the edit and show actions only for the selected option?

nessita:
@thibaudcolas Would you any anu advice from the usability POV?

allcaps:
I wonder: did the radio select widget ever had a _working_ edit related object button? I took a dive.

# Django 1.1.3 (May 17, 2010)
Is the oldest Django installable via PyPi.
It already had the add plus below the radio select.
No edit pencil.
django-1 1 3

# Django 1.7.11 (Nov 24, 2015)
The last version with _only_ the add plus.
django-1 7 11

# Django 1.8 (Apr 1, 2015)
Introduces the edit pencil. 
The pencil doesn't work.
Note, I selected the option.
django-1 8

# Django 1.8.19 (Mar 6, 2018)
Some alignment issues.
The pencil still doesn't work.
django-1 8 19

# Django 1.9 (Dec 2, 2015)
Icons appear behind the _last_ option.
The pencil still doesn't work.
django-1 9

I also tried, Django 1.11, 2.2.28, 3.2, latest. Same story.

# 🤔 

- The pencil on radio select was dead on arrival.
- The pencil on radio select never worked.
- This might not be intentionally. Just a unnoticed side-effect of the pencil implementation? Maybe the authors never considered radio select?
- You'd need to install `django.contrib.redirects` to encounter a radio select in the admin. This app is not enabled by default. Only this app [specifies the radio select widget](https://github.com/django/django/blob/main/django/contrib/redirects/admin.py#L10).
- For admin users, there is little use for editing site objects when creating redirects. Site objects are probably created beforehand, and hardly ever change.
- It seems uncommon to specify the radio select widget in the admin.
- I can understand why there are no complaints about the broken pencil in decades.

Adding the pencil feature to radio select could be a nice addition to Django, but it is more work then I'd like to do. Especially because the JS implementation is how it is. And because it is a feature nobody uses.

So, I'm willing to push forward with the removal of the pencil. If another direction is desired, I'd leave it to others to do the work.

My main take-away: By removing the broken pencil from radio select, we remove something that never worked, and nobody misses.

thibaudcolas:
👍 the rationale makes a lot of sense to me, particularly since as @allcaps mentions:

> For admin users, there is little use for editing site objects when creating redirects

nessita:
Thank you Thibaud for your advise! In order to push this forward, @allcaps we should add at least a test covering the changed logic. Would you have time to work on that? Thanks!

nessita:
Closing due to inactivity. @allcaps please reopen if you find any time to work on this, thanks!

nessita:
Reopening following conversations in #17418

allcaps:
Keeping you posted. I've took anther stab at this. Adding an item won't update the radio select list with the newly created option. So, I've got to give this some more time.

nessita:
> Keeping you posted. I've took anther stab at this. Adding an item won't update the radio select list with the newly created option. So, I've got to give this some more time.

Thank you for the update! No worries if you take your time.

nessita:
Hello @allcaps, would you have time to keep working on this? If yes, could you please rebase onto latest main and fix tests? Thank you!

nessita:
Closing due to inactivity.

PR #17159, opened 2023-08-11
Fixed #34770: Removed autoescape from password_reset templates

Minkey27:
Removed `{% autoescape off %}`  from password_reset templates to safeguard against html injections


nessita:
Closing the PR following my comment in the ticket, we can later reopen if we agree in the ticket that this needs addressing. Thanks for your time!

PR #17160, opened 2023-08-11
failing test for contribute_to_field populating constraints

olivierdalang:
For now this is only a failing test case to run against official CI. Hope it's OK to open such a PR. At some point this will hopefully become a PR to support having fields populating DB constraints in contribute_to_fields.

See https://forum.djangoproject.com/t/dynamically-populate-constraints-in-field-contribute-to-class/22934 for full context.


olivierdalang:
(needs more work, will come back later)

Krozark:
that will be a nice feature to have

PR #17162, opened 2023-08-12
Fixes #32602 -- Clarify wording re: parallel testing and test case vs. test case class

MandadiMahendharReddy:
Ticket link: https://code.djangoproject.com/ticket/32602

### DIFF
--- a/django/test/runner.py
+++ b/django/test/runner.py
@@ -918,9 +918,10 @@ class DiscoverRunner:
         suite = self.test_suite(all_tests)
 
         if self.parallel > 1:
-            subsuites = partition_suite_by_case(suite)
+            subsuites = partition_suite_by_test_class(suite)
             # Since tests are distributed across processes on a per-TestCase
-            # basis, there's no need for more processes than TestCases.
+            # *class* basis, there's no need for more processes than TestCase
+            # *classes*.
             processes = min(self.parallel, len(subsuites))
             # Update also "parallel" because it's used to determine the number
             # of test databases.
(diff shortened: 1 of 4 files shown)


django/test/runner.py
```diff
@@ -1179,7 +1180,7 @@ def reorder_tests(tests, classes, reverse=False, shuffler=None):
         yield from reorder_test_bin(tests, shuffler=shuffler, reverse=reverse)
 
 
-def partition_suite_by_case(suite):
+def partition_suite_by_test_class(suite):
```
felixxm, line 1183: Please revert this change. PR should be limited to docs changes.

docs/ref/django-admin.txt
```diff
@@ -1484,8 +1484,7 @@ override this by passing the desired number of processes, e.g.
 variable.
 
 Django distributes test cases — :class:`unittest.TestCase` subclasses — to
-subprocesses. If there are fewer test cases than configured processes, Django
-will reduce the number of processes accordingly.
```
felixxm, line 1488: Why you removed: _"Django will reduce the number of processes accordingly"_? This ticket is about clarifying wording `test`/`test case`/`test case class`, etc.

MandadiMahendharReddy: Hi @felixxm,as per this
image I replaced that line with "If there are fewer test case classes than configured processes."

django/test/runner.py
MandadiMahendharReddy: Reverted

docs/ref/django-admin.txt
felixxm: Again, this ticket about clarifying wording `test`/`test case`/`test case class`, etc. The sentence that you removed is correct and should be kept.

felixxm: Also, this is only an example from the ticket description, you should audit all docs and check used wording ( `test`/`test case`/`test case class`).

felixxm:
Closing due to inactivity.

PR #17173, opened 2023-08-18
Fixed #34642 -- Added File.open() support for *args and **kwargs.

yvesonline:
Refactored Django File.open to support the full signature of Python builtin open

### DIFF
--- a/django/core/files/base.py
+++ b/django/core/files/base.py
@@ -105,11 +105,11 @@ class File(FileProxyMixin):
     def __exit__(self, exc_type, exc_value, tb):
         self.close()
 
-    def open(self, mode=None):
+    def open(self, mode=None, *args, **kwargs):
         if not self.closed:
             self.seek(0)
         elif self.name and os.path.exists(self.name):
-            self.file = open(self.name, mode or self.mode)
+            self.file = open(self.name, mode or self.mode, *args, **kwargs)
         else:
             raise ValueError("The file cannot be reopened.")
         return self
--- a/docs/ref/files/file.txt
+++ b/docs/ref/files/file.txt
@@ -46,17 +46,22 @@ The ``File`` class
 
         The read/write mode for the file.
 
-    .. method:: open(mode=None)
+    .. method:: open(mode=None, *args, **kwargs)
 
         Open or reopen the file (which also does ``File.seek(0)``).
         The ``mode`` argument allows the same values
-        as Python's built-in :func:`python:open()`.
+        as Python's built-in :func:`python:open()`. ``*args`` and ``**kwargs``
+        are passed after ``mode`` to Python's built-in :func:`python:open`.
 
         When reopening a file, ``mode`` will override whatever mode the file
         was originally opened with; ``None`` means to reopen with the original
         mode.
 
         It can be used as a context manager, e.g. ``with file.open() as f:``.
+
+        .. versionchanged:: 5.0
+
+            Support for passing ``*args`` and ``**kwargs`` was added.
 
     .. method:: __iter__()
 
--- a/tests/files/tests.py
+++ b/tests/files/tests.py
@@ -200,6 +200,21 @@ class FileTests(unittest.TestCase):
             self.assertIs(locks.unlock(f1), True)
             self.assertIs(locks.unlock(f2), True)
 
+    def test_open_supports_full_signature(self):
+        called = False
+
+        def opener(path, flags):
+            nonlocal called
+            called = True
+            return os.open(path, flags)
+
+        file_path = Path(__file__).parent / "test.png"
+        with open(file_path) as f:
+            test_file = File(f)
+
+        with test_file.open(opener=opener):
+            self.assertIs(called, True)
+
 
 class NoNameFileTestCase(unittest.TestCase):
     """
--- a/AUTHORS
+++ b/AUTHORS
@@ -1053,6 +1053,7 @@ answer newbie questions, and generally made Django that much better:
     Yoong Kang Lim <yoongkang.lim@gmail.com>
     Yury V. Zaytsev <yury@shurup.com>
     Yusuke Miyazaki <miyazaki.dev@gmail.com>
+    Yves Weissig <yves@weissig.me>
     yyyyyyyan <contact@yyyyyyyan.tech>
     Zac Hatfield-Dodds <zac.hatfield.dodds@gmail.com>
     Zachary Voase <zacharyvoase@gmail.com>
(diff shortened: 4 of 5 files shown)


django/core/files/base.py
```diff
@@ -105,11 +105,11 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_value, tb):
         self.close()
 
-    def open(self, mode=None):
+    def open(self, mode=None, *args, **kwargs):
```
felixxm, line 108: This change should be [documented](https://docs.djangoproject.com/en/4.2/ref/files/file/#django.core.files.File.open) with a `versionchanged` annotation and release note (`5.0.txt`).

tests/files/tests.py
```diff
@@ -200,6 +200,24 @@ def test_shared_lock(self):
             self.assertIs(locks.unlock(f1), True)
             self.assertIs(locks.unlock(f2), True)
 
+    def test_open_supports_full_signature(self):
+        """
+        Django File.open should support full signature of Python builtin open
+        #34642 File.open to support different encodings
```
felixxm, line 206: This docstring sounds unnecessary, we should at least simplify it, e.g.
```suggestion
        File.open supports full signature of Python's open.
```

felixxm (commented):
@yvesonline Thanks :+1: I left comments.

yvesonline:
> @yvesonline Thanks 👍 I left comments.

Thank you for the review @felixxm !

I applied your suggestion and added to the documentation, please let me know if it's OK like this.

tests/files/tests.py
```diff
@@ -200,6 +200,23 @@ def test_shared_lock(self):
             self.assertIs(locks.unlock(f1), True)
             self.assertIs(locks.unlock(f2), True)
 
+    def test_open_supports_full_signature(self):
+        """
+        File.open supports full signature of Python's open.
+        """
```
shangxiao, line 206: Are we still asking folks to remove these? 🤔

docs/ref/files/file.txt
```diff
@@ -46,18 +46,24 @@ The ``File`` class
 
         The read/write mode for the file.
 
-    .. method:: open(mode=None)
+    .. method:: open(mode=None, *args, **kwargs)
 
         Open or reopen the file (which also does ``File.seek(0)``).
         The ``mode`` argument allows the same values
         as Python's built-in :func:`python:open()`.
+        ``*args`` and ``*kwargs`` are passed to
```
shangxiao, line 54:
```suggestion
        ``*args`` and ``**kwargs`` are passed to
```

docs/ref/files/file.txt
```diff
@@ -46,18 +46,24 @@ The ``File`` class
...
+        Python's built-in :func:`python:open()`.
```
shangxiao, line 55: I wonder if we should say something like "are passed after mode" to be clear that mode is part of the signature, then additional args/kwargs, _except_ for mode, are passed.

tests/files/tests.py
felixxm: Yes, we can remove it. It's the same as the method name.

docs/ref/files/file.txt
yvesonline: Yes, I think that's a good idea.

tests/files/tests.py
```diff
@@ -200,6 +200,23 @@ def test_shared_lock(self):
...
+        mock_file.name = file_path
+        mock_file.mode = "r"
+        mock_file.closed = True
+        file = File(mock_file)
+        with mock.patch(
+            "builtins.open", mock.mock_open(read_data=b"content")
+        ) as mock_open, file.open(encoding="utf-8-sig", errors="strict"):
+            mock_open.assert_called_with(
+                file_path, "r", encoding="utf-8-sig", errors="strict"
+            )
```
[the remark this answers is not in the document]
yvesonline: Is there a reason I'm not aware of why we would want to avoid mocking?

I think mocking in a testing scenario is absolutely fine as it's intended to replace parts of the system under test with mock objects and then make assertions about how they have been used.

I think the `mock_open.assert_called_with` part in particular is very readable. A maintainer immediately knows what the intention was and what is being tested.

I would stick to the mocking approach unless there's reviewer consensus to do so differently.

yvesonline:
Thank you for the review @shangxiao !

I updated the documentation and responded to your review comments, please let me know if it's OK like this.

docs/ref/files/file.txt
```diff
@@ -46,18 +46,24 @@ The ``File`` class
...
         The read/write mode for the file.
 
-    .. method:: open(mode=None)
+    .. method:: open(mode=None, *args, **kwargs)
 
         Open or reopen the file (which also does ``File.seek(0)``).
         The ``mode`` argument allows the same values
         as Python's built-in :func:`python:open()`.
+        ``*args`` and ``**kwargs`` are passed after mode to
+        Python's built-in :func:`python:open()`.
```
shangxiao, line 55:
```suggestion
        Any ``*args`` and ``**kwargs`` passed after mode are forwarded to
        Python's built-in :func:`python:open()`.
```

shangxiao: @felixxm ^ does that sound better? 🤔

tests/files/tests.py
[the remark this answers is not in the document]
shangxiao: @felixxm Thoughts on this? It probably doesn't make much difference in this case but I'm (also) a fan of avoiding mocking when it's equally as easy to prove the actual integration works 👍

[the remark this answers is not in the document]
felixxm: I agree with David, mocking is always the last resort.

felixxm:
@yvesonline Do you have time to keep working on this? If not, I can push remaining edits.

yvesonline:
> @yvesonline Do you have time to keep working on this? If not, I can push remaining edits.

@felixxm I'll have time for this earliest this Sunday (10th of Sep), I'm happy for you to push the remaining edits if you want to close it.

Thanks!

yvesonline:
Thank you again for your review @felixxm & @shangxiao !

I've pushed the remaining edits and it should be ready to go now.

PS: Do you mind pointing me at the "_avoid mock_" documentation? I might have missed a convention and would like to read up!

felixxm:
@yvesonline Thanks :+1: Welcome aboard :boat: 

> PS: Do you mind pointing me at the "_avoid mock_" documentation? I might have missed a convention and would like to read up!

This is not a strict policy, so it's not documented.

shangxiao:
> This is not a strict policy, so it's not documented.

Also if you're genuinely looking for discussion on this head on over to the Django Discord or forum and start a thread. I'll keep an eye out and respond if I get some time 👀

PR #17177, opened 2023-08-21
Improved formset docs by using a set instead of a list in the custom validation example.

lallulli:
In the formset validation official example, we decrease the computational complexity from O(n^2) to O(n log n) simply by using a set instead of a list.

### DIFF
--- a/docs/topics/forms/formsets.txt
+++ b/docs/topics/forms/formsets.txt
@@ -374,14 +374,14 @@ is where you define your own validation that works at the formset level:
     ...         if any(self.errors):
     ...             # Don't bother validating the formset unless each form is valid on its own
     ...             return
-    ...         titles = []
+    ...         titles = set()
     ...         for form in self.forms:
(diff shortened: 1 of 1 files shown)


nessita (approved):
While I don't think the performance in this example is relevant, I do agree that using a `set()` is conceptually more correct.

Regarding performance, it's important to note that what it makes the real difference (if that would be an issue in a real world case), is to find an element in the container:

```
$ python -m timeit -s 'l = list(range(1000))' '101 in l'
200000 loops, best of 5: 1.12 usec per loop
$ python -m timeit -s 's = set(range(1000))' '101 in s'
10000000 loops, best of 5: 31 nsec per loop
```

And not so much the appending/adding:

```
$ python -m timeit -s 'l = list(range(1000))' 'l.append(1)'
10000000 loops, best of 5: 33.6 nsec per loop
$ python -m timeit -s 'l = set(range(1000))' 'l.add(1)'
10000000 loops, best of 5: 39.2 nsec per loop
```

PR #17182, opened 2023-08-22
Refs #34757 -- Moved HTTP redirect logic to django.test.client.ClientMixin.

olibook:
related to https://github.com/django/django/pull/17168#discussion_r1299188864

move test client follow redirect logic to ClientMixin so that it can be used as well for the AsyncClient


felixxm (commented):
@olibook Thanks :+1:

PR #17204, opened 2023-08-27
Fixed #34657 -- Made assert(Not)Contains/assertInHTML display haystacks in error messages.

chinmoy12c:
This PR adds the feature of showing the haystack on failure for these methods. 

  - django.test.SimpleTestCase.assertContains
  - django.test.SimpleTestCase.assertNotContains
  - django.test.SimpleTestCase.assertInHtml

Fixes: [ticket](https://code.djangoproject.com/ticket/34657)

### DIFF
--- a/django/test/testcases.py
+++ b/django/test/testcases.py
@@ -495,6 +495,7 @@ class SimpleTestCase(unittest.TestCase):
             content = b"".join(response.streaming_content)
         else:
             content = response.content
+        content_repr = safe_repr(content)
         if not isinstance(text, bytes) or html:
             text = str(text)
             content = content.decode(response.charset)
@@ -509,7 +510,7 @@ class SimpleTestCase(unittest.TestCase):
                 self, text, None, "Second argument is not valid HTML:"
             )
         real_count = content.count(text)
-        return (text_repr, real_count, msg_prefix)
+        return text_repr, real_count, msg_prefix, content_repr
 
     def assertContains(
         self, response, text, count=None, status_code=200, msg_prefix="", html=False
@@ -521,7 +522,7 @@ class SimpleTestCase(unittest.TestCase):
         If ``count`` is None, the count doesn't matter - the assertion is true
         if the text occurs at least once in the response.
         """
-        text_repr, real_count, msg_prefix = self._assert_contains(
+        text_repr, real_count, msg_prefix, content_repr = self._assert_contains(
             response, text, status_code, msg_prefix, html
         )
 
@@ -529,13 +530,18 @@ class SimpleTestCase(unittest.TestCase):
             self.assertEqual(
                 real_count,
                 count,
-                msg_prefix
-                + "Found %d instances of %s in response (expected %d)"
-                % (real_count, text_repr, count),
+                (
+                    f"{msg_prefix}Found {real_count} instances of {text_repr} "
+                    f"(expected {count}) in the following response\n{content_repr}"
+                ),
             )
         else:
             self.assertTrue(
-                real_count != 0, msg_prefix + "Couldn't find %s in response" % text_repr
+                real_count != 0,
+                (
+                    f"{msg_prefix}Couldn't find {text_repr} in the following response\n"
+                    f"{content_repr}"
+                ),
             )
 
     def assertNotContains(
@@ -546,12 +552,17 @@ class SimpleTestCase(unittest.TestCase):
         successfully, (i.e., the HTTP status code was as expected) and that
         ``text`` doesn't occur in the content of the response.
         """
-        text_repr, real_count, msg_prefix = self._assert_contains(
+        text_repr, real_count, msg_prefix, content_repr = self._assert_contains(
             response, text, status_code, msg_prefix, html
         )
 
         self.assertEqual(
-            real_count, 0, msg_prefix + "Response should not contain %s" % text_repr
+            real_count,
+            0,
+            (
+                f"{msg_prefix}{text_repr} unexpectedly found in the following response"
+                f"\n{content_repr}"
+            ),
         )
 
     def _check_test_client_response(self, response, attribute, method_name):
@@ -884,17 +895,23 @@ class SimpleTestCase(unittest.TestCase):
         real_count = parsed_haystack.count(parsed_needle)
         if msg_prefix:
             msg_prefix += ": "
+        haystack_repr = safe_repr(haystack)
         if count is not None:
             self.assertEqual(
                 real_count,
                 count,
-                msg_prefix
-                + "Found %d instances of '%s' in response (expected %d)"
-                % (real_count, needle, count),
+                (
+                    f"{msg_prefix}Found {real_count} instances of {needle!r} (expected "
+                    f"{count}) in the following response\n{haystack_repr}"
+                ),
             )
         else:
             self.assertTrue(
-                real_count != 0, msg_prefix + "Couldn't find '%s' in response" % needle
+                real_count != 0,
+                (
+                    f"{msg_prefix}Couldn't find {needle!r} in the following response\n"
+                    f"{haystack_repr}"
+                ),
             )
 
     def assertJSONEqual(self, raw, expected_data, msg=None):
--- a/docs/releases/5.1.txt
+++ b/docs/releases/5.1.txt
@@ -201,7 +201,10 @@ Templates
 Tests
 ~~~~~
 
-* ...
+* :meth:`~django.test.SimpleTestCase.assertContains`,
+  :meth:`~django.test.SimpleTestCase.assertNotContains`, and
+  :meth:`~django.test.SimpleTestCase.assertInHTML` assertions now add haystacks
+  to assertion error messages.
 
 URLs
 ~~~~
--- a/docs/topics/testing/tools.txt
+++ b/docs/topics/testing/tools.txt
@@ -1700,6 +1700,10 @@ your test suite.
     attribute ordering is not significant. See
     :meth:`~SimpleTestCase.assertHTMLEqual` for more details.
 
+    .. versionchanged:: 5.1
+
+        In older versions, error messages didn't contain the response content.
+
 .. method:: SimpleTestCase.assertNotContains(response, text, status_code=200, msg_prefix='', html=False)
 
     Asserts that a :class:`response <django.http.HttpResponse>` produced the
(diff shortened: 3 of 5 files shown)


felixxm (commented):
@chinmoy12c Thanks for this patch :+1: I left comments.

django/test/testcases.py
```diff
@@ -524,17 +524,23 @@ def assertContains(
...
-                + "Found %d instances of %s in response (expected %d)"
-                % (real_count, text_repr, count),
+                + "Found %d instances of %s in %s (expected %d)"
+                % (real_count, text_repr, content, count),
             )
         else:
             self.assertTrue(
-                real_count != 0, msg_prefix + "Couldn't find %s in response" % text_repr
+                real_count != 0,
+                msg_prefix + "Couldn't find %s in %s" % (text_repr, content),
```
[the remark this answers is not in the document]
shangxiao: I was thinking this as well… I think there's the potential for printing out far too much content to the console for _extremely_ long strings.

At first glances I'd say we'd need to account for this though not sure how (was thinking we could crop it down but that might not be useful?) Should we provide an option to turn off content? Can we utilise `maxDiff` here?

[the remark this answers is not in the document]
shangxiao: ps what is "|n%s" … I've never seen that before or is it `\n` rendering funny? 🤔

django/test/testcases.py
```diff
@@ -524,17 +524,23 @@ def assertContains(
...
+
         if count is not None:
             self.assertEqual(
                 real_count,
                 count,
                 msg_prefix
-                + "Found %d instances of %s in response (expected %d)"
-                % (real_count, text_repr, count),
+                + "Found %d instances of %s in %s (expected %d)"
+                % (real_count, text_repr, content, count),
```
shangxiao, line 538: I'm testing this out and am getting assertion failure messages like this:

> AssertionError: 0 != 1 : Found 0 instances of 'never' in b'No template used. Sample content: twice once twice. Content ends.' (expected 1)

Do we need to handle byte strings so we don't have the `b` show up?

django/test/testcases.py
```diff
@@ -524,17 +524,23 @@ def assertContains(
             response, text, status_code, msg_prefix, html
         )
 
+        if response.streaming:
+            content = b"".join(response.streaming_content)
+        else:
+            content = response.content
```
shangxiao, line 530:
```suggestion
        content = b"".join(response.streaming_content) if response.streaming else response.content
```

This is merely a suggestion but I've seen reviewers recommend "condition expression" 😊  Not saying change it - just an option!

shangxiao: Sorry to bikeshed, can't help myself 🤣

A possibly more pythonic option could be:

```suggestion
        try:
            content = response.content
        except AttributeError:
            content = b"".join(response.streaming_content)
```

django/test/testcases.py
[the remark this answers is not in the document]
felixxm: > ps what is "|n%s" … I've never seen that before or is it `\n` rendering funny? 🤔

It is `\n` in all examples.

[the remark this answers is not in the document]
chinmoy12c: Hey @felixxm, I've made the changes, let me know whether we should incorporate the feature of turning off content/leverage`maxDiff` in some way.

django/test/testcases.py
```diff
@@ -524,17 +524,26 @@ def assertContains(
...
+                + "Found %d instances of %s (expected %d) in the following response\n%s"
+                % (real_count, text_repr, count, content),
             )
         else:
             self.assertTrue(
-                real_count != 0, msg_prefix + "Couldn't find %s in response" % text_repr
+                real_count != 0,
+                msg_prefix
+                + "Couldn't find %s in the following response\n%s"
+                % (text_repr, content),
```
felixxm, line 546: IMO, we should use `safe_repr(content, True)` in all methods.

felixxm: `safe_repr(..., short=True)` truncates a content so we should use `safe_repr(content)`. It also shows that we need better tests.

django/test/testcases.py
```diff
@@ -889,18 +908,23 @@ def assertInHTML(self, needle, haystack, count=None, msg_prefix=""):
...
         if count is not None:
             self.assertEqual(
                 real_count,
                 count,
-                msg_prefix
-                + "Found %d instances of '%s' in response (expected %d)"
-                % (real_count, needle, count),
+                msg_prefix + "Found %d instances of '%s' (expected %d)"
+                " in the following response\n%s"
+                % (real_count, needle, count, safe_repr(haystack)),
```
thibaudcolas, line 920: Here and below, noting one of my comments from the ticket hasn’t been addressed:

> Additionally for assertInHTML – it’s annoying that needle as displayed in the failure message is the parser’s output, which is therefore broken up over multiple lines. It’d be much nicer if the example above resulted in AssertionError: False is not true : Couldn't find '<b>hey</b>' in response.

This is definitely in the "nice to have" territory, but I can’t think of a better place to address this than now.

To illustrate, with a `self.assertInHTML("<b>hey</b>", "<p>Howdy!</p>")`, Django’s existing output is:

```
AssertionError: False is not true : Couldn't find '<b>
hey
</b>' in response
```

The output proposed here would be:

```
AssertionError: False is not true : Couldn't find '<b>
hey
</b>' in the following response
'<p>
Howdy!
</p>'
```

My preference would be:

```
AssertionError: False is not true : Couldn't find '<b>hey</b>' in the following response
'<p>Howdy!</p>'
```

---

For this to work it’s just a matter of not overriding the `needle` and `haystack` variables at the top of the method, so we print the parameters as-is.

django/test/testcases.py
```diff
@@ -889,18 +908,23 @@ def assertInHTML(self, needle, haystack, count=None, msg_prefix=""):
         haystack = assert_and_parse_html(
             self, haystack, None, "Second argument is not valid HTML:"
         )
+        if msg_prefix:
+            msg_prefix += ": "
```
thibaudcolas, line 912: This isn’t related to the other changes here. I can see why this change would be worthwhile for consistency with other methods, though it seems there are other methods that are missing this, so perhaps worth addressing in another PR that changes this everywhere. 

If we do want to proceed with this as-is, I guess this does need to be acknowledged separately in the release notes.

django/test/testcases.py
```diff
@@ -549,8 +558,18 @@ def assertNotContains(
...
+            if response.streaming
+            else response.content
+        )
+
         self.assertEqual(
-            real_count, 0, msg_prefix + "Response should not contain %s" % text_repr
+            real_count,
+            0,
+            msg_prefix
+            + "%s unexpectedly found in the following response\n%s"
```
thibaudcolas, line 571: Nice that this is now more consistent with Python’s `assertIn` and Django’s `assertTemplateNotUsed`! I wish we used simpler language than "unexpectedly" but consistency matters more.

thibaudcolas (changes requested):
Looking solid! There is an unrelated change I’d recommend booting to a separate PR, and one thing from the ticket that feels like we should address while we’re here.

django/test/testcases.py
chinmoy12c: I think we can keep this one in this PR as is and create a separate ticket for addressing this change in all the methods that do not have this.

thibaudcolas: Sounds good to me, you’ll need to add an entry in the release notes for this as well then.

chinmoy12c: Done 👍

docs/releases/5.0.txt
```diff
@@ -505,6 +505,13 @@ Tests
 
 * The new :option:`test --durations` option allows showing the duration of the
   slowest tests on Python 3.12+.
+* :meth:`~django.test.SimpleTestCase.assertContains`,
```
pauloxnet, line 508:
```suggestion

* :meth:`~django.test.SimpleTestCase.assertContains`,
```
Add an empty line before.

chinmoy12c:
Hey @felixxm, could you review this for the final merge?

shangxiao:
@chinmoy12c Hi, nessita & felixx are busy, please be patient and be assured this is on their todo list – there's no need to tag them personally ;)

django/test/testcases.py
```diff
@@ -524,17 +524,26 @@ def assertContains(
...
                 msg_prefix
-                + "Found %d instances of %s in response (expected %d)"
-                % (real_count, text_repr, count),
+                + "Found %d instances of %s (expected %d) in the following response\n%s"
+                % (real_count, text_repr, count, safe_repr(content)),
             )
         else:
             self.assertTrue(
-                real_count != 0, msg_prefix + "Couldn't find %s in response" % text_repr
+                real_count != 0,
```
ngnpope, line 543: Unrelated to the change, but this would be better using `self.assertNotEqual()` 🤔 (Can be done as a follow up fixing other similar cases.)

django/test/testcases.py
```diff
@@ -883,24 +902,29 @@ def assertHTMLNotEqual(self, html1, html2, msg=None):
...
-                + "Found %d instances of '%s' in response (expected %d)"
-                % (real_count, needle, count),
+                msg_prefix + "Found %d instances of '%s' (expected %d)"
+                " in the following response\n%s"
+                % (real_count, needle, count, safe_repr(haystack)),
             )
         else:
             self.assertTrue(
-                real_count != 0, msg_prefix + "Couldn't find '%s' in response" % needle
+                real_count != 0,
```
ngnpope, line 924: Ditto.

django/test/testcases.py
felixxm: Please remove it for this patch as it's not related with the ticket. We don't mix different features/bugfixes/cleanups in a single commit.

django/test/testcases.py
```diff
@@ -873,24 +892,27 @@ def assertHTMLNotEqual(self, html1, html2, msg=None):
             self.fail(self._formatMessage(msg, standardMsg))
 
     def assertInHTML(self, needle, haystack, count=None, msg_prefix=""):
-        needle = assert_and_parse_html(
+        parsed_needle = assert_and_parse_html(
```
felixxm, line 895: I moved it with a new test to the #17321.

felixxm (commented):
@chinmoy12c Thanks :+1: I left more comments.

django/test/testcases.py
felixxm: I moved it to the #17332.

felixxm:
@chinmoy12c Thanks :+1: I pushed final edits.

thibaudcolas:
Yay! ☺️

PR #17213, opened 2023-08-31
Write a '\n' when count > total_count

zealoussnow:


zealoussnow:
>

zealoussnow:
> Hello @zealoussnow, thank you for your contribution. Could you explain a bit more the use case or bug this PR is addressing/fixing?

I'm sorry, I close the PR by mistake. I write a demo using class ProgressBar, and I found if I pass the variable count big than 100 to update method, a new line will not show. So I think if a new line print at this moment will be better.

```
if done == self.progress_width
```

image

```
if done >= self.progress_width
```
image

shangxiao (commented):
I didn't even know we had a progress bar 😅

At first glances, yes, this would be the more "correct" defensive way to guard here, however… this only fixes 1/2 of the problem.

Playing with the progress bar, if I put in an `update` more than the `total_count` then it prints the progress bar out on separate lines.

The correct thing to do here, if we wanted to even fix this given that it's probably never even used publicly, would be to cap the `perc` variable at 100.

zealoussnow:
@shangxiao Thank for your advise, please review my new changes.

felixxm:
@zealoussnow Thanks :+1: I don't mind making `ProgressBar` more bulletproof (even if it's only a private API), but we need a regression test for this cleanup.
   
Do you have time to keep working on this?

shangxiao:
Oops I missed the notification to review this.

I think the `perc` capping is better 👍  Just tested it and it works better.

There may be other things to do here - like once a `\n` is printed then there's no point in processing anymore input.

Or we could just leave it here as it fixes weird output already.

shangxiao (approved):

shangxiao:
@felixxm I'm ok with changes made so far - anything you wanted to add?

felixxm:
_"... but we need a regression test for this cleanup."_ https://github.com/django/django/pull/17213#issuecomment-1754322353

felixxm:
> @felixxm I'm ok with changes made so far - anything you wanted to add?

_"... but we need a regression test for this cleanup."_ https://github.com/django/django/pull/17213#issuecomment-1754322353

zealoussnow:
> > @felixxm I'm ok with changes made so far - anything you wanted to add?
> 
> _"... but we need a regression test for this cleanup."_ [#17213 (comment)](https://github.com/django/django/pull/17213#issuecomment-1754322353)

Thanks, all tests passed.

felixxm:
> > > @felixxm I'm ok with changes made so far - anything you wanted to add?
> > 
> > 
> > _"... but we need a regression test for this cleanup."_ [#17213 (comment)](https://github.com/django/django/pull/17213#issuecomment-1754322353)
> 
> Thanks, all tests passed.

We need a **new** test, check out [docs](https://docs.djangoproject.com/en/stable/intro/contributing/#writing-some-tests-for-your-ticket).

felixxm:
Closing due to inactivity.

PR #17214, opened 2023-08-31
Fixed #24561 -- Added support for callables on model fields' choices.

nessita:
This is a follow up of PR #16943 ([ticket-31262](https://code.djangoproject.com/ticket/31262)). Most of the ground work was done in that PR, this branch removes the system check and adds some tests to ensure the feature works as expected.

### DIFF
--- a/docs/ref/models/fields.txt
+++ b/docs/ref/models/fields.txt
@@ -115,9 +115,32 @@ human-readable name. For example::
         ("GR", "Graduate"),
     ]
 
+``choices`` can also be defined as a callable that expects no arguments and
+returns any of the formats described above. For example::
+
+    def get_currencies():
+        return {i: i for i in settings.CURRENCIES}
+
+
+    class Expense(models.Model):
+        amount = models.DecimalField(max_digits=10, decimal_places=2)
+        currency = models.CharField(max_length=3, choices=get_currencies)
+
+Passing a callable for ``choices`` can be particularly handy when, for example,
+the choices are:
+
+* the result of I/O-bound operations (which could potentially be cached), such
+  as querying a table in the same or an external database, or accessing the
+  choices from a static file.
+
+* a list that is mostly stable but could vary from time to time or from
+  project to project. Examples in this category are using third-party apps that
+  provide a well-known inventory of values, such as currencies, countries,
+  languages, time zones, etc.
+
 .. versionchanged:: 5.0
 
-    Support for mappings was added.
+    Support for mappings and callables was added.
 
 Generally, it's best to define choices inside a model class, and to
 define a suitably-named constant for each value::
--- a/docs/releases/5.0.txt
+++ b/docs/releases/5.0.txt
@@ -157,14 +157,14 @@ form::
     ]
 
 
-    class Winners(models.Model):
+    class Winner(models.Model):
         name = models.CharField(...)
         medal = models.CharField(..., choices=Medal.choices)
         sport = models.CharField(..., choices=SPORT_CHOICES)
 
-Django 5.0 supports providing a mapping instead of an iterable, and also no
-longer requires ``.choices`` to be used directly to expand :ref:`enumeration
-types <field-choices-enum-types>`::
+Django 5.0 adds support for accepting a mapping or a callable instead of an
+iterable, and also no longer requires ``.choices`` to be used directly to
+expand :ref:`enumeration types <field-choices-enum-types>`::
 
     from django.db import models
 
@@ -177,13 +177,20 @@ types <field-choices-enum-types>`::
     }
 
 
-    class Winners(models.Model):
+    def get_scores():
+        return [(i, str(i)) for i in range(10)]
+
+
+    class Winner(models.Model):
         name = models.CharField(...)
         medal = models.CharField(..., choices=Medal)  # Using `.choices` not required.
         sport = models.CharField(..., choices=SPORT_CHOICES)
+        score = models.IntegerField(choices=get_scores)  # A callable is allowed.
 
 Under the hood the provided ``choices`` are normalized into a list of 2-tuples
-as the canonical form whenever the ``choices`` value is updated.
+as the canonical form whenever the ``choices`` value is updated. For more
+information, please check the :ref:`model field reference on choices
+<field-choices>`.
 
 Minor features
 --------------
--- a/django/db/models/fields/__init__.py
+++ b/django/db/models/fields/__init__.py
@@ -316,9 +316,7 @@ class Field(RegisterLookupMixin):
         if not self.choices:
             return []
 
-        if not is_iterable(self.choices) or isinstance(
-            self.choices, (str, CallableChoiceIterator)
-        ):
+        if not is_iterable(self.choices) or isinstance(self.choices, str):
             return [
                 checks.Error(
                     "'choices' must be a mapping (e.g. a dictionary) or an iterable "
--- a/tests/invalid_models_tests/test_ordinary_fields.py
+++ b/tests/invalid_models_tests/test_ordinary_fields.py
@@ -391,26 +391,6 @@ class CharFieldTests(TestCase):
                     ],
                 )
 
-    def test_choices_callable(self):
-        def get_choices():
-            return [(i, i) for i in range(3)]
-
-        class Model(models.Model):
-            field = models.CharField(max_length=10, choices=get_choices)
-
-        field = Model._meta.get_field("field")
-        self.assertEqual(
-            field.check(),
-            [
-                Error(
-                    "'choices' must be a mapping (e.g. a dictionary) or an iterable "
-                    "(e.g. a list or tuple).",
-                    obj=field,
-                    id="fields.E004",
-                ),
-            ],
-        )
-
     def test_bad_db_index_value(self):
         class Model(models.Model):
             field = models.CharField(max_length=10, db_index="bad")
--- a/tests/migrations/test_writer.py
+++ b/tests/migrations/test_writer.py
@@ -31,6 +31,10 @@ from django.utils.translation import gettext_lazy as _
 from .models import FoodManager, FoodQuerySet
 
 
+def get_choices():
+    return [(i, str(i)) for i in range(3)]
+
+
 class DeconstructibleInstances:
     def deconstruct(self):
         return ("DeconstructibleInstances", [], {})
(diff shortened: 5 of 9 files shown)


docs/releases/5.0.txt
```diff
@@ -150,14 +150,14 @@ form::
...
-    class Winners(models.Model):
+    class Winner(models.Model):
         name = models.CharField(...)
         medal = models.CharField(..., choices=Medal.choices)
         sport = models.CharField(..., choices=SPORT_CHOICES)
 
-Django 5.0 supports providing a mapping instead of an iterable, and also no
-longer requires ``.choices`` to be used directly to expand :ref:`enumeration
-types <field-choices-enum-types>`::
+Django 5.0 added support for accepting a mapping or a callable instead of an
```
ngnpope, line 158: Use present tense when referring to the version being released in its release notes:

```suggestion
Django 5.0 adds support for accepting a mapping or a callable instead of an
```

docs/ref/models/fields.txt
```diff
@@ -115,9 +115,12 @@ human-readable name. For example::
         ("GR", "Graduate"),
     ]
 
+``choices`` can also be defined as a callable that expects no arguments and
+returns any of the formats described above.
```
ngnpope, line 119: I think a small example would be useful somewhere in this documentation, in particular showing that the callable is passed in without being called, i.e. `CharField(choices=get_choices)` and not `CharField(choices=get_choices())`. It may also be worth highlighting why you'd want to use a callable, e.g. delaying I/O operations until first required to avoid slowing down start up.

_(As an aside, this makes me also wonder if it's also worth testing that a callable with, e.g. `@functools.cache`, will work correctly - and perhaps mentioning that this is supported as a way of lazily evaluating the choices once.)_

ngnpope: I noticed that there are no changes to the documentation for `choices` in the forms documentation. Do we need to add anything there or is there already something mentioned?

docs/releases/5.0.txt
```diff
@@ -170,13 +170,20 @@ types <field-choices-enum-types>`::
...
-    class Winners(models.Model):
+    def get_scores():
+        return [(i, str(i)) for i in range(10)]
+
+
+    class Winner(models.Model):
         name = models.CharField(...)
         medal = models.CharField(..., choices=Medal)  # Using `.choices` not required.
         sport = models.CharField(..., choices=SPORT_CHOICES)
+        score = models.IntegerField(choices=get_scores)
```
ngnpope, line 181:
```suggestion
        score = models.IntegerField(choices=get_scores)  # Passing a callable.
```

docs/releases/5.0.txt
```diff
@@ -170,13 +170,20 @@ types <field-choices-enum-types>`::
     }
 
 
-    class Winners(models.Model):
+    def get_scores():
+        return [(i, str(i)) for i in range(10)]
```
ngnpope, line 174: Following on from my comment above about caching. If that works (without affecting the migrations stuff) then we could highlight that here?

ngnpope (changes requested):
Looks generally quite simple 🎉 

I have some comments for you related to documentation.

docs/ref/models/fields.txt
nessita: Thanks Nick for taking a look into this. You have a valid point, we should add a valid use case for using callable choices to the docs. I'll work on that. I have tested and cached callables work OK, so I'll have an example added to the docs.

Regarding no changes for the forms documentation: accepting callables in ChoiceField is a documented feature [since long time ago, it was added in Django 1.8](https://docs.djangoproject.com/en/4.2/ref/forms/fields/#django.forms.ChoiceField.choices) so as far as I understand this work has no impact in form's `ChoiceField`.

docs/releases/5.0.txt
nessita: This does work indeed, though I would not add that detail in this What's New release note (I would not want to obfuscate this section with specific use cases).

Down below the user is invited to read the reference about model field's if they wish to know more, and since I'll be expanding that section as discussed in previous comments, I think we have the caching covered.

ngnpope:
Thanks @nessita. Interesting... So I can see that there are a number of potential reasons why you may want to use a callable:

- Performing I/O to gather the choices (which could potentially be cached), e.g.
  - a query of a table in the same database (which is used for associated metadata for the choice values)
  - a query to a different external database (maybe for integration purposes)
  - accessing the choices from a static file
- Avoiding migration churn for choices that can be added to periodically
  - e,g. using the available timezones in `tzdata` (or previously `pytz`)
- Fully dynamic choices as per the example you provided above

If we cannot solve the case for the fully dynamic case, I think it's still worth adding this support for callables for some of the other use case and we can slap in a chunky warning that more dynamic cases are not supported, i.e. you can return additional choices, but not remove them otherwise some things such as the admin will not behave.

Otherwise I don't know whether we can make a form field show the existing value plus the available values somehow and allow the existing value to validate. It starts to get a bit tricky and would require application specific logic. Also there would be challenges around how to display the value as the display label value would be unknown...

docs/ref/models/fields.txt
```diff
@@ -115,9 +115,33 @@ human-readable name. For example::
...
+
+
+    @functools.cache
+    def fib(n):
+        if n < 2:
+            return n
+        return fib(n - 1) + fib(n - 2)
+
+
+    @functools.cache
```
ngnpope, line 133: It might be worth checking that `@functools.cache` works nicely in the migrations serializer by adding a test...

I think that, as long as decorators use `functools.wraps()` properly we should get the right output from `FunctionTypeSerializer`, but probably worth checking.

nessita: Thank you for raising this. As I think you suspected, the migrations serializer fails with

```
ValueError: Cannot serialize: <functools._lru_cache_wrapper object at 0x7f819ab3b270>
```

I've checked and the Python code does not wraps the function, so I created [a topic in the Python discourse](https://discuss.python.org/t/why-functools-cache-or-lru-cache-does-not-seem-to-wraps-the-passed-function/32965) (though it was held for moderation :eyeglasses:). Depending on what the Python devs say, this behavior may be a bug or not, and we may need to workaround it in our migration serializer.

nessita:
> I can see that there are a number of potential reasons why you may want to use a callable:

I agree on the uses cases you shared and I can add my own: I have a pet project for recording household expenses, and each expense can have one tag. The tag list is a setting, since that list may vary from family to family, and I explicitly avoided made it an M2M because I don't want to have an ever growing table for tags that can be considered static within one deploy.

So right now my `Expense` model has an `CharField` to store the tag, and the choices point to the setting. I had to manually modify the migration introducing this to avoid a new migration every time the setting was changed. If a callable would be allowed, that would have solved my problem.

docs/ref/models/fields.txt
nessita: Turns out we may need to update the migrations serializer to either:
  1. Provide a specific serializer for the `_lru_cache_wrapper` type, or
  2. Provide a serializer for anything that has a `__wrapped__` attribute

Do you have an opinion on the above? Whatever we do, it may need a discussion in the forum and likely a separated ticket/PR.

nessita: I created [ticket 34822](https://code.djangoproject.com/ticket/34822) to provide support for the `_lru_cache_wrapper`, which I think, at this stage, will be a 5.1 feature (if accepted).

nessita:
@ngnpope I have re-worked the docs using your feedback, ideally we can merge this soon so it makes the cut for the feature freeze (which is scheduled for next Monday!).

Let me know your thoughts, thanks!

ngnpope (approved):
Looks straightforward. Only a few small comments, but otherwise I'm happy.

docs/ref/models/fields.txt
```diff
@@ -115,9 +115,41 @@ human-readable name. For example::
         ("GR", "Graduate"),
     ]
 
+``choices`` can also be defined as a callable that expects no arguments and
+returns any of the formats described above. For example::
+
+    import functools
+
+
+    @functools.cache
```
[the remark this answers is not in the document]
ngnpope: Saw you'd created ticket-34822 so I whipped up https://github.com/django/django/pull/17256.

docs/ref/models/fields.txt
```diff
@@ -115,9 +115,41 @@ human-readable name. For example::
...
+    def fib(n):
+        if n < 2:
+            return n
+        return fib(n - 1) + fib(n - 2)
```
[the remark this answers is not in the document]
nessita: Applying suggestion (though I think @felixxm wanted to propose a different example).

I chose Fibonacci because it was a nice example to make good use of caching, but now that the caching has "losr focus", perhaps I should propose an example that returns a setting or a similar "static-ish" list?

docs/ref/models/fields.txt
```diff
@@ -115,9 +115,41 @@ human-readable name. For example::
...
+the choices are:
+
+- the result of I/O-bound operations (which could potentially be cached), such
+  as querying a table in the same or an external database, or accessing the
+  choices from a static file.
+
+- a list that is mostly stable but could vary from time to time or from
+  project to project. Examples in this category are using 3rd party apps that
+  provide well-known inventory of values, such as currencies, countries,
+  languages, timezones, etc.
```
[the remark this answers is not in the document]
nessita: Both applied, thanks!

docs/ref/models/fields.txt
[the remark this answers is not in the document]
nessita: Reviewed, thanks!

docs/ref/models/fields.txt
```diff
@@ -115,9 +115,41 @@ human-readable name. For example::
...
+        size = models.IntegerField(choices=get_choices)
+
+Passing a callable for ``choices`` can be particularly handy when, for example,
+the choices are:
+
+- the result of I/O-bound operations (which could potentially be cached), such
+  as querying a table in the same or an external database, or accessing the
+  choices from a static file.
+
+- a list that is mostly stable but could vary from time to time or from
```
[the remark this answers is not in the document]
ngnpope: Ironic that the section I'm updating in https://github.com/django/django/pull/17256 has `-` instead of `*` (but better to stay consistent there) 😂

ngnpope (approved):
🎉

nessita:
@felixxm I think you wanted to evaluate if we could use a better/different example for callable `choices`. I have proposed a different one, no longer using Fibonacci but something more close to a real life case.
I will land following Nick's approval, but if you want me to change the docs, I'm happy to do that in a follow up branch. Thanks!

PR #17216, opened 2023-09-01
Fixed #34778 -- Avoided importing modules in startapp/startproject.

jacobtylerwalls:
ticket-34778

[Forum discussion](https://forum.djangoproject.com/t/check-for-name-clashes-in-startproject-without-importing-modules/23091)

Avoid side effects from importing modules by using `find_spec()`. That method might still import parent modules if given a dotted path, but startproject checks for dotted paths just a bit earlier than this block.

### DIFF
--- a/django/core/management/templates.py
+++ b/django/core/management/templates.py
@@ -5,7 +5,7 @@ import posixpath
 import shutil
 import stat
 import tempfile
-from importlib import import_module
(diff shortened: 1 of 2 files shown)


felixxm (approved):
@jacobtylerwalls Thanks :+1:

PR #17220, opened 2023-09-03
Fixed #34806 -- Made the cached_db backend resilient to cache backend errors.

sulabhkatila:
[ticket_34806](https://code.djangoproject.com/ticket/34806)

I have added a MemoryError handling to ensure that it doesn't crash and logged the issue to let the user know.

### DIFF
--- a/django/contrib/sessions/backends/cached_db.py
+++ b/django/contrib/sessions/backends/cached_db.py
@@ -2,11 +2,15 @@
 Cached, database-backed sessions.
 """
 
+import logging
+
 from django.conf import settings
 from django.contrib.sessions.backends.db import SessionStore as DBStore
 from django.core.cache import caches
 
 KEY_PREFIX = "django.contrib.sessions.cached_db"
+
+logger = logging.getLogger("django.contrib.sessions")
 
 
 class SessionStore(DBStore):
@@ -52,7 +56,10 @@ class SessionStore(DBStore):
 
     def save(self, must_create=False):
         super().save(must_create)
-        self._cache.set(self.cache_key, self._session, self.get_expiry_age())
+        try:
+            self._cache.set(self.cache_key, self._session, self.get_expiry_age())
+        except Exception:
+            logger.exception("Error saving to cache (%s)", self._cache)
 
     def delete(self, session_key=None):
         super().delete(session_key)
--- a/tests/sessions_tests/tests.py
+++ b/tests/sessions_tests/tests.py
@@ -517,6 +517,22 @@ class CacheDBSessionTests(SessionTestsMixin, TestCase):
         with self.assertRaises(InvalidCacheBackendError):
             self.backend()
 
+    @override_settings(
+        CACHES={"default": {"BACKEND": "cache.failing_cache.CacheClass"}}
+    )
+    def test_cache_set_failure_non_fatal(self):
+        """Failing to write to the cache does not raise errors."""
+        session = self.backend()
+        session["key"] = "val"
+
+        with self.assertLogs("django.contrib.sessions", "ERROR") as cm:
+            session.save()
+
+        # A proper ERROR log message was recorded.
+        log = cm.records[-1]
+        self.assertEqual(log.message, f"Error saving to cache ({session._cache})")
+        self.assertEqual(str(log.exc_info[1]), "Faked exception saving to cache")
+
 
 @override_settings(USE_TZ=True)
 class CacheDBSessionWithTimeZoneTests(CacheDBSessionTests):
--- a/docs/ref/logging.txt
+++ b/docs/ref/logging.txt
@@ -286,6 +286,17 @@ Messages to this logger have ``params`` and ``sql`` in their extra context (but
 unlike ``django.db.backends``, not duration). The values have the same meaning
 as explained in :ref:`django-db-logger`.
 
+.. _django-contrib-sessions-logger:
+
+``django.contrib.sessions``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Log messages related to the :doc:`session framework</topics/http/sessions>`.
+
+* Non-fatal errors occurring when using the
+  :class:`django.contrib.sessions.backends.cached_db.SessionStore` engine are
+  logged as ``ERROR`` messages with the corresponding traceback.
+
 Handlers
 --------
 
--- a/docs/releases/5.1.txt
+++ b/docs/releases/5.1.txt
@@ -115,7 +115,10 @@ Minor features
 :mod:`django.contrib.sessions`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-* ...
+* :class:`django.contrib.sessions.backends.cached_db.SessionStore` now handles
+  exceptions when storing session information in the cache, logging proper
+  error messages with their traceback via the newly added
+  :ref:`sessions logger <django-contrib-sessions-logger>`.
 
 :mod:`django.contrib.sitemaps`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/topics/http/sessions.txt
+++ b/docs/topics/http/sessions.txt
@@ -76,9 +76,17 @@ Once your cache is configured, you have to choose between a database-backed
 cache or a non-persistent cache.
 
 The cached database backend (``cached_db``) uses a write-through cache --
-session writes are applied to both the cache and the database. Session reads
-use the cache, or the database if the data has been evicted from the cache. To
-use this backend, set :setting:`SESSION_ENGINE` to
+session writes are applied to both the database and cache, in that order. If
+writing to the cache fails, the exception is handled and logged via the
+:ref:`sessions logger <django-contrib-sessions-logger>`, to avoid failing an
+otherwise successful write operation.
+
+.. versionchanged:: 5.1
+
+    Handling and logging of exceptions when writing to the cache was added.
+
+Session reads use the cache, or the database if the data has been evicted from
+the cache. To use this backend, set :setting:`SESSION_ENGINE` to
 ``"django.contrib.sessions.backends.cached_db"``, and follow the configuration
 instructions for the `using database-backed sessions`_.
 
--- a/tests/cache/failing_cache.py
+++ b/tests/cache/failing_cache.py
@@ -1,0 +1,7 @@
+from django.core.cache.backends.locmem import LocMemCache
+
+
+class CacheClass(LocMemCache):
+
+    def set(self, *args, **kwargs):
+        raise Exception("Faked exception saving to cache")


django/contrib/sessions/backends/cached_db.py
```diff
@@ -51,8 +51,13 @@ def exists(self, session_key):
...
     def save(self, must_create=False):
-        super().save(must_create)
-        self._cache.set(self.cache_key, self._session, self.get_expiry_age())
+        try:
+            super().save(must_create)
+            self._cache.set(self.cache_key, self._session, self.get_expiry_age())
+        except MemoryError as e:
+            import logging
+
+            logging.error(f"Caching error: {e}")
```
shangxiao, line 60: Unfortunately we can't just log willy-nilly, we have to follow the standard process of getting a named logger and logging to that to allow users of the framework to capture those logs.

I'm not sure myself what the appropriate logger is for this module but nessita or felixx will be able to confirm that.

django/contrib/sessions/backends/cached_db.py
```diff
@@ -51,8 +51,13 @@ def exists(self, session_key):
         )
 
     def save(self, must_create=False):
-        super().save(must_create)
-        self._cache.set(self.cache_key, self._session, self.get_expiry_age())
+        try:
+            super().save(must_create)
+            self._cache.set(self.cache_key, self._session, self.get_expiry_age())
+        except MemoryError as e:
```
shangxiao, line 57: @claudep Is this the correct exception to catch?

shangxiao (changes requested):
Hi & thanks for submitting this PR 😊

There's a change required to the way the logger is created. Also we'll need a new test to confirm the behaviour change.

shangxiao:
> There's a change required to the way the logger is created. Also we'll need a new test to confirm the behaviour change.

Oh and there may be some documentation updates as well to let folks know "from x.xx version caching errors will result in logs"… felix & nessita will confirm this though 👍

django/contrib/sessions/backends/cached_db.py
sulabhkatila: Thank you for the feedback and guidance.

Once I have the information from @felixxm or @nessita, I'll make the necessary updates to the code accordingly.

django/contrib/sessions/backends/cached_db.py
claudep: I'm not sure we can target any specific exception type. In my use case, it was a TooBig error generated by pylibmc. It looks like the error is defined in https://github.com/lericson/pylibmc/blob/97ace274cffd8db/src/_pylibmcmodule.h#L201C12-L201C12 and raised from C code. I'm afraid we have no real choice than catching the broader `Exception` base class.

nessita:
I don't have the answers to the logger questions but I'll check with @felixxm in our next catch up, and get back to you (in about a week).

In the meantime, any chance that the lint errors are fixed?

sulabhkatila:
hi @nessita, can you please lead me in the right direction in terms of adding a test case for this?

shangxiao:
@sulabhkatila Nessita's busy, I can help. You've already added a test? 🤔

shangxiao:
@sulabhkatila A good starting logger may be `django.request`. Have a look at the codebase for existing uses, eg `logger = getLogger("django.request")`. Usually loggers at created at module-level and placed at the top after the imports.

sulabhkatila:
> @sulabhkatila Nessita's busy, I can help. You've already added a test? 🤔

I added the test but it actually doesn't really work. I only added that to show where my head was going with the test. So my idea for the [test](https://github.com/django/django/pull/17220/files#diff-0288ae6614e8e11efb1741a04f1771dfe9b5726d4656b51ea888943243b06aeeR522) was to try to recreate the error and get the test case to catch any error that would be raised (Thought it would be the way from [here](https://github.com/django/django/blob/main/tests/cache/tests.py#L1542)).

But it doesn't really work: the test case passes even when we raise the error [here](https://github.com/django/django/pull/17220/files#diff-b27d2c3d2c7ebac7655d3b878c9a1b904fcfde3b82974a6fe5bd77b99871f741R63) instead of passing (how it is done now).

I am now suspicious if I am even catching the exception properly as if [there](https://github.com/django/django/pull/17220/files#diff-b27d2c3d2c7ebac7655d3b878c9a1b904fcfde3b82974a6fe5bd77b99871f741R54) isn't where the exception is raised??

tests/sessions_tests/tests.py
```diff
@@ -519,6 +519,20 @@ def test_non_default_cache(self):
...
+        key = "some key"
+        val = "a" * (max_value_length + 1)
+        try:
+            self.session[key] = val
+            self.session.save()
+        except Exception as e:
+            exceptions.append(e)
+
+        self.assertEqual(len(exceptions), 0)
+
```
shangxiao, line 535: @sulabhkatila I have a failing test you can use that uses a deliberately failing cache

Add this to `tests/session_tests/cache.py`:

```python
from django.core.cache.backends.locmem import LocMemCache


class FailingCache(LocMemCache):
    def set(self, *args, **kwargs):
        raise Exception("Error saving to cache")
```

Then update your test like so:

```suggestion
    @override_settings(
        CACHES={"default": {"BACKEND": "sessions_tests.cache.FailingCache"}}
    )
    def test_session_cache_failure(self):
        self.session = self.backend()
        self.session.save()
```

django/contrib/sessions/backends/cached_db.py
```diff
@@ -57,10 +57,11 @@ def save(self, must_create=False):
...
             raise
-        except Exception:
-            # import logging
-            # log the error
-            pass
+        except Exception as e:
+            import logging
+
+            logger = logging.getLogger("django.request")
+            logger.error("Error saving to cache: %s", e)
```
shangxiao, line 64: Optionally we can do `logger.exception(<message>)` which captures the exception:

https://docs.python.org/3/library/logging.html#logging.Logger.exception

This hasn't been used in Django yet so let's wait to see what Mariusz thinks.

tests/sessions_tests/tests.py
```diff
@@ -519,19 +519,25 @@ def test_non_default_cache(self):
...
+        error_records = [record for record in cm.records if record.levelname == "ERROR"]
+
+        self.assertTrue(error_records)
+        self.assertTrue(
+            error_records[0].getMessage().startswith("Error saving to cache")
+        )
+
+        self.session["key"] = "val"
+        self.session.save()
+        self.assertEqual(self.session["key"], "val")
```
shangxiao, line 540: There's no need for such complexity in this test (ie the list comp, 2 assert trues)... it's better to keep it simple:

```
        self.session = self.backend()
        self.session["key"] = "val"
        with self.assertLogs("django.request", "ERROR") as cm:
            self.session.save()
        self.assertEqual(
            cm.records[0].message, "Error saving to cache: Error saving to cache"
        )
        self.assertEqual(self.session["key"], "val")
```

django/contrib/sessions/backends/cached_db.py
```diff
@@ -57,10 +57,11 @@ def save(self, must_create=False):
...
         except UpdateError:
             raise
-        except Exception:
-            # import logging
-            # log the error
-            pass
+        except Exception as e:
+            import logging
+
+            logger = logging.getLogger("django.request")
```
shangxiao, line 63: Please move the logger to module level underneath imports. This is the typical convention in Python code and is the more common approach in Django (see other getLogger() calls) … the exception tends to be for security loggers.

shangxiao (changes requested):
Getting better 🎉

sulabhkatila:
Hi @shangxiao, i have made the proposed changes and also tried to update the documentation in this commit.

docs/topics/cache.txt
```diff
@@ -256,6 +256,15 @@ Unlike other cache backends, the database cache does not support automatic
...
 
+.. note::
+
+    In the case of storing the session data in the cache for the cached_db backend,
+    it will not make the whole request crash if storing the data in the cache fails.
+    Instead, a log entry will be created in the "django.request" logger, allowing
+    developers to be aware of the issue without causing request failures. This
+    ensures that even if cache backend exceptions occur, the application continues
+    to function smoothly.
+
```
shangxiao, line 267: I'm not sure a doc update is necessary 🤷‍♂️ We can wait to see what nessita thinks.

shangxiao:
> Hi @shangxiao, i have made the proposed changes and also tried to update the documentation in this commit.

Thanks 🏆  The `logger` still isn't at module-level but nessita may be ok with that.

nessita:
> > Hi @shangxiao, i have made the proposed changes and also tried to update the documentation in this commit.
> 
> Thanks 🏆 The `logger` still isn't at module-level but nessita may be ok with that.

Thanks for the ping, looking into this now!

django/contrib/sessions/backends/cached_db.py
nessita: I did talk with Mariusz about this, and he confirmed that:

* we should use a new logger for `django.contrib.sessions`:
```python
import logging

logging.getLogger("django.contrib.sessions")
```
* we should add a release note about the adding of this logger and a documentation entry in `ref/logging`

nessita:
> > There's a change required to the way the logger is created. Also we'll need a new test to confirm the behaviour change.
> 
> Oh and there may be some documentation updates as well to let folks know "from x.xx version caching errors will result in logs"… felix & nessita will confirm this though 👍

Yes, I think we do need a release note stating this and the adding of the new logger.

nessita:
> > Hi @shangxiao, i have made the proposed changes and also tried to update the documentation in this commit.
> 
> Thanks 🏆 The `logger` still isn't at module-level but nessita may be ok with that.

I am not OK with that :-)

Please let's move the `getLogger` code to right after the imports, at the module level. Thanks!

django/contrib/sessions/backends/cached_db.py
nessita: Agreed.

docs/topics/cache.txt
nessita: I'm not sure either, though if we were to add something, I would definitely ask to have that within the `contrib.sessions` docs and not in the `core.cache` docs. Perhaps a minor/simple note in the `The cached database backend (cached_db) uses a write-through cache – session writes are applied to both the cache and the database.`  paragraph (at `/topics/http/sessions/#using-cached-sessions`)

nessita (changes requested):
Added some comments. Also, please note that the `.DS_Store` binary file that was added, needs to be removed, it does not belong in the repo.

Thank you!

sulabhkatila:
Hi @shangxiao, I have made the changes as discussed by moving the logger, update to the docs in the appropriate file, and removed the .DS_Store binary file.

django/contrib/sessions/backends/cached_db.py
nessita: @sulabhkatila Any chance you missed the comment above? I don't see those changes reflected in the current diff.

django/contrib/sessions/backends/cached_db.py
```diff
@@ -51,8 +54,14 @@ def exists(self, session_key):
         )
 
     def save(self, must_create=False):
-        super().save(must_create)
-        self._cache.set(self.cache_key, self._session, self.get_expiry_age())
+        try:
+            super().save(must_create)
```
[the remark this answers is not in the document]
nessita: @sulabhkatila  This comment is also unaddressed.

nessita (changes requested):
As mentioned in one of the comments, a new logger needs to be defined and documented, and also the call to `save()` inside the try-except block should be moved outside that block.

Thank you for the work so far!

nessita (changes requested):
Hello @sulabhkatila, please read my last comment in full, there are still multiple things pending to implement, such as documenting the new logger and moving the `save()` call to outside the try-except block.

The documenting of the new logger needs to occur in multiple doc pages, I listed those in previous comments I think.

Please only when all those are ready, mark this PR as ready for review in the issue tracker as documented in https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/submitting-patches/#patch-review-checklist

sulabhkatila:
Hi @nessita , I apologize for the oversight.

Regarding the docs, I just added the release note about the behavior in ```docs/releases/5.0.txt```.
Now, the docs related updates have been done on the following files: ```docs/releases/5.0.txt``` and ```docs/topics/cache.txt```. 
Is there any other pages in the docs that I might need to update? 
Regarding the save call, initially I had put ```super().save(must_create())``` call inside the ```try``` block, and after your feedback I figured I had to put that ```save``` call outside and I made the change. And now it looks as such:

```
 def save(self, must_create=False):
        super().save(must_create)
        try:
            self._cache.set(self.cache_key, self._session, self.get_expiry_age())
        except UpdateError:
            raise
        except Exception as e:
            logger.error("Error saving to cache: %s", e)
```

Is there some other ```save()``` calls that I were supposed to touch. Can you please provide more feedback on it.

docs/releases/5.0.txt
```diff
@@ -265,7 +265,8 @@ Minor features
 :mod:`django.contrib.sessions`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-* ...
+* An error message will be logged to the ``django.contrib.sessions`` logger
+  in case a cache storage failure for ``cahced_db``.
```
nessita, line 269: Given that Django 5.0 was already released on December 4th, this release note should be moved to the 5.1.txt release notes, and it should be written in present tense (see the other notes as reference).
Also, there is a typo here:
```suggestion
  in case a cache storage failure for ``cached_db``.
```

nessita (changes requested):
Thank you @sulabhkatila for the updates.

Regarding your question:
> Is there any other pages in the docs that I might need to update? 

Yes, the logging docs also need updating (see docs/ref/logging.txt). Could you also please rebase onto `main` so you incorporate the latest changes from the `main` branch?

nessita:
Hey @sulabhkatila, thank you for your continued work. If this branch is ready for a re-review, please double check that the ticket flags are properly set or unset. [This are the docs describing those details](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/submitting-patches/#patch-review-checklist).

django/contrib/sessions/backends/cached_db.py
```diff
@@ -52,7 +57,12 @@ def exists(self, session_key):
 
     def save(self, must_create=False):
         super().save(must_create)
-        self._cache.set(self.cache_key, self._session, self.get_expiry_age())
+        try:
+            self._cache.set(self.cache_key, self._session, self.get_expiry_age())
+        except UpdateError:
```
nessita, line 62: @shangxiao @sulabhkatila I'm not being able to understand why this `except UpdateError` is needed. I have commented out amd no test failing. I'm also trying to think of a proper test to cover for this code path, but I think it's unreachable: `UpdateError` could be raised when calling `super()` but not from any?/the cache backend.

What's your view on this?

tests/sessions_tests/tests.py
nessita: @shangxiao What's your view on defining this "failingcache" inside the `cache` tests, next to `closeable_cache.py` and `liberal_backend.py`? There's something that unsettles me about placing this file in this folder, but I can't quite pinpoint what it is that bothers me exactly.

shangxiao: @nessita I have no particular view ☺️ Seems like a good place to put it 👍

django/contrib/sessions/backends/cached_db.py
nessita: Thank you! I will push some cleanups and docs updates including the removal of this except block.

tests/sessions_tests/tests.py
nessita: Amazing, thank you! Will do that.

django/contrib/sessions/backends/cached_db.py
sulabhkatila: Hi, it appears I'd added the ```except UpdateError``` without proper consideration at the time. I tested the code without the block, and, indeed, no tests are failing. I apologize for any confusion it may have caused. I'll make sure to be mindful of such changes in the future.

sulabhkatila: > Thank you! I will push some cleanups and docs updates including the removal of this except block.

Thank you, nessita.

nessita:
@claudep Hi there! I pushed some code and docs tweaks, I think it's ready to be merged. Would you fancy doing a final review to this PR? (no worries if you can't!)

nessita (approved):

claudep (approved):
Thanks a lot for working on this! We'll see in the future if other backend operations need some similar safeguards.

PR #17221, opened 2023-09-03
Fixed #34812 -- Raised ImproperlyConfigured if urlpatterns is missing.

TheEpic-dev:
Hello everyone.

As described in [the tracker](https://code.djangoproject.com/ticket/34812), Django users may get slightly confusing errors about circular imports when they try to `include()` URLConf modules that lack the `urlpatterns` attribute.

This PR modifies the `django.urls.include()` function to check if the `arg` argument is a module without a `urlpatterns` attribute, and raises `ImproperlyConfigured` if that is the case.


shangxiao:
Closing per ticket

PR #17223, opened 2023-09-04
Refs #31262 -- Renamed ChoiceIterator to BaseChoiceIterator.

ngnpope:
Some third-party applications, e.g. `django-filter`, already define their own `ChoiceIterator`, so renaming this `BaseChoiceIterator` will be a better fit and avoid any potential confusion. _(It also makes more sense because this doesn't provide any implementation, but is a helper base class for things to extend when they want to avoid premature normalization.)_

See https://github.com/carltongibson/django-filter/pull/1607.


felixxm (approved):
@ngnpope Thanks :+1:

PR #17226, opened 2023-09-04
Fixed #34810 -- Measured test coverage on django-admin commands.

jacobtylerwalls:
Improve code coverage calculation for django-admin commands.

ticket-34810

#### Demo ####
```sh
coverage run ./runtests.py
coverage combine
coverage html
```
Before, this was 12%.
templates-coverage

#### Resources ####
[1] https://coverage.readthedocs.io/en/stable/subprocess.html
Configuration for measuring test coverage on subprocesses: suggests using sitecustomize.py and being defensive in case coverage is not installed. Any folder on the python path can be used, and `tests` meets that criterion.
[2] https://github.com/nedbat/coveragepy/issues/1093#issuecomment-769480418
This coverage.py issue thread describes how to cope with subprocesses run from different working directories: the trick suggested by Ned Batchelder is to use environment variable-substitution in the coveragerc file.

### DIFF
--- a/tests/.coveragerc
+++ b/tests/.coveragerc
@@ -1,7 +1,7 @@
 [run]
 branch = True
 concurrency = multiprocessing,thread
-data_file = .coverages/.coverage
+data_file = ${RUNTESTS_DIR-.}/.coverages/.coverage
 omit =
     */django/utils/autoreload.py
 source = django
--- a/tests/runtests.py
+++ b/tests/runtests.py
@@ -71,6 +71,10 @@ tempfile.tempdir = os.environ["TMPDIR"] = TMPDIR
 # Removing the temporary TMPDIR.
 atexit.register(shutil.rmtree, TMPDIR)
 
+# Add variables enabling coverage to trace code in subprocesses.
+os.environ["RUNTESTS_DIR"] = RUNTESTS_DIR
+os.environ["COVERAGE_PROCESS_START"] = os.path.join(RUNTESTS_DIR, ".coveragerc")
+
 
 # This is a dict mapping RUNTESTS_DIR subdirectory to subdirectories of that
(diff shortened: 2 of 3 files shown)


tests/.coveragerc
```diff
@@ -1,7 +1,7 @@
 [run]
 branch = True
 concurrency = multiprocessing,thread
-data_file = .coverages/.coverage
+data_file = ${RUNTESTS_DIR-.}/.coverages/.coverage
```
nessita, line 4: From the coverage docs you linked, it seems that we also need to set the `parallel` option to True?

*When using this technique, be sure to set the parallel option to true so that multiple coverage.py runs will each write their data to a distinct file.*

nessita (commented):
Overall looks good, will try it IRL now.

nessita:
It's impressive how coverage has grown for some management commands:
Before:
image

After:
image

tests/.coveragerc
jacobtylerwalls: Thanks, I didn't spot that the first time round.

Looking closer now, I'm noticing that as things stand we already have the wanted behavior described at the [docs for parallel option](https://coverage.readthedocs.io/en/7.3.1/config.html#run-parallel): producing under `tests/.coverages` files stamped with the computer name and a random number.

And if I remove `concurrency = multiprocessing,thread`, I just get a single file in that dir called ".coverage", so I think that's the option that's making this work as intended.

I see that the `concurrency` option was added in coverage 4.0, so the admonition about parallel=True may not have kept up with the times.

nessita: Great, thanks for checking. I rebased onto latest `main` and I'll merge once the tests pass. Thanks!

PR #17231, opened 2023-09-07
Update tools.txt

ebram96:


adamchainz (commented):
This change isn't necessary, "isn't much use" is correct grammar.

PR #17232, opened 2023-09-07
docs: amend how F expression works explanation

cikay:


ngnpope:
I think that the existing wording is fine. Changing it to "uses and manipulates" leaves open the implication that it may have been provided to Python, even if Python never did anything with it... Which would be incorrect.

PR #17233, opened 2023-09-07
Added oguzhanakan0 to contributors

oguzhanakan0:
Added myself as a contributor.

Will submit a patch for https://code.djangoproject.com/ticket/34816 if I could be added to the list. 

Thanks in advance!


charettes:
@oguzhanakan0 it usually work the other way around; you submit a patch that includes you in authors. We don't preemptively add contributors to this list before they land changes in the `main` branch.

oguzhanakan0:
Submitted it with the patch in https://github.com/django/django/pull/17234

PR #17238, opened 2023-09-09
Fixed #34824 -- Prevented unnecessary AlterField when ForeignObject.from_fields/to_fields is not a tuple.

RelaxedDong:
Change the from_fields and to_fields of ForeignObject’s existing migration records to tuple types


felixxm (commented):
@RelaxedDong Thanks :+1: I simplified tests.

PR #17241, opened 2023-09-10
Fixed #33647 -- bulk_update silently truncating values for size limited fields

priyank-panchal:
https://code.djangoproject.com/ticket/33647

### DIFF
--- a/django/db/backends/postgresql/features.py
+++ b/django/db/backends/postgresql/features.py
@@ -59,7 +59,7 @@ class DatabaseFeatures(BaseDatabaseFeatures):
             PRIMARY KEY(column_1, column_2)
         )
     """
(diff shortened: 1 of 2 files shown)


shangxiao:
Hi Priyank,

Patches are always appreciated but you must be sure to read through the ticket before submitting them (the entire ticket with all the comments 😉).

[This comment](https://code.djangoproject.com/ticket/33647#comment:3) suggests that the flag was set to `True` for a reason and we can't simply turn it off.

If you want some help with exploring this ticket getting some assistance on the forum or Discord is a good start: https://www.djangoproject.com/community/

priyank-panchal:
Thank you for your suggestion; I will take it into consideration in the future.

PR #17253, opened 2023-09-13
Fixed #34808 -- Documented aggregate default argument in 4.2 docs

lufafajoshua:
Documented the aggregate default argument used to provide a default value to be returned when aggregate functions are used with an empty query-set. Feedback is welcome.


docs/topics/db/aggregation.txt
```diff
@@ -632,3 +632,46 @@ aggregate that author count, referencing the annotation field:
     >>> from django.db.models import Avg, Count
     >>> Book.objects.annotate(num_authors=Count("authors")).aggregate(Avg("num_authors"))
     {'num_authors__avg': 1.66}
+
+Handling empty querysets using aggregate ``default`` 
+----------------------------------------------------
+
+.. versionadded:: 4.0
+
+Aggregation functions return ``None`` when used with an empty ``QuerySet``.
```
shangxiao, line 641: Not just empty querysets but empty groups when grouping 👍  (see my comments in the ticket)

Not just for this sentence  but the paragraph need to keep this in mind.

docs/topics/db/aggregation.txt
```diff
@@ -632,3 +632,46 @@ aggregate that author count, referencing the annotation field:
...
+The aggregate ``default`` is an optional argument that allows specifying a 
+value to use as a ``default`` value when the ``QuerySet`` has no entries,
```
shangxiao, line 643:
```suggestion
value to use as a default value when the ``QuerySet`` has no entries, 
```
Seems like "default value" is referring to the value not the arg 🤔

docs/topics/db/aggregation.txt
```diff
@@ -632,3 +632,46 @@ aggregate that author count, referencing the annotation field:
...
+
+.. code-block:: pycon
+
+    >>> from django.db.models import Sum
+    >>> Book.objects.filter(name__contains="Web").aggregate(Sum('price'))
+    {'price__sum': None}
+
+However, you can pass ``0`` as a ``default`` to be returned if no books can be found.
+
+.. code-block:: pycon
```
shangxiao, line 662: Is this meant to be `python`?

docs/topics/db/aggregation.txt
```diff
@@ -632,3 +632,46 @@ aggregate that author count, referencing the annotation field:
     >>> from django.db.models import Avg, Count
     >>> Book.objects.annotate(num_authors=Count("authors")).aggregate(Avg("num_authors"))
     {'num_authors__avg': 1.66}
+
+Handling empty querysets using aggregate ``default`` 
+----------------------------------------------------
```
shangxiao, line 637: It might be worth explaining that aggregate functions are simply returning what the database is returning. (I'm pretty sure that is the case someone may correct me.)  This pretty much explains all of the behaviour including counts returning 0.

docs/topics/db/aggregation.txt
```diff
@@ -632,3 +632,46 @@ aggregate that author count, referencing the annotation field:
...
+
+However, you can pass ``0`` as a ``default`` to be returned if no books can be found.
+
+.. code-block:: pycon
+
+    >>> from django.db.models import Sum
+    >>> Book.objects.filter(name__contains="Web").aggregate(Sum('price', default=0))
+    {'price__sum': Decimal('0')}
+
+.. admonition:: The aggregate ``default`` argument uses :class:`~django.db.models.functions.Coalesce` under the hood
```
shangxiao, line 668: Just to be clear admonitions are for important notes or warnings where behaviour deviates from the norm.

Since this will be in the current docs for a version of Django that supports `default` we won't need to warn users to use Coalesce instead.

It may be better to just move this up to the top as a helpful note so people understand the query it's performing. (We don't need the coalesce example, that can go).

shangxiao (changes requested):
Thanks for submitting this Lufafa ❤️

There may be a few more passes over the documentation to get it right but for now just submitted feedback on content & structure.

Then we should look at sentence flow & comprehension once all that is sorted 😊

docs/topics/db/aggregation.txt
```diff
@@ -632,3 +632,46 @@ aggregate that author count, referencing the annotation field:
...
     >>> Book.objects.annotate(num_authors=Count("authors")).aggregate(Avg("num_authors"))
     {'num_authors__avg': 1.66}
+
+Handling empty querysets using aggregate ``default`` 
+----------------------------------------------------
+
+.. versionadded:: 4.0
+
+Aggregation functions return ``None`` when used with an empty ``QuerySet``.
+The aggregate ``default`` is an optional argument that allows specifying a
```
ngnpope, line 642:
```suggestion
The optional ``default`` argument for aggregate functions allows specifying a 
```

docs/topics/db/aggregation.txt
ngnpope: Nope. This `pycon` is "python console" and these are examples with the `>>>` prompt.

docs/topics/db/aggregation.txt
```diff
@@ -632,3 +632,46 @@ aggregate that author count, referencing the annotation field:
...
+
+    >>> from django.db.models import Sum
+    >>> Book.objects.filter(name__contains="Web").aggregate(Sum('price'))
+    {'price__sum': None}
+
+However, you can pass ``0`` as a ``default`` to be returned if no books can be found.
+
+.. code-block:: pycon
+
+    >>> from django.db.models import Sum
```
ngnpope, line 664: We don't need to import this again here - we already did so above.

ngnpope (changes requested):

docs/topics/db/aggregation.txt
lufafajoshua: Thanks a lot for the comments friends. I am working on the necessary adjustments.

lufafajoshua:
Hey @shangxiao  and @ngnpope , How about that?

ebaumgartner:
Hi, thanks for doing this.

I don't have anything to add about the content of this PR.

I am a little concerned about the scope of this change compared to the original issue. Is this PR intended to completely fix #34808, or is it one part of the ultimate solution?

The bits I don't see here that were mentioned in #34808 include:

- Adding `default` params to the Cheat Sheet examples.
- Adding some indication within the documentation for Sum, Avg, etc on the querysets page that mention that None is a possible return value and that default can be used to avoid having to handle None.

Both of these bits are about putting information about `default` in places where users who don't yet realize they need to worry about empty querysets will notice it.

docs/ref/models/querysets.txt
```diff
@@ -3903,7 +3903,9 @@ by the aggregate.
 
     * Default alias: ``<field>__avg``
     * Return type: ``float`` if input is ``int``, otherwise same as input
-      field, or ``output_field`` if supplied
+      field, or ``output_field`` if supplied. 
+    * Returns ``None`` for empty querysets or groups.
+
```
ngnpope, line 3908: This isn't quite right - it doesn't return `None`, it returns the value provided in `default`. Also, the "return type" isn't quite correct as it potentially could be `None`.

Perhaps something like the following would be better?

```suggestion
      field, or ``output_field`` if supplied. If the queryset or grouping is
      empty, ``default`` is returned, which is ``None`` by default
```

docs/ref/models/querysets.txt
```diff
@@ -3941,7 +3943,8 @@ by the aggregate.
     Returns the maximum value of the given expression.
 
     * Default alias: ``<field>__max``
-    * Return type: same as input field, or ``output_field`` if supplied
+    * Return type: same as input field, or ``output_field`` if supplied. 
+    * Returns ``None`` for empty querysets or groups.
```
ngnpope, line 3947: Ditto:

```suggestion
    * Return type: same as input field, or ``output_field`` if supplied. If the
      queryset or grouping is empty the value of ``default`` is returned, which
      is ``None`` by default
```

docs/ref/models/querysets.txt
```diff
@@ -3951,7 +3954,8 @@ by the aggregate.
     Returns the minimum value of the given expression.
 
     * Default alias: ``<field>__min``
-    * Return type: same as input field, or ``output_field`` if supplied
+    * Return type: same as input field, or ``output_field`` if supplied. 
+    * Returns ``None`` for empty querysets or groups.
```
ngnpope, line 3958: Ditto:
```suggestion
    * Return type: same as input field, or ``output_field`` if supplied. If the
      queryset or grouping is empty the value of ``default`` is returned, which
      is ``None`` by default
```

docs/ref/models/querysets.txt
```diff
@@ -3962,7 +3966,8 @@ by the aggregate.
 
     * Default alias: ``<field>__stddev``
     * Return type: ``float`` if input is ``int``, otherwise same as input
-      field, or ``output_field`` if supplied
+      field, or ``output_field`` if supplied. 
+    * Returns ``None`` for empty querysets or groups.
```
ngnpope, line 3970: Ditto:
```suggestion
      field, or ``output_field`` if supplied. If the queryset or grouping is
      empty the value of ``default`` is returned, which is ``None`` by default
```

docs/ref/models/querysets.txt
```diff
@@ -3978,7 +3983,8 @@ by the aggregate.
     Computes the sum of all values of the given expression.
 
     * Default alias: ``<field>__sum``
-    * Return type: same as input field, or ``output_field`` if supplied
+    * Return type: same as input field, or ``output_field`` if supplied. 
+    * Returns ``None`` for empty querysets or groups.
```
ngnpope, line 3987: Ditto:
```suggestion
    * Return type: same as input field, or ``output_field`` if supplied. If the
      queryset or grouping is empty the value of ``default`` is returned, which
      is ``None`` by default
```

docs/ref/models/querysets.txt
```diff
@@ -3995,7 +4001,8 @@ by the aggregate.
 
     * Default alias: ``<field>__variance``
     * Return type: ``float`` if input is ``int``, otherwise same as input
-      field, or ``output_field`` if supplied
+      field, or ``output_field`` if supplied. 
+    * Returns ``None`` for empty querysets or groups.
```
ngnpope, line 4005: Ditto:
```suggestion
      field, or ``output_field`` if supplied. If the queryset or grouping is
      empty the value of ``default`` is returned, which is ``None`` by default
```

docs/topics/db/aggregation.txt
```diff
@@ -632,3 +637,34 @@ aggregate that author count, referencing the annotation field:
...
+    {"price__sum": None}
+
+However, you can pass ``0`` as a ``default`` to be returned if no books can be found.
+
+.. code-block:: pycon
+
+    >>> Book.objects.filter(name__contains="Web").aggregate(Sum("price", default=0))
+    {"price__sum": Decimal("0")}
+
+You should note that the ``default`` argument uses :class:`~django.db.models.functions.Coalesce` under the hood.
```
ngnpope, line 670:
```suggestion
The ``default`` argument is implemented under the hood by wrapping the
aggregate function with :class:`~django.db.models.functions.Coalesce`.
```

docs/topics/db/aggregation.txt
```diff
@@ -632,3 +637,34 @@ aggregate that author count, referencing the annotation field:
     >>> from django.db.models import Avg, Count
     >>> Book.objects.annotate(num_authors=Count("authors")).aggregate(Avg("num_authors"))
     {'num_authors__avg': 1.66}
+
+Handling empty querysets using aggregate ``default`` 
+----------------------------------------------------
+
+When an aggregate function is used with an empty ``QuerySet`` or grouping,
+None is returned as a default, this is because aggregate functions result
```
ngnpope, line 645:
```suggestion
``None`` is returned as a default. This is because aggregate functions result 
```

docs/topics/db/aggregation.txt
```diff
@@ -632,3 +637,34 @@ aggregate that author count, referencing the annotation field:
     >>> from django.db.models import Avg, Count
     >>> Book.objects.annotate(num_authors=Count("authors")).aggregate(Avg("num_authors"))
     {'num_authors__avg': 1.66}
+
+Handling empty querysets using aggregate ``default`` 
+----------------------------------------------------
+
+When an aggregate function is used with an empty ``QuerySet`` or grouping,
+None is returned as a default, this is because aggregate functions result 
+in NULL when used with an empty result set. The optional ``default`` argument
```
ngnpope, line 646:
```suggestion
in ``NULL`` when used with an empty result set. The optional ``default`` argument 
```

docs/topics/db/aggregation.txt
```diff
@@ -632,3 +637,34 @@ aggregate that author count, referencing the annotation field:
...
     {'num_authors__avg': 1.66}
+
+Handling empty querysets using aggregate ``default`` 
+----------------------------------------------------
+
+When an aggregate function is used with an empty ``QuerySet`` or grouping,
+None is returned as a default, this is because aggregate functions result 
+in NULL when used with an empty result set. The optional ``default`` argument 
+for aggregate functions allows specifying a value to use as a default when the 
+executed query returns no rows to perform an aggregation. Some aggregate functions
```
ngnpope, line 648: "...query returns no rows to perform an aggregation."

This ending doesn't make sense. Please rephrase this sentence.

docs/topics/db/aggregation.txt
```diff
@@ -632,3 +637,34 @@ aggregate that author count, referencing the annotation field:
...
+
+When an aggregate function is used with an empty ``QuerySet`` or grouping,
+None is returned as a default, this is because aggregate functions result 
+in NULL when used with an empty result set. The optional ``default`` argument 
+for aggregate functions allows specifying a value to use as a default when the 
+executed query returns no rows to perform an aggregation. Some aggregate functions 
+do not support the ``default`` argument because they already handle this, e.g. 
+``Count`` returns ``0``.
+
+For example, say you wanted to get the total price of all books containing
```
ngnpope, line 652:
```suggestion
For example, to get the total price of all books containing 
```

docs/topics/db/aggregation.txt
```diff
@@ -632,3 +637,34 @@ aggregate that author count, referencing the annotation field:
...
+
+For example, say you wanted to get the total price of all books containing 
+"web" in name. Without the ``default`` argument, if there are no books 
+returned by the query, and there are no rows to compute the aggregate 
+``Sum``, None is returned as the computed aggregate.
+
+.. code-block:: pycon
+
+    >>> from django.db.models import Sum
+    >>> Book.objects.filter(name__contains="Web").aggregate(Sum("price"))
```
ngnpope, line 660: In the paragraph above the example you have `"web"` and in the example you have `"Web"` (uppercased first letter).

```suggestion
    >>> Book.objects.filter(name__contains="web").aggregate(Sum("price"))
```

docs/topics/db/aggregation.txt
```diff
@@ -632,3 +637,34 @@ aggregate that author count, referencing the annotation field:
...
+in NULL when used with an empty result set. The optional ``default`` argument 
+for aggregate functions allows specifying a value to use as a default when the 
+executed query returns no rows to perform an aggregation. Some aggregate functions 
+do not support the ``default`` argument because they already handle this, e.g. 
+``Count`` returns ``0``.
+
+For example, say you wanted to get the total price of all books containing 
+"web" in name. Without the ``default`` argument, if there are no books 
+returned by the query, and there are no rows to compute the aggregate 
+``Sum``, None is returned as the computed aggregate.
```
ngnpope, line 655:
```suggestion
``Sum``, ``None`` is returned as the computed aggregate.
```

ngnpope: Also "...is returned as the computed aggregate." isn't correct - no aggregate is computed at all by the database and it returns ``NULL`` (translated to ``None`` in Python). Please rephrase.

ngnpope (changes requested):
Please ensure that documentation is wrapped at 79 characters.

PR #17256, opened 2023-09-13
Fixed #34822 -- Added support for serializing functions decorated with functools.lru_cache in migrations.

ngnpope:
ticket-34822

### DIFF
--- a/docs/releases/5.0.txt
+++ b/docs/releases/5.0.txt
@@ -401,7 +401,9 @@ Management Commands
 Migrations
 ~~~~~~~~~~
 
-* ...
+* Serialization of functions decorated with :func:`functools.cache` or
+  :func:`functools.lru_cache` is now supported without the need to write a
+  custom serializer.
 
 Models
 ~~~~~~
--- a/docs/topics/migrations.txt
+++ b/docs/topics/migrations.txt
@@ -783,7 +783,12 @@ Django can serialize the following:
 - ``LazyObject`` instances which wrap a serializable value.
 - Enumeration types (e.g. ``TextChoices`` or ``IntegerChoices``) instances.
 - Any Django field
-- Any function or method reference (e.g. ``datetime.datetime.today``) (must be in module's top-level scope)
+- Any function or method reference (e.g. ``datetime.datetime.today``) (must be
+  in module's top-level scope)
+
+  - Functions may be decorated if wrapped properly, i.e. using
+    :func:`functools.wraps`
+
 - Unbound methods used from within the class body
 - Any class reference (must be in module's top-level scope)
 - Anything with a custom ``deconstruct()`` method (:ref:`see below <custom-deconstruct-method>`)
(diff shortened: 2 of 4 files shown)


docs/topics/migrations.txt
```diff
@@ -784,6 +784,8 @@ Django can serialize the following:
 - Enumeration types (e.g. ``TextChoices`` or ``IntegerChoices``) instances.
 - Any Django field
 - Any function or method reference (e.g. ``datetime.datetime.today``) (must be in module's top-level scope)
+  - Functions may be decorated if wrapped properly, i.e. using :func:`functools.wraps`
+  - The :func:`functools.cache` and :func:`functools.lru_cache` decorators are also supported
```
nessita, line 788: I'm not sure about this suggestion but I feel it would be a good idea to make it explicit that this support is not a "natural consequence" or implicit:
```suggestion
  - The :func:`functools.cache` and :func:`functools.lru_cache` decorators are explicitly supported
```

nessita (commented):
Looks great, thanks for working on this so quickly!

docs/releases/5.0.txt
```diff
@@ -401,7 +401,8 @@ Management Commands
 Migrations
 ~~~~~~~~~~
 
-* ...
+* Migrations now support serialization of functions decorated with
```
nessita, line 404: On a second thought, I think it would be a good idea to remove "Migrations now" since the section title is specific about Migrations. So:

```suggestion
* Support for serialization of functions decorated with... was added.
```

ngnpope: I copied from the 4.2 release notes where `EnumFlag` was added. Based on that would you still prefer the change?

docs/topics/migrations.txt
nessita: Also this nested items seem to be rendering incorrectly:
image

docs/releases/5.0.txt
ngnpope: I made the suggested change anyway.

docs/topics/migrations.txt
ngnpope: This should be fixed now. _(Forgot the blank lines around nested lists for Restructured Text 🤦🏻)_

nessita (approved):
Amazing, thank you! I will wrap docs to 79cols, rebase and merge afterwards.

docs/releases/5.0.txt
```diff
@@ -401,7 +401,8 @@ Management Commands
 Migrations
 ~~~~~~~~~~
 
-* ...
+* Support for serialization of functions decorated with :func:`functools.cache`
```
felixxm, line 404:
```
`* Migrations now support serialization of functions ...
```

ngnpope: That's what I had until https://github.com/django/django/pull/17256#discussion_r1324732501... I'll let you two fight it out 😂

nessita: hahahaha I won't fight, though I prefer not to duplicate the `Migrations` bit :-)

nessita: Ok, updating

PR #17257, opened 2023-09-13
Ticket 32263

DevilsAutumn:
[Ticket 32263](https://code.djangoproject.com/ticket/32263)

### DIFF
--- a/django/db/migrations/operations/base.py
+++ b/django/db/migrations/operations/base.py
@@ -126,6 +126,9 @@ class Operation:
 
         return router.allow_migrate_model(connection_alias, model)
 
+    def reduce_related(self, operation, app_label):
+        return None
+
     def reduce(self, operation, app_label):
         """
         Return either a list of operations the actual operation should be
--- a/django/db/migrations/operations/models.py
+++ b/django/db/migrations/operations/models.py
@@ -134,41 +134,41 @@ class CreateModel(ModelOperation):
         return False
 
     def reduce_related(self, operation, app_label):
-        if not isinstance(operation, RenameModel):
-            return None
+        if isinstance(operation, RenameModel):
+            impacted_fields = [
+                (_, field)
+                for _, field in self.fields
+                if field.remote_field
+                and field.remote_field.model
+                == f"{app_label}.{operation.old_name_lower}"
+            ]
+            if len(impacted_fields) == 0:
+                return [self]
 
-        impacted_fields = [
-            (_, field)
-            for _, field in self.fields
-            if field.remote_field
-            and field.remote_field.model == f"{app_label}.{operation.old_name_lower}"
-        ]
-        if len(impacted_fields) == 0:
-            return [self]
+            not_impacted_fields = [
+                (_, field)
+                for (_, field) in self.fields
+                if (_, field) not in impacted_fields
+            ]
 
-        not_impacted_fields = [
-            (_, field)
-            for (_, field) in self.fields
-            if (_, field) not in impacted_fields
-        ]
+            fixed_fields = []
 
-        fixed_fields = []
+            for _, impacted_field in impacted_fields:
+                name, path, args, kwargs = impacted_field.deconstruct()
+                kwargs["to"] = f"{app_label}.{operation.new_name_lower}"
+                impacted_field = impacted_field.__class__(*args, **kwargs)
+                fixed_fields.append((_, impacted_field))
 
-        for _, impacted_field in impacted_fields:
-            name, path, args, kwargs = impacted_field.deconstruct()
-            kwargs["to"] = f"{app_label}.{operation.new_name_lower}"
-            impacted_field = impacted_field.__class__(*args, **kwargs)
-            fixed_fields.append((_, impacted_field))
-
-        return [
-            CreateModel(
-                name=self.name,
-                fields=not_impacted_fields + fixed_fields,
-                options=self.options,
-                bases=self.bases,
-                managers=self.managers,
-            ),
-        ]
+            return [
+                CreateModel(
+                    name=self.name,
+                    fields=not_impacted_fields + fixed_fields,
+                    options=self.options,
+                    bases=self.bases,
+                    managers=self.managers,
+                ),
+            ]
+        return super().reduce_related(operation, app_label)
 
     def reduce(self, operation, app_label):
         if (
--- a/django/db/migrations/operations/models.py
+++ b/django/db/migrations/operations/models.py
@@ -133,6 +133,43 @@ class CreateModel(ModelOperation):
                 return True
         return False
 
+    def reduce_related(self, operation, app_label):
+        if not isinstance(operation, RenameModel):
+            return None
+
+        impacted_fields = [
+            (_, field)
+            for _, field in self.fields
+            if field.remote_field
+            and field.remote_field.model == f"{app_label}.{operation.old_name_lower}"
+        ]
+        if len(impacted_fields) == 0:
+            return [self]
+
+        not_impacted_fields = [
+            (_, field)
+            for (_, field) in self.fields
+            if (_, field) not in impacted_fields
+        ]
+
+        fixed_fields = []
+
+        for _, impacted_field in impacted_fields:
+            name, path, args, kwargs = impacted_field.deconstruct()
+            kwargs["to"] = f"{app_label}.{operation.new_name_lower}"
+            impacted_field = impacted_field.__class__(*args, **kwargs)
+            fixed_fields.append((_, impacted_field))
+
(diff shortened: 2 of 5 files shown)


charettes:
@DevilsAutumn, thanks for the patch.

Per [this comment](https://code.djangoproject.com/ticket/32263#comment:7), did you explore ticket-32256 7c18b22e2fa70aa8dcfadb33beb17933abdf7ee4 solution but adapted to `RenameModel` instead of `RenameField`?

It seems that the main issue here is the lack of optimization barrier in `RenameModel.reduce` wrt/ to `self.old_name`?

The large number of changes this requires makes me suspicious that a lot of it might not be necessary here. I think this ticket should focus on making sure that `squashmigration` generate valid output first; having it generate the optimal operations is a different can of worms.

DevilsAutumn:
Pushed changes to handle cases mentioned in ticket only. We can address other cases later.

David-Wobrock:
> @charettes Thanks for taking a look. Yes , I tried similar solution in [7c18b22](https://github.com/django/django/commit/7c18b22e2fa70aa8dcfadb33beb17933abdf7ee4) with `RenameModel` but unfortunately it didn't work. `RenameModel.reduce` is never called as it gets already reduced by `CreateModel` .

Precisely.
Adding this optimization barrier on the `CreateModel.reduce` works also, but it would probably change a few other cases - perhaps some where it's not required.
It would be interesting to see which cases affected by this, and see if we could fine-tune this.

InvalidInterrupt:
(It's been too long since I've looked at the migration optimizer code; please be patient if I'm mistaken about things.)

I think that the minimal change for this might be to add a barrier in `.reduce()` on `CreateModel`, `AddField`, and `AlterField` (i.e. all operations where a relationship may be added). Optimization only needs to be prevented if the op `.reduce()` is called on creates a field referring to the model being renamed. From there it's only a step further to allow the reference creation op to slide past the `RenameModel` while replacing itself with a new op referencing the new model name.

If I remember well, back when I ran out of time to consider this myself originally I was wondering how this all would/should work with migrations _outside_ the app `squashmigrations` is run on. Is manual fixup currently the expected outcome when `squashmigrations` is run on an app, if another app's migration referenced those models somewhere in the middle of their history? Forgive me for asking if that actually works fine (when you don't run into bugs like this one). It's been years since I've done this, and I can't remember too well.

DevilsAutumn:
I think adding just an optimization barrier won't work, i don't see any code which updates the fields of a model referencing other model while reducing operations. It would have been different if there were `AlteField` or `AddField` operations for the referencing field but its just one `CreateModel` whose field is referencing another `CreateModel`. Or maybe i am missing something?

InvalidInterrupt:
I finally made time to test the minimal approach I had talked about; it produces correct output by only modifying a few operations' `.reduce()` methods, with no changes to the core optimizer logic. If you would like to reference it, it's in https://github.com/InvalidInterrupt/django/tree/ticket_32263.

django/db/migrations/operations/base.py
```diff
@@ -146,6 +146,9 @@ def allow_migrate_model(self, connection_alias, model):
 
         return router.allow_migrate_model(connection_alias, model)
 
+    def reduce_related(self, operation, app_label):
+        return None
```
sarahboyce, line 150:
```suggestion
        return [self]
```
Then we don't have to do `op.reduce_related(other, app_label) or [op]` and can do `op.reduce_related(other, app_label)` instead
We might want a doc string here also

sarahboyce (commented):
Thank you for this @DevilsAutumn, I have some minor suggestions

@InvalidInterrupt I had a look into your commit, thank you for sharing
Note that based off the tests written there, it doesn't reduce the migration as much as I was expecting 🤔

sarahboyce (commented):
Thank you for the updates @DevilsAutumn :star: I pushed minor tweaks

To me, this looks good but most of the commentary in the PR appears to be that folks don't feel like this is the right way to approach this. So I am not comfortable landing until we get more input

I am also not sure if @David-Wobrock 's [comment](https://github.com/django/django/pull/17257#issuecomment-1742117627) has been explored. 

I would love for the folks who have been reviewing to say if they feel this is an acceptable solution or explain why they think this is not the right way to solve the issue :heart:

django/db/migrations/operations/base.py
```diff
@@ -157,6 +157,9 @@ def allow_migrate_model(self, connection_alias, model):
 
         return router.allow_migrate_model(connection_alias, model)
 
+    def reduce_related(self, operation, app_label):
+        return [self]
```
InvalidInterrupt, line 161: Is it a good idea for this function to return a list? It feels like it invites subclassess to cause violations of the requirement that each optimization step must not make the list of operations longer (documented here:
https://github.com/django/django/blob/main/django/db/migrations/optimizer.py#L26-L27 )

django/db/migrations/operations/models.py
```diff
@@ -148,6 +148,27 @@ def references_model(self, name, app_label):
...
+                    and field.remote_field.model == remote_field_old_model
+                ):
+                    name, path, args, kwargs = field.deconstruct()
+                    kwargs["to"] = remote_field_new_model
+                    impacted_field = field.__class__(*args, **kwargs)
+                    impacted_fields.append((_, impacted_field))
+                else:
+                    not_impacted_fields.append((_, field))
+
+            return [replace(self, fields=not_impacted_fields + impacted_fields)]
```
InvalidInterrupt, line 169: Why does this re-order fields so that `impacted_fields` appear last? Does re-ordering the fields change the column order in the database?

django/db/migrations/optimizer.py
```diff
@@ -59,6 +59,11 @@ def optimize_inner(self, operations, app_label):
                         # Otherwise keep trying.
                         new_operations.append(operation)
                         break
+                    new_operations = [
+                        reduced_op
+                        for op in new_operations
```
InvalidInterrupt, line 64: This runs on too many operations. Here's a failing test using a contrived example:

```
    def test_rename_model_referenced_by_fk_2(self):
        self.assertOptimizesTo(
            [
                migrations.CreateModel("Author", []),
                migrations.CreateModel("Person", []),
                migrations.CreateModel(
                    "Restaurant",
                    [
                        (
                            "chef",
                            models.ForeignKey("migrations.person", models.CASCADE),
                        ),
                    ],
                ),
                migrations.RunPython(data_migration),
                migrations.RenameModel("Person", "Chef"),
                migrations.RenameModel("Author", "Person"),
                migrations.CreateModel(
                    "Book",
                    [
                        (
                            "author",
                            models.ForeignKey("migrations.person", models.CASCADE),
                        ),
                    ],
                ),
                migrations.RenameModel("Person", "Writer"),
            ],
            [
                migrations.CreateModel("Author", []),
                migrations.CreateModel("Person", []),
                migrations.CreateModel(
                    "Restaurant",
                    [
                        (
                            "chef",
                            models.ForeignKey("migrations.person", models.CASCADE),
                        ),
                    ],
                ),
                migrations.RunPython(data_migration),
                migrations.RenameModel("Person", "Chef"),
                migrations.RenameModel("Author", "Writer"),
                migrations.CreateModel(
                    "Book",
                    [
                        (
                            "author",
                            models.ForeignKey("migrations.writer", models.CASCADE),
                        ),
                    ],
                ),
            ],
        )
```
where `data_migration` is a serializable function which takes no particular action.

`Restaurant.chef` ends up being pointed to `Writer` (while it should remain `Person`), but is setup in an operation before `Author` is renamed to `Writer`.

django/db/migrations/operations/models.py
```diff
@@ -148,6 +148,27 @@ def references_model(self, name, app_label):
...
+                if (
+                    field.remote_field
+                    and field.remote_field.model == remote_field_old_model
+                ):
+                    name, path, args, kwargs = field.deconstruct()
+                    kwargs["to"] = remote_field_new_model
+                    impacted_field = field.__class__(*args, **kwargs)
+                    impacted_fields.append((_, impacted_field))
+                else:
+                    not_impacted_fields.append((_, field))
```
InvalidInterrupt, line 167: Is using a value assigned to `_` considered good style here? I was briefly confused by it.

InvalidInterrupt (changes requested):
Why is only `CreateModel` addressed, and not `AddField` and `AlterField`? Can they not also create an FK that will need to have its target field adjusted?

Also, with `reduce_related()` you may have created a more general tool that its name suggests. I'm not familiar enough with the community here to know if that warrants any additional scrutiny/discussion. (P.S. perhaps it should be named something like `transform_for_left_optimization`?)

PR #17274, opened 2023-09-17
fixed:#34840 Allow the use of partial indexes by not casting text fields in…

shakori999:
fixed:#34840 Correct value format for CharArrayModel in test_isnull_lookup_cast

In the test_isnull_lookup_cast method, the CharArrayModel instance was created with an incorrect value format for the 'field' attribute. This caused a PostgreSQL error when trying to create the instance. 

This commit fixes the issue by using a list to represent an array and encloses the value properly in curly braces, ensuring the correct format for the 'field' attribute.

Resolves: #34840


shangxiao:
[Closing per ticket](url)

PR #17275, opened 2023-09-18
Support async database backends

domingues:
Hello, first of all I apologize if this is not the appropriate place to have this discussion, if not please direct me to the correct place.

This is a try to start discussing the ability yo have asynchronous backends (e.g. psycopg3) and thus being able to use the ORM asynchronously. This proposal is backwards compatible.

Use the `django.db.backends.postgresql_async` backend to try, do not use `runserver`, not all `Query` methods where migrated but these will work:

- `prefetch_related`
- `acount`
- `afirst`
- `alast`
- `async for`

The approach was:
- added foundations: `BaseDatabaseWrapperAsync`, `CursorWrapperAsync`, `CursorDebugWrapperAsync` - same interface but with async methods when needed;
- migrated PostgresSQL backend to psycopg3 async interface;
- migrate `Query` async methods to use the new async interface;
- call `async_to_sync` on sync methods when using an async backend;
- call `sync_to_async` on async methods when using a sync backend;

Does not investigated yet:

- `runserver` shenanigans;
- migrations;

Problems:

- is created one database connection for each operation done on a sync view when using an async backend. That is expected since database connections are "bind" to threads+coroutines. To fix this we will need to find a way to "bind" the connections to a theard+request, that should also facilitate the implementation of connection pools (https://github.com/django/django/pull/16881);
- for the same motives of previous point, the connections are not closed. On async views the workaround, which is not always feasible, is closing the connection at the end of the view;
- ?

Ty for your feedback.


salomvary:
@domingues Have you seen this? https://github.com/django/deps/blob/main/accepted/0009-async.rst#id13

It is a useful source to learn what the maintainers planned for async ORM.

dima-kov:
@salomvary I've read the RFC. but maybe you know more? I can not see any working/implementation plan. What's the progress? How can I contribute? Do we have bunch of small tickets to make it finally work?

PR #17286, opened 2023-09-19
Defined PASSWORD_HASHERS for auth_tests.test_views.ChangelistTests.

LilyFoote:
`auth_tests.test_views.ChangelistTests.test_view_user_password_is_readonly` depends on the password hasher having the three components `algorithm`, `salt` and `hash`.

The default password hasher (`PBKDF2PasswordHasher`) has an extra `iterations` component, breaking the test.

The `MD5PasswordHasher` fits this requirement and is recommended in the `test_sqlite.py` example settings file for test performance.

Override settings for the entire test class because the hasher is used in `setUpTestData`.

### DIFF
--- a/tests/auth_tests/test_views.py
+++ b/tests/auth_tests/test_views.py
@@ -1361,7 +1361,10 @@ def get_perm(Model, perm):
 
 # Redirect in test_user_change_password will fail if session auth hash
 # isn't updated after password change (#21649)
-@override_settings(ROOT_URLCONF="auth_tests.urls_admin")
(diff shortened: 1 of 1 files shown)


LilyFoote:
This test was failing for me locally because I didn't have `PASSWORD_HASHERS` defined in my per-database settings files. I think it should pass independently of the optimisation suggested in `tests.test_sqlite`, so I'm opening this PR.

felixxm (commented):
@LilyFoote Thanks :+1:

PR #17290, opened 2023-09-20
Refs #34808 -- Doc'd that aggregation functions on empty groups can return None.

shangxiao:


nessita:
Hey @shangxiao, thank you for this patch. I like it and I think it can be merged. IIUC, this should `Refs #34808`? :thinking:

shangxiao:
> I like it and I think it can be merged. IIUC, this should `Refs #34808`? 🤔

Yep I think it could ref ticket #34808 👍

nessita (approved):

PR #17307, opened 2023-09-25
Fixed #32406 Support UPDATE RETURNING

aivarsk:
Implement a new QuerySet.update_returning() that executes a QuerySet.update() and returns modified rows on databases that support UPDATE RETURNING (SQLite, PostgreSQL, Oracle)

### DIFF
--- a/django/db/backends/base/features.py
+++ b/django/db/backends/base/features.py
@@ -38,6 +38,7 @@ class BaseDatabaseFeatures:
     can_use_chunked_reads = True
     can_return_columns_from_insert = False
     can_return_rows_from_bulk_insert = False
+    can_return_columns_from_update = False
     has_bulk_insert = True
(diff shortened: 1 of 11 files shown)


django/db/backends/base/features.py
HamaBarhamou, line 1: either False or True, not FalseFalse or TrueTrue

django/db/backends/base/features.py
```diff
@@ -38,6 +38,7 @@ class BaseDatabaseFeatures:
     can_use_chunked_reads = True
     can_return_columns_from_insert = False
     can_return_rows_from_bulk_insert = False
+    can_return_columns_from_update = False
```
HamaBarhamou, line 41: either False or True, not FalseFalse or TrueTrue

PR #17323, opened 2023-09-29
Refs #34233 -- Used strict flag with zip().

ngnpope:
ticket-34233

Now that we're on Python 3.10+ we can use the `strict` flag for the `zip()` builtin.

Using the `strict=True` can help avoid any potential bugs as a result of arguments of differing length being passed to `zip()`. It is also useful to pass `strict=False` to explicitly make clear that it is expected that a call to `zip()` may have arguments of different length.

I've split into fixup commits for review, but they can be squashed later. The first commit sets `strict=True` for all the cases that support it. The first fixup sets `strict=False` for those that (currently) rely on the arguments being of differing length, and the second fixup alters some cases to allow use of `strict=True` by slicing arguments.


django/contrib/admin/helpers.py
```diff
@@ -342,7 +342,9 @@ def __iter__(self):
             )
 
         for form, original in zip(
-            self.formset.initial_forms, self.formset.get_queryset()
+            self.formset.initial_forms,
+            self.formset.get_queryset()[:self.formset.initial_form_count()],
```
ngnpope, line 346: We only need as many results as the number of initial form forms.
This may be more optimal if the queryset were to return a larger number of results.

django/contrib/admin/templatetags/admin_list.py
```diff
@@ -312,7 +312,11 @@ def __init__(self, form, *items):
 
 def results(cl):
     if cl.formset:
-        for res, form in zip(cl.result_list, cl.formset.forms):
+        for res, form in zip(
+            cl.result_list[:cl.formset.total_form_count()],
+            cl.formset.forms,
```
ngnpope, line 317: Similar to the above, in the two cases in this file we only need as many results as there are forms.

tests/expressions_window/tests.py
```diff
@@ -1356,7 +1356,7 @@ def test_distinct_window_function(self):
             {"year": 2011, "sum": 60000},
             {"year": 2013, "sum": 84000},
         ]
-        for idx, val in zip(range(len(results)), results):
+        for idx, val in enumerate(results):
```
ngnpope, line 1359: This made more sense than keeping `zip()`.

tests/forms_tests/widget_tests/test_selectdatewidget.py
```diff
@@ -611,7 +611,7 @@ def test_value_from_datadict(self):
             (("", "12", "1"), "0-12-1"),
             (("2000", "", "1"), "2000-0-1"),
             (("2000", "12", ""), "2000-12-0"),
-            (("", "", "", ""), None),
+            (("", "", ""), None),
```
ngnpope, line 614: It seems this was an unintentional typo when it was introduced.

felixxm (commented):
@ngnpope Thanks :+1: Do we need to add `strict=True` to all `zip()` calls? This seems unnecessary in many cases.

django/contrib/gis/geos/mutable_list.py
```diff
@@ -293,7 +293,7 @@ def _assign_extended_slice(self, start, stop, step, valueList):
                 "to extended slice of size %d" % (len(valueList), len(indexList))
             )
 
-        for i, val in zip(indexList, valueList):
+        for i, val in zip(indexList, valueList, strict=True):
```
felixxm, line 296: `indexList` and `valueList` cannot have different lengths.

django/contrib/gis/geos/mutable_list.py
```diff
@@ -272,7 +272,7 @@ def _assign_extended_slice_rebuild(self, start, stop, step, valueList):
 
         # we're not changing the length of the sequence
         newLen = len(self)
-        newVals = dict(zip(indexList, valueList))
+        newVals = dict(zip(indexList, valueList, strict=True))
```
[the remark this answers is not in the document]
ngnpope: > Do we need to add strict=True to all zip() calls? This seems unnecessary in many cases.

I think that the benefit is that it prevents any inadvertent regressions. Without it, in most cases, there is no check and so if one of the arguments is modified, e.g. an extra value added for list in a test case, it may not actually get handled as expected.

> `indexList` and `valueList` cannot have different lengths, what is the reason for adding `strict=True`? 🤔

I guess in these cases an explicit check was added above so there isn't all that much benefit here aside from preventing potential regressions. _(In an ideal world `zip()` would have always had strict length checking.)_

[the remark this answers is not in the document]
ngnpope: Also, in a similar vein to adding explicit `strict=False`, it signals what the expected behavior is.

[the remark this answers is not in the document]
ngnpope: @felixxm I've added a fixup that reverts these two cases which you can keep or remove depending on what you decide here. I still think there is value, as explained above.

PR #17327, opened 2023-09-30
Add warning about using native date inputs widgets

blue-hexagon:
Inputs that uses the type="date" attribute cannot be localized by Django.

I spent a lot of hours discovering this the hard way which could have been mitigated by a quick notice/warning - hence the addition.

For more info, see last comment of this ticket: https://code.djangoproject.com/ticket/34853

### DIFF
--- a/docs/topics/i18n/formatting.txt
+++ b/docs/topics/i18n/formatting.txt
@@ -12,6 +12,14 @@ templates using the format specified for the current
 Two users accessing the same content may see dates, times and numbers formatted
 in different ways, depending on the formats for their current locale.
 
+.. warning::
+
+    Because of how browsers enforce their own locale on to form inputs such as dates,
+    Django is not capable of formatting native date inputs (inputs that use the 
+    attribute type="date" and similar). This also means that Django does not use
+    native date inputs by default - as there is nothing it can do to format such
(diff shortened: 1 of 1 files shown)


nessita:
Hello @blue-hexagon, thank you for this work.

I've been reviewing the multiple tickets and PRs related to using `input="date"`, and for the doc changes/improvements, it seems that we have two fronts to tackle.

On one hand,  the PR #14905 adds an example on how to override Django's default behavior. That PR hasn't landed (and it needs some rebasing and updating) because it lacks the explanation about why Django uses the `type="text"` default, which related to this branch you are proposing and takes me to the next point. On the other hand, we have this PR which would ideally be a bit more specific on the details about *why*  `type="text"` is the default and the challenges related to localization.

Would you have time to consider pushing the two efforts together to the finish line?

I would require to read the linked django developer email threads and summarize the issues as stated by Carlton [in this comment](https://code.djangoproject.com/ticket/33113#comment:8).

nessita:
Hello @blue-hexagon, did you give any consideration to my comment above? Let me know if you would have any time to work on that. Thank you!

jacobtylerwalls:
Closing due to inactivity.

PR #17328, opened 2023-09-30
Fixes #27929 -- Not removing original (non-hashed) files after processing

theumang100:
I've introduced an option that enables the deletion of the original files after they've been processed, a step that occurs following the "post_process" method. This deletion process involves either renaming the file to incorporate a content-based hash for cache invalidation purposes or making adjustments to files containing references to other files.

To ensure that only files meeting specific criteria are removed, I've implemented certain checks. These checks involve examining the file's creation date and the date of its last modification. Files that haven't been modified or haven't undergone hashing are retained and not subjected to removal.

I will work on test cases and commit it soon.


similato87:
@CodiumAI-Agent /review

nessita:
Hello @theumang100, thank you for this PR.

The failing tests should be fixed, and also the lint errors. Ideally, commits would be squashed into one or two commits and the previous author that started this work should be credited ([using git's `Co-Authored-By`](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors)).

Would you have time to keep working on this?

theumang100:
yes, will work on it. thanks for feedback.

PR #17368, opened 2023-10-15
Fixed #10941 -- Added a template tag to generate query strings.

knyghty:
I was unsure about `query_string` vs `querystring` as most template tags don't have underscores, but I do prefer with the underscore, personally.

We could also simplify the logic a bit if we say we _only_ allow lists (or nothing at all) when setting something to a list, but it seems nicer to allow any iterable to work.


django/template/defaulttags.py
```diff
@@ -1498,3 +1498,44 @@ def do_with(parser, token):
...
+                query_dict[key] = value
+            else:
+                query_dict.setlist(key, value)
+    for key in keys_to_remove:
+        if key in query_dict:
+            del query_dict[key]
+    query_string = query_dict.urlencode()
+    if not query_string:
+        return ""
+    return f"?{query_string}"
```
felixxm, line 1541: Does it work without `mark_safe()`? :thinking:

docs/ref/templates/builtins.txt
```diff
@@ -953,6 +953,63 @@ output (as a string) inside a variable. This is useful if you want to use
...
+instance. If the ``QueryDict`` is empty, then the output will be an empty
+string. Otherwise, the query string will be displayed with a leading "?".
+
+The following example outputs the current query string verbatim. So if the
+query string is ``?color=green&=size=M``, the output would be
+``?color=green&size=M``:
+
+.. code-block:: html+django
+
+    {% query_string request.GET %}
```
felixxm, line 973: We want to implement a new builtin template tag so it should work in Django flow, so we don't need to support custom `QueryDict` which is rather niche. That's why I'd prefer to pass `request` instead of `request.GET` or even avoid passing `request` at all and implement it via a `Node` subclass where `GET` is available in `render()` (`context.request.GET`).

docs/ref/templates/builtins.txt
```diff
@@ -953,6 +953,63 @@ output (as a string) inside a variable. This is useful if you want to use
...
+
+Each keyword argument given will be added to the query string, replacing any
+existing value for that key. With the query string ``?color=blue``, the
+following would result in ``?color=red&size=S``:
+
+.. code-block:: html+django
+
+    {% query_string request.GET color="red" size="S" %}
+
```
felixxm, line 982: Chop unnecessary blank lines.
```suggestion
```

felixxm (commented):
@knyghty Thanks for this patch :+1: I left initial comments.

docs/ref/templates/builtins.txt
knyghty: Hmm, if we don't want to allow a custom QueryDict, then I think it doesn't make much sense to need to pass the request. So I think the second option makes most sense. And I was really hoping to use `simple_tag`, oh well.

felixxm: > And I was really hoping to use simple_tag, oh well.

That's my proposition to make the usage simpler. You can always disagree :disguised_face:

django/template/defaulttags.py
knyghty: It seems to, I think because `simple_tag` is doing some magic.

docs/ref/templates/builtins.txt
knyghty: No, I think you're right. I tried this approach and realised (unless I'm missing something) it still needs the request context processor to access the request... which means I can again use `simple_tag` 🎉. I added the query_dict as an optional argument in case someone still wants to use it with a custom QueryDict and added a test for this. It's just one extra branch but I can still remove it if you think it's unnecessary. I do agree it's quite niche for someone to want to do this, and either not want to enable the request context processor or want to use a custom QueryDict...

Anyway, I pushed this. I also realised I didn't keep alphabetical ordering in the module so I fixed that as well, though it probably messes up the commit diffs a bit.

carstenfuchs: While not supporting custom `QueryDict`s makes it possible to simplify the tag, I'm not sure if custom `QueryDict`s are really so niche:

In my project, I use URLs like `?year=2023&month=12`. Then in the template for the December report I build links to the previous and next month (e. g. `?year=2024&month=1`). As this is not easy to achieve in the template alone, I have the view precompute a custom `QueryDict`, pass it via the template context and then add or modify extra parameters using the `query_string` tag. For example:
```
{# Explicit, custom QueryDict #}
{% query_string precomputed_querydict_for_next_period format=pdf more_extras=abc %}
```

If the `request.GET` was implied, the same could be achieved alternatively like:
```
{# Implied request.GET #}
{% query_string year=next_period_year month=next_period_month format=pdf more_extras=abc %}
```
but I found this a bit long-winded.

This is why in my original proposal supported multiple dictionaries. It allows something like this:
```
{# Multiple dictionaties. `overrides_for_next_period` is a precomputed dict with keys for year and month #}
{% query_string request.GET overrides_for_next_period format=pdf more_extras=abc %}
```

I may be missing something, but so far I have not found a better (short and clear) way to implement this and thus propose to support this use case. 😜

felixxm: > While not supporting custom `QueryDict`s makes it possible to simplify the tag, I'm not sure if custom `QueryDict`s are really so niche:

I'd argue that a template is not the right place for such a complex mechanism of building query strings.

```
{% query_string year=next_period_year month=next_period_month format=pdf more_extras=abc %}
```
This looks fine to me, TBH.

carstenfuchs: > > While not supporting custom `QueryDict`s makes it possible to simplify the tag, I'm not sure if custom `QueryDict`s are really so niche:
> 
> I'd argue that a template is not the right place for such a complex mechanism of building query strings.

Yes, this is why I found the compromise quite appealing: Have the view precompute query parameters that are complex and “view-centric” while allowing the template to refine the details, e. g. other file formats of the view's data, next page, etc.

> ```
> {% query_string year=next_period_year month=next_period_month format=pdf more_extras=abc %}
> ```
> 
> This looks fine to me, TBH.

Yes, it works, too, especially when the `next_period_*` parameters are grouped like this:

```
{% query_string year=next_period.year month=next_period.month format=pdf more_extras=abc %}
```

Hmmm. Would it be feasible if both the `request.GET` was implied *and* passing extra dicts via
```python
@register.simple_tag
def query_string(*args, **kwargs):
    ...
```
was allowed?

knyghty: It's certainly feasible and the extra complexity is low but I'll leave it to the fellows if they think it's worth doing or not.

For me I think it would see limited use and might make the tag a bit harder to understand as the signature would be more complex.

carstenfuchs: > It's certainly feasible and the extra complexity is low but I'll leave it to the fellows if they think it's worth doing or not.
> 
> For me I think it would see limited use and might make the tag a bit harder to understand as the signature would be more complex.

I agree. Just let me mention one more time that also accepting `dict`s would allow the view to pass parameters to the query string that can be left unnamed in the template.

django/template/defaulttags.py
```diff
@@ -1167,6 +1167,54 @@ def now(parser, token):
...
+            if isinstance(value, str):
+                query_dict[key] = value
+            else:
+                query_dict.setlist(key, value)
+    for key in keys_to_remove:
+        if key in query_dict:
+            del query_dict[key]
+    query_string = query_dict.urlencode()
+    if not query_string:
+        return ""
```
felixxm, line 1214: Can we swap these two lines? There is no need to encode a blank string.
```suggestion
    if not query_string:
        return ""
    query_string = query_dict.urlencode()
```

docs/ref/templates/builtins.txt
felixxm: > It's certainly feasible and the extra complexity is low but I'll leave it to the fellows if they think it's worth doing or not.

Maybe the extra complexity is low, but, IMO, it makes API complicated and confusing. Previously it was straightforward , we pass keyword argument that update the current query string. Now it's,
- if you pass a querydict it will be updated with keyword arguments,
- if you don't pass a querydict the current query string is updated with keyword arguments.

For example, Why doesn't passing a querydict update the current query string? (Personally, I'd assume this behavior.) I believe that we should keep this API as simple and straightforward as possible, because templates are not a good place for such a complicated manipulations.

django/template/defaulttags.py
```diff
@@ -1167,6 +1167,54 @@ def now(parser, token):
     return NowNode(format_string, asvar)
 
 
+@register.simple_tag(takes_context=True)
+def query_string(context, query_dict=None, **kwargs):
+    """
+    Add, remove, and change parameters of the request's `QueryDict`,
```
felixxm, line 1173: This docstring is outdated.

docs/ref/templates/builtins.txt
```diff
@@ -961,40 +961,47 @@ output (as a string) inside a variable. This is useful if you want to use
...
 .. code-block:: html+django
 
-    {% query_string request.GET %}
+    {% query_string %}
+
+Using a custom ``QueryDict``:
+
+.. code-block:: html+django
+
+    {% query_string my_query_dict %}
```
felixxm, line 981: Passing a custom query dict complicates this behavior, and this example doesn't explains anything.

docs/ref/templates/builtins.txt
carstenfuchs: I think that your API description is needlessly complex: It updates the items of `request.GET` and returns them as a query string. If a `QueryDict` is given, it uses that instead of `request.GET`.

More importantly, I also think that linking to the next and previous period from month-year report is only marginally more complicated than linking to the next and previous page of a list report. While the latter dominates many examples and there is built-in support in Django for pagination, I believe that linking to a next and previous period of time is a frequent case as well. My suggestion to also support `dict`s as parameters to the `query_string` tag would support that use case, while not supporting it makes the tag useless to me.

My original implementation just took a list of `QueryDict`s and `dict`s, combined them and then modified single items via named parameters. Generic, explicit and still simple, both in API documentation and code. I much like @knyghty 's resulting implementation, but if I could wish, why not just combine explicitly given `dict`s?

felixxm: So, it seems we need to go back to the Forum where we'll reach a wider audience and see what other think, as we will not reach a consensus on use and implementation here.

We have two options:

1. template tag `{% query_string %}` to modify and return the current query string (from `request.GET`) by passing keyword arguments.
2. template tag `{% query_string %}` to modify any query dict by passing keyword argument, other dicts, or query dicts.

I'm against the second option as it's unnecessarily complicated, it seems rather niche, and IMO template tags are not the place for such complex modifications.

felixxm: https://forum.djangoproject.com/t/adding-a-template-tag-to-generate-query-strings/24521/6

django/template/defaulttags.py
```diff
@@ -1167,6 +1167,54 @@ def now(parser, token):
...
+                query_dict[key] = value
+            else:
+                query_dict.setlist(key, value)
+    for key in keys_to_remove:
+        if key in query_dict:
+            del query_dict[key]
+    query_string = query_dict.urlencode()
+    if not query_string:
+        return ""
+    return f"?{query_string}"
```
carltongibson, line 1215: Don't want to bike shed... Can I just ask @knyghty+@felixxm to confirm we **do want** the `?` automatically prefixed? 

Elsewhere we handle `query_dict.urlencode()` without the prefix. 

(Not a biggie, not expressing a negative, just asking for a double check 🎁)

knyghty: I didn't think about it _too_ hard but I do find it nice that a user can just do this:

```html+django
<a href="{% url "foo" %}{% query_string %}">something</a>
```

without needing to know if the query string is empty or not.

carltongibson: Yep, that makes sense. 

(`/foo/?` would be fine though too right? But maybe not as nice.)

knyghty: It would but indeed I think people (including me) would be a bit annoyed at this. It also adds a potential footgun where you don't realise you need to do this and you end up with `/foo/color=blue&size=M`. The other way around, if you double your `??` seems to be handled (by the browser?) as when I make requests like this the duplicate `?` is removed.

django/template/defaulttags.py
knyghty: Thanks, I've updated it.

docs/ref/templates/builtins.txt
knyghty: Do you have any suggestions here? Carlton had an example in the forum thread, though I don't know if you're asking for an example here or just more explanation, but I'm at a bit of a loss as to what more I could explain.

knyghty: Okay, now I understand, I've updated it.

docs/ref/templates/builtins.txt
```diff
@@ -953,6 +953,71 @@ output (as a string) inside a variable. This is useful if you want to use
...
+.. templatetag:: query_string
+
+``query_string``
+----------------
+
+.. versionadded:: 5.1
+
+Outputs the query string from the given :class:`~django.http.QueryDict`
+instance, or ``request.GET`` if not given and the
+``django.template.context_processors.request`` context processor is enabled.
```
ngnpope, line 965: Can we have a test for the behavior when this context processor is not enabled and a custom `QueryDict` is not provided? I presume that it'll currently just crash... Perhaps we want to make it clearer that the context processor needs to be enabled?

docs/ref/templates/builtins.txt
```diff
@@ -953,6 +953,71 @@ output (as a string) inside a variable. This is useful if you want to use
...
+
+    {% query_string my_query_dict %}
+
+Each keyword argument given will be added to the query string, replacing any
+existing value for that key. With the query string ``?color=blue``, the
+following would result in ``?color=red&size=S``:
+
+.. code-block:: html+django
+
+    {% query_string  color="red" size="S" %}
```
ngnpope, line 990:
```suggestion
    {% query_string color="red" size="S" %}
```

docs/ref/templates/builtins.txt
```diff
@@ -953,6 +953,71 @@ output (as a string) inside a variable. This is useful if you want to use
...
+
+It is also possible to remove values by passing ``None`` as a value. With the
```
ngnpope, line 992: Maybe?

```suggestion
It is also possible to remove parameters by passing ``None`` as a value. With the
```

docs/ref/templates/builtins.txt
```diff
@@ -953,6 +953,71 @@ output (as a string) inside a variable. This is useful if you want to use
...
+query string ``?color=blue&size=M``, the following would result in ``?size=M``:
+
+.. code-block:: html+django
+
+    {% query_string color=None %}
+
+If the given variable is a list, the value will remain as a list. If
```
ngnpope, line 999: Again, perhaps?

```suggestion
If the given parameter is a list, the value will remain as a list. If
```

docs/ref/templates/builtins.txt
```diff
@@ -953,6 +953,71 @@ output (as a string) inside a variable. This is useful if you want to use
...
+the current query string is ``?color=blue&size=M&page=3``, the following code
+would output ``?color=blue&size=M&page=4``:
+
+.. code-block:: html+django
+
+    {% query_string page=page.next_page_number %}
+
+You can also store the value in a variable, for example if you need multiple
+links to the same page with syntax such as
+``{% query_string page=page.next_page_number as next_page %}``.
```
ngnpope, line 1019: With the long inline code block this will probably wrap badly?

```suggestion
links to the same page with syntax such as:

.. code-block:: html+django

    {% query_string page=page.next_page_number as next_page_link %}
```

django/template/defaulttags.py
```diff
@@ -1167,6 +1167,55 @@ def now(parser, token):
...
+        except TypeError:
+            query_dict[key] = value
+        else:
+            if isinstance(value, str):
+                query_dict[key] = value
+            else:
+                query_dict.setlist(key, value)
+    for key in keys_to_remove:
+        if key in query_dict:
+            del query_dict[key]
```
ngnpope, line 1212: If we're not iterating and modifying `query_dict` in the loop (we're iterating over `kwargs`) and `kwargs` cannot have duplicated keys anyway, why not remove keys directly? In addition, the try-except-else-and-isinstance is hard to understand at first glance. Would the following be acceptable instead?

```suggestion
    for key, value in kwargs.items():
        if value is None:
            if key in query_dict:
                del query_dict[key]
        elif isinstance(value, Iterable) and not isinstance(value, str):
            query_dict.setlist(key, value)
        else:
            query_dict[key] = value
```

ngnpope (changes requested):
Thanks for this - it'll be helpful to have a nice clean way to manipulate query strings!

docs/ref/templates/builtins.txt
```diff
@@ -953,6 +953,71 @@ output (as a string) inside a variable. This is useful if you want to use
...
+``query_string``
+----------------
+
+.. versionadded:: 5.1
+
+Outputs the query string from the given :class:`~django.http.QueryDict`
+instance, or ``request.GET`` if not given and the
+``django.template.context_processors.request`` context processor is enabled.
+If the ``QueryDict`` is empty, then the output will be an empty string.
+Otherwise, the query string will be displayed with a leading "?".
```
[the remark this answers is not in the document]
knyghty: I went with `displayed` as it's used elsewhere in the docs, perhaps as a way to use simpler language for template developers, but it's unlikely you'd really want to "display" them, so yes, good call I think.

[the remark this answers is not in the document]
ngnpope: > ...but it's unlikely you'd really want to "display" them...

Yup, that was the point - it's not really something that's being "displayed".

django/template/defaulttags.py
```diff
@@ -1167,6 +1167,55 @@ def now(parser, token):
     return NowNode(format_string, asvar)
 
 
+@register.simple_tag(takes_context=True)
+def query_string(context, query_dict=None, **kwargs):
+    """
+    Add, remove, and change parameters of a `QueryDict` and return the result
+    as a querystring. If the `query_dict` argument is not provided,
+    `request.GET` will be used.
```
ngnpope, line 1175: Double backticks as in theory this is reStructuredText?

```suggestion
    Add, remove, and change parameters of a ``QueryDict`` and return the result
    as a query string. If the ``query_dict`` argument is not provided,
    ``request.GET`` will be used.
```

django/template/defaulttags.py
```diff
@@ -1167,6 +1167,55 @@ def now(parser, token):
...
+
+    To remove a key::
+
+        {% query_string foo=None %}
+
+    To use with pagination::
+
+        {% query_string page=page_obj.next_page_number %}
+
+    A custom `QueryDict` can also be used:
```
ngnpope, line 1189:
```suggestion
    A custom ``QueryDict`` can also be used::
```

ngnpope (changes requested):
Sorry - overlooked a few things in the docstring.

django/template/defaulttags.py
```diff
@@ -1167,6 +1168,47 @@ def now(parser, token):
...
+
+        {% query_string my_query_dict foo=3 %}
+    """
+    if query_dict:
+        query_dict = query_dict.copy()
+    else:
+        query_dict = context.request.GET.copy()
```
ngnpope, line 1197: It probably makes sense to check `is None` to avoid a case where an intentionally empty `query_dict` is passed falling back to `request.GET`. We should have a test for this case.

```suggestion
    if query_dict is None:
        query_dict = context.request.GET
    query_dict = query_dict.copy()
```

We should probably also have a another test where we actually pass in a custom, non-empty `QueryDict` instance as we only seem to have `query_string08` which explicitly passes in `request.GET`.

ngnpope (changes requested):

docs/ref/templates/builtins.txt
knyghty: I've added it. I also realised we were missing a test for when using a custom `QueryDict` is supplied without the context processor, which should and does work, so I've added that as well.

I added a warning into the docs as well, though not sure about the wording.

django/template/defaulttags.py
knyghty: I added this test already when addressing the other test.

This change seems reasonable to me, pushed this as well 👍🏻

django/template/defaulttags.py
```diff
@@ -6,6 +6,7 @@
 from datetime import datetime
 from itertools import cycle as itertools_cycle
 from itertools import groupby
+from typing import Iterable
```
ngnpope, line 9: [`typing.Iterable`](https://docs.python.org/3/library/typing.html#typing.Iterable) has been deprecated since Python 3.9...

```suggestion
from collections.abc import Iterable
```

docs/ref/templates/builtins.txt
```diff
@@ -953,6 +953,80 @@ output (as a string) inside a variable. This is useful if you want to use
...
+if provided, or ``request.GET`` if not and the
+``django.template.context_processors.request`` context processor is enabled.
+If the ``QueryDict`` is empty, then the output will be an empty string.
+Otherwise, the query string will be returned with a leading "?".
+
+.. warning::
+    If not using the ``django.template.context_processors.request`` context
+    processor, you must pass either the ``request`` into the template context
+    or a ``QueryDict`` instance into this tag.
+
```
ngnpope, line 973:
```suggestion
```

ngnpope (changes requested):
_Let me know if this iterative reviewing is unhelpful and you'd like me to hold off for a bit - I realise we're practically pairing at this point..._ 🍐

django/template/defaulttags.py
knyghty: Arg! Wish they would make their mind up.

No problem with the short feedback loop, it's keeping me busy!

felixxm (approved):
@knyghty Thanks for updates :+1: 

I pushed small edits to tests and docs.

django/template/defaulttags.py
```diff
@@ -10,6 +10,7 @@
 from django.conf import settings
 from django.utils import timezone
 from django.utils.html import conditional_escape, escape, format_html
+from django.utils.itercompat import is_iterable
```
ngnpope, line 13: I'm curious as to why we still use this over `isinstance(…, collections.abc.Iterable)`? 🤔

Looking at Python's [documentation](https://docs.python.org/3.10/library/collections.abc.html) -- see `3)` in the first section -- it seems as though it should work perfectly fine. Is there something I'm missing?

It looks as though this was introduced in ad077ccbc03f11c3db3951aab110ec98c5907729 for compatibility with Jython which we no longer support… Perhaps it's time to deprecate `is_iterable()`?

kezabelle: `is_iterable` would presumably be `True` for objects using the old `__getitem__` method, which I don't think is equivalent for `isinstance(..., Iterable)`:

> Checking `isinstance(obj, Iterable)` detects classes that are registered as `Iterable` or that have an `__iter__()` method, but it does not detect classes that iterate with the `__getitem__()` method. The only reliable way to determine whether an object is iterable is to call `iter(obj)`.

That may or may not matter these days (or my reading my be wrong 🤷)

felixxm: > It looks as though this was introduced in [ad077cc](https://github.com/django/django/commit/ad077ccbc03f11c3db3951aab110ec98c5907729) for compatibility with Jython which we no longer support… Perhaps it's time to deprecate `is_iterable()`?

Maybe it's time, but we should discuss it and do it separately. I've changed to `is_iterable()` to keep the code consistent for now.

claudep:
That's a great one, thanks a ton! :heart:

django/template/defaulttags.py
ngnpope: I opened ticket-34983 for this.

PR #17369, opened 2023-10-16
Fixed #34798 -- Fixed QuerySet.aggregate() crash when referencing expressions containing subqueries.

charettes:
Regression in 59bea9e, complements e5c844d.

Refs #28477, #34551.

Thanks Haldun Komsuoglu for the report.


felixxm (approved):
@charettes Thanks :+1: I added release notes.

PR #17370, opened 2023-10-16
Fixed #34899 -- Added blank choice to callable choices lazily.

ngnpope:
ticket-34899

So the challenge was that adding the blank choice was performed eagerly. The easiest solution is, therefore, to make the addition of the blank entry lazy. Unfortunately this also means that we need to extract the handling of flattening of choices into a helper function -- `flatten_choices()` (which is done in the first commit). From there, a `BlankChoiceIterator` falls out quite nicely...

...but then we have a handful of issues in testing. It seems that these can be overcome by implementing `__eq__()` and `__getitem__()` on `BaseChoiceIterator`. Alternatively we could use 7e35b749c5add17ba58bc2e40a0183d8b33e4b16 to only use `BlankChoiceIterator` where we encounter a `BaseChoiceIterator` subclass, but then we're duplicating logic which is not so nice... And it also means that comparisons against or subscript access for lazy choices will still fail.

### DIFF
--- a/django/utils/choices.py
+++ b/django/utils/choices.py
@@ -1,10 +1,11 @@
 from collections.abc import Callable, Iterable, Iterator, Mapping
-from itertools import islice, zip_longest
+from itertools import islice, tee, zip_longest
 
 from django.utils.functional import Promise
 
 __all__ = [
     "BaseChoiceIterator",
+    "BlankChoiceIterator",
     "CallableChoiceIterator",
     "flatten_choices",
     "normalize_choices",
@@ -32,6 +33,20 @@ class BaseChoiceIterator:
         raise NotImplementedError(
             "BaseChoiceIterator subclasses must implement __iter__()."
         )
+
+
+class BlankChoiceIterator(BaseChoiceIterator):
+    """Iterator to lazily inject a blank choice."""
+
+    def __init__(self, choices, blank_choice):
+        self.choices = choices
+        self.blank_choice = blank_choice
+
+    def __iter__(self):
+        choices, other = tee(self.choices)
+        if not any(value in ("", None) for value, _ in flatten_choices(other)):
+            yield from self.blank_choice
+        yield from choices
 
 
 class CallableChoiceIterator(BaseChoiceIterator):
--- a/django/utils/choices.py
+++ b/django/utils/choices.py
@@ -1,10 +1,36 @@
 from collections.abc import Callable, Iterable, Iterator, Mapping
+from itertools import islice, zip_longest
 
 from django.utils.functional import Promise
+
+__all__ = [
+    "BaseChoiceIterator",
+    "CallableChoiceIterator",
+    "normalize_choices",
+]
 
 
 class BaseChoiceIterator:
     """Base class for lazy iterators for choices."""
+
+    def __eq__(self, other):
+        if isinstance(other, Iterable):
+            return all(a == b for a, b in zip_longest(self, other, fillvalue=object()))
+        return super().__eq__(other)
+
+    def __getitem__(self, index):
+        if index < 0:
+            # Suboptimally consume whole iterator to handle negative index.
+            return list(self)[index]
+        try:
+            return next(islice(self, index, index + 1))
+        except StopIteration:
+            raise IndexError("index out of range") from None
+
+    def __iter__(self):
+        raise NotImplementedError(
+            "BaseChoiceIterator subclasses must implement __iter__()."
+        )
 
 
 class CallableChoiceIterator(BaseChoiceIterator):
--- a/django/utils/choices.py
+++ b/django/utils/choices.py
@@ -6,6 +6,7 @@ from django.utils.functional import Promise
 __all__ = [
     "BaseChoiceIterator",
     "CallableChoiceIterator",
+    "flatten_choices",
     "normalize_choices",
 ]
 
@@ -41,6 +42,15 @@ class CallableChoiceIterator(BaseChoiceIterator):
 
     def __iter__(self):
         yield from normalize_choices(self.func())
+
+
+def flatten_choices(choices):
+    """Flatten choices by removing nested values."""
+    for value_or_group, label_or_nested in choices or ():
+        if isinstance(label_or_nested, (list, tuple)):
+            yield from label_or_nested
+        else:
+            yield value_or_group, label_or_nested
 
 
 def normalize_choices(value, *, depth=0):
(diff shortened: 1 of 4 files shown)


adamchainz:
I can confirm this fixes [my test project](https://github.com/adamchainz/django-5.0-choices-laziness).

django/utils/choices.py
```diff
@@ -1,11 +1,52 @@
...
+
+class BlankChoiceIterator(BaseChoiceIterator):
+    """Iterator to lazily inject a blank choice."""
+
+    def __init__(self, choices, blank_choice):
+        self.choices = choices
+        self.blank_choice = blank_choice
+
+    def __iter__(self):
+        choices, other = tee(self.choices)
```
adamchainz, line 45: Only iterating once also fixes a bug where a callable returns different results each time ✅

tests/model_forms/tests.py
```diff
@@ -2012,6 +2013,37 @@ def test_runtime_choicefield_populated(self):
...
 
+    def test_callable_choices_are_lazy(self):
+        callable_choices_counter = 0
+
+        def get_animal_choices():
+            nonlocal callable_choices_counter
+            callable_choices_counter += 1
+            return [("LION", "Lion"), ("ZEBRA", "Zebra")]
+
+        class ZooKeeper(models.Model):
```
adamchainz, line 2024: will need `django.test.utils.isolate_apps`

ngnpope:
Thanks for the initial review @adamchainz.

I've restructured a little bit and think this should be pretty much ready now.

tests/model_forms/tests.py
```diff
@@ -2012,6 +2013,38 @@ def test_runtime_choicefield_populated(self):
...
+                choices=get_animal_choices,
+                max_length=5,
+            )
+
+        class ZooKeeperForm(forms.ModelForm):
+            class Meta:
+                model = ZooKeeper
+                fields = ["animal"]
+
+        self.assertEqual(call_count, 0)
```
nessita, line 2037: I was hoping we could use a Mock.mock(wraps=...) here but it seems it does not work with dunder methods so the call is not recorded. Is this why you chose to use a plain integer count?

tests/utils_tests/test_choices.py
```diff
@@ -2,10 +2,40 @@
...
+        msg = "BaseChoiceIterator subclasses must implement __iter__()."
+        with self.assertRaisesMessage(NotImplementedError, msg):
+            iter(InvalidChoiceIterator())
+
+    def test_eq(self):
+        self.assertEqual(
+            SimpleChoiceIterator(),
+            [(1, "Item #1"), (2, "Item #2"), (3, "Item #3")],
+        )
+
```
nessita, line 32: I added this test:

```python
    def test_eq_instances(self):
        self.assertEqual(
            SimpleChoiceIterator(),
            SimpleChoiceIterator(),
        )
```

And it produces this failure:
```
ERROR: test_eq_instances (utils_tests.test_choices.ChoiceIteratorTests.test_eq_instances)
  File "/usr/lib/python3.11/unittest/case.py", line 57, in testPartExecutor
    yield
  File "/usr/lib/python3.11/unittest/case.py", line 623, in run
    ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/unittest/case.py", line 579, in _callTestMethod
    if method() is not None:
    ^^^^^^^^^^^^^^^^^
  File "/home/nessita/fellowship/django/tests/utils_tests/test_choices.py", line 34, in test_eq_instances
    self.assertEqual(
  File "/usr/lib/python3.11/unittest/case.py", line 873, in assertEqual
    assertion_func(first, second, msg=msg)
    ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/unittest/case.py", line 863, in _baseAssertEqual
    if not first == second:
    ^^^^^^^^^^^^^^^^^
  File "/home/nessita/fellowship/django/django/utils/choices.py", line 20, in __eq__
    return type(other)(self) == other
    ^^^^^^^^^^^^^^^^^
TypeError: SimpleChoiceIterator() takes no arguments

```
I think the `__eq__` should work with two instances of the iterator, what do you think?

nessita (commented):
Hey @ngnpope, thanks for the PR. Today I realized that I should've reviewed this sooner since today is Beta release day, but I lost track of time last week during DjangoCon.

Do you have any time to work on the comments or shall I? Thanks again!

tests/utils_tests/test_choices.py
nessita: I haven't had to implement much iterables/iterators before, but after some investigation, this seems to be a better implementation for `__eq__`:

```python
return all(a == b for a, b in zip_longest(self, other, fillvalue=object()))
```

@adamchainz @ngnpope what do you think?

nessita:
I pushed a rebase onto main plus some extra tests for the `__eq__` definition, using the proposed implementation to fix the test failures I was having. I also added tests for the `IndexError` case.

While tests run for these changes, I will work on adding some tests for the extracted `flatten_choices` helper.

tests/model_forms/tests.py
ngnpope: Honestly I did this because I was checking at what point things got called after definition, etc and it was a simple next step from the provided reproducer that used a boolean. If it can be made to use the approach you mentioned, that's fine, but this is also simple and self-contained.

tests/utils_tests/test_choices.py
ngnpope: This might be a bit magical? But I guess relying on the fact that we need to iterate is ok.

ngnpope:
Thanks for looking 😊

> Do you have any time to work on the comments or shall I? Thanks again!

Sorry, but I won't be able to do anything until next week. Would you be able to take it over the finishing line for me?

tests/model_forms/tests.py
nessita: I tried to change to use mock but it fails as if the mock was never called (but it is!)
After some investigation, this may be related with https://github.com/python/cpython/issues/69783

So I think that leaving it as proposed is OK. Thanks!

tests/utils_tests/test_choices.py
nessita: I guess is a matter of taste, it took me 5 iterations of prints to understand what the original code was doing, hahaha

The fact that this nicely works with any iterable instance, makes me prefer this option.

nessita:
> Thanks for looking 😊
> 
> > Do you have any time to work on the comments or shall I? Thanks again!
> 
> Sorry, but I won't be able to do anything until next week. Would you be able to take it over the finishing line for me?

Yes, absolutely! I have made one final commit with tests for `flatten_choices` and I will merge once the CI run is green. After that, beta release! :tada: 

Thank you for the PR!

PR #17384, opened 2023-10-19
Fixed #15578 -- Stated the processing order of fixtures in the fixtures docs.

dtraleigh:
Documentation update for open ticket.

https://code.djangoproject.com/ticket/15578


docs/topics/db/fixtures.txt
```diff
@@ -126,11 +126,31 @@ You could also write a decorator to encapsulate this logic::
...
+
+.. code-block:: shell
+
+    manage.py loaddata foo bar baz
+
+or (in a test case)
+
+.. code-block:: shell
+
+    Fixtures = ['foo', 'bar', 'baz']
```
evildmp, line 139: If I were writing this, I might use more real-sounding names, like `people`, `places`, `events` - I find that helps me imagine an example better.

evildmp (approved):
This looks good to me.

lsaffre (changes requested):
Thank you so far, but what happens when we have two apps a and b (installed in that order) and a has a fixture bar while b has a fixture foo? IOW is the ordering rather (fixture, app) or rather (app, fixture)?

docs/topics/db/fixtures.txt
nessita: I agree with this. Following the existing examples, we could do: `mydata`, `otherdata`, `thirdset` or similar.

docs/topics/db/fixtures.txt
```diff
@@ -126,11 +126,31 @@ You could also write a decorator to encapsulate this logic::
 Just be aware that this logic will disable the signals whenever fixtures are
 deserialized, not just during ``loaddata``.
 
-Note that the order in which fixture files are processed is undefined. However,
-all fixture data is installed as a single transaction, so data in
-one fixture can reference data in another fixture. If the database backend
-supports row-level constraints, these constraints will be checked at the
-end of the transaction.
+Fixtures can be loaded in a specified order. For example:
```
nessita, line 129: I think we need to out this in a subsection, with a title along the lines `In which order are the fixtures loaded?` (I'm not sure I like all the title-questions, but my proposal would be consistent with the existing style) @evildmp any opinion on the "titles as questions"?

docs/topics/db/fixtures.txt
nessita: I also think that this example should include the class definition, otherwise is unclear where the line would show up in a test case.

docs/topics/db/fixtures.txt
```diff
@@ -126,11 +126,31 @@ You could also write a decorator to encapsulate this logic::
...
+    manage.py loaddata foo bar baz
+
+or (in a test case)
+
+.. code-block:: shell
+
+    Fixtures = ['foo', 'bar', 'baz']
+
+In this example, ``foo`` refers to any fixture named ``foo`` in any application.
+This would load ``foo`` followed by ``bar`` and then ``baz``.
```
nessita, line 142:
```suggestion
This would first load all the ``foo``-named fixtures from all applications (in
the order in which applications are defined), followed by all the ``bar`` and
then all the ``baz`` fixtures.
```

docs/topics/db/fixtures.txt
```diff
@@ -126,11 +126,31 @@ You could also write a decorator to encapsulate this logic::
...
+
+Where there are several fixtures with the same name, the order in which
+applications are defined in ``INSTALLED_APPS`` is also taken into account.
+The application listed first will be the one whose fixture is loaded first.
```
nessita, line 146: Only after submitting the suggestion above, I read this paragraph. I think it would be idea to come up with a unified paragraphs with the combined information for fixtures order + app order.

nessita (commented):
Looks good, I added some comments and suggestions! Thanks

docs/topics/db/fixtures.txt
```diff
@@ -126,11 +126,31 @@ You could also write a decorator to encapsulate this logic::
...
+
+Be aware that if the database backend supports row-level constraints, these
+constraints will be checked at the end of the transaction. Any relationships
+across fixtures may result in a load error if the database configuration does
+not support deferred constraint checking. (See `MySQL`_ for an example)
```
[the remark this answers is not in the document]
evildmp: I may be late to this particular party, but I'd be quite keen to remove slangy language like "check out" from the documentation if we can. I'd prefer "See ..." or "Refer to ...".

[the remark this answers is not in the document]
felixxm: I've been trying to avoid `see ...` because it seems less inclusive than `check out...`

docs/topics/db/fixtures.txt
lsaffre: To summarize above comments, I suggest something like this as the code example:

.. code-block:: python

  class AnimalTestCase(TestCase):
      fixtures = ["mammals", "birds", "insects"]

dtraleigh:
> Thank you so far, but what happens when we have two apps a and b (installed in that order) and a has a fixture bar while b has a fixture foo? IOW is the ordering rather (fixture, app) or rather (app, fixture)?

Appreciate the question. From the comments in the ticket, it looks like it would be fixture, app. So with your example, if I ran the command:
`manage.py loaddata foo bar`
the data load order would be (assuming these exist):
1. foo from application a
2. foo from application b
3. bar from application a
4. bar from application b

Having said that, I do believe this is communicated starting on line 142 but do please let me know if we need to tweak it further.

docs/topics/db/fixtures.txt
```diff
@@ -126,11 +126,34 @@ You could also write a decorator to encapsulate this logic::
...
-Note that the order in which fixture files are processed is undefined. However,
-all fixture data is installed as a single transaction, so data in
-one fixture can reference data in another fixture. If the database backend
-supports row-level constraints, these constraints will be checked at the
-end of the transaction.
+Fixtures can be loaded in a specified order. For example:
+
+.. code-block:: shell
+
+    manage.py loaddata foo bar baz
```
lsaffre, line 133: I suggest to change this to "manage.py loaddata mammals birds insects"

docs/topics/db/fixtures.txt
```diff
@@ -126,11 +126,34 @@ You could also write a decorator to encapsulate this logic::
...
+        fixtures = ["mammals", "birds", "insects"]
+
+In this example, ``mammals`` refers to any fixture named ``mammals`` in any
+application. This would first load all the ``mammals``-named fixtures from all
+applications (in the order in which applications are defined), followed by
+all the ``birds`` and then all the ``insects`` fixtures.
+
+Where there are several fixtures with the same name, the order in which
+applications are defined in ``INSTALLED_APPS`` is also taken into account.
+The application listed first will be the one whose fixture is loaded first.
```
lsaffre, line 149: I suggest to replace lines 142-149 by the following single sentence, which answers my question:

These examples would first load all the fixtures named ``mammals`` from all applications (in the order in which applications are defined in ``INSTALLED_APPS``), followed by all the ``birds`` and then all the ``insects`` fixtures.

docs/topics/db/fixtures.txt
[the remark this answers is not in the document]
nessita: I suggest we use "Refer to" since is more clear in intent and less informal. @dtraleigh Could you make this change please?

docs/topics/db/fixtures.txt
nessita: @dtraleigh Is this is something you could work on? I do think we should have the "order of fixtures" in its own subsection now that we are expanding it.

docs/topics/db/fixtures.txt
nessita: I think this is a good proposal. @dtraleigh do you agree? Could you incorporate this?

dtraleigh: Concise. Yes, I'll add.

docs/topics/db/fixtures.txt
dtraleigh: @nessita How does this subsection sound to you?

dtraleigh:
@nessita Final changes pushed. I believe this addresses all feedback. Ready to tweak as needed by you or anyone else. 😃

nessita:
> @nessita Final changes pushed. I believe this addresses all feedback. Ready to tweak as needed by you or anyone else. 😃

Thank you! I did some minor changes, most notably I moved the newly added subsection a little upwards. Will merge as soon as the CI finishes the checks.

Thank you for your contribution! For many more :clinking_glasses:

PR #17388, opened 2023-10-20
Fixed #34909 -- Associated links in admin navigation sidebar with row descriptions.

erosselli:
## Problem
Fixed [#34909](https://code.djangoproject.com/ticket/34909). 

In the Admin home page that lists your apps and their models (see screenshot below), there is an `Add` and a `Change` button for each model. All these buttons had the same accessible name, `"Add"` and `"Change"` respectively. This could be confusing for users using screen readers, since all the buttons would be called the same. It's better if the screen reader reads out something like `"Add <model-name>"` instead of just `"Add"`, so that it's clearer for the user _what_ they're adding or changing. See [accessible naming guidelines](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/#composingeffectiveanduser-friendlyaccessiblenames) for more information on this topic.

## Alternative option
Adding something like `aria-label="Add <model-name>"` to the buttons would fix the issue for screen reader users, but could impact the experience of voice control users. An example of this is that a user could use voice control saying something like "Click Add", but because the `Add` button now has a "hidden" name that's different (`Add <model-name>`), the voice control might not recognize this button when the user says "Click Add". This kind of issue is explained in more detail in [this post about voice control usability considerations](https://www.smashingmagazine.com/2022/06/voice-control-usability-considerations-partially-visually-hidden-link-names/). 

## Implemented solution
The implemented fix was to use `aria-describedby` with a reference to the model name, so that screen reader users can hear something like "Add" followed by the model name after a brief pause, while the accessible name of the button remains `Add` and voice control users are not affected. This fix was tested using both the Mac VoiceOver utility and the Mac Voice Control utility (both in Ventura 13.4).

The Admin screens that were updated are the following:
Screenshot of Django Admin homepage
Screenshot of Django Admin page that lists all models for a specific app

### DIFF
--- a/django/contrib/admin/templates/admin/app_list.html
+++ b/django/contrib/admin/templates/admin/app_list.html
@@ -8,29 +8,33 @@
           <a href="{{ app.app_url }}" class="section" title="{% blocktranslate with name=app.name %}Models in the {{ name }} application{% endblocktranslate %}">{{ app.name }}</a>
         </caption>
         {% for model in app.models %}
-          <tr class="model-{{ model.object_name|lower }}{% if model.admin_url in request.path|urlencode %} current-model{% endif %}">
-            {% if model.admin_url %}
-              <th scope="row"><a href="{{ model.admin_url }}"{% if model.admin_url in request.path|urlencode %} aria-current="page"{% endif %}>{{ model.name }}</a></th>
-            {% else %}
-              <th scope="row">{{ model.name }}</th>
-            {% endif %}
+          {% with model_name=model.object_name|lower %}
+            <tr class="model-{{ model_name }}{% if model.admin_url in request.path|urlencode %} current-model{% endif %}">
+              <th scope="row">
+                {% if model.admin_url %}
+                  <a href="{{ model.admin_url }}"{% if model.admin_url in request.path|urlencode %} aria-current="page"{% endif %}>{{ model.name }}</a>
+                {% else %}
+                  {{ model.name }}
+                {% endif %}
+              </th>
 
-            {% if model.add_url %}
-              <td><a href="{{ model.add_url }}" class="addlink">{% translate 'Add' %}</a></td>
-            {% else %}
-              <td></td>
-            {% endif %}
-
-            {% if model.admin_url and show_changelinks %}
-              {% if model.view_only %}
-                <td><a href="{{ model.admin_url }}" class="viewlink">{% translate 'View' %}</a></td>
+              {% if model.add_url %}
+                <td><a href="{{ model.add_url }}" class="addlink">{% translate 'Add' %}</a></td>
               {% else %}
-                <td><a href="{{ model.admin_url }}" class="changelink">{% translate 'Change' %}</a></td>
+                <td></td>
               {% endif %}
-            {% elif show_changelinks %}
-              <td></td>
-            {% endif %}
-          </tr>
+
+              {% if model.admin_url and show_changelinks %}
+                {% if model.view_only %}
+                  <td><a href="{{ model.admin_url }}" class="viewlink">{% translate 'View' %}</a></td>
+                {% else %}
+                  <td><a href="{{ model.admin_url }}" class="changelink">{% translate 'Change' %}</a></td>
+                {% endif %}
+              {% elif show_changelinks %}
+                <td></td>
+              {% endif %}
+            </tr>
+          {% endwith %}
         {% endfor %}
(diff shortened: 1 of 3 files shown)


nessita:
Thank you @erosselli for this work!

I have one question: when a user using the voice control system says `Click Add`, and there are multiple `Add` links, which one does the system chooses?

Separately, I would suggest that one or two extra tests are added covering the new structure of the `th` and the adding of the new `aria-describedby` attributes.

I'll be marking the ticket as "patch needs improvement", but please once you push these changes, please unset the ticket flags "patch needs improvement" and "patch needs tests" so it shows again in the review queue.

erosselli:
Hi @nessita , when using voice control, all actionable items will be numbered. When you say something like "Click Add", it will number the buttons that matched the name (in this case "Add"), and you can then say the number of the button you want to click. See screenshot below
I'll work on adding the missing tests.
Screenshot 2023-10-30 at 17 44 53

erosselli:
@nessita I have added a test for the accessible button names

nessita:
> @nessita I have added a test for the accessible button names

Thank you @erosselli! The branch looks great. Question: could we use a more verbose `th` id such as `"{{ app.app_label }}-{{ model.object_name|lower }}"`? Would this affect the general outcome in any way?

(From my understanding of how `aria-describedby` works, this should work, but I wanted to double check with you.)

erosselli:
Hi @nessita , I think it should work, as long as `model.object_name` does not have any whitespaces, which I think it shouldn't right? If it does have spaces, that would affect the meaning of `aria-describedby` , since it would be `aria-described-by="first_word second_word"` and that would be interpreted as two separate `id`s .

nessita:
I'm unable to reproduce the Selenium issue locally and it seems quite unrelated to this change.

django/contrib/admin/templates/admin/app_list.html
```diff
@@ -8,29 +8,33 @@
           <a href="{{ app.app_url }}" class="section" title="{% blocktranslate with name=app.name %}Models in the {{ name }} application{% endblocktranslate %}">{{ app.name }}</a>
         </caption>
         {% for model in app.models %}
-          <tr class="model-{{ model.object_name|lower }}{% if model.admin_url in request.path|urlencode %} current-model{% endif %}">
-            {% if model.admin_url %}
-              <th scope="row"><a href="{{ model.admin_url }}"{% if model.admin_url in request.path|urlencode %} aria-current="page"{% endif %}>{{ model.name }}</a></th>
-            {% else %}
-              <th scope="row">{{ model.name }}</th>
-            {% endif %}
+          {% with model_name=model.object_name|lower %}
```
felixxm, line 11: As far as I'm aware it's the same as `model.model_name`, so with is unnecessary.

nessita: I suggested that to maintain the existing definition in the `tr` row (otherwise we would be changing semantics from what's in `main`). Do you think it's safe to remove lower?

felixxm: I didn't suggest to remove `lower`. I suggested to use `model_name` instead of `object_name.lower`:

https://github.com/django/django/blob/0eaad0bb593512bedbc96e52bdcf03213d691ab3/django/db/models/options.py#L178-L179

felixxm: Nvd, we have here `model_dict`, not `model._meta` so `model_name` is not available.

felixxm:
@erosselli Thanks :+1: Welcome aboard ⛵ 

I made small edits to tests, pushed template refactoring to a separate commit.

nessita:
@felixxm yesterday I've pinged the Accessibility team for a review and they will complete it soon, so the merging should wait until then. Thanks!

felixxm:
> @felixxm yesterday I've pinged the Accessibility team for a review and they will complete it soon, so the merging should wait until then. Thanks!

I know, have you seen my comment on Discord? I don't think we need another round of manual accessibility tests, I'm going to merge it today or tomorrow.

nessita:
> > @felixxm yesterday I've pinged the Accessibility team for a review and they will complete it soon, so the merging should wait until then. Thanks!
> 
> I know, have you seen my comment on Discord?

I didn't before, now I did :-)

> I don't think we need another round of manual accessibility tests, I'm going to merge it today or tomorrow.

Ok, thanks.

PR #17392, opened 2023-10-20
Fixed #27403 -- Doc'd that QuerySet.prefetch_related() doesn't guarantee transactional consistency.

timb07:
https://code.djangoproject.com/ticket/27403

Added a note about the potential race condition in prefetch_related() that could produce an inconsistent result, one that does not correspond to any point in the database history.

The scenario and doc change were discussed with @evildmp at the DjangoCon US 2023 sprints.

### DIFF
--- a/docs/ref/models/querysets.txt
+++ b/docs/ref/models/querysets.txt
@@ -1211,6 +1211,19 @@ results; these ``QuerySets`` are then used in the ``self.toppings.all()`` calls.
 The additional queries in ``prefetch_related()`` are executed after the
 ``QuerySet`` has begun to be evaluated and the primary query has been executed.
 
+Note that there is no mechanism to prevent another database query from altering
+the items in between the execution of the primary query and the additional
+queries, which could produce an inconsistent result. For example, if a
+``Pizza`` is deleted after the primary query has executed, its toppings will
+not be returned in the additional query, and it will seem like the pizza has no
+toppings:
+
+.. code-block:: pycon
+
+    >>> Pizza.objects.prefetch_related("toppings")
+    #  "Hawaiian" Pizza was deleted in another shell.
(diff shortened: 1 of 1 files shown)


charettes:
It should block this PR but this problem is even more common when lazily accessing foreign key due to a lack of `select_related` usage.

The reason why it's more common than when `prefetch_related` is used is that the latter at least minimises the time between both queries and thus reduces the possibilities of data drifts. It's not the case for foreign key relationships that are preemptively fetched with `select_related` as a lot of time can be spent before they are lazily accessed.

e.g.

```python
book = Book.objects.get(pk=book_pk)
# do something with book
# ...
# eventually try to access book.author
author_name = book.author.name
# Nothing guarantees that `book.author` won't raise a `DoesNotExist` here of that
# it's coherent with what book was when it was fetched
```

What I'm trying to say here is that while problem exists for `prefetch_related` it is generalized to any lack of fetching a relationship graph in multiple queries while under `READ COMMITED` (recommended isolation level for Django). I do understand that users might expect this not to be case though hence why this documentation addition is worthy.

As for how we could limit this data drift at the ORM level we could rely on JSON support to help to some extent here. For example, if we added support for custom `SelectRelated` to `QuerySet.select_related` (ticket-23854) we could imagine that support for multi-valued relationship could be implemented as well which should ensure snapshot integrity

```python
Author.objects.select_related("books")
```

```sql
SELECT
   author.*,
   COALESCE((
       SELECT json_agg(json_build_object('id', id, 'title', title, 'author_id', author_id))
       FROM book
       WHERE author_id=author.id
    ), '[]'::json) books 
FROM author
```

Supporting all types of fields would require defining json serde (e.g. binary -> base64) but it doesn't seem like an inconceivable endeavour.

felixxm (commented):
@timb07 Thanks :+1:

PR #17418, opened 2023-10-28
34624 related field widget wrapper

allcaps:
@nessita I'd like to re-open https://github.com/django/django/pull/17158 but not sure how to do that. So, here is another PR.

I hope the test is sufficient. If not, let me know.


allcaps:
I see this fails and will pick it up again, when I have the time.

Strategy is to stop using Sites and Redirects, but create a ModelAdmin with a radio select. And test against that.

nessita:
Hello @allcaps, thank you for re-trying! I can re-open the other PR for you, will close this one since it's failing and most of the conversation is in the original PR.

PR #17419, opened 2023-10-28
Refs #34038 -- Improve link readability and identifiability

yushanfans2233:
#### Before
before-light

#### After
after-light

#### Before
before-dark

#### After
after-dark

### DIFF
--- a/django/contrib/admin/static/admin/css/base.css
+++ b/django/contrib/admin/static/admin/css/base.css
@@ -363,6 +363,7 @@ thead th {
 
 thead th a:link, thead th a:visited {
     color: var(--body-quiet-color);
+    text-decoration: none;
 }
 
 thead th.sorted {
@@ -707,6 +708,7 @@ div.breadcrumbs {
 
 div.breadcrumbs a {
     color: var(--breadcrumbs-link-fg);
+    text-decoration: none;
 }
 
 div.breadcrumbs a:focus, div.breadcrumbs a:hover {
@@ -720,9 +722,19 @@ div.breadcrumbs a:focus, div.breadcrumbs a:hover {
     background: url(../img/icon-viewlink.svg) 0 1px no-repeat;
 }
 
+.viewlink:link, .viewlink:visited, .inlineviewlink:link, .inlineviewlink:visited {
+    color: var(--body-fg);
+    text-decoration: none;
+}
+
 .hidelink {
     padding-left: 16px;
     background: url(../img/icon-hidelink.svg) 0 1px no-repeat;
+}
+
+.hidelink:link, .hidelink:visited{
+    color: var(--body-fg);
+    text-decoration: none;
 }
 
 .addlink {
@@ -730,14 +742,29 @@ div.breadcrumbs a:focus, div.breadcrumbs a:hover {
     background: url(../img/icon-addlink.svg) 0 1px no-repeat;
 }
 
+.addlink:link, .addlink:visited{
+    color: var(--body-fg);
+    text-decoration: none;
+}
+
 .changelink, .inlinechangelink {
     padding-left: 16px;
     background: url(../img/icon-changelink.svg) 0 1px no-repeat;
 }
 
+.changelink:link, .changelink:visited, .inlinechangelink:link, .inlinechangelink:visited {
+    color: var(--body-fg);
+    text-decoration: none;
+}
+
 .deletelink {
     padding-left: 16px;
     background: url(../img/icon-deletelink.svg) 0 1px no-repeat;
+}
+
+.deletelink:link, .deletelink:visited {
+    color: var(--body-fg);
+    text-decoration: none;
 }
 
 a.deletelink:link, a.deletelink:visited {
@@ -746,6 +773,16 @@ a.deletelink:link, a.deletelink:visited {
 
 a.deletelink:focus, a.deletelink:hover {
     color: #993333; /* XXX Probably unused? */
+    text-decoration: none;
+}
+
+.historylink:link, .historylink:visited{
+    color: var(--body-fg);
+    text-decoration: none;
+}
+
+.closelink:link, .closelink:visited{
+    color: var(--body-fg);
     text-decoration: none;
 }
 
@@ -928,6 +965,7 @@ a.deletelink:focus, a.deletelink:hover {
 
 #header a:link, #header a:visited, #logout-form button {
     color: var(--header-link-color);
+    text-decoration: none;
 }
 
 #header a:focus , #header a:hover {
@@ -1112,7 +1150,6 @@ a.deletelink:focus, a.deletelink:hover {
 .paginator a:link, .paginator a:visited {
     padding: 2px 6px;
     background: var(--button-bg);
-    text-decoration: none;
     color: var(--button-fg);
 }
 
--- a/django/contrib/admin/static/admin/css/base.css
+++ b/django/contrib/admin/static/admin/css/base.css
@@ -1148,6 +1148,7 @@ a.deletelink:focus, a.deletelink:hover {
 }
 
 .paginator a:link, .paginator a:visited {
+    text-decoration: none;
     padding: 2px 6px;
     background: var(--button-bg);
     color: var(--button-fg);
@@ -1157,6 +1158,10 @@ a.deletelink:focus, a.deletelink:hover {
     border: none;
     background: none;
     color: var(--link-fg);
+}
+
+.paginator a.showall:link, .paginator a.showall:visited{
+    text-decoration: underline;
 }
 
 .paginator a.showall:focus, .paginator a.showall:hover {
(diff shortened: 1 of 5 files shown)


sarahboyce:
If you like, you can add a screenshot within the `test_user_change_password` test (something like `self.take_screenshot("user_change")`) which will also add this page with light dark mode etc in our tests 👍

yushanfans2233:
@sarahboyce done, PTAL 😄

sarahboyce:
Thank you @yushanfans2233! This looks right but for some reason I can't see another screenshot generated 🤔 can you run the new selenium test with screenshots locally (https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/unit-tests/#running-the-selenium-tests) and see if it created screenshots?

yushanfans2233:
I ran the command `.\runtests.py --parallel=1 --selenium=chrome --screenshots`.
Then I see the following screenshots created:
- test_user_change_password_dark-user_change.png
- test_user_change_password_desktop_size-user_change.png
- test_user_change_password_mobile_size-user_change.png
- test_user_change_password_rtl-user_change.png

sarahboyce (approved):
Needs a review from the accessibility team but this looks good to me ⭐

yushanfans2233:
Thanks for your help, really appreciate it 😸 .

tests/admin_views/tests.py
```diff
@@ -6717,6 +6717,18 @@ def test_redirect_on_add_view_continue_button(self):
...
+    def test_user_change_password(self):
+        self.admin_login(
+            username="super", password="secret", login_url=reverse("admin:index")
+        )
+
+        change_url = reverse("admin:auth_user_change", args=(self.superuser.pk,))
+        with self.wait_page_loaded():
+            self.selenium.get(self.live_server_url + change_url)
+
+        self.take_screenshot("user_change")
```
felixxm, line 6730: This test doesn't have any assertions. Instead I'd add `take_screenshot` to one of existing tests, e.g. `UserAdminTest.test_save_button()`.

django/contrib/admin/static/admin/css/base.css
```diff
@@ -102,8 +102,8 @@ body {
 /* LINKS */
 
 a:link, a:visited {
-    color: var(--body-fg);
-    text-decoration: none;
+    color: var(--link-fg);
+    text-decoration: underline;
```
sabderemane, line 106: @knyghty @thibaudcolas I have one concern about this change, I haven't tested this change yet and can't see the screenshot on my phone, but it got me thinking about titles that can also be links, are they also considered underlined or just the content?

sabderemane (commented):
Looks good! Thank you for your contribution @yushanfans2233 😄
I just add a comment for the accessibility team

django/contrib/admin/static/admin/css/base.css
felixxm: As far as I'm aware, all links in the admin will be underlined, so this is a big UI change.

sabderemane: Argh, sorry I should have rephrase, I wanted to say : "do we want them also underlined or just the content?"
I don't know if it's already the case according the style all links in the admin will be underlined as you said, but I think we should agree on what we want before to merge this PR

knyghty (commented):
Thanks for the work here, it's definitely needed. Also sorry for the delay.

Clearly the accessibility of the links in the body is improved. I think we do need some overrides though. The main one is that buttons that act as links are currently underlined, e.g. the "delete" button and the "Add Foo" button on the list page as well as the view on site and history buttons.

There are some where I'm less certain. For example the title text (default "Django administration" probably doesn't _need_ to be underlined. I'd also argue that the links in the navigation sidebar don't really need to be underlined, either. But again, interested in a second opinion here.

I also see a link that should be underlined, but isn't:

Screenshot 2023-11-02 at 17 41 30

I think unrelated to this PR, but I also notice the log out link in the header has a fainter underline than the rest of the links, because it's using `border-bottom` rather than `text-decoration`:

Screenshot 2023-11-02 at 17 46 06

yushanfans2233:
Here are two alternative solutions:

### Solution 1
Apply an underline and blue color to all `<a>` tags, and use a CSS class to remove this style from specific locations where it is not needed.

- Pros: Ensures that no links requiring improvement are missed, even if new ones are added in the future.
- Cons: May require modifying a significant number of places.

### Solution 2
Introduce a new CSS class called `.a11y-link` and apply it to the links that require improved readability and identifiability.

- Pros: Provides more precise targeting without affecting existing elements.
- Cons: There is a potential risk of missing some existing links, as well as future additions, when attempting to identify and modify them all.

knyghty:
@yushanfans2233 personally I'd give option 1 a go. It's better to accidentally add an underline to something than it is to accidentally forget one, and if there are any they can be fixed over time as they're noticed.

yushanfans2233:
Hello everyone, I have made some necessary fixes based on the following principles:

1. To indicate its functionality in navigating to a new page or an anchor within the current page, the `<a>` element should be styled with color and underlining.

2. When the `<a>` element is utilized as a button or appears similar to a button, it should not be styled with color and underlining.

tests/admin_views/tests.py
yushanfans2233: `UserAdminTest` is not a `AdminSeleniumTestCase`, I can't take screenshot in it.

knyghty (commented):
One thing I'm not sure about. Maybe it's fine as it is or maybe not.

At the moment the filters don't have underline, and the only blue ones are the active filters. Clearly we need to show which filter is active. I'm just not sure which is best. Maybe it's fine how it is, but it is a bit hard to discern the text there are links.

Screenshot 2023-11-05 at 16 43 58

The Today and Now controls also need _something_ but not sure what. At the moment it's very unclear you can interact with them. But they don't really act like links to other pages, they are more like buttons. Again, not sure the best way forward here:

Screenshot 2023-11-05 at 16 47 40

yushanfans2233:
> One thing I'm not sure about. Maybe it's fine as it is or maybe not.
> 
> At the moment the filters don't have underline, and the only blue ones are the active filters. Clearly we need to show which filter is active. I'm just not sure which is best. Maybe it's fine how it is, but it is a bit hard to discern the text there are links.
> 
> Screenshot 2023-11-05 at 16 43 58
> The Today and Now controls also need _something_ but not sure what. At the moment it's very unclear you can interact with them. But they don't really act like links to other pages, they are more like buttons. Again, not sure the best way forward here:
> 
> Screenshot 2023-11-05 at 16 47 40

Yes, I completely agree with your viewpoint. It is important to enhance the visibility of these elements and make it clear to users that they are clickable functional elements. However, I believe we should create a new ticket to address this improvement as it is not directly related to the changes made in this PR.

knyghty:
For the sidebar filter, I can maybe agree, however there is a regression for the datetime inputs as this PR removes the colour. See this example from 4.2:

Screenshot 2023-11-06 at 14 40 19

yushanfans2233:
Yes, there is a regression on this link, but I don't believe it was introduced by my PR. I am running an admin site with the latest code from the Django main branch, and it does not display as blue.

image

If you want, I can fix it in this PR.

Keramblock:
Any news on this PR? We also need that =(

nessita:
> Any news on this PR? We also need that =(

As per the latest activity, it seems that this PR needs a review from the @django/accessibility team.

knyghty (changes requested):
We had a good chat about this at our latest @django/accessibility meeting.

We agreed that for now, we should just fix the colours and not add the underline.

We should still do the underline part, but as a new ticket, because:
- The changes can easily be done separately.
- Because of the difficulty in deciding what should and shouldn't be underlined, it's better to get the less contentious change out first.

nessita:
> We had a good chat about this at our latest @django/accessibility meeting.
> 
> We agreed that for now, we should just fix the colours and not add the underline.
> 
> We should still do the underline part, but as a new ticket, because:
> 
>     * The changes can easily be done separately.
> 
>     * Because of the difficulty in deciding what should and shouldn't be underlined, it's better to get the less contentious change out first.

Thank you @knyghty. I agree that separating the link color change from the underlining work is a good path forward. Therefore I recommend:

* File a new ticket for the color link regression, so I can triage it as a release blocker and we can provide the minimal fix suitable for backporting, and
* Potentially re-title the original [ticket #34038](https://code.djangoproject.com/ticket/34038) to focus on the underlines.

What do you think?

knyghty:
@nessita sounds good on my end

yushanfans2233:
Dear Django community,

I regret to inform you that due to my current busy schedule, I am unable to continue working on this PR for the time being. However, please know that I remain eager to assist if needed on any other project within the community.
So I suggest to close this PR and transfer the ticket to others.

Thank you for your understanding.

nessita:
> Dear Django community,
> 
> I regret to inform you that due to my current busy schedule, I am unable to continue working on this PR for the time being. However, please know that I remain eager to assist if needed on any other project within the community. So I suggest to close this PR and transfer the ticket to others.
> 
> Thank you for your understanding.

No worries @yushanfans2233, we appreciate everything you have done so far! Thank you for letting us know.

PR #17438, opened 2023-11-02
Fixed #34830 -- Added request to csrf_failure view's template context.

prakhar599:
Ticket: https://code.djangoproject.com/ticket/34830
Added 'request' context processor to **_'csrf_failure'_** function

### DIFF
--- a/tests/view_tests/tests/test_csrf.py
+++ b/tests/view_tests/tests/test_csrf.py
@@ -131,3 +131,7 @@ class CsrfViewTests(SimpleTestCase):
         with mock.patch.object(Path, "open") as m:
             csrf_failure(mock.MagicMock(), mock.Mock())
             m.assert_called_once_with(encoding="utf-8")
+
+    def test_csrf_response_has_request_context_processor(self):
+        response = self.client.post("/")
+        self.assertIs(response.wsgi_request, response.context.get("request"))
--- a/django/views/csrf.py
+++ b/django/views/csrf.py
@@ -64,6 +64,7 @@ def csrf_failure(request, reason="", template_name=CSRF_FAILURE_TEMPLATE_NAME):
         "DEBUG": settings.DEBUG,
         "docs_version": get_docs_version(),
         "more": _("More information is available with DEBUG=True."),
+        "request": request,
     }
     try:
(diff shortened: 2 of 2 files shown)


smithdc1:
@prakhar599 -- Thanks for the patch. 👍 

Please can you also write a regression test for this? I'd suggest adding tests to https://github.com/django/django/blob/main/tests/view_tests/tests/test_csrf.py

prakhar599:
@smithdc1 Thanks for response and I've just added a regression test commit to my patch.

tests/view_tests/tests/test_csrf.py
```diff
@@ -131,3 +131,7 @@ def test_template_encoding(self):
         with mock.patch.object(Path, "open") as m:
             csrf_failure(mock.MagicMock(), mock.Mock())
             m.assert_called_once_with(encoding="utf-8")
+
+    def test_csrf_response_has_request_context_processor(self):
+        response = self.client.post("/")
+        self.assertIn("request", response.context)
```
nessita, line 137: Thank you for adding a test. I would suggest we make a stronger assertion here, something like:
```suggestion
        self.assertIs(response.wsgi_request, response.context.get("request"))
```

Which ensures both that the `request` key is provided:
```
AssertionError: <WSGIRequest: POST '/'> is not None
```
and that it has the correct value:
```
AssertionError: <WSGIRequest: POST '/'> is not <object object at 0x7f28ced59fe0>
```

nessita (commented):
Looks good, thanks!

Besides the minor test fix I suggested, I was wondering if we should document this somewhere, though I couldn't find the right place. The setting documentation for `CSRF_FAILURE_VIEW` is about customizing the *view* itself, and the CSRF reference does not mention the ability to customize only the template, so considering that the existing context content is undocumented, I think this does not need a doc change.

prakhar599:
@nessita Yes I agree `assertIs`  will ensure that the "request" variable in the response context is the same object as response.wsgi_request, This assertion `(self.assertIs(response.wsgi_request, response.context.get("request"))` looks more appropriate. However, I was also looking for appropriate place to document it but couldn't find any. Thanks for the Test fix though.

PR #17444, opened 2023-11-05
Clarified a few points about DATABASES.TIME_ZONE

shangxiao:
(as per the commit message)

I felt the `DATABASE.TIME_ZONE` description was leaving out a few important points:

1. The default used is UTC if the value is set to `None` & `USE_TZ=True`. This is in the time zone section of the docs but would be good to explicitly state it here.
2. When set & using postgres, it sets the database connection's `TIME ZONE` setting (I got that we were trying to be general in the description so I just said "database connection's time zone")
3. In my experience I found it's wise to set this to the same as the general `TIME_ZONE` if a.) your `TIME_ZONE` is something other than UTC and b.) you're using date/time processing in raw sql
4. While you can use `date_trunc()` with UTC set as it has a param to tell it what time zone to use, there's no such equivalent param for the `generate_series()` function. I found if you generate a series transitioning daylight savings then the only way to get it to work 100% is to set the connection's `TIME ZONE`. I had this happen to me when generating daily totals of data that went across a daylight savings transition 🙄

### DIFF
--- a/docs/ref/settings.txt
+++ b/docs/ref/settings.txt
@@ -663,8 +663,10 @@ A string representing the time zone for this database connection or ``None``.
 This inner option of the :setting:`DATABASES` setting accepts the same values
 as the general :setting:`TIME_ZONE` setting.
 
-When :setting:`USE_TZ` is ``True`` and this option is set, reading datetimes
-from the database returns aware datetimes in this time zone instead of UTC.
+When :setting:`USE_TZ` is ``True``, reading datetimes from the database
+returns aware datetimes with the timezone set to this option's value if not
+``None``, or to UTC otherwise.
+
 When :setting:`USE_TZ` is ``False``, it is an error to set this option.
 
 * If the database backend doesn't support time zones (e.g. SQLite, MySQL,
@@ -687,13 +689,18 @@ When :setting:`USE_TZ` is ``False``, it is an error to set this option.
     third-party systems connect to the same database and expect to find
     datetimes in local time, then you must set this option.
 
-* If the database backend supports time zones (e.g. PostgreSQL), the
-  ``TIME_ZONE`` option is very rarely needed. It can be changed at any time;
-  the database takes care of converting datetimes to the desired time zone.
+* If the database backend supports time zones (e.g., PostgreSQL), then the
+  database connection's time zone is set to this value.
 
-  Setting the time zone of the database connection may be useful for running
-  raw SQL queries involving date/time functions provided by the database, such
-  as ``date_trunc``, because their results depend on the time zone.
+  Although setting the ``TIME_ZONE`` option is very rarely needed, there are
+  situations where it becomes necessary. Specifically, it's recommended to
+  match the general :setting:`TIME_ZONE` setting when dealing with raw queries
+  involving date/time functions like PostgreSQL's ``date_trunc()`` or
+  ``generate_series()``, especially when generating time-based series that
+  transition daylight savings.
+
+  This value can be changed at any time, the database will handle the
+  conversion of datetimes to the configured time zone.
 
   However, this has a downside: receiving all datetimes in local time makes
   datetime arithmetic more tricky — you must account for possible offset


docs/ref/settings.txt
```diff
@@ -687,13 +689,16 @@ When :setting:`USE_TZ` is ``False``, it is an error to set this option.
     third-party systems connect to the same database and expect to find
     datetimes in local time, then you must set this option.
 
-* If the database backend supports time zones (e.g. PostgreSQL), the
-  ``TIME_ZONE`` option is very rarely needed. It can be changed at any time;
-  the database takes care of converting datetimes to the desired time zone.
+* If the database backend supports time zones (e.g. PostgreSQL) then the
+  database connection's time zone set to this value.
```
smithdc1, line 693:
```suggestion
* If the database backend supports time zones (e.g., PostgreSQL), then the
  database connection's time zone is set to this value.
```

TIL.
e.g. meaning for example is always followed by a comma.

shangxiao: Thanks missed a couple of other bits too 👍

docs/ref/settings.txt
```diff
@@ -665,6 +665,8 @@ as the general :setting:`TIME_ZONE` setting.
 
 When :setting:`USE_TZ` is ``True`` and this option is set, reading datetimes
 from the database returns aware datetimes in this time zone instead of UTC.
+When this option is ``None`` then UTC will be used.
```
nessita, line 668: I value the clarification but I still have questions (I think I now the answers, but I fear other readers may be confused):

1. UTC will be used and the returned timezones will be aware or not?
2. `this option` refers to `USE_TZ` or `TIME_ZONE`?

So I would suggest instead:

```suggestion
When :setting:`USE_TZ` is ``True``, reading datetimes from the database
returns aware datetimes.  If this option is set, the returned datetimes will
be in this time zone, otherwise UTC will be used.
```

docs/ref/settings.txt
```diff
@@ -687,13 +689,16 @@ When :setting:`USE_TZ` is ``False``, it is an error to set this option.
     third-party systems connect to the same database and expect to find
     datetimes in local time, then you must set this option.
 
-* If the database backend supports time zones (e.g. PostgreSQL), the
-  ``TIME_ZONE`` option is very rarely needed. It can be changed at any time;
-  the database takes care of converting datetimes to the desired time zone.
+* If the database backend supports time zones (e.g., PostgreSQL), then the
```
nessita, line 692: I personally think that this comma does belong here, though I see that in some other places this pattern is used. Though from 981 "e.g." in the `django` folder, 41 are followed by a comma, and from 457 in the `docs` folder, 84 have the following comma.

Is this an English [from a specific region?] style pattern that I'm missing?

```suggestion
* If the database backend supports time zones (e.g. PostgreSQL), then the
```

docs/ref/settings.txt
```diff
@@ -687,13 +689,16 @@ When :setting:`USE_TZ` is ``False``, it is an error to set this option.
...
-  Setting the time zone of the database connection may be useful for running
-  raw SQL queries involving date/time functions provided by the database, such
-  as ``date_trunc``, because their results depend on the time zone.
+  Although setting the ``TIME_ZONE`` option is very rarely needed it may be
+  necessary to set this to the same value as the general :setting:`TIME_ZONE`
+  setting when raw queries that involve date/time functions such as
+  PostgreSQL's ``date_trunc()`` or ``generate_series()`` when generating
+  time-based series that transition daylight savings. It can be changed at any
+  time; the database takes care of converting datetimes to the desired time
+  zone.
```
nessita, line 701: This fragment seems like it has a grammatical issue: "_when raw queries that involve date/time functions ... when generating time-based series that transition daylight savings_"

Did you mean "_when raw queries that involve date/time functions ... generate time-based series that transition daylight savings_"?

What about:
```suggestion
  Although setting the ``TIME_ZONE`` option is very rarely needed, there are
  situations where it becomes necessary. Specifically, it should match the
  general :setting:`TIME_ZONE` setting when dealing with raw queries involving
  date/time functions like PostgreSQL's ``date_trunc()`` or
  ``generate_series()``, especially when generating time-based series that
  transition daylight savings.

  This value can be changed it at any time, the database will handle
  the conversion of datetimes to the configured time zone.
```

nessita (commented):
Thank you David! I made some suggestions, if you can't work on this I'm happy to push edits and merge. Let me know!

docs/ref/settings.txt
shangxiao: 1. Yep with `USE_TZ` then returned timestamps will be aware 
2. `TIME_ZONE`

The suggestion is ok but I'm not sure I think it's "better" 🤔 because it brings the "aware" part to the front which is secondary.

I guess the challenge here is explaining this pseudocode in natural language:

```
if USE_TZ = False:
   error!

else:
   timezone = DATABASES.TIME_ZONE
   if timezone is None:
        timezone = 'UTC'
```

The fact that timestamps are "aware" is secondary/implied if that makes sense? 🤔 

If you think that the suggestion makes that ^ pseudocode clear then I'm ok.

docs/ref/settings.txt
shangxiao: I don't see that pattern but it could just be my country🤷‍♂️

docs/ref/settings.txt
shangxiao: lol I just re-read my sentence and it definitely is a challenge because it's too long and has "… when … when …" which is bad.

I like how you split it up 👍 

Re "Specifically, it should match … " I tried to avoid "should" style because it's not a hard rule.  I was going for "may be necessary" or "recommended". Otherwise your suggestion is perfect!

docs/ref/settings.txt
nessita: Thank you! I gotta confess, as I web service developer, I think I first try to confirm whether timestamps will be aware or not, and then, if "yes", I seek information about what timezones they will be in. But, OTOH, the first question should be answered (or seek answers in) the details of the `USE_TZ` setting, so I can see where you are coming from :-).

How about:

```suggestion
When :setting:`USE_TZ` is ``True``, reading datetimes from the database
returns aware datetimes with the timezone set to the value specified in
this option if not ``None``, or to ``UTC`` otherwise.
```

docs/ref/settings.txt
nessita: Ok! So, how about:

```suggestion
  Although setting the ``TIME_ZONE`` option is very rarely needed, there are
  situations where it becomes necessary. Specifically, it's recommended to
  match the general :setting:`TIME_ZONE` setting when dealing with raw queries
  involving date/time functions like PostgreSQL's ``date_trunc()`` or
  ``generate_series()``, especially when generating time-based series that
  transition daylight savings.

  This value can be changed it at any time, the database will handle
  the conversion of datetimes to the configured time zone.
```

nessita:
@shangxiao @smithdc1 
I pushed a rebase with a separated commit with my suggestions, we can keep iterating or if you both are happy, I can squash and merge. Thank you!

nessita:
Fixed a minor grammar error, rebasing and merging once CI is green.

PR #17448, opened 2023-11-05
Fixed #34457 -- Restored output for makemigrations --check.

adamchainz:
ticket-34457

### DIFF
--- a/tests/migrations/test_commands.py
+++ b/tests/migrations/test_commands.py
@@ -2392,20 +2392,35 @@ class MakeMigrationsTests(MigrationTestBase):
                 "makemigrations", "migrations", "--name", "invalid name", "--empty"
             )
 
-    def test_makemigrations_check(self):
+    def test_makemigrations_check_with_changes(self):
         """
         makemigrations --check should exit with a non-zero status when
         there are changes to an app requiring migrations.
         """
+        out = io.StringIO()
         with self.temporary_migration_module() as tmpdir:
-            with self.assertRaises(SystemExit):
-                call_command("makemigrations", "--check", "migrations", verbosity=0)
-            self.assertFalse(os.path.exists(tmpdir))
+            with self.assertRaises(SystemExit) as cm:
+                call_command(
+                    "makemigrations",
+                    "--check",
+                    "migrations",
+                    stdout=out,
+                )
+            self.assertEqual(os.listdir(tmpdir), ["__init__.py"])
+        self.assertEqual(cm.exception.code, 1)
+        self.assertIn("Migrations for 'migrations':", out.getvalue())
 
+    def test_makemigrations_check_no_changes(self):
+        """
+        makemigrations --check should exit with a zero status when there are no
+        changes.
+        """
+        out = io.StringIO()
         with self.temporary_migration_module(
             module="migrations.test_migrations_no_changes"
         ):
-            call_command("makemigrations", "--check", "migrations", verbosity=0)
+            call_command("makemigrations", "--check", "migrations", stdout=out)
+        self.assertEqual("No changes detected in app 'migrations'\n", out.getvalue())
 
     def test_makemigrations_migration_path_output(self):
         """
--- a/django/core/management/commands/makemigrations.py
+++ b/django/core/management/commands/makemigrations.py
@@ -72,7 +72,7 @@ class Command(BaseCommand):
             dest="check_changes",
             help=(
                 "Exit with a non-zero status if model changes are missing migrations "
-                "and don't actually write them."
+                "and don't actually write them. Implies --dry-run."
             ),
         )
         parser.add_argument(
@@ -114,6 +114,8 @@ class Command(BaseCommand):
             raise CommandError("The migration name must be a valid Python identifier.")
         self.include_header = options["include_header"]
         check_changes = options["check_changes"]
+        if check_changes:
+            self.dry_run = True
         self.scriptable = options["scriptable"]
         self.update = options["update"]
         # If logs and prompts are diverted to stderr, remove the ERROR style.
@@ -251,12 +253,12 @@ class Command(BaseCommand):
                 else:
                     self.log("No changes detected")
         else:
-            if check_changes:
-                sys.exit(1)
             if self.update:
                 self.write_to_last_migration_files(changes)
             else:
                 self.write_migration_files(changes)
+            if check_changes:
+                sys.exit(1)
 
     def write_to_last_migration_files(self, changes):
         loader = MigrationLoader(connections[DEFAULT_DB_ALIAS])
--- a/docs/ref/django-admin.txt
+++ b/docs/ref/django-admin.txt
@@ -742,7 +742,7 @@ Generate migration files without Django version and timestamp header.
 .. django-admin-option:: --check
 
 Makes ``makemigrations`` exit with a non-zero status when model changes without
-migrations are detected.
+migrations are detected. Implies ``--dry-run``.
 
 .. django-admin-option:: --scriptable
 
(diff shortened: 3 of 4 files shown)


jacobtylerwalls (approved):
Thanks, this is an elegant solution for the original intent to avoid writing the migration files to disk. Sorry @shangxiao for letting the discussion die off on your PR!

pauloxnet (approved):
LGTM

shangxiao:
In my PR I added "implies --dry-run" to the check help text, is that something we may want to consider?

Also… do we need a doc update?  (because atm it just says "Makes makemigrations exit with a non-zero status when model changes without migrations are detected." might be worth saying it doesn't change anything or "implies dry-run")

I was thinking whether merging `--check` & `--dry-run` was an option now because they're identical except for return value but there may still be use cases where `--dry-run` is used but the context requires a 0 (ie shell scripts with set -e or whatever)

tests/migrations/test_commands.py
```diff
@@ -2397,15 +2397,27 @@ def test_makemigrations_check(self):
...
+                "makemigrations",
+                "--check",
+                "migrations",
+                stdout=out,
+            )
+        self.assertEqual(cm.exception.code, 1)
+        self.assertIn("Migrations for 'migrations':", out.getvalue())
+        self.assertFalse(os.path.exists(tmpdir))
 
+        out = io.StringIO()
```
nessita, line 2415: While I acknowledge that the current `main` have both checks in the same unit test, I would very much like us to use this opportunity to split the tests in two:
 * test_makemigrations_check_with_changes
 * test_makemigrations_check_no_changes

What do you think @adamchainz?

tests/migrations/test_commands.py
```diff
@@ -2397,15 +2397,27 @@ def test_makemigrations_check(self):
...
         with self.temporary_migration_module(
             module="migrations.test_migrations_no_changes"
         ):
-            call_command("makemigrations", "--check", "migrations", verbosity=0)
+            call_command("makemigrations", "--check", "migrations", stdout=out)
+        self.assertEqual(out.getvalue(), "No changes detected in app 'migrations'\n")
```
nessita, line 2420: Could we please check that the status code here to be `0`?

nessita (commented):
Thank you Adam, this looks good.

I added comments and agree with @shangxiao that it's important to mention in the `--check` help message that the `--dry-run` flag is implied. This provides clarity and assures Django users that the command is safe and won't have any side effects, given the universal understanding of `dry-run` as a safe operation.

tests/migrations/test_commands.py
adamchainz: Agree, I like to split unit tests.

tests/migrations/test_commands.py
adamchainz: aye aye

adamchainz: Actually, there’s no way to add an explicit assertion: the lack of a `SystemExit` implies the command would exit with a zero status code on the command line.

adamchainz:
Okay, made the update. I added “Implies `--dry-run`.” to both the help text and documentation, and split the tests.

tests/migrations/test_commands.py
```diff
@@ -2392,20 +2392,37 @@ def test_makemigrations_with_invalid_custom_name(self):
...
+        ):
+            call_command(
+                "makemigrations",
+                "--check",
+                "migrations",
+                stdout=out,
+            )
+        self.assertEqual(cm.exception.code, 1)
+        self.assertIn("Migrations for 'migrations':", out.getvalue())
+        self.assertFalse(os.path.exists(tmpdir))
```
nessita, line 2413: This one got me puzzled for a while: while conducting the review, I did the "usual" and I changed minimal parts of the diff to ensure a test would fail. Specifically, I changed:
```diff
--- a/django/core/management/commands/makemigrations.py
+++ b/django/core/management/commands/makemigrations.py
@@ -115,7 +115,7 @@ class Command(BaseCommand):
         self.include_header = options["include_header"]
         check_changes = options["check_changes"]
         if check_changes:
-            self.dry_run = True
+            self.dry_run = False
         self.scriptable = options["scriptable"]
         self.update = options["update"]
         # If logs and prompts are diverted to stderr, remove the ERROR style.
```
and I ran the test suite, confidently expecting that the `test_makemigrations_check_with_changes` would fail. IT DID NOT! :scream: 

After the shock, I started debugging and noticed that `self.assertFalse(os.path.exists(tmpdir))` will always hold after exiting the `self.temporary_migration_module` context manager because the `tmpdir` folder is deleted on exit (this is explained by the usage of `tempfile.TemporaryDirectory()` as context manager in the implementation of `temporary_migration_module`).

So I said "well, we *just* need to move the assert to inside the `with`  block and we are golden. Guess what? it was not enough! I also had to split the multiple `with` managers into two nested calls, I think this is because the raising of the `SystemExit` is preventing the execution of the assert. So in the end I think this is the correct test:

```diff
--- a/tests/migrations/test_commands.py
+++ b/tests/migrations/test_commands.py
@@ -2398,19 +2398,17 @@ class MakeMigrationsTests(MigrationTestBase):
         there are changes to an app requiring migrations.
         """
         out = io.StringIO()
-        with (
-            self.temporary_migration_module() as tmpdir,
-            self.assertRaises(SystemExit) as cm,
-        ):
-            call_command(
-                "makemigrations",
-                "--check",
-                "migrations",
-                stdout=out,
-            )
+        with self.temporary_migration_module() as tmpdir:
+            with self.assertRaises(SystemExit) as cm:
+                call_command(
+                    "makemigrations",
+                    "--check",
+                    "migrations",
+                    stdout=out,
+                )
+            self.assertFalse(os.path.exists(tmpdir))
         self.assertEqual(cm.exception.code, 1)
         self.assertIn("Migrations for 'migrations':", out.getvalue())
-        self.assertFalse(os.path.exists(tmpdir))
 
     def test_makemigrations_check_no_changes(self):
         """
```

nessita (changes requested):
Thanks! Found an issue with the test, added details as inline comments.

tests/migrations/test_commands.py
adamchainz: Yes, that’s absolutely correct. Thanks for checking. I’ve applied your patch and added you as coauthor.

adamchainz: One adjustment was actually needed, since `MigrationWriter.basedir` always creates the directory and `__init__.py`, even in dry-run mode:

```diff
diff --git tests/migrations/test_commands.py tests/migrations/test_commands.py
index dece54f509..d9ca30b002 100644
--- tests/migrations/test_commands.py
+++ tests/migrations/test_commands.py
@@ -2406,7 +2406,7 @@ def test_makemigrations_check_with_changes(self):
                     "migrations",
                     stdout=out,
                 )
-            self.assertFalse(os.path.exists(tmpdir))
+            self.assertEqual(os.listdir(tmpdir), ["__init__.py"])
         self.assertEqual(cm.exception.code, 1)
         self.assertIn("Migrations for 'migrations':", out.getvalue())

```

Perhaps we should consider this a separate bug, since it could prevent a clean run of `makemigrations --check` in the niche scenario where the migration directory doesn't exist.

nessita: Thanks @adamchainz! I agree that the creation of the folder + `__init__.py` file is unexpected when running with `dry-run`. Do you have time to the ticket?

nessita (approved):
Thank you! Approving.

I pushed a small edit to be consistent in the param ordering when asserting over the returned output (using `self.assertEqual(expected, out.getvalue())`)

PR #17449, opened 2023-11-06
Arrange code so we do not have temporarily undefined classes

AlanCoding:
We have been hitting a core dump as uWSGI expires workers, where the stack trace hints at a "double free" type of error. This only happens with the default preforking behavior of uWSGI, and in some other environmental conditions that remain unclear to me.

This is one of a few changes I've identified that, combined, resolve this. Another one is https://github.com/jazzband/django-oauth-toolkit/pull/1357

This also looks like better practice to me. While I realize there are some times we have to use a variable that won't be defined until later in the file, I cannot find a reason that needs to be the case here.

### DIFF
--- a/django/contrib/postgres/fields/hstore.py
+++ b/django/contrib/postgres/fields/hstore.py
@@ -10,6 +10,26 @@ from django.utils.translation import gettext_lazy as _
 __all__ = ["HStoreField"]
 
 
+class KeyTransform(Transform):
+    output_field = TextField()
+
+    def __init__(self, key_name, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.key_name = key_name
+
+    def as_sql(self, compiler, connection):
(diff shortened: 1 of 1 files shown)


charettes:
I'm not sure how this is meant to help at all. 

`HStoreField.get_transform` cannot be called until the module is fully loaded anyway so neither `KeyTransformFactory` or `KeyTransform` are _temporary_ by any means. I don't see how organizing it this way will help unless there's a pretty critical bug in uWSGI in how it loads modules. In other words, `KeyTransformFactory` is resolved at run time and is not _bound_ at module definition time so the ordering should not matter.

The source of Django contains tens (if not hundreds) of instances of objects being defined this way and Python doesn't offer any way to test against this pattern so even if we wanted to merge this change we can't have a guarantee that we wouldn't regress somehow. There are cases where it just impossible to even attempt such as when dealing with cycles of references between objects.

PR #17452, opened 2023-11-07
Fixed #34952 -- Copied dir list when processing locale folders to avoid missing entries during os.walk traversal.

cordery:
Fix for ticket #34952 https://code.djangoproject.com/ticket/34952#ticket

Also referenced by ticket #34925 https://code.djangoproject.com/ticket/34925.

### DIFF
--- a/django/core/management/commands/compilemessages.py
+++ b/django/core/management/commands/compilemessages.py
@@ -91,7 +91,8 @@ class Command(BaseCommand):
 
         # Walk entire tree, looking for locale directories
         for dirpath, dirnames, filenames in os.walk(".", topdown=True):
-            for dirname in dirnames:
+            # As we may modify dirnames, iterate through a copy of it instead
+            for dirname in list(dirnames):
                 if is_ignored_path(
                     os.path.normpath(os.path.join(dirpath, dirname)), ignore_patterns
                 ):
--- a/tests/i18n/test_compilation.py
+++ b/tests/i18n/test_compilation.py
@@ -195,6 +195,64 @@ class IgnoreDirectoryCompilationTests(MessageCompilationTests):
         self.assertNoneExist(self.CACHE_DIR, ["en", "fr", "it"])
         self.assertNoneExist(self.NESTED_DIR, ["en", "fr", "it"])
 
+    def test_no_dirs_accidentally_skipped(self):
+        os_walk_results = [
+            # To discover .po filepaths, compilemessages uses with a starting list of
+            # basedirs to inspect, which in this scenario are:
+            #   ["conf/locale", "locale"]
+            # Then os.walk is used to discover other locale dirs, ignoring dirs matching
+            # `ignore_patterns`. Mock the results to place an ignored directory directly
+            # before and after a directory named "locale".
+            [("somedir", ["ignore", "locale", "ignore"], [])],
+            # This will result in three basedirs discovered:
+            #   ["conf/locale", "locale", "somedir/locale"]
+            # os.walk is called for each locale in each basedir looking for .po files.
+            # In this scenario, we need to mock os.walk results for "en", "fr", and "it"
+            # locales for each basedir:
+            [("exclude/locale/LC_MESSAGES", [], ["en.po"])],
+            [("exclude/locale/LC_MESSAGES", [], ["fr.po"])],
+            [("exclude/locale/LC_MESSAGES", [], ["it.po"])],
+            [("exclude/conf/locale/LC_MESSAGES", [], ["en.po"])],
+            [("exclude/conf/locale/LC_MESSAGES", [], ["fr.po"])],
+            [("exclude/conf/locale/LC_MESSAGES", [], ["it.po"])],
+            [("exclude/somedir/locale/LC_MESSAGES", [], ["en.po"])],
+            [("exclude/somedir/locale/LC_MESSAGES", [], ["fr.po"])],
+            [("exclude/somedir/locale/LC_MESSAGES", [], ["it.po"])],
+        ]
+
+        module_path = "django.core.management.commands.compilemessages"
+        with mock.patch(f"{module_path}.os.walk", side_effect=os_walk_results):
+            with mock.patch(f"{module_path}.os.path.isdir", return_value=True):
+                with mock.patch(
+                    f"{module_path}.Command.compile_messages"
+                ) as mock_compile_messages:
+                    call_command("compilemessages", ignore=["ignore"], verbosity=4)
+
+        expected = [
+            (
+                [
+                    ("exclude/locale/LC_MESSAGES", "en.po"),
+                    ("exclude/locale/LC_MESSAGES", "fr.po"),
+                    ("exclude/locale/LC_MESSAGES", "it.po"),
+                ],
+            ),
+            (
+                [
+                    ("exclude/conf/locale/LC_MESSAGES", "en.po"),
+                    ("exclude/conf/locale/LC_MESSAGES", "fr.po"),
+                    ("exclude/conf/locale/LC_MESSAGES", "it.po"),
+                ],
+            ),
+            (
+                [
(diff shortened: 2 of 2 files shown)


tests/i18n/test_compilation.py
```diff
@@ -195,6 +195,66 @@ def test_ignores_based_on_pattern(self):
...
+        locale_dir = base_dir / "locale"
+        copytree("canned_locale", locale_dir)
+
+        # Check that the '_._skipme' directory comes just before the 'locale' directory
+        dirlist = [x.name for x in os.scandir(base_dir)]
+        if dirlist.index("locale") - dirlist.index(skip_dir.name) != 1:
+            raise SkipTest(
+                "Could not generate a directory that appears just before `locale`, "
+                "cannot perform test."
+            )
```
shangxiao, line 246: Yeah this is a bit tricky innit. Mariusz & Natalia will have the final say but this may just be one of the rare times I recommend mocking in order to guarantee a deterministic test 🤔

django/core/management/commands/compilemessages.py
```diff
@@ -91,7 +91,8 @@ def handle(self, **options):
 
         # Walk entire tree, looking for locale directories
         for dirpath, dirnames, filenames in os.walk(".", topdown=True):
-            for dirname in dirnames:
+            # As we may modify dirnames, iterate through a copy of it instead
+            for dirname in list(dirnames):
```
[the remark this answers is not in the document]
cordery: My thoughts exactly :)

django/core/management/commands/compilemessages.py
```diff
@@ -91,7 +91,8 @@ def handle(self, **options):
 
         # Walk entire tree, looking for locale directories
         for dirpath, dirnames, filenames in os.walk(".", topdown=True):
-            for dirname in dirnames:
+            # As we may modify dirnames, iterate through a copy of it instead
```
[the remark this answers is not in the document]
cordery: Thanks, caught it when I was investigating the bug that actually bit me from the other ticket.  Wondered if there were more examples and found this after some simple regex through the django codebase.

tests/i18n/test_compilation.py
cordery: Yeah, I notice that it Skipped when run on the windows CI check, meaning it couldn't create a folder than came before locale.   Not sure if there is some other name trick that would reliably come first on windows.  

I could mock it, but at some point I'm essentially just asserting that this is how lists work in python:

```python
In [1]: x=[1,2,3,4,5,6,7,8,9,10]

In [2]: for i in x:
   ...:      print(i)
   ...:      x.remove(i)
   ...: 
1
3
5
7
9
```

To be honest, I feel like I'm attempting to explain and write tests to prove something that is essentially a fundamental programming mistake (a common and easy one to make).  The effort involved feels like it exceeds the 'risk' of this kind of patch in my opinion.  It is an obvious bug, I have proved that it does in fact cause real world consequences, the solution is trivial and requires no change to the actual logic of the function.  

I don't think it's enormously worth having somewhat fragile tests that spend a good deal of effort proving why you shouldn't remove items from the list you are iterating through.  Likewise, you are having to read and review all of this to approve a safe 6 character obvious fix to an obvious problem.  The sort of thing that a linter rule could probably provide an auto fix for (if the pylint rule actually worked).  That said, I'm happy to make whatever reasonable changes to this PR are required.

nessita: It took me a while but here is a test that I find readable while ensuring exactly that the whole list of dirnames is traversed (removing the call to `list()` makes the test fail). I'm not proud of using 3 mock calls, perhaps we could create the relevant folder to avoid mocking `isdir`, but the other two IMHO are necessary to keep the test simple-ish:

(EDIT: I re-worked the test a bit after submitting this msg, so the test in the GH email will differ from this one)

```python
    def test_no_dirs_accidentally_skipped(self):
        os_walk_results = [
            # First os.walk call, used to build `basedirs` searching for `locale` dirs
            # and ignoring paths based on `ignore_patterns`
            [("somedir", ["ignore", "locale", "ignore"], [])],
        ]

        locales = ["en", "fr", "it"]
        os_walk_results.extend(
            # Remaining three os.walk calls, used from within the loop iterating over
            # `basedirs`, which in this scenario has 3 entries:
            # exclude/locale, exclude/conf/locale, exclude/somedir/locale
            [(f"dirpath-{d}", [], [f"{locale}.po"])]
            for d in range(3)
            for locale in locales
        )

        import_module = "django.core.management.commands.compilemessages"
        with mock.patch(f"{import_module}.os.walk", side_effect=os_walk_results):
            with mock.patch(f"{import_module}.os.path.isdir", return_value=True):
                with mock.patch(
                    f"{import_module}.Command.compile_messages"
                ) as mock_compile_messages:
                    call_command("compilemessages", ignore=["ignore"], verbosity=4)

        expected = [
            mock.call([(f"dirpath-{d}", f"{locale}.po") for locale in locales])
            for d in range(3)
        ]
        self.assertEqual(mock_compile_messages.mock_calls, expected)
```
@shangxiao @cordery Opinions?

nessita (commented):
Looks good, added comments and suggestions!

tests/i18n/test_compilation.py
shangxiao: Nice 👍  💯 re "the key to the test that we need to add is ensuring that if someone accidentally removes the list() call, a test barfs"

Mocking seems like the only way to achieve that here. We're always asking people to avoid mocking so this is one of those exceptional "Although practicality beats purity." type moments.

Aside from that, this is just my personal opinion but I usually recommend to colleagues that meta-programming be avoided in tests when you can simply statically declare the fixtures as it can improve readability (unless they're too big to read):

```python
        os_walk_results = [
            # First os.walk call, used to build `basedirs` searching for `locale` dirs
            # and ignoring paths based on `ignore_patterns`
            [("somedir", ["ignore", "locale", "ignore"], [])],
            # Remaining three os.walk calls, used from within the loop iterating over
            # `basedirs`, which in this scenario has 3 entries:
            # exclude/locale, exclude/conf/locale, exclude/somedir/locale    
            [('dirpath-1', [], ['en.po'])],
            [('dirpath-1', [], ['fr.po'])],
            [('dirpath-1', [], ['it.po'])],
            [('dirpath-2', [], ['en.po'])],
            [('dirpath-2', [], ['fr.po'])],
            [('dirpath-2', [], ['it.po'])]]
        ]
```
This way people don't need to turn on their brain's Python interpreter to read the test :)

nessita: > Nice 👍 💯 re "the key to the test that we need to add is ensuring that if someone accidentally removes the list() call, a test barfs"

First of all, sorry for my confusion, I meant that "a test barks" not that "a test barfs" :nauseated_face: 
Though perhaps one could create a somewhat disgusting analogy between a test failure and the latter :rofl: 

> Aside from that, this is just my personal opinion but I usually recommend to colleagues that ~meta-programming~ dynamic fixture construction be avoided in tests when you can simply statically declare the fixtures

I generally agree with this, though there is a delicate balance between readability and making the test as straightforward as possible. In this case, I can't fully decide, defining the lists through expansion will make the test very long :thinking:. I guess we could try to:
* remove the mock for `isdir` and only create the `somedir` folder (so other calls to `isdir` are False and then we'd need to mock less results for `os.walk`)
* have the `os.walk` results defined through expansion

@cordery Would you have time to work on this? If not I'll pick it up soon-ish. Thanks everyone!

cordery: > First of all, sorry for my confusion, I meant that "a test barks" not that "a test barfs" 🤢
> Though perhaps one could create a somewhat disgusting analogy between a test failure and the latter 🤣
lol, I didn't even blink, it seemed like a totally valid expression to use for a test failure!

Sure thing, thank you for doing the lions share already!  You are both heroes.  I'll rework it this evening.

nessita (approved):

PR #17457, opened 2023-11-08
Docs: Added a hyphen to "database-generated"

bcail:


nessita:
Hey @bcail, "database generated" is a correct term, I don't see a gain in making the change you are suggesting. And changing strings at this point would increase the translators works, so I'll be closing for now, but let me know if there is another reason for this change that I'm missing.

PR #17466, opened 2023-11-11
Fixed #34964 -- Avoided migrations when combining or inverting the order of CheckConstraint conditions.

jacobtylerwalls:
ticket-34964

### DIFF
--- a/django/db/models/query_utils.py
+++ b/django/db/models/query_utils.py
@@ -157,7 +157,7 @@ class Q(tree.Node):
     def identity(self):
         path, args, kwargs = self.deconstruct()
         identity = [path, *kwargs.items()]
-        for child in args:
+        for child in sorted(args, key=str):
             if isinstance(child, tuple):
                 arg, value = child
                 value = make_hashable(value)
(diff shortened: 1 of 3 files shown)


django/db/models/query_utils.py
```diff
@@ -157,7 +157,7 @@ def deconstruct(self):
     def identity(self):
         path, args, kwargs = self.deconstruct()
         identity = [path, *kwargs.items()]
-        for child in args:
+        for child in sorted(args, key=str):
```
charettes, line 160: Any thoughts about using `hash` instead?

charettes (commented):
As brought up on the ticket my only concerns here are that this changes the identify of all usage of `Q` on only for `CheckConstraint.check` where we're pretty confident that the ordering won't matter. I'm not entirely sure this is the case for `Index.condition` and when dealing with complex index expressions though.

I guess an alternative would be to have `CheckConstraint` _sort_ solely its `.check` instead.

django/db/models/query_utils.py
jacobtylerwalls: Sure, I could do that by moving the sort operation later, after all `make_hashable()` conversions. 👍

jacobtylerwalls:
Closing per ticket. Thanks y'all for the advice & research!

PR #17482, opened 2023-11-16
Refs #21183 -- feat(admin): add toggle button for list filters

jsuchan-reef:
- reusing some parts of `nav-sidebar` without doing too much unnecessary changes in said element to keep the diff small at this stage of PR.

- forum thread https://forum.djangoproject.com/t/ticket-21183-reopening/25269

before
before

after (expanded / collapsed)
after

### DIFF
--- a/django/contrib/admin/static/admin/css/sidebar.css
+++ b/django/contrib/admin/static/admin/css/sidebar.css
@@ -4,16 +4,14 @@
     max-height: 100vh;
 }
 
-.toggle-nav-sidebar {
+.toggle-sidebar {
     z-index: 20;
-    left: 0;
     display: flex;
     align-items: center;
     justify-content: center;
     flex: 0 0 23px;
     width: 23px;
     border: 0;
-    border-right: 1px solid var(--hairline-color);
     background-color: var(--body-bg);
     cursor: pointer;
     font-size: 1.25rem;
@@ -21,13 +19,23 @@
     padding: 0;
 }
 
+#toggle-nav-sidebar {
+    left: 0;
+    border-right: 1px solid var(--hairline-color);
+}
+
+#toggle-filter-sidebar {
+    right: 0;
+    border-left: 1px solid var(--hairline-color);
+}
+
 [dir="rtl"] .toggle-nav-sidebar {
     border-left: 1px solid var(--hairline-color);
     border-right: 0;
 }
 
-.toggle-nav-sidebar:hover,
-.toggle-nav-sidebar:focus {
+.toggle-sidebar:hover,
+.toggle-sidebar:focus {
     background-color: var(--darkened-bg);
 }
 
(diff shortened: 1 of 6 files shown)


django/contrib/admin/static/admin/css/sidebar.css
```diff
@@ -51,24 +59,36 @@
...
 
-.main.shifted .toggle-nav-sidebar::before {
+#toggle-filter-sidebar::before,
+.main.nav-sidebar-expanded #toggle-nav-sidebar::before {
+    /* arrow left */
     content: '\00AB';
 }
 
+.main > #changelist-filter {
+    display: none;
```
jsuchan-reef, line 75: using `display` instead of `visibility` (as `#nav-sidebar` does) because it doesnt need additional justifications via in `flex` and `margin` (as `#nav-sidebar` does)

knyghty (commented):
Thanks for the work here.

I made a first pass, though I didn't look at the code yet at all.

With [`enable_nav_sidebar=False`](https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#django.contrib.admin.AdminSite.enable_nav_sidebar), the open/close button is still visible without styling:

Screenshot 2023-11-24 at 19 03 28

It also looks like it should be openable and closable because of the full height, which may or may not be okay.

A horizontal scrollbars appear at all screen widths <= 1185px (Mac OS, Firefox). Example at tablet size:

Screenshot 2023-11-24 at 19 13 01

This makes the mobile experience unusable:

https://github.com/django/django/assets/3871354/f66cfcf8-e576-4ceb-99ec-4eb88595d215

When you do scroll to the right, the button obscures the sidebar:

Screenshot 2023-11-24 at 19 13 47

This really needs to be fixed for the bigger screen sizes. For smaller screen sizes, it may be best to place it underneath as currently happens.

Two things we need to clear up:

1. Should this be independent of `enable_nav_sidebar`?
2. What should the bahaviour be on mobile?

knyghty:
One other point of UX:

For the nav sidebar, if you have the sidebar closed, you still know where you are because there are breadcrumbs and headers and so.

For the filter sidebar, there is no way to know if there are filters applied or not when the sidebar is closed, other than looking at the URL in the address bar. This seems like an oversight. I think there should be some indicator for the user that they are viewing a filtered list.

jsuchan-reef:
thanks for the feedback, I'll look into it.

jsuchan-reef:
@knyghty I believe I addressed all the issues mentioned, although I couldnt reproduce some of them (maybe fixed as a byproduct of some other changes?).

mjurbanski-reef:
@knyghty anything we can do to get this moving along?

knyghty:
> @knyghty anything we can do to get this moving along?

Before anyone is going to review this it probably needs:
- To not be in draft
- To have an accepted ticket (I don't think referencing a wontfix ticket is enough here)
- New tests, not just fixing the existing ones
- It will also need selenium tests
- And at least a release note, not sure personally about other docs.

It's also not clear to me if you addressed the point I made above about indicating if the page is filtered.

PR #17489, opened 2023-11-19
Fixed #34977: Make change password form reachable through button

fsbraun:
https://code.djangoproject.com/ticket/34977

To improve usability and accessibility, this PR replaces the link to change a user's password within the help text of the password field of the change user form by a button.

image

Besides an improved user experience, I would expect this PR to also improve accessibility. But since I am anything but an expert on accessibility, I would like to ask the accessibility team to review and give hints on hot to improve accessibility.

### DIFF
--- a/django/contrib/auth/forms.py
+++ b/django/contrib/auth/forms.py
@@ -36,10 +36,9 @@ class ReadOnlyPasswordHashWidget(forms.Widget):
 
     def get_context(self, name, value, attrs):
         context = super().get_context(name, value, attrs)
+        usable_password = value and not value.startswith(UNUSABLE_PASSWORD_PREFIX)
         summary = []
-        if not value or value.startswith(UNUSABLE_PASSWORD_PREFIX):
-            summary.append({"label": gettext("No password set.")})
-        else:
+        if usable_password:
             try:
                 hasher = identify_hasher(value)
             except ValueError:
@@ -53,7 +52,12 @@ class ReadOnlyPasswordHashWidget(forms.Widget):
             else:
                 for key, value_ in hasher.safe_summary(value).items():
                     summary.append({"label": gettext(key), "value": value_})
+        else:
+            summary.append({"label": gettext("No password set.")})
         context["summary"] = summary
+        context["button_label"] = (
+            _("Reset password") if usable_password else _("Set password")
+        )
         return context
 
     def id_for_label(self, id_):
@@ -253,9 +257,8 @@ class UserChangeForm(forms.ModelForm):
     password = ReadOnlyPasswordHashField(
         label=_("Password"),
         help_text=_(
-            "Raw passwords are not stored, so there is no way to see this "
-            "user’s password, but you can change or unset the password using "
-            '<a href="{}">this form</a>.'
+            "Raw passwords are not stored, so there is no way to see "
+            "the user’s password."
         ),
     )
 
@@ -271,11 +274,8 @@ class UserChangeForm(forms.ModelForm):
             if self.instance and not self.instance.has_usable_password():
                 password.help_text = _(
                     "Enable password-based authentication for this user by setting a "
-                    'password using <a href="{}">this form</a>.'
+                    "password."
                 )
-            password.help_text = password.help_text.format(
-                f"../../{self.instance.pk}/password/"
-            )
         user_permissions = self.fields.get("user_permissions")
         if user_permissions:
             user_permissions.queryset = user_permissions.queryset.select_related(
--- a/django/contrib/auth/templates/auth/widgets/read_only_password_hash.html
+++ b/django/contrib/auth/templates/auth/widgets/read_only_password_hash.html
@@ -1,5 +1,8 @@
 <div{% include 'django/forms/widgets/attrs.html' %}>
-{% for entry in summary %}
-<strong>{{ entry.label }}</strong>{% if entry.value %}: <bdi>{{ entry.value }}</bdi>{% endif %}
-{% endfor %}
+  <p>
+    {% for entry in summary %}
+    <strong>{{ entry.label }}</strong>{% if entry.value %}: <bdi>{{ entry.value }}</bdi>{% endif %}
+    {% endfor %}
+  </p>
+  <p><a class="button" href="{{ password_url|default:"../password/" }}">{{ button_label }}</a></p>
 </div>
--- a/docs/releases/5.1.txt
+++ b/docs/releases/5.1.txt
@@ -57,6 +57,12 @@ Minor features
   :func:`~.django.contrib.auth.decorators.user_passes_test` decorators now
   support wrapping asynchronous view functions.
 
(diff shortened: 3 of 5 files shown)


nessita:
I would go a different route to solve this issue, specially because I believe that the best practice for help texts is not to use markup at all (more so for easing translations).

So my proposal is to:

* Re-phrase the help texts so they are complete sentences that make sense but without any actionable links nor buttons.
* Add the action button next to the help text.

This would also have the (nice) side effect to separate "form concerns" (ie help texts) from UI elements that want to be displayed together in a template.

Initial proposal (to be discussed with the accessibility team):
```diff
diff --git a/django/contrib/auth/forms.py b/django/contrib/auth/forms.py
index 5398af4dad..30d28cbe00 100644
--- a/django/contrib/auth/forms.py
+++ b/django/contrib/auth/forms.py
@@ -179,9 +179,8 @@ class UserChangeForm(forms.ModelForm):
     password = ReadOnlyPasswordHashField(
         label=_("Password"),
         help_text=_(
-            "<p>Raw passwords are not stored, so there is no way to see this "
-            "user’s password, but you can change the password here:</p>"
-            '<p><a class="button" role="button" href="{}">Change password</a></p>'
+            "Raw passwords are not stored, so there is no way to see this "
+            "user’s password."
         ),
     )
 
diff --git a/django/contrib/auth/templates/auth/widgets/read_only_password_hash.html b/django/contrib/auth/templates/auth/widgets/read_only_password_hash.html
index c73042b18f..f88d2e1ba4 100644
--- a/django/contrib/auth/templates/auth/widgets/read_only_password_hash.html
+++ b/django/contrib/auth/templates/auth/widgets/read_only_password_hash.html
@@ -1,5 +1,8 @@
+{% load i18n %}
+
 <div{% include 'django/forms/widgets/attrs.html' %}>
 {% for entry in summary %}
 <strong>{{ entry.label }}</strong>{% if entry.value %}: <bdi>{{ entry.value }}</bdi>{% endif %}
 {% endfor %}
+<a class="button" role="button" href="../password/">{% trans "Change password" %}</a>
 </div>
```

Draft UI:
image

fsbraun:
Good idea! Once we're touching `ReadOnlyPasswordHashField` I'd also like to change the formatting that the button always will be on the right. Depending on viewport width, the button will break into the next line.

nessita:
@fsbraun Would you have time to update this PR by rebasing onto main and ensuring is ready for re-review? Thanks!

@django/accessibility Would you have any opinion on my proposal above?

/remind me to re-check the status of this PR in 3 weeks

nessita:
> @nessita I've rebased and rewritten to reflect your first comment:
> 
>     * Help texts do not contain markup or buttons
> 
>     * Widget now contains the button for the link
> 
>     * In private exchange with @sabderemane we concluded that a `role` attribute for the button is not advisable from a accessibility viewpoint

Thank you @fsbraun ! Please remember to update the ticket flags accordingly.

django/contrib/admin/static/admin/css/widgets.css
```diff
@@ -591,3 +591,10 @@ ul.timelist, .timelist li {
     width: 600px;
     height: 400px;
 }
+
+/* READ-ONLY-PASSWORD WIDGET */
+
+.read-only-password  > div + div {
```
nessita, line 597:
```suggestion
.read-only-password > div + div {
```

django/contrib/auth/forms.py
```diff
@@ -54,11 +54,17 @@ def get_context(self, name, value, attrs):
                 for key, value_ in hasher.safe_summary(value).items():
                     summary.append({"label": gettext(key), "value": value_})
         context["summary"] = summary
+        context["usable_password"] = len(summary) > 1
```
nessita, line 57: I believe that is cleaner to define using `value.startswith(UNUSABLE_PASSWORD_PREFIX)`. So, something like:
```python
        usable_password = not value.startswith(UNUSABLE_PASSWORD_PREFIX)
        if not value or not usable_password:
            summary.append({"label": gettext("No password set.")})
        else:
            # ...
        context["usable_password"] = usable_password
```

django/contrib/auth/forms.py
```diff
@@ -54,11 +54,17 @@ def get_context(self, name, value, attrs):
...
         return context
 
     def id_for_label(self, id_):
         return None
 
+    class Media:
+        css = {
+            "all": ("admin/css/widgets.css",),
+        }
+
```
nessita, line 67: Is this really needed? I have removed it and I still see the same formatting in the admin :thinking:

django/contrib/auth/forms.py
nessita: On a second thought, since this is the Widget class, shall we instead add to the context the label and action of the button? My initial thinking is that approach would be cleaner and simpler in the template. What do you think?

nessita (commented):
Overall looks good!

Added some comments, and also please note that this work needs documentation updates, specifically we should review if any of the existing docs mention the link, and also we need release notes in the `5.1.txt` file.

Thank you @fsbraun !

django/contrib/auth/forms.py
fsbraun: @nessita What's the right way to get the admin namespace from within the widget? I would need it to reverse the set password URL. The template uses a `../bla` which indeed is a bit hacky. The reason might be that the admin namespace is not readily available.

Admin forms get the admin site passed to their `__init__` function. But that pattern would require to change the signature of the `UserChangeForm` which I would like to avoid. 

I have moved the button label from the template to the `get_context` method, however.

django/contrib/auth/forms.py
fsbraun: Other widgets on the user change form load the `widgets.css`, so removing it here will indeed not change anything.

I would still like to keep it, since the widget needs it for margins. If we removed the css from the widget it will fail to display properly if other widgets on the change from do not load the `widgets.css` for whatever reason.

fsbraun: Alternative: Have your own `auth/css/widgets.css`?

fsbraun:
@nessita Thank you for your comments!

I added a bullet point to the release notes. I searched the docs and did not find a reference to the link. Have I missed something?

I also implemented some of your comments where I did not see any roadblocks. I added comments to the others. Getting the URL of the change password action from within the widget seems tricky to me.

fsbraun:
Building docs fails apparently since it cannot locate the class `django.contrib.auth.forms.ReadOnlyPasswordHashWidget `. I am not sure why. What am I missing?

django/contrib/auth/forms.py
nessita: Thanks for the extra details. My advice then is as follows (I'll make a new commit with these):

* Change the HTML to make use of semantic elements such as `<p>`
* Remove the new CSS classes and custom CSS rule
* Let users of this widget define their own CSS by including the widget in whichever CSS context is relevant for them.

nessita (commented):
Looks good, thanks! I will make a few tweaks and merge. :trophy:

PR #17500, opened 2023-11-21
Refs #34986 -- Confirmed support for PyPy 3.10 with PostgreSQL.

ngnpope:
ticket-34986

### DIFF
--- a/tests/backends/postgresql/test_server_side_cursors.py
+++ b/tests/backends/postgresql/test_server_side_cursors.py
@@ -5,6 +5,8 @@ from contextlib import contextmanager
 
 from django.db import connection, models
 from django.test import TestCase
+from django.test.utils import garbage_collect
+from django.utils.version import PYPY
 
 from ..models import Person
 
@@ -88,14 +90,21 @@ class ServerSideCursorsPostgres(TestCase):
         persons = Person.objects.iterator()
         next(persons)  # Open a server-side cursor
         del persons
+        garbage_collect()
         cursors = self.inspect_cursors()
         self.assertEqual(len(cursors), 0)
 
+    @unittest.skipIf(
+        PYPY,
+        reason="Cursor not closed properly due to differences in garbage collection.",
+    )
     def test_server_side_cursors_setting(self):
         with self.override_db_setting(DISABLE_SERVER_SIDE_CURSORS=False):
             persons = Person.objects.iterator()
             self.assertUsesCursor(persons)
             del persons  # Close server-side cursor
 
+        # On PyPy, the cursor is left open here and attempting to force garbage
+        # collection breaks the transaction wrapping the test.
         with self.override_db_setting(DISABLE_SERVER_SIDE_CURSORS=True):
             self.assertNotUsesCursor(Person.objects.iterator())
--- a/.github/workflows/schedule_tests.yml
+++ b/.github/workflows/schedule_tests.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Run tests
         run: python tests/runtests.py -v2
 
-  pypy:
+  pypy-sqlite:
     runs-on: ubuntu-latest
     name: Ubuntu, SQLite, PyPy3.10
     continue-on-error: true
@@ -56,7 +56,44 @@ jobs:
         run: python -m pip install --upgrade pip setuptools wheel
       - run: python -m pip install -r tests/requirements/py3.txt -e .
       - name: Run tests
-        run: python tests/runtests.py -v2
+        run: python -Wall tests/runtests.py --verbosity=2
+
+  pypy-postgresql:
+    runs-on: ubuntu-latest
+    name: Ubuntu, PostgreSQL, PyPy3.10
+    continue-on-error: true
+    services:
+      postgres:
+        image: postgres:13-alpine
+        env:
+          POSTGRES_DB: django
+          POSTGRES_USER: user
+          POSTGRES_PASSWORD: postgres
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: pypy-3.10-nightly
+          cache: 'pip'
+          cache-dependency-path: 'tests/requirements/py3.txt'
+      - name: Install libmemcached-dev for pylibmc
+        run: sudo apt-get install libmemcached-dev
+      - name: Install and upgrade packaging tools
+        run: python -m pip install --upgrade pip setuptools wheel
+      - run: python -m pip install -r tests/requirements/py3.txt -r tests/requirements/postgres.txt -e .
+      - name: Create PostgreSQL settings file
+        run: mv ./.github/workflows/data/test_postgres.py.tpl ./tests/test_postgres.py
+      - name: Run tests
+        run: python -Wall tests/runtests.py --settings=test_postgres --verbosity=2
 
   javascript-tests:
     runs-on: ubuntu-latest
--- a/tests/requirements/postgres.txt
+++ b/tests/requirements/postgres.txt
@@ -1,1 +1,2 @@
-psycopg[binary]>=3.1.8
+psycopg>=3.1.14; implementation_name == 'pypy'
+psycopg[binary]>=3.1.8; implementation_name != 'pypy'


mgorny:
Thanks a lot for doing this! Could you also include something like:

```diff
diff --git a/tox.ini b/tox.ini
index 2e0a3b421a..c635a129b2 100644
--- a/tox.ini
+++ b/tox.ini
@@ -26,7 +26,7 @@ setenv =
     PYTHONDONTWRITEBYTECODE=1
 deps =
     -e .
-    py{3,310,311,312}: -rtests/requirements/py3.txt
+    py{3,310,311,312,py3}: -rtests/requirements/py3.txt
     postgres: -rtests/requirements/postgres.txt
     mysql: -rtests/requirements/mysql.txt
     oracle: -rtests/requirements/oracle.txt
```

ngnpope:
@mgorny Sure. I'll see if there is anything else that makes sense while I'm at it.

ngnpope:
Currently down to 2 failures and 1 error running on SQLite 😎 

```
auth_tests.test_auth_backends.AuthenticateTests.test_aauthenticate_sensitive_variables
model_forms.tests.OtherModelFormTests.test_prefetch_related_queryset
queries.tests.CloneTests.test_evaluated_queryset_as_argument
```

django/utils/encoding.py
```diff
@@ -9,15 +9,14 @@
 
 
 class DjangoUnicodeDecodeError(UnicodeDecodeError):
-    def __init__(self, obj, *args):
-        self.obj = obj
+    def __init__(self, *args):
```
timgraham, line 12: perhaps this method is obsolete now?

ngnpope: Perhaps, not sure. Pickle stuff with exception subclasses has always been pretty weird.

tests/forms_tests/field_tests/test_urlfield.py
```diff
@@ -136,7 +137,10 @@ def test_urlfield_strip_on_none_value(self):
         self.assertIsNone(f.clean(None))
 
     def test_urlfield_unable_to_set_strip_kwarg(self):
-        msg = "__init__() got multiple values for keyword argument 'strip'"
+        if PYPY:
+            msg = "got multiple values for keyword argument 'strip'"
+        else:
+            msg = "__init__() got multiple values for keyword argument 'strip'"
```
felixxm, line 143: I would remove `__init__()` from the message instead:
```suggestion
        msg = "got multiple values for keyword argument 'strip'"
```

django/utils/choices.py
```diff
@@ -73,36 +73,41 @@ def normalize_choices(value, *, depth=0):
...
-            # separately.
-            pass
-        case Iterable() if depth < 2 and not any(
-            isinstance(x, (Promise, bytes, str)) for x in value
-        ):
-            # String-like types are iterable, so the guard above ensures that
-            # they're handled by the default case below.
-            pass
-        case Callable() if depth == 0:
+    if isinstance(value, (BaseChoiceIterator, Promise, bytes, str)):
```
felixxm, line 76: I would wait for the PyPy Team before implementing any temporary solution here.

ngnpope:
For `auth_tests.test_auth_backends.AuthenticateTests.test_aauthenticate_sensitive_variables` the snag is that PyPy executes async stuff differently...

I get the following traceback for CPython:

```pytb
  File "/usr/local/lib/python3.10/site-packages/asgiref/sync.py", line 277, in __call__
    return call_result.result()
  File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 451, in result
    return self.__get_result()
  File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "/usr/local/lib/python3.10/site-packages/asgiref/sync.py", line 353, in main_wrap
  File "/tests/django/django/test/utils.py", line 434, in inner
    return await func(*args, **kwargs)
  File "/tests/django/tests/auth_tests/test_auth_backends.py", line 773, in test_aauthenticate_sensitive_variables
    await aauthenticate(
  File "/tests/django/django/contrib/auth/__init__.py", line 99, in aauthenticate
    return await sync_to_async(authenticate)(request, **credentials)
  File "/usr/local/lib/python3.10/site-packages/asgiref/sync.py", line 479, in __call__
    ret: _R = await loop.run_in_executor(
  File "/usr/local/lib/python3.10/site-packages/asgiref/current_thread_executor.py", line 40, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/local/lib/python3.10/site-packages/asgiref/sync.py", line 538, in thread_handler
    return func(*args, **kwargs)
  File "/tests/django/django/views/decorators/debug.py", line 78, in sensitive_variables_wrapper
  File "/tests/django/django/contrib/auth/__init__.py", line 79, in authenticate
    user = backend.authenticate(request, **credentials)
  File "/tests/django/django/views/decorators/debug.py", line 78, in sensitive_variables_wrapper
  File "/tests/django/tests/auth_tests/test_auth_backends.py", line 717, in authenticate
    raise TypeError
TypeError
```

And the following for PyPy:

```pytb
  File "/opt/pypy/lib/pypy3.10/site-packages/asgiref/sync.py", line 277, in __call__
    return call_result.result()
  File "/opt/pypy/lib/pypy3.10/concurrent/futures/_base.py", line 451, in result
    return self.__get_result()
  File "/opt/pypy/lib/pypy3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "/opt/pypy/lib/pypy3.10/asyncio/tasks.py", line 304, in __wakeup
    future.result()
  File "/opt/pypy/lib/pypy3.10/asyncio/futures.py", line 201, in result
    raise self._exception.with_traceback(self._exception_tb)
  File "/opt/pypy/lib/pypy3.10/site-packages/asgiref/current_thread_executor.py", line 40, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/opt/pypy/lib/pypy3.10/site-packages/asgiref/sync.py", line 538, in thread_handler
    return func(*args, **kwargs)
  File "/opt/pypy/lib/pypy3.10/_contextvars.py", line 44, in run
    return callable(*args, **kwargs)
  File "/tests/django/django/views/decorators/debug.py", line 78, in sensitive_variables_wrapper
  File "/tests/django/django/contrib/auth/__init__.py", line 79, in authenticate
    user = backend.authenticate(request, **credentials)
  File "/tests/django/django/views/decorators/debug.py", line 78, in sensitive_variables_wrapper
  File "/tests/django/tests/auth_tests/test_auth_backends.py", line 717, in authenticate
    raise TypeError
TypeError
```

I suspect we get away with all the `__traceback_hide__` stuff that was added to `asgiref` under CPython, but in PyPy we get a `_contextvars.Context.run()` call where `callable` is a `functools.partial()`. And that exposes the following:

```
name = 'callable'
value = functools.partial(<function authenticate at 0x0000000002c99f60>, None, username='testusername', password='mypassword')
```

We could extend `SafeExceptionReporterFilter._cleanse_special_types()` to handle `functools.partial()`, but that the problem here is that we had `@sensitive_variables("credentials")` but the `**credentials` makes it unpacked to `username=..., password=..." in the `functools.partial()` call.

Blergh. Not sure where to go from here.

django/utils/choices.py
ngnpope: Sure. Being familiar with the code, it was straightforward.

Was helpful to put this in for now to help see the wood for the trees...

mgorny:
CC @mattip, @cfbolz; could you particularly take a look at https://github.com/django/django/pull/17500#issuecomment-1821767664 and advise?

django/utils/choices.py
ngnpope: This is in a separate commit, so we can remove it easily if PyPy gets a fix.

cfbolz:
I'm currently fixing the pattern matching bug in pypy.

@ngnpope missing context to help here a little bit. is the problem that CPython's `functools.partial` instances are less inspectable/have different reprs? or the problem that `_contextvars` is an uninspectable C module in CPython, but just a regular Python file in PyPy? we could certainly add `__tracebackhide__` to `_contextvars.Context.run()` if that is helpful.

ngnpope:
Thanks for the [pattern matching fix](https://foss.heptapod.net/pypy/pypy/-/commit/461b1615847a16570ec820b6562e188f80e8d73c), @cfbolz. I'll give that a whirl when the next nightly is out 🎉

I'll try and work out a little more detail of what is going on and get back to you regarding this last test failure.

django/utils/encoding.py
ngnpope: Have confirmed that we don't need this `__init__()`. Ta.

I also tampered with the expected message in `utils_tests.test_encoding.TestEncodingUtils.test_force_str_DjangoUnicodeDecodeError` to double check that the original was causing the pickling issue in parallel test runs. It appears that this was also an issue with CPython, so not strictly related to this PyPy issue, but became visible because of it.

django/utils/choices.py
felixxm: We can remove it now as it's fixed in PyPY ([logs](https://github.com/django/django/actions/runs/6976283957/job/18985932640)).

django/utils/version.py
```diff
@@ -6,6 +6,9 @@
 
 from django.utils.regex_helper import _lazy_re_compile
 
+# Private, stable API for detecting the Python implementation.
+PYPY = sys.implementation.name == "pypy"
```
mattip, line 10: small nit: this is not really "private", it is [PEP 421](https://peps.python.org/pep-0421/) and was added for python>=3.3. `sys.implementation` does not exist on python2

felixxm: "Private and stable" apply here to the `PYPY` variable, not to the `sys.implementation`.

django/utils/choices.py
ngnpope: I've confirmed the fix with a local nightly build too. Have dropped the commit.

tests/model_enums/tests.py
```diff
@@ -317,11 +317,17 @@ class Boolean(bool, models.Choices):
...
-        msg = "type 'datetime.timezone' is not an acceptable base type"
-        with self.assertRaisesMessage(TypeError, msg):
+        if PYPY:
+            exc = AttributeError
+            msg = "'timezone' object has no attribute '_value_'"
+        else:
+            exc = TypeError
+            msg = "type 'datetime.timezone' is not an acceptable base type"
+
+        with self.assertRaisesMessage(exc, msg):
```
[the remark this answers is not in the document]
ngnpope: Here's a simple reproducer which requires Python 3.12 (as the pure-Python `_pydatetime` wasn't available before):

```pycon
>>> import datetime, _pydatetime
>>> class timezone(_pydatetime.timezone): pass  # Works fine...
... 
>>> class timezone(datetime.timezone): pass  # Fails to subclass...
... 
TypeError: type 'datetime.timezone' is not an acceptable base type
```

I'm not sure what the expected behaviour is here but it ought to be consistent...

Perhaps it's intentional that subclassing `datetime.timezone` shouldn't be allowed?

However, the case that we have here does seem to work perfectly fine using `_pydatetime`:

```pycon
>>> import enum
>>> class Timezone(_pydatetime.timezone, enum.Enum):
...     UTC = _pydatetime.timedelta(0), "UTC"
... 
>>> Timezone.UTC
<Timezone.UTC: datetime.timezone(datetime.timedelta(0), 'UTC')>
>>> Timezone.UTC == _pydatetime.timezone.utc
True
```

It seems that PyPy does support subclassing `datetime.timezone`, but I think there is another bug when defining enum members that doesn't occur when using `_pydatetime.timezone` in CPython...

```pycon
>>>> import datetime
>>>> class timezone(datetime.timezone): pass  # Works fine...
>>>> import enum
>>>> class Timezone(datetime.timezone, enum.Enum): pass  # Also works fine...
>>>> class Timezone(datetime.timezone, enum.Enum):
....     UTC = datetime.timedelta(0), "UTC"  # Something isn't right here...
....     
  File "/opt/pypy3/lib/pypy3.10/enum.py", line 289, in __new__
    enum_member = __new__(enum_class, *args)
  File "/opt/pypy3/lib/pypy3.10/datetime.py", line 2292, in __new__
    return cls._create(offset, name)
  File "/opt/pypy3/lib/pypy3.10/datetime.py", line 2296, in _create
    self = tzinfo.__new__(cls)
  File "/opt/pypy3/lib/pypy3.10/datetime.py", line 1196, in __new__
    return super(tzinfo, cls).__new__(cls)
TypeError: Enum.__new__() missing 1 required positional argument: 'value'
>>>> 
```

tests/queries/tests.py
```diff
@@ -2572,7 +2573,7 @@ def test_evaluated_queryset_as_argument(self):
         list(n_list)
         # Make one of cached results unpickable.
         n_list._result_cache[0].lock = Lock()
-        with self.assertRaises(TypeError):
+        with self.assertRaises(pickle.PicklingError if PYPY else TypeError):
             pickle.dumps(n_list)
```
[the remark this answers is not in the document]
ngnpope: Here's a simple reproducer in Python 3.10:

```pycon
>>> import pickle, threading
>>> lock = threading.Lock()
>>> pickle.dumps(lock)  # Using C
TypeError: cannot pickle '_thread.lock' object
>>> pickle._dumps(lock)  # Using Python
  File "/usr/lib/python3.10/pickle.py", line 1757, in _dumps
    buffer_callback=buffer_callback).dump(obj)
  File "/usr/lib/python3.10/pickle.py", line 487, in dump
    self.save(obj)
  File "/usr/lib/python3.10/pickle.py", line 578, in save
    rv = reduce(self.proto)
TypeError: cannot pickle '_thread.lock' object
```

And here's the same with PyPy:

```pycon
>>>> import pickle, threading
>>>> lock = threading.Lock()
>>>> pickle.dumps(lock)
  File "/opt/pypy3/lib/pypy3.10/pickle.py", line 1769, in _dumps
    _Pickler(f, protocol, fix_imports=fix_imports,
  File "/opt/pypy3/lib/pypy3.10/pickle.py", line 500, in dump
    self.save(obj)
  File "/opt/pypy3/lib/pypy3.10/pickle.py", line 616, in save
    self.save_reduce(obj=obj, *rv)
  File "/opt/pypy3/lib/pypy3.10/pickle.py", line 700, in save_reduce
    save(cls)
  File "/opt/pypy3/lib/pypy3.10/pickle.py", line 573, in save
    f(self, obj)  # Call unbound method with explicit self
  File "/opt/pypy3/lib/pypy3.10/pickle.py", line 1142, in save_type
    return self.save_global(obj)
  File "/opt/pypy3/lib/pypy3.10/pickle.py", line 1084, in save_global
    raise PicklingError(
pickle.PicklingError: Can't pickle <class '_thread.lock'>: it's not found as _thread.lock
```

Note the difference - it would seem that we're falling back to one of the non-pickleable cases raised out of [`Objects/typeobject.c`](https://github.com/python/cpython/blob/1b8aa7aafd2ac07836777707fd9ba5ffb353eb0c/Objects/typeobject.c) for CPython, but we don't seem to hit some equivalent in PyPy, but something else. This isn't therefore an issue with the `pickle` module, but some deeper down implementation difference.

Interestingly, I think the underlying issue here is that the locks show themselves as being `_thread.lock` which just isn't a thing!

CPython:

```pycon
>>> from _thread import lock
ImportError: cannot import name 'lock' from '_thread' (unknown location)
```

PyPy:

```pycon
>>>> from _thread import lock
ImportError: cannot import name 'lock' from '_thread' (unknown location)
```

ngnpope:
> @ngnpope missing context to help here a little bit. is the problem that CPython's `functools.partial` instances are less inspectable/have different reprs? or the problem that `_contextvars` is an uninspectable C module in CPython, but just a regular Python file in PyPy? we could certainly add `__tracebackhide__` to _contextvars.Context.run() if that is helpful.

@cfbolz Sorry for the delay getting back to you. This is the last failure, so we're making progress!

I've looked at the tracebacks and commented out the common bits between CPython and PyPy.

Here is the one for CPython:

```pytb
# Traceback (most recent call last):
#   File "/usr/local/lib/python3.10/site-packages/asgiref/sync.py", line 277, in __call__
#     return call_result.result()
#   File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 451, in result
#     return self.__get_result()
#   File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
#     raise self._exception

    File "/usr/local/lib/python3.10/site-packages/asgiref/sync.py", line 353, in main_wrap
    File "/django/source/django/test/utils.py", line 434, in inner
      return await func(*args, **kwargs)
    File "/django/source/tests/auth_tests/test_auth_backends.py", line 773, in test_aauthenticate_sensitive_variables
      await aauthenticate(
    File "/django/source/django/contrib/auth/__init__.py", line 99, in aauthenticate
      return await sync_to_async(authenticate)(request, **credentials)
    File "/usr/local/lib/python3.10/site-packages/asgiref/sync.py", line 479, in __call__
      ret: _R = await loop.run_in_executor(

#   File "/usr/local/lib/python3.10/site-packages/asgiref/current_thread_executor.py", line 40, in run
#     result = self.fn(*self.args, **self.kwargs)
#   File "/usr/local/lib/python3.10/site-packages/asgiref/sync.py", line 538, in thread_handler
#     return func(*args, **kwargs)

#   File "/django/source/django/views/decorators/debug.py", line 73, in sensitive_variables_wrapper
#     return func(*func_args, **func_kwargs)
#   File "/django/source/django/contrib/auth/__init__.py", line 79, in authenticate
#     user = backend.authenticate(request, **credentials)
#   File "/django/source/django/views/decorators/debug.py", line 73, in sensitive_variables_wrapper
#     return func(*func_args, **func_kwargs)
#   File "/django/source/tests/auth_tests/test_auth_backends.py", line 717, in authenticate
#     raise TypeError
# TypeError
```

And the one for PyPy:

```pytb
# Traceback (most recent call last):
#   File "/opt/pypy/lib/pypy3.10/site-packages/asgiref/sync.py", line 277, in __call__
#     return call_result.result()
#   File "/opt/pypy/lib/pypy3.10/concurrent/futures/_base.py", line 451, in result
#     return self.__get_result()
#   File "/opt/pypy/lib/pypy3.10/concurrent/futures/_base.py", line 403, in __get_result
#     raise self._exception

    File "/opt/pypy/lib/pypy3.10/asyncio/tasks.py", line 304, in __wakeup
      future.result()
    File "/opt/pypy/lib/pypy3.10/asyncio/futures.py", line 201, in result
      raise self._exception.with_traceback(self._exception_tb)

#   File "/opt/pypy/lib/pypy3.10/site-packages/asgiref/current_thread_executor.py", line 40, in run
#     result = self.fn(*self.args, **self.kwargs)
#   File "/opt/pypy/lib/pypy3.10/site-packages/asgiref/sync.py", line 538, in thread_handler
#     return func(*args, **kwargs)

    File "/opt/pypy/lib/pypy3.10/_contextvars.py", line 44, in run
      return callable(*args, **kwargs)

#   File "/django/source/django/views/decorators/debug.py", line 73, in sensitive_variables_wrapper
#     return func(*func_args, **func_kwargs)
#   File "/django/source/django/contrib/auth/__init__.py", line 79, in authenticate
#     user = backend.authenticate(request, **credentials)
#   File "/django/source/django/views/decorators/debug.py", line 73, in sensitive_variables_wrapper
#     return func(*func_args, **func_kwargs)
#   File "/django/source/tests/auth_tests/test_auth_backends.py", line 717, in authenticate
#     raise TypeError
# TypeError
```

_I'm not sure why there is a difference with the first block - perhaps something different happens in PyPy when stitching frames together in relation to async code? I seem to be able to get some of that output if I fiddle with the way that exceptions are caught, etc. But I think we don't need to worry about this bit too much._

The issue is, as you thought, that `_contextvars` is opaque in CPython, whereas in PyPy `callable` is coming out as:

```python
functools.partial(<function authenticate at 0x00007fec19875b00>, None, username='testusername', password='mypassword'```
```

So we could add `__traceback_hide__ = True` to `_contextvars.Context.run()` in PyPy which should solve this. See https://github.com/django/asgiref/pull/383 for similar work in `asgiref`.

An alternative is to add something to always mask `functools.partial()` to always hide it's arguments:

```diff
diff --git a/django/views/debug.py b/django/views/debug.py
index c1265bfe6b..b280f96c66 100644
--- a/django/views/debug.py
+++ b/django/views/debug.py
@@ -195,6 +195,14 @@ class SafeExceptionReporterFilter:
                     multivaluedict[param] = self.cleansed_substitute
         return multivaluedict
 
+    def get_cleansed_partial(self, request, value):
+        """Replace all arguments to partial functions."""
+        return type(value)(
+            value.func,
+            *(None if v is None else self.cleansed_substitute for v in value.args),
+            **{key: self.cleansed_substitute for key in value.keywords},
+        )
+
     def get_post_parameters(self, request):
         """
         Replace the values of POST parameters marked as sensitive with
@@ -234,6 +242,10 @@ class SafeExceptionReporterFilter:
         if is_multivalue_dict:
             # Cleanse MultiValueDicts (request.POST is the one we usually care about)
             value = self.get_cleansed_multivaluedict(request, value)
+
+        if isinstance(value, (functools.partial, functools.partialmethod)):
+            value = self.get_cleansed_partial(request, value)
+
         return value
 
     def get_traceback_frame_variables(self, request, tb_frame):

```

This fixed the test for me. What are your thoughts on that @felixxm?

felixxm:
Merged two commits in 174369a990c00d75cf40e0a06ab82bec9a89a7ef and 9baaf89eed42f9e747c774bc79543f47ea9697f8. This should help us avoid general issues when running tests. Let's wait for the next daily build.

felixxm:
> This fixed the test for me. What are your thoughts on that @felixxm?

I'm skeptical about fixing something in Django that is not Django's fault.

mattip:
PyPy has a `__pypy__.hidden_applevel` decorator that can be used to do the same as `__traceback_hide__`. We already use this in [functools.cmp_to_key](https://foss.heptapod.net/pypy/pypy/-/blob/branch/py3.9/lib-python/3/functools.py#L208) or [functools.reduce](https://foss.heptapod.net/pypy/pypy/-/blob/branch/py3.9/lib-python/3/functools.py#L240) which in CPython are usually c-extension function. We can add this to the appropriate [pure-python `_contextvars.py`](https://foss.heptapod.net/pypy/pypy/-/blob/branch/py3.9/lib_pypy/_contextvars.py) functions and classes that should be hidden from tracebacks. Which are appearing in tracebacks and should not?

ngnpope:
@mattip That sounds perfect! If you could add that it'd be great.

mattip:
I hid `_contextvars.Context.run` [to pypy3.9](https://foss.heptapod.net/pypy/pypy/-/commit/aecc058ec0e7dc29467f249aa7e0c378301f4e5a)

> I've also provided the minimal reproducers you requested above.

Thanks, I opened https://github.com/python/cpython/issues/112451

ngnpope:
Thanks @mattip. Could you also merge that into the pypy3.10 branch please?

mattip:
Pushed to py3.10 too

felixxm:
I merged "non-controversial" changes in the 6089230d3ec580f2f44e85f23962c14905fefcfd, 051dbb53884eb202131c27dfdeac7c3ddd7b1072, and baf705f34a8c8977d042ce43c71f508f9ca4f8ce.

@ngnpope Please rebase.

ngnpope:
Thanks @felixxm. I've rebased.

Running over the remaining commits:

- [Refs #34986 -- Added configuration to tox.ini for PyPy.](https://github.com/django/django/pull/17500/commits/2c20e86cef800543744e4c215f75b6a44737e089)
  _This was requested above by @mgorny to make it easier to use tox with PyPy. It seemed reasonable._
- [Refs #34986 -- Used non-binary build of psycopg for PyPy.](https://github.com/django/django/pull/17500/commits/b8b277d89e491f547eb2a7af19dec38b5e0c2ca1)
  _This I added because the C extension for `psycopg>=3` doesn't work with PyPy. To be honest, things don't work out all that well with the basic `ctypes` implementation either, but something that ought to work is better than something that definitely won't work. Performance will be solved by an HPy (or CFFI) implementation in `psycopg`, and perhaps this will come for "free" via support in Cython. See https://github.com/psycopg/psycopg/issues/154._
- [Refs #34986 -- Fixed some test assertions for PyPy.](https://github.com/django/django/pull/17500/commits/1686dfefce260782c4c688e84071885a8a8d1caf)
  _These are the two "controversial" cases. The subclassing of `datetime.timezone` is a curious one. In theory there is no reason why this should be prohibited and, as such, is likely a problem with CPython. Maybe this'll get fixed in https://github.com/python/cpython/issues/112451. I think there is still an issue here in PyPy though when using it as a base type of an enum which doesn't happen if we use the pure Python `_pydatetime` available as of CPython 3.12+. As for the other case related to pickling, this seems to also be something related to PyPy's implementation of `_thread` and how things get pickled. I don't know whether this is fixable._
- [Refs #34822, #34986 -- Fixed migrations serializer support for functools.lru_cache().](https://github.com/django/django/pull/17500/commits/e469a35f06ae585d590fe63189142e53e216cb70)
  _I think this one should be fine to merge. This isn't really PyPy's fault. In theory it's possible to have the pure Python version in CPython which would also be a function rather than the class type from the C module._

felixxm:
> Running over the remaining commits:
>  * [Refs #34986 -- Added configuration to tox.ini for PyPy.](https://github.com/django/django/pull/17500/commits/2c20e86cef800543744e4c215f75b6a44737e089)
>  _This was requested above by @mgorny to make it easier to use tox with PyPy. It seemed reasonable._

I think we should add it once we manage to pass all the test, as we do with a new Python version.

> * [Refs #34986 -- Used non-binary build of psycopg for PyPy.](https://github.com/django/django/pull/17500/commits/b8b277d89e491f547eb2a7af19dec38b5e0c2ca1)
> _This I added because the C extension for `psycopg>=3` doesn't work with PyPy. To be honest, things don't work out all that well with the basic `ctypes` implementation either, but something that ought to work is better than something that definitely won't work. Performance will be solved by an HPy (or CFFI) implementation in `psycopg`, and perhaps this will come for "free" via support in Cython. See [HPy support psycopg/psycopg#154](https://github.com/psycopg/psycopg/issues/154)._

Let me check this more carefully.

> * [Refs #34986 -- Fixed some test assertions for PyPy.](https://github.com/django/django/pull/17500/commits/1686dfefce260782c4c688e84071885a8a8d1caf)
> _These are the two "controversial" cases. The subclassing of `datetime.timezone` is a curious one. In theory there is no reason why this should be prohibited and, as such, is likely a problem with CPython. Maybe this'll get fixed in [C-extension `datetime.timezone` is not acceptable as a base class python/cpython#112451](https://github.com/python/cpython/issues/112451). I think there is still an issue here in PyPy though when using it as a base type of an enum which doesn't happen if we use the pure Python `_pydatetime` available as of CPython 3.12+. As for the other case related to pickling, this seems to also be something related to PyPy's implementation of `_thread` and how things get pickled. I don't know whether this is fixable._

Let's wait with that.

> * [Refs #34822, #34986 -- Fixed migrations serializer support for functools.lru_cache().](https://github.com/django/django/pull/17500/commits/e469a35f06ae585d590fe63189142e53e216cb70)
> _I think this one should be fine to merge. This isn't really PyPy's fault. In theory it's possible to have the pure Python version in CPython which would also be a function rather than the class type from the C module._

Merged.

ngnpope:
> Let me check this more carefully.

Regarding using PostgreSQL, with the fix in https://github.com/psycopg/psycopg/pull/686, and the non-binary version I only get the following errors:

```
FAIL: test_closed_server_side_cursor (backends.postgresql.test_server_side_cursors.ServerSideCursorsPostgres)
  File "/opt/pypy/lib/pypy3.10/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/opt/pypy/lib/pypy3.10/unittest/case.py", line 591, in run
  File "/opt/pypy/lib/pypy3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/django/source/tests/backends/postgresql/test_server_side_cursors.py", line 92, in test_closed_server_side_cursor
    self.assertEqual(len(cursors), 0)
  File "/opt/pypy/lib/pypy3.10/unittest/case.py", line 845, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/opt/pypy/lib/pypy3.10/unittest/case.py", line 838, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: 1 != 0

FAIL: test_server_side_cursors_setting (backends.postgresql.test_server_side_cursors.ServerSideCursorsPostgres)
  File "/opt/pypy/lib/pypy3.10/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/opt/pypy/lib/pypy3.10/unittest/case.py", line 591, in run
  File "/opt/pypy/lib/pypy3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/django/source/tests/backends/postgresql/test_server_side_cursors.py", line 101, in test_server_side_cursors_setting
    self.asserNotUsesCursor(Person.objects.iterator())
  File "/django/source/tests/backends/postgresql/test_server_side_cursors.py", line 57, in asserNotUsesCursor
    self.assertUsesCursor(queryset, num_expected=0)
  File "/django/source/tests/backends/postgresql/test_server_side_cursors.py", line 49, in assertUsesCursor
    self.assertEqual(len(cursors), num_expected)
  File "/opt/pypy/lib/pypy3.10/unittest/case.py", line 845, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/opt/pypy/lib/pypy3.10/unittest/case.py", line 838, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: 1 != 0
```

For which we could do the following 😖 

```diff
diff --git a/tests/backends/postgresql/test_server_side_cursors.py b/tests/backends/postgresql/test_server_side_cursors.py
index 705e798c23..712fff8371 100644
--- a/tests/backends/postgresql/test_server_side_cursors.py
+++ b/tests/backends/postgresql/test_server_side_cursors.py
@@ -1,3 +1,4 @@
+import gc
 import operator
 import unittest
 from collections import namedtuple
@@ -5,6 +6,7 @@ from contextlib import contextmanager
 
 from django.db import connection, models
 from django.test import TestCase
+from django.utils.version import PYPY
 
 from ..models import Person
 
@@ -88,9 +90,12 @@ class ServerSideCursorsPostgres(TestCase):
         persons = Person.objects.iterator()
         next(persons)  # Open a server-side cursor
         del persons
+        if PYPY:
+            gc.collect()
         cursors = self.inspect_cursors()
         self.assertEqual(len(cursors), 0)
 
+    @unittest.skipIf(PYPY, reason="Forced garbage collection breaks test.")
     def test_server_side_cursors_setting(self):
         with self.override_db_setting(DISABLE_SERVER_SIDE_CURSORS=False):
             persons = Person.objects.iterator()

```

tests/model_enums/tests.py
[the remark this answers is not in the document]
felixxm: @ngnpope What do you think about removing this test? We are not testing any Django-specific behavior here and we already have tests that show some classes are not supported. I don't see the point in keeping it.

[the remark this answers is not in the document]
ngnpope: Yeah, scrub it 🙂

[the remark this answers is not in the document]
ngnpope: I've added a commit to remove it.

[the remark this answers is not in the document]
felixxm: Merged in 06c5cb1284557edc05f963afb6f6bcdd55b00911.

tests/queries/tests.py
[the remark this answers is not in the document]
felixxm: Can any of you open an issue in Python and share your findings?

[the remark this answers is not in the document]
ngnpope: I think this is something to be fixed in PyPy, or  maybe we need to change the test. Ultimately we just want something unpicklable here.

Yes, `_thread.lock` isn't a thing, but that's to be expected - it's been this way forever and this is a very low level implementation thing that isn't meant to be used directly.

The discrepancy seems to be either due to the way that PyPy handles or implements the low level thread stuff which is not part of the standard library, or the way that it implemented the equivalent of stuff from `typeobject.c` which is where the `TypeError` is coming from on CPython.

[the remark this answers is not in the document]
ngnpope: For further consideration related to this, I get the following if I attempt to pickle `dict_items`...

CPython:

```pycon
>>> import pickle
>>> pickle.dumps({}.items())
TypeError: cannot pickle 'dict_items' object
```

PyPy:

```pycon
>>>> import pickle
>>>> pickle.dumps({}.items())
  File "/opt/pypy3/lib/pypy3.10/pickle.py", line 1769, in _dumps
    _Pickler(f, protocol, fix_imports=fix_imports,
  File "/opt/pypy3/lib/pypy3.10/pickle.py", line 500, in dump
    self.save(obj)
  File "/opt/pypy3/lib/pypy3.10/pickle.py", line 616, in save
    self.save_reduce(obj=obj, *rv)
  File "/opt/pypy3/lib/pypy3.10/pickle.py", line 700, in save_reduce
    save(cls)
  File "/opt/pypy3/lib/pypy3.10/pickle.py", line 573, in save
    f(self, obj)  # Call unbound method with explicit self
  File "/opt/pypy3/lib/pypy3.10/pickle.py", line 1142, in save_type
    return self.save_global(obj)
  File "/opt/pypy3/lib/pypy3.10/pickle.py", line 1084, in save_global
    raise PicklingError(
pickle.PicklingError: Can't pickle <class 'dict_items'>: it's not found as builtins.dict_items
```

In CPython, when we look at `static PyObject *reduce_newobj(PyObject *obj)` in `Objects/typeobject.c` we can see:

```c
    if (Py_TYPE(obj)->tp_new == NULL) {
        PyErr_Format(PyExc_TypeError,
                     "cannot pickle '%.200s' object",
                     Py_TYPE(obj)->tp_name);
        return NULL;
    }
```

And looking at the implementation for `dict_items` in `Objects/dictobject.c` we can see that `PyTypeObject PyDictItems_Type` does not implement `tp_new`. In CPython we also have the following in `static PyObject *
type_call(PyTypeObject *type, PyObject *args, PyObject *kwds)`:

```c
    if (type->tp_new == NULL) {
        _PyErr_Format(tstate, PyExc_TypeError,
                      "cannot create '%s' instances", type->tp_name);
        return NULL;
    }
```

And so, in CPython:

```pycon
>>> from threading import Lock
>>> type(Lock())()
TypeError: cannot create '_thread.lock' instances
>>> type({}.items())({"a": 1})
TypeError: cannot create 'dict_items' instances
```

Compared to PyPy:

```pycon
>>>> from threading import Lock
>>>> type(Lock())()
TypeError: object.__new__(lock) is not safe, use lock.__new__()
>>>> type({}.items())({"a": 1})
dict_items([('a', 1)])
```

@cfbolz @mattip Any ideas on how to proceed here? I don't think that CPython is wrong - these are just types that cannot be created this way, yet PyPy is able to do so in some cases. I had a look in `pypy/objspace/std/objectobject.py`, but don't really know where to begin.

@felixxm Maybe, as mentioned above, we should just use a custom thing, e.g.

CPython:

```pycon
>>> class Unpickleable:
...     def __reduce__(self):
...         raise Exception("Cannot pickle.")
... 
>>> import pickle
>>> pickle.dumps(Unpickleable())
  File "<stdin>", line 3, in __reduce__
Exception: Cannot pickle.
```

PyPy:

```pycon
>>>> class Unpickleable:
....     def __reduce__(self):
....         raise Exception("Cannot pickle.")
....         
>>>> import pickle
>>>> pickle.dumps(Unpickleable())
  File "/opt/pypy3/lib/pypy3.10/pickle.py", line 1769, in _dumps
    _Pickler(f, protocol, fix_imports=fix_imports,
  File "/opt/pypy3/lib/pypy3.10/pickle.py", line 500, in dump
    self.save(obj)
  File "/opt/pypy3/lib/pypy3.10/pickle.py", line 591, in save
    rv = reduce(self.proto)
  File "<stdin>", line 3, in __reduce__
Exception: Cannot pickle.
```

Looks like it does the job!

ngnpope:
Updated using an alternative unpickleable object as mentioned in https://github.com/django/django/pull/17500#discussion_r1414188768 and added some fixes for PostgreSQL. I've also added a commit to extend the scheduled testing to include PostgreSQL and add the ability to execute tests for PyPy using a label on a PR.

felixxm:
> Updated using an alternative unpickleable object as mentioned in [#17500 (comment)](https://github.com/django/django/pull/17500#discussion_r1414188768) and added some fixes for PostgreSQL. I've also added a commit to extend the scheduled testing to include PostgreSQL and add the ability to execute tests for PyPy using a label on a PR.

I don't think we want more CI tests for PyPy. I don't see much value in labeled CI tests for PyPy, because how could we decide if something might not be compatible with PyPy? Daily builds seems to be sufficient in such rare cases.

ngnpope:
> I don't think we want more CI tests for PyPy. I don't see much value in labeled CI tests for PyPy, because how could we decide if something might not be compatible with PyPy? Daily builds seems to be sufficient in such rare cases.

Ok. I'll revert that one to only add a daily run with PostgreSQL.

tests/queries/tests.py
```diff
@@ -2571,8 +2575,8 @@ def test_evaluated_queryset_as_argument(self):
         # Evaluate the Note queryset, populating the query cache
         list(n_list)
         # Make one of cached results unpickable.
-        n_list._result_cache[0].lock = Lock()
-        with self.assertRaises(TypeError):
+        n_list._result_cache[0].lock = UnpickleableError()
```
felixxm, line 2578: This is not a lock anymore so I'll rename it.

tests/backends/postgresql/test_server_side_cursors.py
```diff
@@ -31,17 +31,20 @@ def inspect_cursors(self):
...
-                self.addCleanup(operator.delitem, connection.settings_dict, setting)
+                cleanups.append((operator.delitem, settings, name))
+            settings[name] = value
+
+        yield
 
-            connection.settings_dict[setting] = kwargs[setting]
-            yield
+        for operation, *args in reversed(cleanups):
+            operation(*args)
```
felixxm, line 49: I don't see how this change is related with PyPy :thinking:

ngnpope: This was something I looked at while trying to see if I could get the `test_server_side_cursors_setting` working. Things were behaving strangely and I saw that this context manager did things in a strange way as detailed in the message for that commit:

> The `yield` should be after all settings changes have been applied (not that this context manager is used with more than one setting override) and the clean up should happen when the context manager exits, not after the end of the test. This is because the context manager is used multiple times within the same test.

Ultimately this didn't help with the issue I had around cursor destruction - so it's not related - but I left the commit because it seemed more correct. If you don't think that makes sense, feel free to drop it.

felixxm: Let's move this to a separate PR.

ngnpope: Extracted to #17578.

felixxm (approved):
@ngnpope Thanks :+1:

mgorny:
Thanks a lot for all your hard work! Will these changes make it to 5.0.x?

felixxm:
> Will these changes make it to 5.0.x?

Unfortunately not.

ngnpope:
@mgorny Most of the changes were to Django's tests to handle minor implementation differences. If you're not running those, then there are very few changes.

Note that you'll need to run using a nightly build of PyPy right now as there were a few tweaks there over PyPy v7.3.13. And we've only tested for SQLite and PostgreSQL (which requires `psycopg>=3.1.14` without the binary build).

I think the only one that'll stop PyPy working for 5.0 is 5f9e5c1b0d5680f793ba7646d52ffab9e2c3623f to fix a crash in serializing for migrations. _(This was related to something newly added in 5.0, but arguably wasn't a release blocker because PyPy support had stagnated anyway.)_

And, if you want to use Django's autoreloader while using Python's `-X` options, you may need to backport 051dbb53884eb202131c27dfdeac7c3ddd7b1072. But this is probably quite a niche thing.

mgorny:
No problem. I've been able to backport them easily to 5.0, 4.2 and 4.1 branches in Gentoo. Thanks a lot, again!

felixxm:
@ngnpope GitHub action for PyPy and PostgreSQL [crashes](https://github.com/django/django/actions/runs/7217307991/job/19666401476).

ngnpope:
Thanks @felixxm. This was a presumption that the existing settings file was correct. Fixed in https://github.com/django/django/pull/17612.

PR #17514, opened 2023-11-23
Fixed #34991 -- Fixed pagination links and input layout in admin's change list page when using list_editable.

knyghty:
https://code.djangoproject.com/ticket/34991

Okay, it doesn't look perfectly the same as before, but IMO it looks better so I won't lose sleep over it.

I also fixed a small pre-existing RTL bug you can see in the first RTL screenshot (margin on the last pagination link is in the wrong direction).

# Before

## LTR

Screenshot 2023-11-23 at 11 13 55

## RTL

Note the gap also is between 3 and 4, should be between 4 and "318 releases".

Screenshot 2023-11-23 at 11 14 19

# After

## LTR

Screenshot 2023-11-23 at 11 14 41

## RTL

Screenshot 2023-11-23 at 11 15 07

## Worst I could make it look

Smallest tablet size, sidebar open, with filter sidebar to right.

Screenshot 2023-11-23 at 11 15 57

### DIFF
--- a/django/contrib/admin/static/admin/css/base.css
+++ b/django/contrib/admin/static/admin/css/base.css
@@ -1100,6 +1100,9 @@ a.deletelink:focus, a.deletelink:hover {
 /* PAGINATOR */
 
 .paginator {
+    display: flex;
+    align-items: center;
+    gap: 4px;
     font-size: 0.8125rem;
     padding-top: 10px;
     padding-bottom: 10px;
@@ -1141,6 +1144,10 @@ a.deletelink:focus, a.deletelink:hover {
 .paginator a:focus, .paginator a:hover {
     color: white;
     background: var(--link-hover-color);
+}
+
+.paginator input {
+    margin-left: auto;
 }
 
 .base-svgs {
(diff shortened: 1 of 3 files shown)


django/contrib/admin/templates/admin/pagination.html
```diff
@@ -1,6 +1,6 @@
 {% load admin_list %}
 {% load i18n %}
-<p class="paginator">
+<div class="paginator">
```
knyghty, line 3: Not sure if this is necessary or not. It could probably be removed. But it's not a paragraph or phrasing content so it seemed reasonable to make this semantic and remove any default styling while I was looking around.

nessita: I wouldn't change this (just yet?) because there are other templates using the same layout with `<p>` (tests/admin_views/templates/admin/admin_views/article/pagination.html, django/contrib/admin/templates/admin/object_history.html), so changing in one place and not the others feels incomplete.

nessita (commented):
Thank you Tom, looks good. Could you please add a release note?

knyghty:
@nessita thanks, done. I added it to the 4.2.8 release notes, not sure if it needs to be elsewhere.

nessita (approved):
Excellent, thank you! Will merge when CI is green.

nessita:
> @nessita thanks, done. I added it to the 4.2.8 release notes, not sure if it needs to be elsewhere.

4.2.8 release notes is enough, thanks!

PR #17517, opened 2023-11-23
Fixed #34995 -- Fixed position of related widget's add link on admin pages for smaller screens.

knyghty:
https://code.djangoproject.com/ticket/34995

A little tricky. Had to reposition the selector div so the add links are in the right place.

Will wait for the ticket to be accepted before working on it more.

# Before

Screenshot 2023-11-23 at 16 13 01

# After

Screenshot 2023-11-23 at 17 40 37

Screenshot 2023-11-23 at 17 40 45

Screenshot 2023-11-23 at 17 40 57

Screenshot 2023-11-23 at 17 41 32

### DIFF
--- a/django/contrib/admin/static/admin/css/responsive.css
+++ b/django/contrib/admin/static/admin/css/responsive.css
@@ -237,22 +237,6 @@ input[type="submit"], button {
         padding: 7px;
     }
 
-    /* Related widget */
-
-    .related-widget-wrapper {
-        float: none;
-    }
-
-    .related-widget-wrapper-link + .selector {
-        max-width: calc(100% - 30px);
-        margin-right: 15px;
-    }
-
-    select + .related-widget-wrapper-link,
-    .related-widget-wrapper-link + .related-widget-wrapper-link {
-        margin-left: 10px;
-    }
-
     /* Selector */
 
     .selector {
@@ -270,7 +254,7 @@ input[type="submit"], button {
     }
 
     .selector .selector-filter input {
-        width: auto;
+        width: 100%;
         min-height: 0;
         flex: 1 1;
     }
@@ -292,7 +276,6 @@ input[type="submit"], button {
         width: 26px;
         height: 52px;
         padding: 2px 0;
-        margin: auto 15px;
         border-radius: 20px;
         transform: translateY(-10px);
     }
@@ -336,7 +319,6 @@ input[type="submit"], button {
         width: 52px;
         height: 26px;
         padding: 0 2px;
-        margin: 15px auto;
         transform: none;
     }
 
@@ -684,19 +666,10 @@ input[type="submit"], button {
         align-self: center;
     }
 
-    select + .related-widget-wrapper-link,
-    .related-widget-wrapper-link + .related-widget-wrapper-link {
-        margin-left: 15px;
-    }
-
     /* Selector */
 
     .selector {
         flex-direction: column;
-    }
-
-    .selector > * {
-        float: none;
     }
 
     .selector-available, .selector-chosen {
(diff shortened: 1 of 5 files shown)


nessita:
Thank you @knyghty ! Functionality looks great, I tested small, medium and wide screens. I'm not sure I understand the need to add the extra `div` to the selector, could you please explain a bit more? is it possible/does it make sense to add a selenium test for it?

knyghty:
@nessita There isn't an extra element involved. `quickElement` appends the `selector` div to the end of the parent. However, we need it to be at the beginning so that the add link is on the right, rather than the left, when using flex. So the extra line moves it to the beginning.

I pushed a comment explaining this as it's not very obvious. I'm not sure (and hope) a Selenium test is needed because the behaviour is covered with existing tests, I'm not sure exactly what test I would add that doesn't already exist.

sarahboyce:
If you need a regression test, I think we can do this in the Javascript tests:
```
--- a/js_tests/admin/SelectFilter2.test.js
+++ b/js_tests/admin/SelectFilter2.test.js
@@ -5,9 +5,11 @@ QUnit.module('admin.SelectFilter2');

 QUnit.test('init', function(assert) {
     const $ = django.jQuery;
-    $('<form><select id="id"></select></form>').appendTo('#qunit-fixture');
+    $('<form id="test"><label for="id_id">Test</label><div class="helptext">This is helpful.</div><select id="id"></select></form>').appendTo('#qunit-fixture');
     $('<option value="0">A</option>').appendTo('#id');
     SelectFilter.init('id', 'things', 0);
+    assert.equal($('#test').children().first().prop("tagName"), "DIV");
+    assert.equal($('#test').children().first().attr("class"), "selector");
     assert.equal($('.selector-available h2').text().trim(), "Available things");
     assert.equal($('.selector-chosen h2').text().trim(), "Chosen things");
     assert.equal($('.selector-chosen select')[0].getAttribute('multiple'), '');
```

nessita:
@knyghty I can't reproduce locally the Selenium failure, could you take a look if you have some time?

knyghty:
@nessita Of the test failures in the CI, the only one I can reproduce is `admin_views.test_autocomplete_view.SeleniumTests.test_select` and then only in Firefox. So I am not really sure what more I can do.

sarahboyce:
> @nessita Of the test failures in the CI, the only one I can reproduce is `admin_views.test_autocomplete_view.SeleniumTests.test_select` and then only in Firefox. So I am not really sure what more I can do.

`admin_views.test_autocomplete_view.SeleniumTests.test_select` is a natorious flakey test, if you can figure that one out amazing but I think it will likely pass on a rerun.
With the other failures it is suspicious that they fail on the first click of the page. It might be that with the extra html manipulation, the page is not yet "ready" to be clicked and we need to add in some waits. I am having a play and seeing if I can get them passing on my fork 👍

sarahboyce:
https://github.com/sarahboyce/django/pull/6/commits/b9130404c3fa7ac314ebddf09a3122acabae66e6

Thrown the kitchen sink at it (aka tried a whole bunch of things and hoped for the best), this seems to work 🤷‍♀️ 
Going to re-run 2 more times as a "flakiness" check 👍

sarahboyce (approved):
https://www.selenium.dev/selenium/docs/api/py/webdriver/selenium.webdriver.common.action_chains.html
I agree the `ActionChains` with the `move_to_element` functionality resolves when elements are outside the window in small windows and makes these tests more rubust to the ui changes.
Great job!

nessita:
As I said in [Discord](https://discord.com/channels/856567261900832808/1177242175852007464/1179845176596893756):

> Ok, all done, proper authors, CI fully green, isolated commit
@knyghty  I am definitely landing this in main, then I'll do a local backport and see how things behave
because we've been landing so many changes in the admin CSS, that I wanna be sure that this change "in isolation" in 5.0 is an improvement (I think so but want to triple check)

PR #17518, opened 2023-11-23
Fixed #34982 -- Fixed helptext and password widget alignment on tablet.

knyghty:
https://code.djangoproject.com/ticket/34982

# Before

Screenshot 2023-11-23 at 18 30 29

Screenshot 2023-11-23 at 18 27 21

# After

Screenshot 2023-11-23 at 18 27 46

Screenshot 2023-11-23 at 18 27 58

### DIFF
--- a/tests/admin_views/tests.py
+++ b/tests/admin_views/tests.py
@@ -5754,6 +5754,7 @@ class SeleniumTests(AdminSeleniumTestCase):
         and with stacked and tabular inlines.
         Refs #13068, #9264, #9983, #9784.
         """
+        from selenium.webdriver import ActionChains
         from selenium.webdriver.common.by import By
 
         self.admin_login(
@@ -5766,6 +5767,8 @@ class SeleniumTests(AdminSeleniumTestCase):
 
         # Main form ----------------------------------------------------------
         self.selenium.find_element(By.ID, "id_pubdate").send_keys("2012-02-18")
+        status = self.selenium.find_element(By.ID, "id_status")
+        ActionChains(self.selenium).move_to_element(status).click(status).perform()
         self.select_option("#id_status", "option two")
         self.selenium.find_element(By.ID, "id_name").send_keys(
             " the mAin nÀMë and it's awεšomeıııİ"
@@ -5784,6 +5787,10 @@ class SeleniumTests(AdminSeleniumTestCase):
         self.selenium.find_element(
             By.ID, "id_relatedprepopulated_set-0-pubdate"
         ).send_keys("2011-12-17")
+        status = self.selenium.find_element(
+            By.ID, "id_relatedprepopulated_set-0-status"
+        )
+        ActionChains(self.selenium).move_to_element(status).click(status).perform()
         self.select_option("#id_relatedprepopulated_set-0-status", "option one")
         self.selenium.find_element(
             By.ID, "id_relatedprepopulated_set-0-name"
@@ -5815,6 +5822,10 @@ class SeleniumTests(AdminSeleniumTestCase):
         self.selenium.find_element(
             By.ID, "id_relatedprepopulated_set-1-pubdate"
         ).send_keys("1999-01-25")
+        status = self.selenium.find_element(
+            By.ID, "id_relatedprepopulated_set-1-status"
+        )
+        ActionChains(self.selenium).move_to_element(status).click(status).perform()
         self.select_option("#id_relatedprepopulated_set-1-status", "option two")
         self.selenium.find_element(
             By.ID, "id_relatedprepopulated_set-1-name"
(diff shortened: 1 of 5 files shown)


nessita:
Overall looks good but the adding of `min-width` is breaking the M2M widgets :sob: :see_no_evil::
image

knyghty:
Hmm, okay. the min-width is necessary to fix the password widget. Not sure what we should do in this case. It's ugly, but not completely broken.

nessita:
> Hmm, okay. the min-width is necessary to fix the password widget. Not sure what we should do in this case. It's ugly, but not completely broken.

I understand how the `min-width`  is necessary for when there is a visible label. The thing is that for non visible labels, width should be zero (ish). This change affects all screen sizes, it looks weird and, IMHO, for mobile it looks bad because the text entry overflows the container (see my screenshot above).

nessita:
Maybe we can customize the readonly widget itself, which I think it's the only one suffering from the label (reduced) width?

knyghty:
I don't think we should customise the widget as it doesn't live in `contrib.admin`, it could (I guess?) be used outside the admin. `UserChangeForm` also tends to be overridden a lot so not sure about setting it there either.

I also don't see an obvious way to easily target it with CSS in the admin. This issue is caused by removing `flex-wrap: wrap` from `.flex-container`. Without this we get the current behaviour where it wraps to the next line instead.

I scraped together a hacky fix by fixing the label on the selector. If this is the best we can do, at least we shouldn't need if we can get the stacked forms merged.

knyghty:
I'm open to any other ideas though, not really sure if this one works that well.

nessita:
> I don't think we should customise the widget as it doesn't live in `contrib.admin`, it could (I guess?) be used outside the admin. `UserChangeForm` also tends to be overridden a lot so not sure about setting it there either.

Makes sense.

> I also don't see an obvious way to easily target it with CSS in the admin. This issue is caused by removing `flex-wrap: wrap` from `.flex-container`. Without this we get the current behaviour where it wraps to the next line instead.
> 
> I scraped together a hacky fix by fixing the label on the selector. If this is the best we can do, at least we shouldn't need if we can get the stacked forms merged.

The "hacky fix" would be commit 3?

knyghty:
> The "hacky fix" would be commit 3?

Yes, "Fix related widget".

sarahboyce:
I think I may not understand exactly what is required from the ticket. Do we want the helptext to be aligned to the label?
<details>
<summary>(if so I think we can get rid of some of these margins but I feel like I misunderstood)</summary>

```
--- a/django/contrib/admin/static/admin/css/forms.css
+++ b/django/contrib/admin/static/admin/css/forms.css
@@ -126,8 +126,6 @@ form .aligned div.radiolist {
 form .aligned p.help,
 form .aligned div.help {
     margin-top: 0;
-    margin-left: 160px;
-    padding-left: 10px;
 }

 form .aligned p.date div.help.timezonewarning,
@@ -144,16 +142,6 @@ form .aligned div.help:last-child {
     padding-bottom: 0;
 }

-form .aligned input + p.help,
-form .aligned textarea + p.help,
-form .aligned select + p.help,
-form .aligned input + div.help,
-form .aligned textarea + div.help,
-form .aligned select + div.help {
-    margin-left: 160px;
-    padding-left: 10px;
-}
-
 form .aligned ul li {
     list-style: none;
 }
diff --git a/django/contrib/admin/static/admin/css/rtl.css b/django/contrib/admin/static/admin/css/rtl.css
index 307606f156..e6de33d6b1 100644
--- a/django/contrib/admin/static/admin/css/rtl.css
+++ b/django/contrib/admin/static/admin/css/rtl.css
@@ -149,12 +149,6 @@ form ul.inline li {
     padding-left: 7px;
 }

-form .aligned p.help,
-form .aligned div.help {
-    margin-right: 160px;
-    padding-right: 10px;
-}
-
 form div.help ul,
 form .aligned .checkbox-row + .help,
 form .aligned p.date div.help.timezonewarning,
```
</details>

nessita:
> I think I may not understand exactly what is required from the ticket. Do we want the helptext to be aligned to the label?
> (if so I think we can get rid of some of these margins but I feel like I misunderstood)

@sarahboyce I understand that the main goal of the ticket is to not have the password read only widget be in a second line under the label. The help text does not seem like a regression since the 160px margin is there since years ago.

nessita:
@knyghty I'm rebasing onto main to fix conflicts and trigger selenium tests. What you refer to as "hacky solution" looks good to me, but perhaps @sarahboyce can help with the review as well :-)

This branch also nicely fixes a misalignment of the help text for the M2M widgets! Which I think is a nice side effect. See main:
image

And this branch:
image

tests/admin_views/tests.py
```diff
@@ -5841,7 +5841,9 @@ def test_prepopulated_fields(self):
         element = self.selenium.find_element(
             By.ID, "id_relatedprepopulated_set-2-0-status"
         )
-        self.selenium.execute_script("window.scrollTo(0, %s);" % element.location["y"])
+        self.selenium.execute_script(
+            "window.scrollTo(200, %s);" % element.location["y"]
+        )
```
sarahboyce, line 5846:
```suggestion
        self.selenium.execute_script(
            "window.scrollTo(%s, %s);"
            % (status_element.location["x"], status_element.location["y"])
        )
```

knyghty: Oh, I forgot the one I was testing with, thanks!

nessita (approved):
Looks great! Rebasing, squashing and then merging.

tests/admin_views/tests.py
```diff
@@ -6,6 +6,8 @@
 from unittest import mock
 from urllib.parse import parse_qsl, urljoin, urlparse
 
+from selenium.webdriver import ActionChains
```
felixxm, line 9: `selenium` is an optional requirement, please move this import to the test method.

nessita: Great catch, thank you! fixing and rebasing and squashing and merging :-)

PR #17531, opened 2023-11-27
Refs #21442 -- Improved requests test coverage

smithdc1:
This PR improves test coverage for requests. It includes additional coverage for some missing lines and also aims to test additional scenarios for `request.POST`. I can split these out if that would be helpful. 

The additional `request.POST` scenarios include:

- Data being provided for other methods other than `POST`.
- `CONTENT_TYPE` = `application/json`
- Mutipart with JSON,
- Multipart with JSON + name/value, 
- Multipart with JSON + name/value + CSV, 
- Multipart with JSON + name/value + FILE

@carltongibson  -- did you have any other scenarios on your list 📓


carltongibson (approved):
Hey @smithdc1 — this looks great. 

I think it captures everything we need for regression test purposes. 

_Maybe_ I have an example or two that we might want to look at when adding support for other content types but I don't think they're any different in structure from the ones you have for capturing the existing behaviour. (≈_You get a string_ 😜)

felixxm (commented):
@smithdc1 Thanks :+1: 

I pushed tiny edits.

PR #17553, opened 2023-11-29
Warn about persistent connection "feature" in Django 4.0+

retupmoca:
Persistent database connections do not work in Django 4.0+ when running under ASGI; this caused a "too many database connections" problem that was difficult for me to find any information on.

Since the ticket (https://code.djangoproject.com/ticket/33497) is marked as a feature and has been open for almost 2 years, I think it would be worth calling this scenario out in the documentation for anyone else who might hit it.

### DIFF
--- a/docs/ref/databases.txt
+++ b/docs/ref/databases.txt
@@ -40,6 +40,12 @@ database connection at the end of each request. To enable persistent
 connections, set :setting:`CONN_MAX_AGE` to a positive integer of seconds. For
 unlimited persistent connections, set it to ``None``.
 
+.. warning::
+
+    Persistent connections :ticket:`do not work <33497>` when running Django
+    under ASGI. Setting :setting:`CONN_MAX_AGE` to anything other than ``0``
+    may cause you to exceed your database connection limit.
+
 Connection management
 ~~~~~~~~~~~~~~~~~~~~~
 


docs/ref/databases.txt
```diff
@@ -40,6 +40,12 @@ database connection at the end of each request. To enable persistent
 connections, set :setting:`CONN_MAX_AGE` to a positive integer of seconds. For
 unlimited persistent connections, set it to ``None``.
 
+.. warning::
+
+    Persistent connections :ticket:`do not work <33497>` when running Django
```
lhmzhou, line 45: Would it make sense to add more context to this warning? I suggest adding some actionable items, too:

Spaghetti thoughts:

When running Django under ASGI, be cautious with persistent connections, as they may encounter issues, as documented in `[this ticket]()`. A crucial consideration is the setting `CONN_MAX_AGE`, which, when set to a value other than 0, might lead to exceeding your database connection limit.

To address this:

1. Check Ticket 33497: Refer to `[Ticket 33497]()` for updates and community discussions regarding persistent connection issues in Django ASGI.

2. Review CONN_MAX_AGE Setting: Evaluate the necessity of setting CONN_MAX_AGE to a non-zero value in your specific use case. Consider potential impacts on database connection limits.

3. Monitor Connection Limits: Keep a close eye on your database connection limit, especially when utilizing persistent connections. Frequent monitoring helps you stay within the allowable limits and prevent disruptions.

PR #17554, opened 2023-11-29
Fixed #28586 -- Added model field lazy loading modes.

adamchainz:


adamchainz:
@charettes I’ve been slowly but steadily chipping away at this. Would you like to make an initial review, particularly of the public API in the release notes, and the way that the callable interface works?

charettes:
@adamchainz sure! Can it wait this weekend? I'm taking a long break for Easter so I should have plenty of focus time.

adamchainz:
Of course, take your time.

charettes (commented):
👋 @adamchainz I just want to start by saying thank you for the continued work here, this is really looking promising 🙇 

I discussed an idea on how defining a form of fetching _protocol_ (like how _resolvable_ w/ `resolve_expression` and _compilable_ w/ `as_sql` work) to avoid unnecessary closures and keep fetcher modes to the strict minimum which is to receive an instance, a fetcher, and route the call to the fetcher.

There also seems to be a missing usage of fetch mode in `ReverseOneToOneDescriptor.__get__` but maybe this is expected for now as this is still a WIP.

Lastly I left some comments with regards to naming. I'm sure others will have opinions as well but it felt like `django.db.models.fetching` and `fetching_mode` vs `lazy` might be a better .

django/db/models/fields/related_descriptors.py
```diff
@@ -509,16 +517,34 @@ def __get__(self, instance, cls=None):
...
+                    if (peer := weakref_peer()) is not None and not self.is_cached(peer)
+                ]
+                if len(peers) > 1:
+                    # If the instance was fetched with other instances then
+                    # prefetch the field for all of them
+                    qs = self.get_queryset()
+                    prefetch_related_objects(
+                        peers, Prefetch(self.related.get_accessor_name(), queryset=qs)
+                    )
+                    rel_obj = self.related.get_cached_value(instance)
```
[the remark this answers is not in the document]
adamchainz: Yes, I just didn’t update this yet from Andreas’ original PR copying from django-auto-prefetch, yet.

django/db/models/lazy.py
```diff
@@ -0,0 +1,63 @@
+from contextlib import contextmanager
```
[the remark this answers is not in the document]
adamchainz: I tried describing the feature name a few different ways before settling on “lazy-loading modes” as in the release notes. But “lazy-fetching modes” could also be a candidate.

Happy to rename the module, but I’d prefer to keep it aligned with the name of the feature.

I agree that “lazy” is used in many ways - translations, “lazy relationships” for passing a string to Foreign Keys, `SimpleLazyObject`... But it does also accurately describes how the data is lazily fetched by a field on access. Naming things is hard.

django/db/models/lazy.py
```diff
@@ -0,0 +1,63 @@
...
+    fetch_for_instances(instances)
+
+
+class LazyFieldAccess(Exception):
+    """Blocked lazy access of a model field."""
+
+    pass
+
+
+def RAISE(instance, field, fetch_for_instances, **kwargs):
```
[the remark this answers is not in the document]
adamchainz: I was thinking of that, but then maybe users would want “fetch peers” + warnings, or log messages directly rather than warnings. Maybe something we can instead document the pattern for? Like:

```python
def warn_then_fetch_peers(...):
    logger.warning("fetching peers!")
    return FETCH_PEERS(...)

set_default_lazy_mode(warn_then_fetch_peers)
```

django/db/models/query_utils.py
```diff
@@ -203,7 +204,22 @@ def __get__(self, instance, cls=None):
...
+                        value_by_pk = {
+                            pk: value
+                            for pk, value in self.field.model._base_manager.db_manager()
+                            .filter(pk__in={i.pk for i in instances})
+                            .values_list("pk", attname)
+                        }
+                        for instance in instances:
+                            setattr(instance, attname, value_by_pk[instance.pk])
+
+                get_lazy_mode()(instance, self.field, fetch_for_instances)
```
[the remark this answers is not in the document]
adamchainz: Yeah, that would be good. I originally tried an interface but it was hard for some reason. At least, it was confusing whether methods should live on the descriptors or fields. I will give it another try, making the descriptors the “fetchers”.

Fetch one and many have separate implementations here only for backwards compatibility. Since its introduction, `refresh_from_db()` [has been documented](https://docs.djangoproject.com/en/5.0/ref/models/instances/#django.db.models.Model.refresh_from_db) as the backing method for deferred attributes:

> Note that when deferred fields are accessed, the loading of the deferred field’s value happens through this method.

There are more backwards compatibility concerns for related descriptors, but there won’t be for `GenericForeignKey` (TODO), so maybe we should only have one method.

adamchainz (commented):
Thank you Simon, thoughtful as always. (P.S. just finished your Django Chat episode yesterday, nice one!). I’ve replied inline to your comments and will work on an update.

> There also seems to be a missing usage of fetch mode in `ReverseOneToOneDescriptor.__get__` but maybe this is expected for now as this is still a WIP.

Yes, this needs doing, as does `GenericForeignKey`.

django/db/models/lazy.py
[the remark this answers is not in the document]
adamchainz: I’ve tried that for now, and updated the release notes to avoid the word lazy. I went for “on-demand fetching”, which makes sense to me. WDYT?

adamchainz:
> Yes, this needs doing, as does `GenericForeignKey`.

I tried `GenericForeignKey` just now, but it’s complicated:

https://github.com/django/django/blob/ec8552417df51df8482df61b8ad78a7002634011/django/contrib/contenttypes/fields.py#L234-L267

A `fetch()` function would need to perform all this validation per-instance. I think for ease of coding and review, we should tackle it in a follow-up PR as an optimization.

django/db/models/lazy.py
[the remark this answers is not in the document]
charettes: I see, that makes sense to separate non-terminal _surfacing_ of _leaks_ from the actual fetching implementation. I wonder if there is anything that we could provide to avoid such boiler plate though.

django/db/models/query_utils.py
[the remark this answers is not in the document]
charettes: That looks great! Any reason for going with a single `fetch` method over two distinct `fetch_(one|many)`?

I don't see it as a hard blocker but the fact that all `fetch` implementations have an `if` branch that is unnecessary from the caller's perspective (the mode handler knows which method it should call) seemed odd to me.

django/db/models/lazy.py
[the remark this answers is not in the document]
charettes: I think your proposed structure and wording under `fields.fetching` makes a lot of sense. Thank you for doing that!

charettes:
This is looking great @adamchainz thanks for pushing this through and sorry for my slow replies.

If you'd like I could give a shot at implementing the `GenericForeignKey` part in a separate branch that targets this one. Even if it doesn't land in the same PR it would be great to validate that our approach works properly and possibly dog-food the newly introduce [GenericPrefetch](https://docs.djangoproject.com/en/5.0/ref/contrib/contenttypes/#django.contrib.contenttypes.prefetch.GenericPrefetch).

charettes:
Completely untested code but from eye balling the other `fetch` methods the following should work without too much issues?

```
diff --git a/django/contrib/contenttypes/fields.py b/django/contrib/contenttypes/fields.py
index 770f88265c..2b7597623b 100644
--- a/django/contrib/contenttypes/fields.py
+++ b/django/contrib/contenttypes/fields.py
@@ -255,16 +255,30 @@ def __get__(self, instance, cls=None):
                 return rel_obj
             else:
                 rel_obj = None
-        if ct_id is not None:
-            ct = self.get_content_type(id=ct_id, using=instance._state.db)
-            try:
-                rel_obj = ct.get_object_for_this_type(
-                    using=instance._state.db, pk=pk_val
-                )
-            except ObjectDoesNotExist:
-                pass
-        self.set_cached_value(instance, rel_obj)
-        return rel_obj
+
+        get_fetching_mode()(self, instance)
+        return self.get_cached_value(instance)
+
+    def fetch(self, instances):
+        f = self.model._meta.get_field(self.ct_field)
+        if len(instances) == 1:
+            ct_id = getattr(instance, f.attname, None)
+            pk_val = getattr(instance, self.fk_field)
+            instance = instances[0]
+            if ct_id is not None:
+                ct = self.get_content_type(id=ct_id, using=instance._state.db)
+                try:
+                    rel_obj = ct.get_object_for_this_type(
+                        using=instance._state.db, pk=pk_val
+                    )
+                except ObjectDoesNotExist:
+                    pass
+            self.set_cached_value(instance, rel_obj)
+        else:
+            prefetch_related_objects(
+                instances, f.name
+            )
+
```

django/db/models/fields/fetching.py
```diff
@@ -0,0 +1,64 @@
...
+    fetcher.fetch(instances)
+
+
+class LazyFieldAccess(Exception):
+    """Blocked lazy access of a model field."""
+
+    pass
+
+
+def RAISE(fetcher, instance, **kwargs):
```
charettes, line 30: Should we get rid of the `**kwargs`? Were they meant to be used for something?

charettes:
Gave a shot at it here https://github.com/adamchainz/django/pull/4

django/db/models/fields/fetching.py
```diff
@@ -0,0 +1,64 @@
...
+
+def set_default_fetching_mode(mode):
+    global _default
+    if not callable(mode):  # TODO: verify signature
+        raise TypeError("mode must be callable.")
+    _default = mode
+
+
+@contextmanager
+def lazy_mode(mode):
```
charettes, line 48: Should we rename `fetching_mode`?

django/db/models/fields/fetching.py
adamchainz: I left them to leave space for future extensions, if users wrap the calls. But I guess we don’t have that for the callables in `django.db.models.deletion`, so we can drop them.

tolomea:
Thanks Adam for working to push this thing to it's final form. I wonder when the 10 year anniversary of our first version is.

I've had a skim, I plan to look at the details a bit more.

I quite like the fetching modes mechanism and name.

I don't know if this has been covered somewhere but separate modes for deferred fields and FK(and friends) makes a lot of sense to me and is probably what I will look to do in my own code.
With the current API I guess it's possible by doing isinstance checks inside a custom mode.
But that's not very obvious or discoverable.

My rationale here is when you fetch an object from the DB, unless you do stuff to say otherwise all the primitive fields are fetched and none of the FK's.
Absence of an FK is the default it means no one explicitly asked for it.
Absence of a primitive field is unusual it means someone explicitly said don't load this.
So I would want a config that automatically fetches the FK's (in bulk) and errors on the primitive fields.

django/db/models/fields/fetching.py
tolomea: I'm not that familiar with the back compat policy but the kwargs feels unnecessary
if I had a custom fetch mode I would understand if a feature release broke it due to signature change and it would break loud and fast

django/db/models/fields/related_descriptors.py
```diff
@@ -254,13 +255,9 @@ def __get__(self, instance, cls=None):
             else:
                 rel_obj = None
             if rel_obj is None and has_value:
-                rel_obj = self.get_object(instance)
-                remote_field = self.field.remote_field
-                # If this is a one-to-one relation, set the reverse accessor
-                # cache on the related object to the current instance to avoid
-                # an extra SQL query if it's accessed later on.
-                if not remote_field.multiple:
-                    remote_field.set_cached_value(rel_obj, instance)
```
tolomea, line 263: did this move? or go away?

tolomea:
Also further to the above the doco encourages overridding refresh_from_db which makes setting FETCH_PEERS for deferred fields in your project a third party library compatibility risk as they might done that, although it does feel unlikely.

I wonder if it could / should detect a custom refresh_from_db in FETCH_PEERS and... warn? fall back to FETCH_ONE?

Likewise with the other backward compatibility things.

FWIW I looked though old notes and it seems like the OG version was sometime around September 2016 (later than I remember) so this might still make it into core before the 10th anniversary.

django/db/models/fields/fetching.py
adamchainz: Yes, but I went with `fetch_mode` to keep it shorter.

adamchainz:
Hello all! @pelme pushed me to continue working on this 👍  I have just pushed a new version with these changes:

1. Rebased onto `main` with updates to target Django 6.0.
2. Merged in the branch from @charettes for `GenericForeignKey` support, plus a tweak to use a separate `GenericForeignKeyFetcher` class so it works with the `RAISE` method. (We could also split a descriptor class from `GenericForeignKey` to achieve this, but I’m concerned about backwards-compatibility, despite it not being public API.)
3. Renamed things to cover “fetch modes” (not *fetching*) consistently.
4. Added a database topic guide on fetch modes.

> I don't know if this has been covered somewhere but separate modes for deferred fields and FK(and friends) makes a lot of sense to me and is probably what I will look to do in my own code.
> With the current API I guess it's possible by doing isinstance checks inside a custom mode.
> But that's not very obvious or discoverable.

I think I‘d rather not split into two separate fetch modes/features. That would be harder to explain and maintain. I would rather just allow custom modes to do this, and document how to. But that would require covering lots of internal APIs right now, which is why I’ve held back on it.

> FWIW I looked though old notes and it seems like the OG version was sometime around September 2016 (later than I remember) so this might still make it into core before the 10th anniversary.

🤞 

---

Review would be appreciated. I’m going to try testing this branch against some real projects to check for edge cases.

django/db/models/query_utils.py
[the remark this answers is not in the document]
adamchainz: Now using `fetch_one` and `fetch_many`! No need for a branch.

django/db/models/fields/fetching.py
adamchainz: I got rid of `**kwargs`. If we document “make your own mode”, we can encourage using them there.

pelme (commented):
Amazing work! This looks great! Me, @Majsvaffla, @nip3o and @saeldion have made this review together. We are very much looking forward having fetch modes in Django.

django/db/models/fields/fetch_modes.py
```diff
@@ -0,0 +1,63 @@
...
+
+    orig = getattr(_local, "mode", None)
+    _local.mode = mode
+    try:
+        yield
+    finally:
+        if orig is None:
+            del _local.mode
+        else:
+            _local.mode = orig
```
charettes, line 59: I've brought it before but the fact the fetch mode is a global variable as opposed to attributed to `QuerySet` instances is likely going to make ergonomics quite weird when wanting to mix modes for distinct querysets or enable it only in some cases.

I believe the way database routing is implemented for example where both `QuerySet.using` and routers can be used to determine what database should be used is battle proven approach. If you then want to rely on global state decorators to advise your router decision you can but Django has lived quite well without having core support for that in database routers.

The lack of queryset method / hook prevents usage of this feature in declarative scenarios such as admins, CBV, or anything that takes a `QuerySet` as a parameter (form, serializer).

e.g.

```python
class BookView(ListView):
   queryset = Book.objects.fetch_mode(FETCH_PEERS)


class AuthorManager(Manager):
   def get_queryset(self):
       return super().get_queryset().fetch_mode(RAISE)

class Author(Model):
    objects = AuthorManager()
```

django/db/models/query.py
```diff
@@ -142,6 +145,16 @@ def __iter__(self):
                 else:
                     setattr(obj, field.name, rel_obj)
 
+            # Create peers, but only for >1 result, to save memory.
+            if first_obj is None:
+                first_obj = obj
+            else:
```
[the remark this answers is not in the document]
charettes: If the fetching mode was bound to the current queryset instance this whole logic could be delegated to the fetcher itself. That would allow the `FetchPeer` logic to handle the `list[weakref]` creation only when needed while avoiding its creation when dealing with other fetching modes.

charettes (commented):
I left a few comments about  the global state fetcher approach, `__set_name__` manual calls, systematic creation of `peers: list[weakref]`.

docs/topics/db/fetch-modes.txt
```diff
@@ -0,0 +1,83 @@
...
+
+* .. attribute:: FETCH_PEERS
+
+    Fetches the missing field for all instances that came from the same
+    ``QuerySet``. This behavior is based on the assumption that if you need a
+    field for one instance, you probably need it for all instances in the same
+    batch, because you’re likely looping over them.
+
+    This mode reduces most cases of the “N+1 queries problem” to two queries.
+
```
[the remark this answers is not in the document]
adamchainz: I added the detail in the above sentence:

```rst
Such “peer” instances are tracked in a weak reference list.
```

I don't think we need to use an example, but I agree it's an important detail, at least to allay concerns about memory leaking.

docs/topics/db/fetch-modes.txt
```diff
@@ -0,0 +1,83 @@
...
+Fetch modes
+===========
+
+.. module:: django.db.models.fields.fetch_modes
+
+.. currentmodule:: django.db.models
+
+When accessing model fields that were not loaded as part of the original query,
+Django will fetch that field’s data from the database. You can customize the
+behavior of this fetching with a **fetch mode**.
```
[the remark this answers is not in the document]
adamchainz: I have added a list of where field modes apply. They do not apply to related managers or `ManyToManyField` fields. There may be a way to get them to work with `.all()`, because that can be prefetched, but I'm not sure about anything else.

adamchainz:
> I've brought it before but the fact the fetch mode is a global variable as opposed to attributed to `QuerySet` instances is likely going to make ergonomics quite weird when wanting to mix modes for distinct querysets or enable it only in some cases.

I have just pushed a new version that is implemented solely as a `QuerySet` method.

After working through it, I agree with you, @charettes, that this cuts with the grain of Django. While the global state is convenient, it could easily mess up some expectations in third-party apps.

I thought it would be nice to allow swapping the default fetch mode, but from some experience testing this PR, I’m not 100% sure it’s a great idea. On one project that I tested, using the `RAISE` fetch mode broke migrations because some migration `RunSQL` operations in Wagtail relied on the default `FETCH_ONE`-style behaviour. Opting in individual `QuerySet`s, or models through a custom manager is the safer option, if a little more involved.

I will respond to the other review comments later today, and I also intend to work on some further changes:

- [ ] Nesting: copy a model instance’s fetch mode to related objects fetched through the fetch mode
- [ ] Nesting: copy a model instance’s fetch mode through its related managers
- [ ] Update `select_related()` and `prefetch_related()` docs to refer to fetch modes

docs/topics/db/optimization.txt
```diff
@@ -219,6 +219,15 @@ them:
 * in view code or other layers, possibly making use of
```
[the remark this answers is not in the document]
adamchainz: Done, and restructured the whole section.

docs/topics/db/optimization.txt
```diff
@@ -219,6 +219,15 @@ them:
...
 
+Use the ``FETCH_PEERS`` fetch mode
+----------------------------------
+
+Use the :attr:`~django.db.models.FETCH_PEERS` fetch mode to reduce the number
+of queries when fetching fields on-demand for multiple instances. This can
+mitigate a lack of :meth:`~django.db.models.query.QuerySet.select_related` or
+:meth:`~django.db.models.query.QuerySet.prefetch_related` usage when it’s hard
+to predict which fields will be accessed.
+
```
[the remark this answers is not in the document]
adamchainz: Done and I tried to emphasize that `FETCH_PEERS` should be considered first.

django/db/models/query.py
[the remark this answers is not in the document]
adamchainz: This is done.

I also added peers during `RawModelIterable`, making fetch modes work for instances retrieved through raw queries.

docs/releases/6.0.txt
```diff
@@ -37,6 +37,56 @@ compatible with Django 6.0.
...
+
+    from django.db.models import fetch_mode, RAISE
+
+
+    def dashboard(request):
+        ...
+        # Raise a FieldFetchBlocked exception for any on-demand queries
+        # attempted during template rendering
+        with fetch_mode(RAISE):
+            return render(request, "dashboard.html", context)
```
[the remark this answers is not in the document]
adamchainz: Outdated now as the context manager/decorator and global state is gone, in favour of the per-QS method.

docs/topics/db/fetch-modes.txt
```diff
@@ -0,0 +1,83 @@
...
+    default mode.
+
+* .. attribute:: FETCH_PEERS
+
+    Fetches the missing field for all instances that came from the same
+    ``QuerySet``. This behavior is based on the assumption that if you need a
+    field for one instance, you probably need it for all instances in the same
+    batch, because you’re likely looping over them.
+
+    This mode reduces most cases of the “N+1 queries problem” to two queries.
```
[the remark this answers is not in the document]
adamchainz: Done.

docs/topics/db/fetch-modes.txt
```diff
@@ -0,0 +1,83 @@
...
+
+        from django.db.models import fetch_mode, RAISE
+
+
+        def dashboard(request):
+            ...
+            # Raise a FieldFetchBlocked exception for any on-demand queries
+            # attempted during template rendering
+            with fetch_mode(RAISE):
+                return render(request, "dashboard.html", context)
```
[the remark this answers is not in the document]
adamchainz: Now irrelevant with the queryset-based implementation.

docs/topics/db/fetch-modes.txt
```diff
@@ -0,0 +1,83 @@
...
+
+            def ready(self):
+                set_default_fetch_mode(FETCH_PEERS)
+
+.. function:: fetch_mode(mode)
+
+    A function decorator and context manager that sets the fetch mode for a
+    section of code. This mode will be used for all field fetching within the
+    decorated function or context, within the current thread or asynchronous
+    task.
```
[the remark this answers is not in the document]
adamchainz: I think the new queryset-based API makes it clear which objects and fetches the fetch mode applies to.

docs/topics/db/fetch-modes.txt
```diff
@@ -0,0 +1,83 @@
...
+
+When accessing model fields that were not loaded as part of the original query,
+Django will fetch that field’s data from the database. You can customize the
+behavior of this fetching with a **fetch mode**.
+
+Available modes
+===============
+
+Django provides three modes:
+
```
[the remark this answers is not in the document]
adamchainz: I’ve updated the new docs with an example that’s explained for each built-in mode.

adamchainz:
Re @tolomea 

> I don't know if this has been covered somewhere but separate modes for deferred fields and FK(and friends) makes a lot of sense to me and is probably what I will look to do in my own code.

I understand the rationale for splitting behaviour between FKs and deferred fields, but I think actually splitting them in the API would be too fine-grained. Fetching deferred fields in bulk is still useful, such as for [model classes that defer some fields by default](https://adamj.eu/tech/2023/12/11/django-always-defer-a-field/).

> With the current API I guess it's possible by doing isinstance checks inside a custom mode. But that's not very obvious or discoverable.

Yes, this would be the way. We could document a public API around combining fetch modes like this, but maybe in a follow-up PR.

> Also further to the above the doco encourages overridding refresh_from_db which makes setting FETCH_PEERS for deferred fields in your project a third party library compatibility risk as they might done that, although it does feel unlikely.

Thankfully, this risk is gone by adopting a `QuerySet`-based API. Instead, third-party libraries will need to opt into using a different fetch mode.

django/db/models/fields/__init__.py
```diff
@@ -954,7 +954,9 @@ def contribute_to_class(self, cls, name, private_only=False):
         self.model = cls
         cls._meta.add_field(self, private=private_only)
         if self.column:
-            setattr(cls, self.attname, self.descriptor_class(self))
+            descriptor = self.descriptor_class(self)
+            setattr(cls, self.attname, descriptor)
+            descriptor.__set_name__(cls, self.name)
```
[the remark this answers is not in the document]
adamchainz: The aim here was to have the "fetcher protocol" include a consistent way to see the related field's name available, so the `RAISE` mode can rely on it.

Descriptors did not consistently point at fields. At least `GenericForeignKey` is, oddly, its own descriptor. Adding a recursive `field` attribute there felt wrong, but it seemed to make sense to call [`__set_name__`](https://docs.python.org/3/reference/datamodel.html#object.__set_name__) here because it's the same point that Python would do so if the descriptor were placed in the class definition.

In retrospect, it's probably less invasive and more sensible to make `field` consistently available, and fix GFK to have a separate descriptor class. I'll give that a try now.

django/db/models/fields/related_descriptors.py
```diff
@@ -133,6 +134,9 @@ class Child(Model):
     def __init__(self, field_with_rel):
         self.field = field_with_rel
 
+    def __set_name__(self, owner, name):
+        self.name = name
+
```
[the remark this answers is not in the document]
adamchainz: Per above - trying with field.

https://github.com/django/django/pull/17554#discussion_r2042249701

adamchainz:
I've just pushed an update with a new commit that does the "nesting", copying the fetch mode to related objects to apply to a whole tree of objects unless explicitly replaced.

I'm still going to work a bit on the `select_related` and `prefetch_related` docs, but I think we're nearly there on the `QuerySet`-based design.

django/db/models/fields/__init__.py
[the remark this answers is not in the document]
adamchainz: Okay, I just pushed a fix. It required a new commit that splits a descriptor class from `GenericForeignKey`, which I also put into a new PR at #19381. Another minor hack it required is a new property in `ReverseOneToOneDescriptor` for `field`, with a docstring explaining the ducktyping.

adamchainz:
I just fixed the test failure. (I wrote *behaviour* instead of *behavior*, failing the spellcheck!)

docs/ref/models/querysets.txt
```diff
@@ -1149,12 +1160,14 @@ that is that ``select_related('foo', 'bar')`` is equivalent to
 
 .. method:: prefetch_related(*lookups)
 
-Returns a ``QuerySet`` that will automatically retrieve, in a single batch,
-related objects for each of the specified lookups.
+Returns a ``QuerySet`` that will automatically retrieve the given lookups, each
+in one extra batch query. Prefetching is a way to optimize database access
+when you know you'll be accessing related objects later, so you can avoid
+triggering triggering the on-demand loading behavior of the model instances’
```
Zerotask, line 1166:
```suggestion
triggering the on-demand loading behavior of the model instances’
```

adamchainz: Thank you for spotting!

django/db/models/query_utils.py
```diff
@@ -241,6 +242,20 @@ def _check_parent_chain(self, instance):
...
 
+    def fetch_one(self, instance):
+        instance.refresh_from_db(fields=[self.field.attname])
+
+    def fetch_many(self, instances):
+        attname = self.field.attname
+        value_by_pk = {
+            pk: value
+            for pk, value in self.field.model._base_manager.db_manager()
+            .filter(pk__in={i.pk for i in instances})
```
g-nie, line 253: I wonder if the set comprehension here could have some performance impact for a large number of `instances`. I'm not sure what else could be done instead though.

g-nie (commented):
Thanks @adamchainz this looks great! I left a small thought

django/db/models/query_utils.py
adamchainz: It shouldn't take a meaningful amount of time. Yes, it's for N instances, but gathering their in-memory PKs will be way faster than the actual query.

docs/topics/db/optimization.txt
```diff
@@ -196,28 +196,45 @@ thousands of records are returned. The penalty will be compounded if the
...
-* in :doc:`managers and default managers </topics/db/managers>` where
-  appropriate. Be aware when your manager is and is not used; sometimes this is
-  tricky so don't make assumptions.
+Use ``prefetch_related_objects()``
+----------------------------------
 
-* in view code or other layers, possibly making use of
-  :func:`~django.db.models.prefetch_related_objects` where needed.
+Where :attr:`~django.db.models.query.QuerySet.prefetch_related` would be useful
+“after the fact”, use :func:`~django.db.models.prefetch_related_objects` to
```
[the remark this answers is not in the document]
adamchainz: Yes, that's much better, thanks!

docs/topics/db/queries.txt
```diff
@@ -1665,8 +1665,14 @@ you, but we point it out here for the curious.)
...
 Django also creates API accessors for the "other" side of the relationship --
 the link from the related model to the model that defines the relationship.
-For example, a ``Blog`` object ``b`` has access to a list of all related
-``Entry`` objects via the ``entry_set`` attribute: ``b.entry_set.all()``.
+For example, a ``Blog`` object ``b`` has a manager that returns all related
+``Entry`` objects in the ``entry_set`` attribute: ``b.entry_set.all()``.
+
+These accessors may be pre-fetched by the ``QuerySet`` methods
+:meth:`~django.db.models.query.QuerySet.select_related` or
+:meth:`~django.db.models.query.QuerySet.prefetch_related`. If not pre-fetched,
```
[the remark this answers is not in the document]
adamchainz: Done, thanks!

docs/ref/models/querysets.txt
```diff
@@ -1149,12 +1160,14 @@ that is that ``select_related('foo', 'bar')`` is equivalent to
...
+in one extra batch query. Prefetching is a way to optimize database access
+when you know you'll be accessing related objects later, so you can avoid
+triggering the on-demand loading behavior of the model instances’
+:doc:`fetch mode </topics/db/fetch-modes>`.
 
-This has a similar purpose to ``select_related``, in that both are designed to
-stop the deluge of database queries that is caused by accessing related objects,
-but the strategy is quite different.
+This method has a similar purpose to :meth:`select_related`, in that both are
+designed to eagerly fetch related objects. However, they work in different ways.
```
jacobtylerwalls, line 1170: I would expect chaining `.prefetch_related(None)` to clear a FETCH_PEERS fetch mode, but it doesn't -- it continues to fetch peers. What is your view on what should happen?

tolomea: currently the two API's are manipulating independent state, it'd feel weird to clear it like this instead of using fetch_mode
of course your other suggestion does couple them in another way

docs/topics/db/optimization.txt
```diff
@@ -196,28 +196,45 @@ thousands of records are returned. The penalty will be compounded if the
...
-query that is executed in a loop, and could therefore end up doing many database
-queries, when only one was needed. So:
+Use the :attr:`~django.db.models.FETCH_PEERS` fetch mode to make on-demand
+field access more efficient with bulk-fetching. Enable all it for all usage of
+your models :ref:`with a custom manager <fetch-modes-custom-manager>`.
+
+Using this fetch mode is easier than declaring fields to fetch with
+:meth:`~django.db.models.query.QuerySet.select_related` or
+:meth:`~django.db.models.query.QuerySet.prefetch_related`, especially when it's
+hard to predict which fields will be accessed.
```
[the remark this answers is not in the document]
tolomea: that select_related() call is almost always a bad idea, it's very prone to over fetching and to doing so in a way that is fairly hostile to the database, I'd be more in favor of removing that behaviour

docs/ref/models/querysets.txt
jacobtylerwalls: Makes sense. (To my ear the docs advertised this as the successor to `prefetch_related()`, but looking again the docs clarify that there is a separate use case for `prefetch_related()`, e.g. in managers, so I'm convinced these are really two APIs after all.)

adamchainz: Yeah, they are intended as two separate APIs. I tried to clarify that fetch modes apply to “on-demand fetches”, the existing behaviour where accessing an attribute fetches data. I think it’s clear enough, but if you have any suggested docs tweaks, I’m happy to see them.

docs/topics/db/optimization.txt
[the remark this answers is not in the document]
charettes: Strong +1 to deprecating `select_related()` without a mask as a follow up that's almost always a bad idea.

django/contrib/contenttypes/fields.py
```diff
@@ -157,11 +156,19 @@ def get_content_type(self, obj=None, id=None, using=None, model=None):
             # This should never happen. I love comments like this, don't you?
             raise Exception("Impossible arguments to GFK.get_content_type!")
 
+
+class GenericForeignKeyDescriptor:
```
charettes, line 165: Just to make sure I understand the rationale behind this change.

The same thing could be achieved by adding a `field` property on `GenericForeignKey` that returns `self` but a distinct class is deemed preferable as it makes `GenericForeignKey` follow the pattern used by other related fields.

It might be worth _explaining_ the rationale behind these changes in the commit message and the benefits it provides for the followed commits in the patch set for future reference.

docs/topics/db/queries.txt
```diff
@@ -1649,14 +1649,14 @@ a join with an ``F()`` object, a ``FieldError`` will be raised:
...
 ===============
 
-When you define a relationship in a model (i.e., a
+When you define a relationship in a model (with
 :class:`~django.db.models.ForeignKey`,
 :class:`~django.db.models.OneToOneField`, or
-:class:`~django.db.models.ManyToManyField`), instances of that model will have
-a convenient API to access the related object(s).
+:class:`~django.db.models.ManyToManyField`), instances of the model class gain
+accessor attributes for the related object(s).
```
charettes, line 1690: Again sensible chain but a commit description would go a long way explain _why_ this is being done here for future reference.

django/contrib/contenttypes/fields.py
```diff
@@ -257,7 +267,10 @@ def __get__(self, instance, cls=None):
             except ObjectDoesNotExist:
                 pass
         self.field.set_cached_value(instance, rel_obj)
-        return rel_obj
+
+    def fetch_many(self, instances):
+        missing_instances = [i for i in instances if not self.field.is_cached(i)]
```
charettes: Minor but it might be worth using a local reference for `self.field.is_cached` to avoid two attribute lookups per instance in a potentially large loop

```python
is_cached = self.field.is_cache
missing_instances = [i for i in instances if not is_cached(i)]
```

django/db/models/base.py
```diff
@@ -473,6 +482,14 @@ class ModelState:
     # on the actual save.
     adding = True
     fields_cache = ModelStateFieldsCacheDescriptor()
+    fetch_mode = ModelStateFetchModeDescriptor()
+    peers = ()
+
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        state.pop("fetch_mode", None)
+        state.pop("peers", None)
```
charettes: Just to make sure I understand, does this mean that a pickled model instances, and by definition a queryset of such objects, looses its fetch mode affinity? For example would that mean that caching a queryset though pickling and then unpickling it would not preserve the specified mode (e.g. raising errors on attributes that would require fetching).

Maybe that's by design but the lack of comments makes me wonder the purpose behind it.

django/db/models/fetch_modes.py
```diff
@@ -0,0 +1,52 @@
...
+class Raise(FetchMode):
+    __slots__ = ()
+
+    def fetch(self, fetcher, instance):
+        klass = instance.__class__.__qualname__
+        field_name = fetcher.field.name
+        raise FieldFetchBlocked(f"Fetching of {klass}.{field_name} blocked.")
+
+
+RAISE = Raise()
```
charettes, line 61: Really like how this turned out, thanks for the tweaks!

django/db/models/fields/related_descriptors.py
```diff
@@ -441,16 +447,20 @@ def RelatedObjectDoesNotExist(self):
     def is_cached(self, instance):
         return self.related.is_cached(instance)
 
-    def get_queryset(self, **hints):
-        return self.related.related_model._base_manager.db_manager(hints=hints).all()
+    def get_queryset(self, *, instance):
+        return self.related.related_model._base_manager.db_manager(
+            hints={"instance": instance}
```
charettes, line 465: I appreciate that a clearer signature is used but it should be done in a distinct commit as it might be used under other circumstances in the wild.

django/db/models/fields/related_descriptors.py
```diff
@@ -441,16 +447,20 @@ def RelatedObjectDoesNotExist(self):
...
         if querysets and len(querysets) != 1:
             raise ValueError(
                 "querysets argument of get_prefetch_querysets() should have a length "
                 "of 1."
             )
-        queryset = querysets[0] if querysets else self.get_queryset()
+        queryset = (
+            querysets[0] if querysets else self.get_queryset(instance=instances[0])
+        )
         queryset._add_hints(instance=instances[0])
```
charettes: Aren't these two `instance` hint passing redundant?

charettes (commented):
Thanks for pushing this forward and sorry for the delay in review @adamchainz.

I left a few comments mainly requesting clarifications and more verbose commit message but in essence the state of this PR is looking great to me 🥇 🙇

django/db/models/fetch_modes.py
```diff
@@ -0,0 +1,52 @@
...
+FETCH_PEERS = FetchPeers()
+
+
+class Raise(FetchMode):
+    __slots__ = ()
+
+    def fetch(self, fetcher, instance):
+        klass = instance.__class__.__qualname__
+        field_name = fetcher.field.name
+        raise FieldFetchBlocked(f"Fetching of {klass}.{field_name} blocked.")
```
jacobtylerwalls: Thanks for this tool, currently using it to debug a project. I noticed that with `RAISE`, the exception raised here will include context like:

```py
  File "/Users/jwalls/django/django/db/models/fields/related_descriptors.py", line 231, in __get__
    rel_obj = self.field.get_cached_value(instance)
  File "/Users/jwalls/django/django/db/models/fields/mixins.py", line 21, in get_cached_value
    return instance._state.fields_cache[self.cache_name]
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
KeyError: 'nodegroup'

During handling of the above exception, another exception occurred:
```

What do you think about raising this error `from None` to suppress that irrelevant context?

docs/topics/db/fetch-modes.txt
```diff
@@ -0,0 +1,140 @@
...
+Fetch modes are copied to related objects when they are fetched, so they can
+apply to a whole tree of relationships, not just the top-level model in the
+initial ``QuerySet``.
+
+.. admonition:: Referencing fetch modes
+
+    Fetch modes are defined in ``django.db.models.fetch_modes``, but for
+    convenience they're imported into :mod:`django.db.models`. The standard
+    convention is to use ``from django.db import models`` and refer to the
+    fetch modes as ``models.<mode>``.
```
blighj, line 45: I like this admonition, but the examples in the docs or the tests with this PR don't use this pattern.
If we are going to encourage this ``models.<mode>`` pattern, I think it should be used in the examples and tests too.

docs/topics/db/optimization.txt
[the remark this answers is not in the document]
nessita: I'm also +1 to deprecating `select_related()`, I've been badly bitten by this a few years ago.

docs/topics/db/fetch-modes.txt
```diff
@@ -0,0 +1,140 @@
...
+Fetch modes apply to:
+
+* :class:`~django.db.models.ForeignKey` fields
+* :class:`~django.db.models.OneToOneField` fields and their reverse accessors
+* Fields deferred with :meth:`.QuerySet.defer` or :meth:`.QuerySet.only`
+* :ref:`generic-relations`
+
+Fetch modes are copied to related objects when they are fetched, so they can
+apply to a whole tree of relationships, not just the top-level model in the
+initial ``QuerySet``.
```
jacobtylerwalls: To my ear, the "whole tree" wording enables the wishful thinking that the fetch mode gets copied everywhere, not merely to related objects, but _to those related objects' related managers_, e.g. this test showing "wishful thinking" fails:

```diff
diff --git a/tests/many_to_one/tests.py b/tests/many_to_one/tests.py
index 3c7f8bbd66..d471a3ec69 100644
--- a/tests/many_to_one/tests.py
+++ b/tests/many_to_one/tests.py
@@ -933,6 +933,10 @@ class ManyToOneTests(TestCase):
             a1.reporter
         with self.assertNumQueries(0):
             a2.reporter
+        with self.assertNumQueries(1):
+            for reporter in [a1.reporter, a2.reporter]:
+                for article in reporter.article_set.all():
+                    print(article)
 
     def test_fetch_mode_raise_forward(self):
         a = Article.objects.fetch_mode(RAISE).get(pk=self.a.pk)
```
```
AssertionError: 2 != 1 : 2 queries executed, 1 expected
Captured queries were:
1. SELECT "many_to_one_article"."id", "many_to_one_article"."headline", "many_to_one_article"."pub_date", "many_to_one_article"."reporter_id" FROM "many_to_one_article" WHERE "many_to_one_article"."reporter_id" = 1 ORDER BY "many_to_one_article"."headline" ASC
2. SELECT "many_to_one_article"."id", "many_to_one_article"."headline", "many_to_one_article"."pub_date", "many_to_one_article"."reporter_id" FROM "many_to_one_article" WHERE "many_to_one_article"."reporter_id" = 2 ORDER BY "many_to_one_article"."headline" ASC
```
(It isn't very realistic to traverse Article -> Reporter -> Article, but you could imagine starting from a third table, I just didn't bother to update the test models.)

Is there a way we can avoid leaving this impression?

This might also help fill in the missing details in the part that describes if peers mode "is not appropriate or efficient enough" -- it could feel like some useful information is being withheld.

jacobtylerwalls: By the way, what do you think is the level of effort for copying the fetch mode to the related descriptors? Or would it be a bad idea?

Peers mode is nice because you can shave off a query if the relation is empty, but if the next level relation has to be prefetched because you haven't implemented a manager on the related model, you have to use prefetch_related, and there goes the lazy savings from peers mode.

docs/topics/db/optimization.txt
```diff
@@ -196,28 +196,45 @@ thousands of records are returned. The penalty will be compounded if the
 database lives on a separate server, where network overhead and latency also
 play a factor.
 
-Retrieve everything at once if you know you will need it
-========================================================
+Retrieve related objects efficiently
+====================================
+
+Generally, accessing the database multiple times to retrieve different parts
+of a single 'set' of data is less efficient than retrieving it all in one
```
jacobtylerwalls:
```suggestion
of a single "set" of data is less efficient than retrieving it all in one
```

django/db/models/query.py
```diff
@@ -291,6 +308,7 @@ def __init__(self, model=None, query=None, using=None, hints=None):
         self._prefetch_done = False
         self._known_related_objects = {}  # {rel_field: {pk: rel_obj}}
         self._iterable_class = ModelIterable
+        self._fetch_mode = FETCH_ONE
```
jacobtylerwalls, line 313: When debugging a project earlier this summer, it was enormously helpful to be able to just monkey-patch this to `RAISE` and comb through failures.

One _interesting_ failure I wasn't able to solve looked like:

```py
request.user.userprofile
```
```py
django.core.exceptions.FieldFetchBlocked: Fetching of User.userprofile blocked.
```

To fix that, I'd need:
- a custom user model
- a custom model manager for that custom model that overrides `get_queryset()` to chain `.fetch_mode(models.FETCH_PEERS)`

My questions are:
- that's the easiest way, right?
- Is there a future where `django.contrib.auth` would ever use `.fetch_mode(models.FETCH_PEERS)` in `get_user()`? Since it's a contrib app? Or would that violate the mental model of "django defaults to FETCH_ONE"?

One query for a `.userprofile` per request isn't so bad, but the fact that I can't "globally" turn on RAISE mode without my tests immediately failing on this is actually what impacts me more.

docs/releases/6.0.txt
```diff
@@ -90,6 +90,50 @@ Notably, the return type of the :class:`EmailMessage.message()
 :py:class:`email.message.EmailMessage`. This supports the same API as the
 previous ``SafeMIMEText`` and ``SafeMIMEMultipart`` return types, but is not an
 instance of those now-deprecated classes.
+Model field fetch modes
```
jacobtylerwalls:
```suggestion

Model field fetch modes
```

jacobtylerwalls (commented):
I'm looking forward to this one @adamchainz. Couple of usage / roadmap questions for you.

django/db/models/query.py
```diff
@@ -1835,6 +1854,12 @@ def using(self, alias):
         clone._db = alias
         return clone
 
+    def fetch_mode(self, fetch_mode):
+        """Set the fetch mode for the QuerySet."""
+        clone = self._chain()
```
jacobtylerwalls, line 1938: `.fetch_mode(None)` fails pretty far from the source of the problem, after the qs is evaluated:
```py
--> 134     if fetch_mode.track_peers:
    135         peers.append(weak_ref(obj))
    136         obj._state.peers = peers

AttributeError: 'NoneType' object has no attribute 'track_peers'
```

Could we fail immediately inside this method, similar to how `.order_by(None)` immediately fails (rather than during query execution)?

django/db/models/query_utils.py
```diff
@@ -281,6 +282,20 @@ def _check_parent_chain(self, instance):
...
```
jacobtylerwalls: Somebody in the wild might have non-hashable primary keys, which as far as I know work just fine (outside of `in_bulk()` for obvious reasons).

The test model for custom pks even lacked a `__hash__` method until 8fd21b0da35697591e86f4eab0035c4360a45144.

Should we use different data structures here to account for this?

docs/topics/db/optimization.txt
```diff
@@ -196,28 +196,45 @@ thousands of records are returned. The penalty will be compounded if the
...
-:meth:`~django.db.models.query.QuerySet.prefetch_related` thoroughly, and use
-them:
+When the :attr:`~django.db.models.FETCH_PEERS` fetch mode is not appropriate or
+efficient enough, use :meth:`~django.db.models.query.QuerySet.select_related`
+and :meth:`~django.db.models.query.QuerySet.prefetch_related`. Understand their
+documentation thoroughly and apply them where needed.
+
+It may be useful to apply these methods in :doc:`managers and default managers
+</topics/db/managers>`. Be aware when your manager is and is not used;
+sometimes this is tricky so don't make assumptions.
```
jacobtylerwalls, line 230: I'm not thrilled about saying something is tricky and immediately moving on. Is there something concrete we can add or link to?

docs/topics/db/fetch-modes.txt
```diff
@@ -0,0 +1,140 @@
...
+    books = Book.objects.fetch_mode(FETCH_PEERS)
+
+Fetch modes apply to:
+
+* :class:`~django.db.models.ForeignKey` fields
+* :class:`~django.db.models.OneToOneField` fields and their reverse accessors
+* Fields deferred with :meth:`.QuerySet.defer` or :meth:`.QuerySet.only`
+* :ref:`generic-relations`
+
+Fetch modes are copied to related objects when they are fetched, so they can
```
jacobtylerwalls: Can we also copy the fetch mode to objects created via `QuerySet.create()`?

django/db/models/base.py
```diff
@@ -595,7 +612,7 @@ def __init__(self, *args, **kwargs):
         post_init.send(sender=cls, instance=self)
 
     @classmethod
-    def from_db(cls, db, field_names, values):
+    def from_db(cls, db, field_names, values, *, fetch_mode=None):
```
jacobtylerwalls, line 615: Should we document the ability to set `._state.fetch_mode` directly? I found this useful when I was downstream of a `RAISE`, but I needed to loosen the mode back to `FETCH_PEERS`, because there was no way to "fix" the underlying N+1 problem if the object I was dealing with came from `create()` (since there is no one-shot `create()` + `select_related()`).

docs/topics/db/optimization.txt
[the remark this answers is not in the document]
adamchainz: Ticket: https://code.djangoproject.com/ticket/36593

django/contrib/contenttypes/fields.py
adamchainz: Yes, precisely. I added this explanation over in the separate PR (#19381):

> This makes GenericForeignKey more similar to other fields which act as
descriptors, preparing it to add “fetcher protocol” support in a clear and
consistent way.

I hope that is sufficient.

django/contrib/contenttypes/fields.py
adamchainz: Done, and similarly in `ForwardManyToOneDescriptor` and `ReverseOneToOneDescriptor` as well.

django/db/models/fetch_modes.py
adamchainz: Done, and expanded the tests to check for this with extra assertions like:

```python
with self.assertRaisesMessage(FieldFetchBlocked, msg) as cm:
    ...
self.assertIsNone(cm.exception.__cause__)
self.assertTrue(cm.exception.__suppress_context__)
```

docs/topics/db/optimization.txt
adamchainz: I totally agree, but this wording was pre-existing, and this PR is already giant. Can we punt this one?

jacobtylerwalls: Happy to handle in a follow-up.

django/db/models/fields/related_descriptors.py
adamchainz: Good point, I have split the signature change into a new commit.

django/db/models/fields/related_descriptors.py
adamchainz: Well spotted, thanks, removed.

django/db/models/base.py
adamchainz: I think we should leave it as a private API for now. Let's allow the tinkerers who read the source to play with it for the first release.

But at least we've documented two attributes of `Model._state` ([link](https://docs.djangoproject.com/en/5.2/ref/models/instances/#django.db.models.Model._state)) so there's precedent to document it more.

django/db/models/query.py
adamchainz: I'm not sure if it's worth special-casing `None` here, given it's not documented… can we push this to a follow-up enhancement?

django/db/models/query_utils.py
adamchainz: After a quick search, I think Django already doesn't support unhashable primary keys. The deletion collector performs a similar set comprehension:

https://github.com/django/django/blob/2722cb61ccae84f593e6d2c28814e3c628743994/django/db/models/deletion.py#L490

Similarly, model forms rely on hashing PKs:

https://github.com/django/django/blob/2722cb61ccae84f593e6d2c28814e3c628743994/django/forms/models.py#L720

Also, because values must be unique in the database, that implies a hash function can be written.

So, I think this isn't a use case we need to worry about.

django/db/models/query.py
adamchainz: > that's the easiest way, right?

I think it would be slightly easier to subclass `ModelBackend` and replace `get_user` with a version that adds the `.fetch_mode()` call.

> Is there a future where `django.contrib.auth` would ever use `.fetch_mode(models.FETCH_PEERS)` in `get_user()`? Since it's a contrib app? Or would that violate the mental model of "django defaults to FETCH_ONE"?

I think there could be such a future. I'd hope we could also change the admin to use `models.FETCH_PEERS` by default.

> One query for a `.userprofile` per request isn't so bad, but the fact that I can't "globally" turn on RAISE mode without my tests immediately failing on this is actually what impacts me more.

Yeah, I found a similar issue in a project I tried the same trick on. That one failed during migrations, because a data migration in a third-party app was relying on the default `FETCH_ONE` behaviour.

I don't think there's a good API we can provide to allow enabling `RAISE` globally without breaking lots of hard-to-fix stuff. That's why I've come around to Simon's suggestion of going with the `QuerySet`-based API. We can always address adding a supported global switch later, but for now having the targeted API seems best.

django/db/models/query.py
jacobtylerwalls: For sure.

django/db/models/query_utils.py
jacobtylerwalls: Appreciate the research, that's reassuring!

django/db/models/fetch_modes.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class FetchOne(FetchMode):
+    __slots__ = ()
+
+    def fetch(self, fetcher, instance):
+        fetcher.fetch_one(instance)
+
+
+FETCH_ONE = FetchOne()
```
jacobtylerwalls, line 23: Last Q: did you consider making this an enum? (can totally punt...)

docs/topics/db/fetch-modes.txt
adamchainz: Done, in a new commit. I wasn't sure if it was worth documenting, since `Model.objects.fetch_mode(...).create(...)` seems like an obvious pattern that should work, and the use case seems kinda niche.

django/db/models/fetch_modes.py
adamchainz: I think initial designs used an enum or constants, but @charettes suggested to use callables, copying the `on_delete` modes like `CASCADE`. Not only is this more consistent, it opens the door for adding custom logic, like including logging or conditionally blocking fetches.

docs/topics/db/fetch-modes.txt
adamchainz: Yes, totally! Thanks for spotting this, fixed.

django/db/models/base.py
adamchainz: Yes, these state removals do mean that pickled instances lose their fetch modes.

I added these lines because weakrefs aren't pickleable, copying from:

https://github.com/adamchainz/django-auto-prefetch/blob/7f3b7aed5b1b029b7ddeb4d5fa9c0d76281b4fe4/src/auto_prefetch/__init__.py#L118-L126

However, it's a good point that unpickled instances shouldn't lose their fetch modes. I guess the best we can do is drop peers only, so `FETCH_PEERS` won't work for unpickled instances. However, it will be propagated, so at least there's that.

I'll change this and add a test for an unpickled instnace.

docs/topics/db/fetch-modes.txt
adamchainz: > By the way, what do you think is the level of effort for copying the fetch mode to the related descriptors? Or would it be a bad idea?

I think the level of effort would be very hard, with significant refactoring. Right now fetch modes are focused on fields, not related managers.

It would also be even more complicated if we wanted to support further operations on related managers, like `reporter.article_set.filter(...)`. This could even introduce the chance of fetching unexpected data, since the chained operations may not be safe to combine across multiple related managers, such as if annotations are used.

> …you have to use prefetch_related, and there goes the lazy savings from peers mode.

I think you can at least use [`prefetch_related_objects()`](https://devdocs.io/django~5.2/ref/models/querysets#django.db.models.prefetch_related_objects) in some cases.

> Is there a way we can avoid leaving this impression?

I'll try to improve the wording here.

adamchainz: New wording:

```rst
Django copies the fetch mode of an instance to any related objects it fetches,
so the mode applies to a whole tree of relationships, not just the top-level
model in the initial ``QuerySet``. This copying is also done in related
managers, even though fetch modes don't affect such managers' queries.
```

I think this should be sufficient, given the list above (“Fetch modes apply to”) also doesn't mention related manager queries.

docs/releases/6.0.txt
```diff
@@ -37,6 +37,56 @@ compatible with Django 6.0.
 What's new in Django 6.0
 ========================
 
+Model field fetch modes
+-----------------------
+
+The on-demand fetching behavior of model fields is now configurable with
+:doc:`fetch modes </topics/db/fetch-modes>`. Django provides three fetch
+modes:
```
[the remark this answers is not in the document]
adamchainz: I've tried editing this paragraph in various ways, but to me, it reads a bit clunky to mention "N+1 queries" here and then very shortly after. I hope we can trust users' attention span to reach point 2, plus we can advertise the feature in other ways, such as blog posts and conference talks.

docs/topics/db/queries.txt
```diff
@@ -1653,8 +1653,14 @@ you, but we point it out here for the curious.)
...
-For example, a ``Blog`` object ``b`` has access to a list of all related
-``Entry`` objects via the ``entry_set`` attribute: ``b.entry_set.all()``.
+For example, a ``Blog`` object ``b`` has a manager that returns all related
+``Entry`` objects in the ``entry_set`` attribute: ``b.entry_set.all()``.
+
+These accessors may be pre-fetched by the ``QuerySet`` methods
+:meth:`~django.db.models.query.QuerySet.select_related` or
+:meth:`~django.db.models.query.QuerySet.prefetch_related`. If not pre-fetched,
+access will trigger an on-demand fetch through the current
+:doc:`fetch mode </topics/db/fetch-modes>`.
```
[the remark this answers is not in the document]
adamchainz: Yes, this is done.

django/db/models/base.py
adamchainz: I ended up adding `__reduce__` methods to fetch modes to make them pickle more efficiently, in a new commit 30c72657ff9d3747eb5d4b59aacce9bac0acdbad.

adamchainz:
Alright, I think I handled all review comments!

django/db/models/fetch_modes.py
jacobtylerwalls: I'm down with the callables, but I just meant wrapping them in an enum like this, mostly just for discoverability:
```py
class FetchModes(enum.Enum):
    FETCH_ONE = FETCH_ONE
    FETCH_PEERS = FETCH_PEERS
    RAISE = RAISE
```

adamchainz: Oh, right… I'm not super convinced, but maybe others would be.

Can we defer this one, too? I think we'd want the same pattern for the deletion modes as well, so it could be done as a cleanup affecting both.

jacobtylerwalls: No problem. Just picking your brain 👍

django/db/models/query_utils.py
jacobtylerwalls: Any interest in observing database parameter limits, similar to how `in_bulk()` [does it](https://github.com/django/django/blob/0ddbe12ea99a2dc1b757dc2015ba8bb6bfd9d653/django/db/models/query.py#L1195), or factoring out some reusable private helper?

charettes: Kind of a shame that `in_bulk` doesn't support `values` / `values_list` in the first place as that could have been

```python
value_by_pk = (
    self.field.model._base_manager
    .values_list(attname, flat=True).in_bulk({i.pk for i in instances})
)
```

in all cases it seems that the `db_manager()` call is unnecessary?

jacobtylerwalls: > Kind of a shame that in_bulk doesn't support values / values_list

I'll open a new features ticket and link to the [sketched](https://github.com/django/django/pull/18497) implementation, given that we have this additional use in Django itself now.

jacobtylerwalls: https://github.com/django/new-features/issues/74

charettes: Also, not sure if the dict literal is faster but the `dict` function can be used as well

```python
        value_by_pk = dict(
            self.field.model._base_manager
            .filter(pk__in={i.pk for i in instances})
            .values_list("pk", attname)
        )

**Comment by adamchainz** [comment]
> I checked `refresh_from_db()`, and I think I would expect this to work:

I *wouldn't* expect that to work. My model of `refresh_from_db()` is that it refetches the field data from the database into the current instance, but doesn't copy *everything* from a newly fetched instance (from `from_queryset`) into this one. For example, `Model.refresh_from_db(from_queryset=Model.objects.annotate(...))` doesn't copy the annotated attributes into the current instance, either.

File: django/db/models/query_utils.py (hunk 41 above)
Comment (adamchainz) [reply django/db/models/query_utils.py]: > Any interest in observing database parameter limits, similar to how `in_bulk()` [does it](https://github.com/django/django/blob/0ddbe12ea99a2dc1b757dc2015ba8bb6bfd9d653/django/db/models/query.py#L1195), or factoring out some reusable private helper?

Well, yes, that seems great. I will take a look at how easy it would be to revive that old PR.

Parameter limit support is necessary, right? Otherwise, this will crash when there are enough instances to fetch, yes?

> in all cases it seems that the `db_manager()` call is unnecessary?

Yup, good shout, updating.

> Also, not sure if the dict literal is faster but the `dict` function can be used as well

`dict()` is indeed faster:

```
$ uvx -p 3.14 ipython
...
Python 3.14.0rc1 (main, Aug  8 2025, 16:52:21) [Clang 20.1.4 ]
...

In [1]: items = [(i, str(i)) for i in range(1000)]

In [2]: %timeit {k: v for k, v in items}
22.1 μs ± 208 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [3]: %timeit dict(items)
15.7 μs ± 52.9 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

I'll update.

**Comment by jacobtylerwalls** [comment]
My mental model of `refresh_from_db()` in general matches yours regarding annotations. However, my understanding of `from_queryset` is that it adds nuance. It's doc'd as providing "more control over how the model is reloaded", and the [examples](https://docs.djangoproject.com/en/5.2/ref/models/instances/#django.db.models.Model.refresh_from_db) extend beyond syncing field values, e.g. locking rows.

@charettes, any thoughts on whether to propagate the fetch mode from the `from_queryset`?

**Comment by charettes** [comment]
@jacobtylerwalls my understanding of what should happen when `refresh_from_db` is passed a `from_queryset` is the following.

1. The model instance `refresh_from_db` is called on preserves it's existing `fetch_mode` affinity. That is if a model was fetched with `FETCH_PEERS` and `from_queryset` fetch mode is different the instance remains `FETCH_PEERS`.
2. The model instances (instance**s** as there can be select related of prefetched objects) created from using `from_queryset` to pluck `fields` attribute from use `from_queryset` fetch mode.

In summary I think the way work by default is sensible and allow `from_queryset.fetch_mode` to be used efficiently to catch potential overfetching?

File: django/db/models/fetch_modes.py [hunk 48]
```diff
@@ -0,0 +1,61 @@
+from django.core.exceptions import FieldFetchBlocked
+
+
+class FetchMode:
```
tim-mccurrach, line 4: I really love this pattern because it allows users to subclass and define their own fetch mode really easily. There are many variations that users will inevitably want:
 - fetch-many and then log
 - fetch-one but log if there are peers to fetch
 - add sentry event (ot whatever observability platform you're using) and do each of the above.
 - etc.

And this pattern allows for all of the above without an ever-growing list of fetch-modes. Is it worth documenting this? We could add a simple how-to with an example.

django/db/models/fetch_modes.py
```diff
@@ -0,0 +1,61 @@
...
+    __slots__ = ()
+
+    track_peers = True
+
+    def fetch(self, fetcher, instance):
+        instances = [
+            peer
+            for peer_weakref in instance._state.peers
+            if (peer := peer_weakref()) is not None
+        ]
```
tim-mccurrach, line 36: In light of the comment above, I think the logic within`FetchPeers` will be useful for custom fetch-modes. I can imagine users only wanting to log/do-something in the case when `len(instances) > 1`. 

It would be great to make `FetchPeers` more subclass-able without users needing to copy code. Two ideas that occur are:
 - Maybe give the `fetch` method a return value of `len(instances)`, so that you can call `super().fetch(...)` and vary the behaviour accordingly.
 - Alternatively, move lines 32-36 to a separate method, `get_peers`. This would make overwriting the `fetch` method easier. (This method could even go in the base `FetchMode` class).

jacobtylerwalls:
> In summary I think the way work by default is sensible and allow from_queryset.fetch_mode to be used efficiently to catch potential overfetching?

Cool, thanks for satisfying my concern here. To me it's not clear why we would [only copy some of the `_state`](https://github.com/django/django/blob/8956ee3ce393f143696fa6c50a1153c0623003ad/django/db/models/base.py#L787) from the fetched instance and not all of it, but this is a fairly advanced use case and can be worked around with just setting `_state` directly as I mentioned [above](https://github.com/django/django/pull/17554#discussion_r2319233714) instead of inheriting it from the `from_queryset`, and we don't have strong consensus to change it.

django/db/models/query.py
```diff
@@ -122,10 +125,17 @@ def __iter__(self):
...
             obj = model_cls.from_db(
-                db, init_list, row[model_fields_start:model_fields_end]
+                db,
+                init_list,
+                row[model_fields_start:model_fields_end],
+                fetch_mode=fetch_mode,
             )
+            if fetch_mode.track_peers:
+                peers.append(weak_ref(obj))
+                obj._state.peers = peers
```
tim-mccurrach, line 138: This is really pernickety (sorry), but `peers` starts off life as a tuple and then changes to a list here. This makes it more difficult to use, since there are fewer assumptions you can make at the point of using it. Having said this, I appreciate that this is not part of the public API, so maybe that need not be a concern.

django/db/models/fetch_modes.py
adamchainz: Yes, I'd like to document the pattern. I was thinking of leaving it out of the initial PR, but perhaps that's not worth it at this point!

django/db/models/query.py
adamchainz: There are good reasons for this discrepancy…

1. We keep `peers` as a class-level variable in `ModelState` to avoid creating any values for new instances that don't need.
2. For class-level variables, we want them to be immutable, hence a tuple.
3. When peers are collected, though, they must be a list, so we can grow it as we instantiate the model instances here.

If we were using type hints, we would define `peers: Sequence[weak_ref[T]]`, so we'd know that code relied only on the common methods of a sequence, not any of the specifics of tuple or list.

django/db/models/fetch_modes.py
adamchainz: I think I will leave this as a follow-up, I haven't found the time to really hash out the details, and the 6.0 merge deadline is here… Also, it's probably best to leave it for tinkerers to explore for one version before we commit to a certain pattern.

I'd also like to make another follow-up, to use `FETCH_PEERS` by default in the admin.

django/db/models/fetch_modes.py
adamchainz: I think best to leave this for a follow-up where we document how to make a custom fetch mode.

jacobtylerwalls:
Hey @adamchainz thanks for being on top of all the feedback. Regrettably this won't make the 6.0 freeze, which is happening as we speak, but this is first on my list for 6.1. Weeks, not months. I'll start with #19842. Then, merging this one so early in the cycle will allow us plenty of time to address the follow-ups you mentioned above.

django/db/models/query_utils.py
adamchainz: I've just rebased and made this bit use `in_bulk()` now that support was added in 1820d35b17f0a95f4ce888971b9ca0c7a3697c83.

adamchainz:
Rebased targeting Django 6.1!

django/db/models/query_utils.py
jacobtylerwalls: Nice. Just checking: did you plan to use `dict()` here after all?

jacobtylerwalls: Whoops, that's irrelevant now that we got `in_bulk` working :-), resolving

docs/topics/db/optimization.txt
jacobtylerwalls: ticket-36665

django/db/models/query_utils.py
```diff
@@ -281,6 +282,17 @@ def _check_parent_chain(self, instance):
...
         return None
 
+    def fetch_one(self, instance):
+        instance.refresh_from_db(fields=[self.field.attname])
+
+    def fetch_many(self, instances):
+        attname = self.field.attname
+        value_by_pk = self.field.model._base_manager.values_list(attname).in_bulk(
+            {i.pk for i in instances}
+        )
```
jacobtylerwalls: Earlier you removed a no-argument call to `db_manager()`, but now I wonder if we need to provide a db alias here.

With only an `"other"` connection, I see this works: the deferred field is fetched from `"other"`:
```py
In [18]: qs = ResourceInstance.objects.defer('principaluser').using('other')

In [19]: for el in qs:
    ...:     print(el.principaluser)
    ...: 
admin
```
versus this failure showing we tried to go to the default alias:
```py
In [20]: qs = ResourceInstance.objects.defer('principaluser').using('other').fetch_mode(models.FETCH_PEERS)

In [21]: for el in qs:
    ...:     print(el.principaluser)

File ~/django/django/db/models/query_utils.py:290, in DeferredAttribute.fetch_many(self, instances)
    288 def fetch_many(self, instances):
    289     attname = self.field.attname
--> 290     value_by_pk = self.field.model._base_manager.values_list(attname).in_bulk(
    291         {i.pk for i in instances}
    292     )
    293     for instance in instances:
    294         setattr(instance, attname, value_by_pk[instance.pk])
...
File ~/django/django/db/backends/dummy/base.py:21, in complain(*args, **kwargs)
     20 def complain(*args, **kwargs):
---> 21     raise ImproperlyConfigured(
     22         "settings.DATABASES is improperly configured. "
     23         "Please supply the ENGINE value. Check "
     24         "settings documentation for more details."
     25     )

ImproperlyConfigured: settings.DATABASES is improperly configured. Please supply the ENGINE value. Check settings documentation for more details.
```

How would the user provide an alias here?

charettes: @jacobtylerwalls we can safely assume that all instances are originating from the same database (otherwise they wouldn't be queryset peers) so doing

```python
db = instances[0]._state.db
value_by_pk = self.field.model._base_manager.using(db).values_list(attname).in_bulk(
    {i.pk for i in instances}
)
```

should do here

adamchainz: Yeah, that looks good. @jacobtylerwalls since you set "ready to merge", will you take care of this edit?

jacobtylerwalls: Done! (added a test)

jacobtylerwalls:
Thanks to all involved! Will merge when checks finish.

PR #17516, opened 2023-11-23
Fixed #34994 -- Fixed checkbox layout on narrow screen widths.

knyghty:
https://code.djangoproject.com/ticket/34994

# Before

Screenshot 2023-11-23 at 15 45 06

# After

Screenshot 2023-11-23 at 15 45 26

PR #17560, opened 2023-11-30
[RFC/WIP] Refs #????? -- Switched to logical properties in CSS.

ngnpope:
Over the past few months there have been a lot of regressions and subsequent fixes for various issues with RTL CSS.

For example: #17561, #17558, #17556, https://github.com/django/django/pull/17516#issuecomment-1824560923, #17514, #16685, #16684, #16675... And that's just this year alone.

I'm wondering if it's just time to switch as much as possible of this over from physical to logical properties?

Support is generally pretty good now. We're probably just waiting a few months for the support percentage to creep up for Chrome/Edge for `float`, `clear` and some others which were unflagged in v118, but I think it's worth discussing now. We can also make use of `@supports` if necessary -- which is widely supported -- to check for some cases and fall back to physical properties for a period of time. A notable exception is lack of support for logical `background-position`, but we could perhaps be clever with use of CSS variables for this (which I haven't investigated here).

I've knocked up a script that can generate a bunch of commits so that we can easily re-run it. There will likely be a few manual things to tweak also, e.g. related to splitting shorthand properties that don't support logical dimensions. We'd also need to ensure that the documentation is updated to require use of logical properties (and perhaps add a linting check to avoid them creeping in).

Anyway... Is this a good idea worth pursuing? Pinging @knyghty and @felixxm for their input as they've fixed most of these issues.

### DIFF
--- a/django/contrib/admin/static/admin/css/responsive_rtl.css
+++ b/django/contrib/admin/static/admin/css/responsive_rtl.css
@@ -1,45 +1,10 @@
 /* TABLETS */
 
 @media (max-width: 1024px) {
-    [dir="rtl"] .colMS {
-    }
-
-    [dir="rtl"] #user-tools {
-    }
-
-    [dir="rtl"] #changelist .actions label {
-    }
-
-    [dir="rtl"] #changelist .actions select {
-    }
-
-    [dir="rtl"] .change-list .filtered .results,
-    [dir="rtl"] .change-list .filtered .paginator,
-    [dir="rtl"] .filtered #toolbar,
-    [dir="rtl"] .filtered div.xfull,
-    [dir="rtl"] .filtered .actions,
-    [dir="rtl"] #changelist-filter {
-    }
-
     [dir="rtl"] .inline-group ul.tools a.add,
     [dir="rtl"] .inline-group div.add-row a,
     [dir="rtl"] .inline-group .tabular tr.add-row td a {
         background-position: calc(100% - 8px) 9px;
-    }
-
-    [dir="rtl"] .related-widget-wrapper-link + .selector {
-    }
-
-    [dir="rtl"] .selector .selector-filter label {
-    }
-
-    [dir="rtl"] .object-tools li {
-    }
-
-    [dir="rtl"] .object-tools li + li {
-    }
-
-    [dir="rtl"] .dashboard .module table td a {
     }
 
     [dir="rtl"] .selector-add {
@@ -56,22 +21,5 @@
 
     [dir="rtl"] .active.selector-remove:focus, .active.selector-remove:hover {
         background-position: 0 -140px;
-    }
-}
-
-/* MOBILE */
-
-@media (max-width: 767px) {
-    [dir="rtl"] .aligned .related-lookup,
-    [dir="rtl"] .aligned .datetimeshortcuts {
-    }
-
-    [dir="rtl"] .aligned ul,
-    [dir="rtl"] form .aligned ul.errorlist {
-    }
-
-    [dir="rtl"] #changelist-filter {
-    }
-    [dir="rtl"] .aligned .vCheckboxLabel {
     }
 }
--- a/django/contrib/admin/static/admin/css/rtl.css
+++ b/django/contrib/admin/static/admin/css/rtl.css
@@ -1,14 +1,5 @@
 /* GLOBAL */
 
-th {
-}
-
-.module h2, .module caption {
-}
-
-.module ul, .module ol {
-}
-
 .viewlink, .addlink, .changelink, .hidelink {
     background-position: 100% 1px;
 }
@@ -17,102 +8,7 @@ th {
     background-position: 100% 1px;
 }
 
-.object-tools {
-}
-
-thead th:first-child,
-tfoot td:first-child {
-}
-
-/* LAYOUT */
-
-#user-tools {
-}
-
-div.breadcrumbs {
-}
-
-#content-main {
-}
-
-#content-related {
-}
-
-.colMS {
-}
-
-/* SORTABLE TABLES */
-
-table thead th.sorted .sortoptions {
-}
-
-thead th.sorted .text {
-}
-
-/* dashboard styles */
-
-.dashboard .module table td a {
-}
-
-/* changelists styles */
-
-.change-list .filtered table {
-}
-
-#changelist-filter {
-}
-
-#changelist-filter li.selected {
-}
-
-#changelist table tbody td:first-child, #changelist table tbody th:first-child {
-}
-
-.paginator .end {
-}
-
-.paginator input {
-}
-
 /* FORMS */
-
-.aligned label {
-}
-
-.submit-row a.deletelink {
-}
-
-.vDateField, .vTimeField {
-}
-
-.aligned .form-row input {
-}
-
-form .aligned ul {
-}
-
-form ul.inline li {
-}
-
-form .aligned p.help,
-form .aligned div.help {
-}
-
-form div.help ul,
-form .aligned .checkbox-row + .help,
-form .aligned p.date div.help.timezonewarning,
-form .aligned p.datetime div.help.timezonewarning,
-form .aligned p.time div.help.timezonewarning {
-}
-
-form .wide p.help, form .wide div.help {
-}
-
-.submit-row {
-}
-
-fieldset .fieldBox {
-}
 
 .errorlist li {
     background-position: 100% 12px;
@@ -141,15 +37,6 @@ fieldset .fieldBox {
 .calendarbox .calendarnav-next:focus,
 .calendarbox .calendarnav-next:hover {
     background-position: 0 -15px;
-}
-
-.calendar caption, .calendarbox h2 {
-}
-
-.selector {
-}
-
-.selector .selector-filter {
 }
 
 .selector-add {
@@ -184,23 +71,6 @@ a.active.selector-clearall:focus, a.active.selector-clearall:hover {
     background-position: 0 -176px;
 }
 
-.inline-deletelink {
-}
-
 form .form-row p.datetime {
     overflow: hidden;
-}
-
-.related-widget-wrapper {
-}
-
-/* MISC */
-
-.inline-related h2, .inline-group h2 {
-}
-
-.inline-related h3 span.delete {
-}
-
-.inline-related h3 span.delete label {
 }


knyghty (commented):
Some (very) initial thoughts:

1. Great idea!
2. I wonder how much of this code is used or necessary as we've moved more and more to flex.
3. I didn't know these existed at all! I worry a bit that these selectors are less well known and a bit harder for people to immediately understand.

From 2:
- Is it better to do an audit and clean up unused styles first or do this first? Or no substantial difference?
- I'd like to move more things to flex (it's on my list...), which would also likely reduce the need for RTL-specific styles.

thibaudcolas:
This is great :)

The need for this is pretty clear-cut to me, it’s just not productive to maintain separate RTL styles when logical properties and values are possible, but there are enough specific implementation details that need careful consideration so I’d recommend proceeding with a DEP so there’s room for people to validate specific details of the implementation plan.

I’d like to see specifically:

- Whether the browser support is good enough for use in Django (Chrome/Edge since October/November 2023 for logical `float: inline-start` and `inline-end`, Safari 15 in September 2021 for border radius properties, Safari 14.1 from 2021 for lots of stuff)
- Whether we really do need to use logical properties and values for both the horizontal and vertical axis, or whether we could do horizontal only (so LTR <-> RTL mirroring).
- A plan for linting. I don’t see usage of logical properties and values being enforced without linting. We’ve just started [discussing linting for CSS on the forum](https://forum.djangoproject.com/t/adding-a-formatter-for-css-js/25754), but only in passing. I’ve already written Stylelint configs to enforce usage of logical properties and values if it helps, but we’d need to agree to use Stylelint first.
- A plan for contributor docs. This stuff is simple once you know it but if you don’t it’s really surprising.
- A plan for `background-position` and similar CSS transforms that only have physical values, with a CSS variable reversing the direction ideally rather than `:dir` or `[dir]` style overrides

If possible – I don’t think Django does this well currently but I’d also like to see:

- An assessment of how much JS and Python code would need adapting for RTL languages (hopefully not too much)
- Some ideas about how to rework Django’s icons so they’re mirrored.
- How we will engage with people who actually use right-to-left languages to work on this and QA the work.

And ideally as "stretch goals":

- A plan for code cleanup / moving more things to Flexbox or Grid.
- A QA plan involving a visual regression testing setup, as this will be a case where for LTR languages there should be no changes at all. So this would be very simple to test with visual regression tools.

I wouldn’t recommend the cleanup being a pre-condition for this work. There’s plenty enough benefits to switching even unused code to logical properties and values :)

ngnpope:
> ...but there are enough specific implementation details that need careful consideration so I’d recommend proceeding with a DEP so there’s room for people to validate specific details of the implementation plan.

Ok. Not sure there is all that much that it would require a DEP, but I guess we can look into doing one... Has the when to DEP or not thing been worked out yet - I know there was some discussion recently? It's never really been clear, and I'm concerned I'd spend time on something that nobody will engage with.

> - Whether the browser support is good enough for use in Django (Chrome/Edge since October/November 2023 for logical `float: inline-start` and `inline-end`, Safari 15 in September 2021 for border radius properties, Safari 14.1 from 2021 for lots of stuff)

So currently it looks like we're at [82.92%](https://caniuse.com/mdn-css_properties_float_flow_relative_values) for the logical float/clear stuff. Am sure this has jumped up since last week. Based on our [browser support](https://docs.djangoproject.com/en/4.2/faq/admin/#what-browsers-are-supported-for-using-the-admin), there is some leeway with this, but I expect we wouldn't be looking at landing something like this for a few months yet, but we can still plan.

> - Whether we really do need to use logical properties and values for both the horizontal and vertical axis, or whether we could do horizontal only (so LTR <-> RTL mirroring). I’d have thought the horizontal axis would be enough, as I can’t imagine anyone wanting to use Django in a TTB+RTL layout.

The point is that the vertical stuff should largely come for free (except for the `background-position` situation). We may not explicitly choose to support it, but if the writing mode is changed to be vertical and we're using logical properties/values, etc. it should all just flow as expected. The idea is that we shouldn't need to think about different directions as distinct things.

> - A plan for linting. I don’t see usage of logical properties and values being enforced without linting. We’ve just started [discussing linting for CSS on the forum](https://forum.djangoproject.com/t/adding-a-formatter-for-css-js/25754), but only in passing. I’ve already written Stylelint configs to enforce usage of logical properties and values if it helps, but we’d need to agree to use Stylelint first.

For sure. We'd want to make sure we don't regress. This is quite trivial, but we definitely will need something in place. Have seen the discussions and it seems positive, which is great.

> - A plan for contributor docs. This stuff is simple once you know it but if you don’t it’s really surprising.

This will be the biggest thing. Linting will help highlight stuff, but we effectively need to define our policy on this so that it's clear what the expectation is when contributing. Alas, visibility of this always seems to be the challenge - people will always contribute without reading the guidelines 🤷🏻 

When we do this, we'll likely call it out in the top section of the release notes that the admin has changed to using logical properties with a link to the details in the conributing guide. The good news is that, where people have overridden stuff using physical properties, they've probably largely only considered LTR anyway, and things should mostly continue to work as before in those situations with no backward compatibility issues. Even where people have overridden for RTL with physical properties should work in general, but there may be small issues here and there.

> - A plan for background-position and similar CSS transforms that only have physical values, with a CSS variable reversing the direction ideally rather than `:dir` or `[dir]` style overrides

This is what I want to investigate next. We'll need one of those overrides at the top level to define the variable(s), but the idea is that everything else flows from that.

> - An assessment of how much JS and Python code would need adapting for RTL languages (hopefully not too much)

A good shout. I don't think there is much, but this reminds me that we should also look at the various error pages, etc. I've mainly focused on the admin at this point.

> - Some ideas about how to rework Django’s icons so they’re mirrored.

Also interesting. Some of this is handled by using multiple icons, but are there situations where we can flip and/or rotate instead? 🤔 

> - How we will engage with people who actually use right-to-left languages to work on this and QA the work.

Another excellent point. I have an idea of a few people we could ask, but it'd be good to put out a call on the forum, perhaps when we have progressed with an initial DEP or such...

> - A plan for code cleanup / moving more things to Flexbox or Grid.

While I'm all for this, I think we should keep this firmly as a separate goal. I don't see that switching properties from physical to logical would be an impediment to this in any way - in fact, I think it may even simplify the task. It's likely that we can even delete some of the logical styles when migrating things to flex or grid as those layout models take flow directions into consideration by design.

> - A QA plan involving a visual regression testing setup, as this will be a case where for LTR languages there should be no changes at all. So this would be very simple to test with visual regression tools.

We've added the capability to take screenshots, but there isn't support for easy comparison yet. I guess we could look into something like percy.io or equivalent - maybe some generous company will donate it as a freebie? 🤞🏻

> I wouldn’t recommend the cleanup being a pre-condition for this work. There’s plenty enough benefits to switching even unused code to logical properties and values :)

Agreed. Clean up is good irrespective of when it comes. If some things come before, that's just a bonus of smaller diffs for review.

thibaudcolas:
@ngnpope if you’re hesitant to start the DEP I’d happily volunteer to do that, at least a first draft which you can add to at a later date? @knyghty what do you think, does this "require a DEP"?

As far as engaging with it once opened – there are a lot of glaring issues with Django’s RTL support currently, and @django/accessibility has been struggling a lot with the state of Django’s CSS, which this work would help in improving quite a bit. So personally I’d suspect this will be smooth-sailing to take through.

---

Re browser support – sounds good to me!

---

> The point is that the vertical stuff should largely come for free (except for the background-position situation). We may not explicitly choose to support it, but if the writing mode is changed to be vertical and we're using logical properties/values, etc. it should all just flow as expected. The idea is that we shouldn't need to think about different directions as distinct things.

It’s not really for free though, as it means people who write code will need to think of left-right mirroring _and_ of the "vertical" axis direction. Personally I doubt _anyone_ wants to see the Django admin in a TTB-RTL layout, so I’d rather we skipped this entirely.

---

> > Some ideas about how to rework Django’s icons so they’re mirrored.

> Also interesting. Some of this is handled by using multiple icons, but are there situations where we can flip and/or rotate instead? 🤔

I would recommend it’d be better for maintenance if we had a policy to only use icons in the UI that are either non-directional, or that can safely be mirrored. That way you don’t have to write so many overrides on "which icon to use in which variant of the admin" – you just write individual icon definitions with the LTR perspective, and then have some way to define which icons are and aren’t mirrored as a "global".

A note on icons mirroring – I think this should also be in the "separate goal" / "not MVP" area, because Django’s icon definitions are currently really hard to work with for other reasons, and I think it’d be better to fix this tech debt _before_ changing the code for RTL support.

> > A plan for code cleanup / moving more things to Flexbox or Grid.

> While I'm all for this, I think we should keep this firmly as a separate goal. 

Same. I think it’d be nice to cover this in the DEP nonetheless, but not as something that must happen as part of this switch to logical properties.

> We've added the capability to take screenshots, but there isn't support for easy comparison yet. I guess we could look into something like percy.io or equivalent - maybe some generous company will donate it as a freebie? 🤞🏻

Yeah Percy would be great. It does have a free tier, but I’m not sure the free tier would be enough for what we’d need here.

I can recommend [BackstopJS](https://github.com/garris/BackstopJS) for one-off efforts like this, as something we’d run locally. I’ve used it extensively so happy to expand further on what this would look like.

thibaudcolas:
There’s overwhelming support for setting up Prettier for CSS in the forum discussion about [Adding a formatter for CSS & JS](https://forum.djangoproject.com/t/adding-a-formatter-for-css-js/25754). Based on discussions here and on that forum thread, do people think we should start a forum discussion about _linting_ CSS already? Or wait for more to happen with either the auto-formatting or this logical properties and values refactoring?

ngnpope:
I think it's worth starting the discussion. Things can be done in phases - just having something in place to begin with would be great and then adapting the rules for logical properties later would be best.

ngnpope:
Regarding the DEP - sounds fine if there is enough interest in it. Happy to put something together if we want to take that path. I just wasn't sure whether this was big enough a change to warrant it.

> I can recommend [BackstopJS](https://github.com/garris/BackstopJS) for one-off efforts like this, as something we’d run locally.

Nice. I'm also currently working on overhauling `django-docker-box` to make it easier to work with and resolve many longstanding issues. One thing will be making it easy to work with linting and coverage and get the output/reports. This could be something we wire in to that too to reduce the barrier to contribution.

knyghty:
> @knyghty what do you think, does this "require a DEP"?

I think it's easier to start with a forum post and then move to a DEP if there's no clear consensus, personally. But it's also unclear on what requires one or not.

thibaudcolas:
👍 thanks! Main reason I suggested a DEP over a forum post is there there are a lot of aspects to this that I think it’d be easier to validate by reviewing a documented implementation plan, where there can be multiple drafts, rather than a long thread of replies. But if others think it’s not needed, happy to follow suite!

thibaudcolas:
Hi @ngnpope, happy new year :) I thought I’d check in and ask what your plans are to move this along?

I see right-to-left support as a great way to get new people involved with core development, as there are lots of obvious gaps. I’m aware of a few development sprints coming up in the first half of 2024, as well as programs dedicated to new contributors – where this work would make for lovely contributions.

What are your plans? Do you still intend to publish a forum post about this?

ngnpope:
Hi @thibaudcolas. I haven't really had a chance to look into this yet. Might do later in the month.

ngnpope:
I think there's no need to rush here anyway... I get that it'd be a nice clean up prior to flex stuff, but if you want to focus on that, then I think that should happen independently.

> So currently it looks like we're at 82.92% for the logical float/clear stuff.

This is now [85.57%](https://caniuse.com/mdn-css_properties_float_flow_relative_values) with Chrome < 118 at 5.28% being the biggest detriment. In a few months time I expect we'll be > 90% and it'll mostly be mobile browsers, e.g. Samsung Internet, that affect this.

This will be one of the things that I want to consider - do we need to go down the `@supports` route or do we just hold on a little longer? Also - do we care about mobile browsers like Samsung Internet and Opera Mobile, or not, if they haven't gained support for these rules?

> What are your plans? Do you still intend to publish a forum post about this?

I think I'll do a forum post, but I'd like to consider some more things first, e.g. an approach to handling background images and icons.

thibaudcolas:
👍 thanks :) 

Samsung Internet is a Chromium derivative, I don’t know how many versions behind Google Chrome they’re at but we could do the math and I wouldn’t be surprised if they did land support by the time this work would be released for Django.

For Opera Mobile, be worth double checking Django’s exact browser support rules but I wouldn’t expect it’s one we want to support.

fsbraun:
@sakhawy and I have solved the browser support issue for django CMS by adding a "hardcoded" RTL css document similar to django admin's `rtl.css` that contains physical styles ("patch") for the insufficiently supported logical styles, like `:dir()` or `float: inline-start`. 

The idea is to transition the main styles to logical ones and separating out support for legacy browsers, which then hopefully can be removed soon as browser support improves.

I'd be happy to contribute some sort of compatibility layer here.

knyghty:
@fsbraun I think the current proposal is to first add stylelint, and then do this, with a stylelint config to enforce them. But I think adding stylelint to the project without specific configuration for this is probably the first step. Separately, I'm also working on using prettier (or so) for formatting of JS and CSS, but I don't think there should be much overlap here.

MHLut:
I have used [stylelint-use-logical-spec](https://www.npmjs.com/package/stylelint-use-logical-spec/v/3.2.0) in my projects and liked how it works. It will update every relevant CSS property, which will be tricky to check. I also do not know how it plays with third-party CSS using old-style properties.

I haven't read the previous comments in detail, but trust that the comments are correct.

PS. We might tag Moe, who has started talk about RTL support for Django, but I don't know his GitHub handle (will send him this thread in Discord)

jacobtylerwalls:
Thanks @thibaudcolas for a passing mention of this in another PR. Can I help with waking this up in any way?

> But I think adding stylelint to the project without specific configuration for this is probably the first step.

I agree that's the first step here. We could do a little DEP for the rest in parallel?

PR #17562, opened 2023-11-30
Fixed selenium tests in Firefox.

pbratkowski:
Unlike Chromium-based browsers, Firefox does not automatically scroll elements into view when using the Actions API with `move_to_element`. This calls the `scrollIntoView` function explicitly, which fixes some selenium tests when not running in headless mode.

Namely, the `admin_views.tests.SeleniumTests.test_prepopulated_fields` test has a page larger than the viewport, and some elements are hidden, leading the test failing with an Exception like:
```
selenium.common.exceptions.MoveTargetOutOfBoundsException: Message: Move target (570, 968) is out of bounds of viewport dimensions (1280, 955)
```

### DIFF
--- a/tests/admin_views/tests.py
+++ b/tests/admin_views/tests.py
@@ -5852,6 +5852,9 @@ class SeleniumTests(AdminSeleniumTestCase):
         status = self.selenium.find_element(
             By.ID, "id_relatedprepopulated_set-2-0-status"
         )
+        # Fix for Firefox which does not scroll to clicked elements automatically with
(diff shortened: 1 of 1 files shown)


ahmedasar00:
Thanks for the work! @sarahboyce  
I noticed there is an issue with Selenium tests on Firefox, where the browser does not auto-scroll elements that are out of view.
I will start working on this problem to improve test stability.
I will refer to this issue as the Firefox auto-scroll Selenium issue.

PR #17567, opened 2023-12-01
Fixed #35002 -- Made UniqueConstraints with fields respect nulls_distinct.

peterthomassen:


tests/schema/tests.py
```diff
@@ -3510,6 +3510,47 @@ def test_unique_constraint_nulls_distinct_unsupported(self):
...
+        with connection.schema_editor() as editor:
+            editor.create_model(Author)
+        constraint = UniqueConstraint(
+            fields=("height", "weight"), name="constraint", nulls_distinct=False
+        )
+        with connection.schema_editor() as editor:
+            self.assertIsNone(editor.add_constraint(Author, constraint))
+        constraints = self.get_constraints(Author._meta.db_table)
+        self.assertNotIn(constraint.name, constraints)
+
```
peterthomassen, line 3553: I'm not sure this is needed. Happy to remove if preferred; I added it for completeness.

Note that the assertion is different than in the corresponding part of `test_unique_constraint_nulls_distinct_unsupported`, because adding/removing the constraint actually *does* trigger queries. They are no-op:
```
AssertionError: 8 != 0 : 8 queries executed, 0 expected
Captured queries were:
1. CREATE TABLE "new__schema_author" ("id" integer NOT NULL PRIMARY KEY AUTOINCREMENT, "name" varchar(255) NOT NULL, "height" integer unsigned NULL CHECK ("height" >= 0), "weight" integer NULL, "uuid" char(32) NULL)
2. INSERT INTO "new__schema_author" ("id", "name", "height", "weight", "uuid") SELECT "id", "name", "height", "weight", "uuid" FROM "schema_author"
3. DROP TABLE "schema_author"
4. ALTER TABLE "new__schema_author" RENAME TO "schema_author"
5. CREATE TABLE "new__schema_author" ("id" integer NOT NULL PRIMARY KEY AUTOINCREMENT, "name" varchar(255) NOT NULL, "height" integer unsigned NULL CHECK ("height" >= 0), "weight" integer NULL, "uuid" char(32) NULL)
6. INSERT INTO "new__schema_author" ("id", "name", "height", "weight", "uuid") SELECT "id", "name", "height", "weight", "uuid" FROM "schema_author"
7. DROP TABLE "schema_author"
8. ALTER TABLE "new__schema_author" RENAME TO "schema_author"
```
Ditto for removing the constraint.

Not sure whether it's worth doing anything about that (but probably not in this PR anyway).

sarangkkl (approved):

tests/schema/tests.py
charettes: yeah I think we can do without!

peterthomassen:
Done.

@charettes Thanks, I didn't know. I had read up on it before I pushed, and the [docs](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/working-with-git/#working-on-a-ticket) say that when "working for a fix on the 1.4 branch", one would base off stable/1.4.x. That implied to me that I should start from the stable/5.0.x branch.

tests/schema/tests.py
```diff
@@ -3504,6 +3504,47 @@ def test_unique_constraint_nulls_distinct_unsupported(self):
             self.assertIsNone(editor.add_constraint(Author, constraint))
             self.assertIsNone(editor.remove_constraint(Author, constraint))
 
+    @skipUnlessDBFeature("supports_nulls_distinct_unique_constraints")
+    def test_create_unique_nulls_distinct(self):
```
charettes, line 3508: Maybe this test should be renamed to mention the fact that this will result in a constraint and the other test in an index as it relies on expressions?

charettes (commented):
Thanks for the target adjustment and the quick patch @peterthomassen!

I think we can do without the second test given `test_unique_constraint_nulls_distinct_unsupported` already covers this code path here.

I think that the patch is in a good enough shape to have mergers eyes on it at this point. I suspect we'll want to rename `test_unique_constraint_nulls_distinct` to `test_unique_constraint_index_nulls_distinct` and your one `test_unique_constraint_nulls_distinct` but we'll see what they'll think!

charettes:
> the [docs](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/working-with-git/#working-on-a-ticket) say that when "working for a fix on the 1.4 branch", one would base off stable/1.4.x. That implied to me that I should start from the stable/5.0.x branch.

That's interesting, that would explain why we've seen similar submission in the past. I wonder if we'd want to nuance the suggestion to mention that if the bug is still present in `main` you likely want to base of that instead.

peterthomassen:
> I think that the patch is in a good enough shape to have mergers eyes on it at this point.

Ok, I take it that this means you folks are going to make any subsequent changes, so I can go off for dinner. :)

> I suspect we'll want to rename `test_unique_constraint_nulls_distinct` to `test_unique_constraint_index_nulls_distinct` and your one `test_unique_constraint_nulls_distinct` but we'll see what they'll think!

I like that!

PR #17573, opened 2023-12-04
Deprecated django.contrib.gis.geoip2.GeoIP2.open().

ngnpope:
Another bit of preparation, deprecating `GeoIP2.open()` which adds nothing over the constructor.


ngnpope:
_Test failure is unrelated._

PR #17576, opened 2023-12-05
Fixed #13314 -- Make FileField validate the full filepath against max_length on ModelForms

anorthall:
Ticket: https://code.djangoproject.com/ticket/13314

This ticket has been laying about for over a decade and I decided I'd have a stab at closing it.

## The problem
Django forms do not validate the full file path against the `FileField.max_length` attribute, only against the file name of the file uploaded by the user. The file name of the file uploaded by the user and the actual file path provided to the database can be wildly different in length, due to the impact of the`models.fields.FileField.upload_to` attribute.

## My approach
It makes sense that this validation can occur automatically in case of a `ModelForm`, where Django is able to easily link the `forms.fields.FileField` to the `models.fields.FileField` and access the `generate_filename()` method on it, to determine what the true length of the file path will be for any given file.

I have added two new attributes to `forms.fields.FileField`:
- `forms.fields.FileField._model_instance`
- `forms.fields.FileField._model_field_name`

The `ModelForm` constructor will automatically set these attributes when generating fields for a `ModelForm`, and as such proper validation will take place 'automatically'.

## Concerns
1. It is not possible to provide a helpful error message to the user of the form as it is not possible to determine how long their file name should be due to the dynamic nature of file path generation. For now I have settled with `The filename is too long. Rename the file to something shorter.`, but perhaps something even more generic such as `There was an issue uploading the file.` should be used?
2. Translation of any new error message will be required.


felixxm:
@anorthall Thanks for this patch :+1: IMO, the only reasonable way to solve this is to introduce a new `FileField` kwarg (e.g. `max_upload_to_length`) where user could declare a length. Like `max_digits` and `decimal_places` work for `DecimalField`. There are of course backward compatibility concerns. I'm not sure it's worth fixing. Maybe we could just add a warning to docs that you should take the length of `upload_to` into account and limit `forms.FileField.max_length` :thinking:.

anorthall:
> @anorthall Thanks for this patch 👍 IMO, the only reasonable way to solve this is to introduce a new `FileField` kwarg (e.g. `max_upload_to_length`) where user could declare a length. Like `max_digits` and `decimal_places` work for `DecimalField`. There are of course backward compatibility concerns. I'm not sure it's worth fixing. Maybe we could just add a warning to docs that you should take the length of `upload_to` into account and limit `forms.FileField.max_length` 🤔.

I agree that it doesn't seem worth the added complexity of doing it 'automatically' (but you don't know until you try!) and that just highlighting the issue in the docs would likely be the best solution here.

I won't be coding for a few days now but I will come back to this and unless anyone has any other thoughts I will close this PR and submit a new one with a docs change.

anorthall:
Submitted changes removing all previous work and just updating documentation.

docs/ref/forms/fields.txt
```diff
@@ -681,6 +681,24 @@ For each field, we describe the default widget used if you don't specify
...
+    :exc:`~django.db.IntegrityError` will be raised.
+
+    You should ensure that the ``max_length`` argument of the file field is
+    set to a value that is less than or equal to the maximum length of the
+    filename in the database, including any location specified in the
+    :attr:`~django.db.models.FileField.upload_to` attribute.
+
+    In cases where the filename is dynamically generated after upload, you
+    will need to ensure that the case where a filename longer than that
+    which is allowed by the database is properly handled.
```
browniebroke, line 700: I wonder if this block should be dedented? (Not saying it should, I genuinely don't know)

<details>
<summary>Current</summary>

image

</details>

<details>
<summary>Dedented</summary>

image

</details>

anorthall: Thanks @browniebroke - I'm quite neutral on both options. I tried to find some other location with a similar style to copy and culd not.

browniebroke: I gave this one more thought and it occurred to me that perhaps it should be turned into an [`admonition` directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-admonition), which would make it look like:

image

You can search for `.. admonition::` for examples.

anorthall: This is a great idea, thanks. Will do this later today.

anorthall: Thanks again, I've both added the admonition and (not insignificantly) rephrased the text to be more clear and concise (hopefully...). I think this version is much better, but let me know what you think.

CleanShot 2025-04-27 at 14 19 33@2x

anorthall: CleanShot 2025-04-27 at 14 41 17@2x
I did some more rephrasing - this is where we are at now. It's really quite difficult to cleanly explain!

docs/ref/forms/fields.txt
```diff
@@ -677,9 +677,32 @@ For each field, we describe the default widget used if you don't specify
...
+        You should ensure that the ``max_length`` argument of the form field
+        is set to a value low enough to prevent this from happening. If an
+        :attr:`~django.db.models.FileField.upload_to` location is specified on
+        the model field then this will also become part of the filename and
+        you must account for this.
+
+        If the filename is dynamically generated after upload (such as by a
+        callable :attr:`~django.db.models.FileField.upload_to` on the model
+        field), then you must ensure that you handle the possibility of a
+        filename that is longer than is allowed by the database.
```
browniebroke, line 705:
```suggestion
        filename longer than the database field.
```

browniebroke (approved):
Looks good, left a few small suggestions. I don't feel strongly about them, but I find that they break and shorten sentences.

anorthall:
> Looks good, left a few small suggestions. I don't feel strongly about them, but I find that they break and shorten sentences.

Thanks again - all applied, look sensible!

browniebroke (approved):

sarahboyce:
Thank you for this PR :star: 

I see you've chosen to put the admonition within forms [FileField](https://docs.djangoproject.com/en/5.2/ref/forms/fields/#django.forms.FileField) rather than models [FileField](https://docs.djangoproject.com/en/5.2/ref/models/fields/#django.db.models.FileField). 
I was wondering whether this might be clearer within a model FileField, perhaps specifically within `FileField.upload_to` as this also has example models and we could show an appropriate max_length for the model form? :thinking: 
What do you think?

PR #17584, opened 2023-12-07
Fixed #34742 -- added tests when calling DELETE endpoint without trailing slash

avaneeshk098:
I have added the tests to check if the request method is DELETE. If a RuntimeError is not thrown an AssertionError would be called.


django/middleware/common.py
```diff
@@ -83,7 +83,7 @@ def get_full_path_with_slash(self, request):
         new_path = request.get_full_path(force_append_slash=True)
         # Prevent construction of scheme relative urls.
         new_path = escape_leading_slashes(new_path)
-        if settings.DEBUG and request.method in ("POST", "PUT", "PATCH"):
+        if settings.DEBUG and request.method in ("POST", "PUT", "PATCH", "DELETE"):
```
ngnpope, line 86: Please also update the docstring above.

avaneeshk098:
[Ticket #34742](https://code.djangoproject.com/ticket/34742)

felixxm (commented):
@avaneeshk098 Thanks :+1: Welcome aboard ⛵️

I pushed small edits.

PR #17593, opened 2023-12-11
Config.py and registry.py comment and doctstring updates

alesiram:
Updated comments, docstrings within the config.py and resgistry.py files.


shangxiao:
Hi. Sorry but we don't include comments that simply describe what the code is doing like "These are imports" & "These are global variables", as a maintainer can read the code instead of reading the comment 🤷.

PR #17594, opened 2023-12-11
Refs #33497 -- Added connection pool support for PostgreSQL.

sarahboyce:
@apollo13 has done all the hard work in this PR: https://github.com/django/django/pull/16881 hoping to push it over the line.

Also hope the tests are testing the right thing, tried to take inspiration from `psycopg_pool` tests: https://github.com/psycopg/psycopg/blob/52ed68a9d699c1e43e07a4d04441534197552de5/tests/pool/test_pool.py#L103

But let me know what I'm missing - very new space for me 👍


apollo13:
Lovely, thank you for picking up my work.

django/db/backends/postgresql/base.py
```diff
@@ -287,36 +350,43 @@ def get_new_connection(self, conn_params):
...
         # can be the case when using temporary or ephemeral credentials.
-        commit_role = self.ensure_role()
+        role_name = self.settings_dict.get("OPTIONS", {}).get("assume_role")
+        commit_role = ensure_role(connection, self.ops, role_name)
+
+        return commit_role or commit_tz
+
+    def _close(self):
+        if self.connection is not None:
+            # NOTE: `wrap_database_error` only works for `putconn` as long as there
```
apollo13, line 373: typo, should be `wrap_database_errors` (plural)

django/db/backends/postgresql/base.py
```diff
@@ -287,36 +350,43 @@ def get_new_connection(self, conn_params):
...
+            # is not `reset` function set in the pool because that is defered into
```
apollo13, line 374: s/not/no

django/db/backends/postgresql/base.py
```diff
@@ -179,6 +197,45 @@ class DatabaseWrapper(BaseDatabaseWrapper):
...
+                # Ensure we run in autocommit, Django properly sets it later on
+                connect_kwargs["autocommit"] = True
+                pool = ConnectionPool(
+                    kwargs=connect_kwargs,
+                    open=False,  # Do not open the pool during startup
+                    configure=self._configure_connection,
+                    **pool_options,
+                )
+
+            # NOTE: `setdefault` ensures that multiple threads don't set this in
```
bluetech, line 232: One subtlety with this trick that I checked is that `pool.open` is thread safe; luckily it is :+1:

django/db/backends/postgresql/base.py
```diff
@@ -179,6 +197,45 @@ class DatabaseWrapper(BaseDatabaseWrapper):
...
+            if pool_options is not None:
+                # Verify that we are not running with persistent connections
+                if self.settings_dict.get("CONN_MAX_AGE", 0) != 0:
+                    raise ImproperlyConfigured(
+                        "Pooling doesn't support persistent connections"
+                    )
+                if pool_options is True:  # simply sets the default options
+                    pool_options = {}
+
+                from psycopg_pool import ConnectionPool
```
bluetech, line 220: This import requires `psycopg[pool]` to be installed, I think it will be helpful to users to catch `ModuleNotFoundError` here and provide some helpful error explaining that `psycopg[pool]` needs to be installed. It would also be good to mention in the docs.

django/db/backends/postgresql/base.py
```diff
@@ -272,7 +330,12 @@ def get_new_connection(self, conn_params):
...
                 )
-        connection = self.Database.connect(**conn_params)
+        if self.pool:
+            # If nothing else has opened the pool, open it now
+            self.pool.open()
+            connection = self.pool.getconn()
+        else:
+            connection = self.Database.connect(**conn_params)
         if set_isolation_level:
             connection.isolation_level = self.isolation_level
```
bluetech, line 340: This line gave me pause; it modifies an attribute on a connection that (being part of a pool) is now reused. It seems like something better done in `configure_connection`. But, I guess this was already the case with persistent connections and if it's fine there, probably also fine here.

docs/ref/databases.txt
```diff
@@ -244,6 +244,30 @@ database configuration in :setting:`DATABASES`::
...
     }
 
+.. _postgresql-pool:
+
+Connection Pool
+---------------
+
+To use a connection pool with `psycopg`_ 3.1.8+, you can either set ``"pool"``
+in the :setting:`OPTIONS` part of your database configuration in
+:setting:`DATABASES` to be ``True`` or as a dict to be passed to
```
bluetech, line 254: Maybe worth mentioning that `True` means the psycopg defaults are used?

django/db/backends/postgresql/base.py
```diff
@@ -179,6 +197,45 @@ class DatabaseWrapper(BaseDatabaseWrapper):
...
     # PostgreSQL backend-specific attributes.
     _named_cursor_idx = 0
+    _connection_pools = {}
+
+    @property
+    def pool(self):
+        if not is_psycopg3:
+            return None
+
+        if self.alias not in self._connection_pools:
```
bluetech, line 207: In my code, I used `(alias, dbname, conn_params)` as the key, not just the alias. I don't remember why, but I don't think it was to fix something in particular.

One thing that using only `alias` does mean is that any changes made to the `DATABASES` entry after the fact will be completely ignored, although this is probably not something that Django condones officially (maybe in tests?).

bluetech (commented):
Thanks for working on this @sarahboyce. I left some minor comments, but it looks good to me.

django/db/backends/postgresql/base.py
apollo13: Mhm, yes we should probably connect this to `setting_changed` signal and then clean up `self._connection_pools`. But aside from that I think the alias should be enough, changing settings is really really really not supported ;)

django/db/backends/postgresql/base.py
sarahboyce: Are you referring to `connection.isolation_level = self.isolation_level`? I believe that was done here 76356d963c131252e1ce4285083fd21fd0bdedd9 as part of [#24318](https://code.djangoproject.com/ticket/24318) if that helps 👍

sarahboyce: Ah sorry, I think I get what you mean now... true 🤔

django/db/backends/postgresql/base.py
sarahboyce: Added to the docs - great catch thank you!
I can't find other places that we are catching a `ModuleNotFoundError` to add a more helpful message. So tempted not to (I would also need to add a test to cover it I think).

docs/ref/databases.txt
sarahboyce: Good suggestion thank you, added in ⭐

docs/ref/databases.txt
```diff
@@ -244,6 +244,31 @@ database configuration in :setting:`DATABASES`::
...
 
+.. _postgresql-pool:
+
+Connection Pool
+---------------
+
+To use a connection pool with `psycopg`_ 3.1.8+, you can either set ``"pool"``
+in the :setting:`OPTIONS` part of your database configuration in
+:setting:`DATABASES` to be a dict to be passed to `ConnectionPool`_ or to
+``True`` to use the `ConnectionPool`_ defaults::
```
ngnpope, line 255: It should be possible to use intersphinx links here... Maybe something like:
```suggestion
:setting:`DATABASES` to be a dict to be passed to :class:`~psycopg:psycopg_pool.ConnectionPool` or to
``True`` to use the ``ConnectionPool`` defaults::
```

django/db/backends/postgresql/base.py
sarahboyce: I added a test `test_connect_pool_with_isolation_level` to try and cover when it gets set "twice". I don't think it's doing any harm?

docs/ref/databases.txt
```diff
@@ -244,6 +244,32 @@ database configuration in :setting:`DATABASES`::
...
+            "OPTIONS": {
+                "pool": True,
+            },
+        },
+    }
+
+This option requires ``psycopg[pool]`` or ``psycopg_pool`` to be installed and
+is ignored with ``psycopg2``.
+
+.. _ConnectionPool: https://www.psycopg.org/psycopg3/docs/api/pool.html#psycopg_pool.ConnectionPool
```
ngnpope, line 271:
```suggestion
```

django/db/backends/postgresql/base.py
bluetech: One production scenario where this is relevant:

I (used to) use ["AWS RDS IAM authentication"](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.Connecting.Python.html) (similar in [Google](https://cloud.google.com/sql/docs/postgres/iam-logins#logging-in-as-a-user), [Microsoft](https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/how-to-connect-with-managed-identity#retrieve-the-access-token-from-the-azure-instance-metadata-service), [Vault](https://developer.hashicorp.com/vault/docs/secrets/databases/postgresql)), basically it means you call an external service and get a temporary connection password that is valid for 15 minutes. The way people usually implement this it is by overriding `get_connection_params` and setting `params['password']`, taking care to refresh the password when needed.

I sent a question about this to psycopg https://github.com/psycopg/psycopg/discussions/500 but since I stopped using this method I haven't followed up unfortunately. However the comment by Daniele does suggest that changing the connection params on the pool after the fact can be officially supported if the work is done to ensure it doesn't regress.

django/db/backends/postgresql/base.py
bluetech: >  I can't find other places that we are catching a ModuleNotFoundError to add a more helpful message. So tempted not to (I would also need to add a test to cover it I think).

You will get some hits if you search for `ImportError`, as `ModuleNotFoundError` is a newer more targeted subclass of `ImportError` and not often used yet. But a doc comment is good enough too.

django/db/backends/postgresql/base.py
bluetech: From looking at the code (haven't verified in practice), I think the problem can happen in the following scenario:

- `isolation_level` is *not* configured in options (so `set_isolation_level` is false). In this case READ_COMMITTED is expected.
- User gets a connection from the pool.
- User sets `conn.isolation_level = SERIALIZABLE`
- User closes the connection (returns to the pool)
- User gets a connection from the pool (gets same reused connection from before)
- Expected: isolation level is READ_COMMITTED
- Actual: isolation level is SERIALIZABLE

I should note that the same issue happens with persistent connections, and I also see now that it is clearly documented as a caveat here https://docs.djangoproject.com/en/5.0/ref/databases/#caveats:

> If you modify parameters such as the connection’s isolation level or time zone, you should either restore Django’s defaults at the end of each request, force an appropriate value at the beginning of each request, or disable persistent connections.

So perhaps the fix is to have a similar note for the connection pool. That said, given that the pool always goes through the `DatabaseBackend` it *is* possible make this "just work" by making sure to never skip setting `isolation_level`, at least when the pool is used.

tests/backends/postgresql/tests.py
```diff
@@ -223,6 +223,82 @@ def test_connect_non_autocommit(self):
...
+        from psycopg_pool import PoolTimeout
+
+        new_connection = connection.copy()
+        self.assertIsNone(new_connection.pool)
+
+        new_connection.settings_dict["OPTIONS"]["pool"] = {
+            "min_size": 2,
+            "timeout": 0.1,
+        }
+        new_connection._connection_pools = {}
```
bluetech, line 237: This line clears all of the process-level connection pools without closing them, which might trigger some ResourceWarnings in the Django test suite, in case you care about those.

I don't know if the DJango test suite has the notion of running specific tests in isolated processes. If it does, then this is a good candidate for it.

@sarahboyce I'm curious if you tried to run the Django test suite itself with a connection pool configuration. It will be a good exercise to see what breaks if you do this (ideally everything passes...).

django/db/backends/postgresql/base.py
```diff
@@ -179,6 +197,45 @@ class DatabaseWrapper(BaseDatabaseWrapper):
     ops_class = DatabaseOperations
     # PostgreSQL backend-specific attributes.
     _named_cursor_idx = 0
+    _connection_pools = {}
```
bluetech, line 200: Another regular safety hazard with mutable global resources aside from thread safety is "fork safety" - what happens with the resource in the child process after a fork. AFAICT psycopg pool is not "fork safe" (https://github.com/psycopg/psycopg/issues/651#issuecomment-1738659965). There are of course many WSGI servers using fork. What I'm not sure about is whether Django tries to be compatible with this scenario:

- Process opens a connection
- Process forks
- Child opens a connection

or if it requires the master process to not do any DB connections before forking.

(With existing connections/persistent connections code this is OK because the connections are stored in a thread local which gets reset on fork.)

bluetech (commented):
Thanks for the update, I left a few more replies and comments.

django/db/backends/postgresql/base.py
apollo13: That is fair, but how gets `self._connection_pools` ever cleaned up if the `conn_params` change every X minutes? Maybe we should have it as `{alias: (dbname, conn_params, pool)}` and verify that dbname/conn_params didn't change?

django/db/backends/postgresql/base.py
apollo13: Is it really okay with persistent connections? After fork if the thread local gets reset wouldn't that mean that the connection is garbage collected which could possibly close the tcp connection? I don't think anything in Django is written with fork safety in mind. While WSGI servers use fork it generally only works if the fork early enough I think (preferably before loading the WSGI app)

tests/backends/postgresql/tests.py
sarahboyce: I have tweaked the code and improved the after test cleanup (maybe check again and see if you think that's better?)

> I'm curious if you tried to run the Django test suite itself with a connection pool configuration.
I hadn't but that is a good idea. It initially failed immediately with:
```
Found 16748 test(s).
Cloning test database for alias 'default'...
Got an error creating the test database: source database "test_django" is being accessed by other users
DETAIL:  There are 4 other sessions using the database.
```
Which I was able to get around by setting `parallel` to 1, but I think that means there is an issue with `_clone_test_db` which tries to close connections (and maybe needs to close the pool? Not sure yet)
Then 16k tests not parallized takes a little time, looks like a handful of failures (a couple are expected based off the way I wrote the pool tests) but mostly passing.
Going to try to look into `_clone_test_db`

django/db/backends/postgresql/base.py
sarahboyce: Updated

tests/backends/postgresql/tests.py
sarahboyce: 4244d71d829a717a1a871d52bdda910d173bb6b1 this means I can now run them in parallel
There's quite a lot to look into here, mostly timezone tests are failing. But this is a simplified list of the failing tests
```
db_functions.datetime.test_extract_trunc.DateFunctionWithTimeZoneTests
timezones.tests.LegacyDatabaseTests
timezones.tests.NewFormsTests.test_model_form
cache.tests.DBCacheWithTimeZoneTests

backends.postgresql.tests.Tests.test_connect_and_rollback
backends.postgresql.tests.Tests.test_connect_custom_cursor_factory
backends.postgresql.tests.Tests.test_connect_role
backends.tests.BackendTestCase.test_is_usable_after_database_disconnects
```

sarahboyce: 80f11f5ec3938bad12a97b6be447fb65ca63ef7b this takes care of the timezone tests and then only 
Only `backends.postgresql.tests.Tests.test_connect_custom_cursor_factory` I think is maybe worth looking into now and I think that is due to the settings changes during the test itself
So basically there

django/db/backends/postgresql/base.py
sarahboyce: > isolation_level is not configured in options (so set_isolation_level is false). In this case READ_COMMITTED is expected.

I think `connection.isolation_level` is None in this case as `set_isolation_level` is False and it will then take the Database default, which is probably `READ_COMMITTED`.

> User sets conn.isolation_level = SERIALIZABLE

How do they do this sorry?

django/db/backends/postgresql/base.py
sarahboyce: What are you expecting if the dbname/conn_params have changed? Should I update them in the dictionary, close the existing pool and open a new one?

django/db/backends/postgresql/base.py
bluetech: > How do they do this sorry?

Looking now, there does not seem to be a sanctioned way to do it from Django. And 3rd party libraries like https://github.com/Opus10/django-pgtransaction do it with an SQL statement, not setting the psycopg connection property. So it doesn't seems like a scenario that Django needs to concern itself with.

django/db/backends/postgresql/base.py
bluetech: As @apollo13 said, including the params in the key is definitely no good, it will cause `cls._connection_pools` to grow unbounded in the "new password every X minutes" scenario.

As I mentioned in the psycopg discussion link, I think the only viable solution is to update the pool itself on `settings_changed` such that it will use the new params when creating new connections. Currently it's actually possible to do this, but Daniele says it's accidental, and before it can be relied upon, needs a regression test in psycopg.

I think that for now, we should say that changing settings dynamically is not going to work with `pool`.

> What are you expecting if the dbname/conn_params have changed? Should I update them in the dictionary, close the existing pool and open a new one?

I can't see this working; existing connections using the old password/setting continue to work, closing the pool will close all existing connections (even ones currently in use by ongoing requests), which is very disruptive.

django/db/backends/postgresql/base.py
bluetech: > Is it really okay with persistent connections? After fork if the thread local gets reset wouldn't that mean that the connection is garbage collected which could possibly close the tcp connection?

libpq (the underlying C lib used by psycopg) is using `O_CLOEXEC` (since [a long time](https://github.com/postgres/postgres/commit/7627b91cd5d) ago at least) so the socket isn't inherited in fork.

>  I don't think anything in Django is written with fork safety in mind. While WSGI servers use fork it generally only works if the fork early enough I think (preferably before loading the WSGI app)

I myself really only use thread and gevent workers so I don't have much operation experience with how process workers work. I found the following description in uWSGI docs ([link](https://uwsgi-docs.readthedocs.io/en/latest/articles/TheArtOfGracefulReloading.html#preforking-vs-lazy-apps-vs-lazy)):

"By default uWSGI loads the whole application in the first process and after the app is loaded it does fork() itself multiple times."

My guess is that it at least means it imports the wsgi entry point, which in turns does `django.setup()`. Is it guaranteed that `django.setup()` doesn't cause any queries to run? (The warning [here](https://docs.djangoproject.com/en/5.0/ref/applications/#django.apps.AppConfig.ready) certainly advises against it...)

tests/backends/postgresql/tests.py
bluetech: Hmm regarding the change in 80f11f5: removing the `and not self.pool` condition from `init_connection_state` is odd because it means that `_configure_connection` is now runs twice, once by the pool's `configure` callback, and once by `init_connection_state`. Were you able to understand why it's needed?

django/db/backends/postgresql/base.py
sarahboyce: Thank you for looking into it 🙇‍♀️

tests/backends/postgresql/tests.py
sarahboyce: So by not doing this for pool, the timezone changes were never commited and then broke tests when using a pool setup.
I have now removed the logic refactor creating `_configure_connection`.

django/db/backends/postgresql/base.py
sarahboyce: > Is it guaranteed that django.setup() doesn't cause any queries to run? (The warning [here](https://docs.djangoproject.com/en/5.0/ref/applications/#django.apps.AppConfig.ready) certainly advises against it...)

Since Django 5.0, this raises a Runtimewarning. So not guaranteed not to happen but actively discouraged.

tests/backends/postgresql/tests.py
apollo13: I am not sure that the new refactor is correct. Now we configure the connection every time on pool checkout this is certainly wrong! FWIW if the tests fail with the pool it might be that they also fail currently with persistent connections.

apollo13: Let me expand on my previous comment to explain why I think that the latest refactor is wrong:

* With default connections and non-persistent connections we close a connection every request and reopen it every request. As such we need one query to set the timezone every request
* With persistent connections the connection stays open and the timezone is only set once per connection. The assumption is that the view doesn't change the timezone in a bad way.
* With a pool we are basically back to the first case, as far as Django is concerned we are closing the connection every request (but in reality put it back to the pool) and once we check it out again it should not net to apply a new timezone (ie it behaves like a persistent connection)

To test the behavior I would recommend the following: 

 * Disable pooling, set a timezone in the settings that is != your default timezone and enable query logging in postgres. Verify that the timezone is set every request.
 * Now enable persistent connections and verify that the timezone is only set once per connection
 * Now disable persistent connections and enable a pool with a maxsize of 1, verify that the timezone is set once per connection

sarahboyce: I investigated and `cache.tests.DBCacheWithTimeZoneTests` was failing because it was running `createcachetable` management command on test setUp (which then creates a database connection in the pool). If I add similar logic to `_close_all_connections` which I am running on on the test database cloning/creating methods, then this all passes.

I think these issues are more around test state clean up which I guess I'm still trying to figure out.

I can add tests around the timezone setting so I don't remove the logic again 😁

apollo13:
Fwiw, while it is nice if the test suite runs with a pool enabled I don't think it is a requirement to get this merged. The test suite is rather special after all. Nevertheless it might also show us bugs in the implementation.

django/db/backends/postgresql/base.py
bluetech: This is good enough for me!

django/db/backends/postgresql/base.py
apollo13: I will be perfectly honest. I am not sure I 100% understand what the transaction isolation code is doing here and why it is written as it is. It should be possible to move this into `_configure_connection` as well. But maybe not since the base backend calls:
```
        self.connection = self.get_new_connection(conn_params)
        self.set_autocommit(self.settings_dict["AUTOCOMMIT"])
        self.init_connection_state()
```
so we would have to investigate whether changing the transcation level after the autocommit settings will work. In hindsight I now realize that we also change the timezone/role earlier when using the pool as opposed to the old behavior where it is done in `init_connection_state`. Can this cause problems? I think no, but…

apollo13: Thinking more about this, the comment above:
```
        # - before calling _set_autocommit() because if autocommit is on, that
        #   will set connection.isolation_level to ISOLATION_LEVEL_AUTOCOMMIT.
```
is no longer true (https://www.psycopg.org/docs/connection.html#connection.isolation_level):
```
This value is now entirely separate from the [autocommit](https://www.psycopg.org/docs/connection.html#connection.autocommit) property: in previous version, if autocommit was set to True this property would have returned [ISOLATION_LEVEL_AUTOCOMMIT](https://www.psycopg.org/docs/extensions.html#psycopg2.extensions.ISOLATION_LEVEL_AUTOCOMMIT); it will now return the server isolation level.
```
so we might be able to adjust this as needed. Then again we probably have to set this before setting autocommit to off because otherwise it would affect only the current transaction? Needs more thought.

HamaBarhamou:
> @apollo13a fait tout le travail acharné dans ce PR : #16881 dans l'espoir de le faire franchir la ligne.
> 
> J'espère également que les tests testent la bonne chose, j'ai essayé de m'inspirer des `psycopg_pool`tests : https://github.com/psycopg/psycopg/blob/52ed68a9d699c1e43e07a4d04441534197552de5/tests/pool/test_pool.py#L103
> 
> Mais faites-moi savoir ce qui me manque - un tout nouvel espace pour moi 👍
> 
> Billet associé : https://code.djangoproject.com/ticket/33497

really glad to see that y can continue where x left off

docs/ref/databases.txt
```diff
@@ -244,6 +244,34 @@ database configuration in :setting:`DATABASES`::
...
+            },
+        },
+    }
+
+This option requires ``psycopg[pool]`` or ``psycopg_pool`` to be installed and
+is ignored with ``psycopg2``.
+
+.. warning::
+
+    Connection pools do not handle dynamically updated database settings.
```
felixxm, line 273: As far as I'm aware, this is generally not supported (and never has been). I don't think we need this warning.

docs/ref/databases.txt
```diff
@@ -244,6 +244,34 @@ database configuration in :setting:`DATABASES`::
         },
     }
 
+.. _postgresql-pool:
+
+Connection Pool
+---------------
+
+To use a connection pool with `psycopg`_ 3.1.8+, you can either set ``"pool"``
```
felixxm, line 252: 3.1.8 is the minimum supported version, IMO, we don't need to mention it here
```suggestion
To use a connection pool with `psycopg`_, you can either set ``"pool"``
```

tests/backends/postgresql/tests.py
```diff
@@ -223,6 +223,90 @@ def test_connect_non_autocommit(self):
...
+            conn_params = new_connection.get_connection_params()
+            connection_1 = new_connection.get_new_connection(conn_params)
+            new_connection.get_new_connection(conn_params)
+            with self.assertRaises(PoolTimeout):
+                # The pool has a maximum of 2 connections.
+                new_connection.get_new_connection(conn_params)
+
+            new_connection.pool.putconn(connection_1)
+            connection_3 = new_connection.get_new_connection(conn_params)
+            # Confirm reused the first connection when one is available.
```
felixxm, line 248: To add more explanation, comments should state the expected behavior and omit prefixes like "Confirm" since all tests confirm things. This guideline is from [Python coding style](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#python-style).


```suggestion
            # Confirm reused the first connection when one is available.
```

django/db/backends/postgresql/base.py
```diff
@@ -272,7 +336,12 @@ def get_new_connection(self, conn_params):
                     f"Invalid transaction isolation level {isolation_level_value} "
                     f"specified. Use one of the psycopg.IsolationLevel values."
                 )
-        connection = self.Database.connect(**conn_params)
+        if self.pool:
+            # If nothing else has opened the pool, open it now
```
felixxm, line 340:
```suggestion
            # If nothing else has opened the pool, open it now.
```

felixxm:
I tried to run our test suite with `"pool": True` with no luck :disappointed: :
```
FAILED (failures=103, errors=14, skipped=506, expected failures=7)
```

felixxm:
Also, it crashes when destroying test databases:
```
$./runtests.py backends
....
FAILED (failures=7, errors=2, skipped=82)
discarding closed connection: <psycopg.Connection [BAD] at 0x7fdc8b99dff0>
  File "django/django/db/backends/utils.py", line 103, in _execute
    return self.cursor.execute(sql)
  File "python3.10/site-packages/psycopg/cursor.py", line 737, in execute
    raise ex.with_traceback(None)
psycopg.OperationalError: consuming input failed: decryption failed or bad record mac
```

sarahboyce:
The backend postgres tests are not designed to pass with a pool as they for example check that there is no pool and change the settings. The test cleanup would also not work as it's relying on the tests being ran not as a pool.
Were your test failures mostly timezone tests (if you ignore the backend tests)?

sarahboyce:
```
FAILED (failures=102, errors=13, skipped=380, expected failures=7)
```
I have basically the same. I can try to rewrite the backend tests that they would pass with the change of underlying settings (but might be quite a lot of test changes). 
This 80f11f5ec3938bad12a97b6be447fb65ca63ef7b toke care of the timezone tests but then I was warned that getting the test suite to pass with a connection pool might not be what we should aim for. So I don't really know if I should continue with it 🤔 
@felixxm what do you think?

felixxm:
> The backend postgres tests are not designed to pass with a pool as they for example check that there is no pool and change the settings.

As for me this is unexpected :thinking: Cannot we adjust them to work? by using a connection copy and changing `OPTIONS`.

felixxm:
> I have basically the same. I can try to rewrite the backend tests that they would pass with the change of underlying settings (but might be quite a lot of test changes). This [80f11f5](https://github.com/django/django/commit/80f11f5ec3938bad12a97b6be447fb65ca63ef7b) toke care of the timezone tests but then I was warned that getting the test suite to pass with a connection pool might not be what we should aim for. So I don't really know if I should continue with it 🤔 @felixxm what do you think?

Not sure what the plan was, but are we going to document that users cannot test their apps when connection pooling is enabled? They can encounter the same issues.

Some failures are really suspicious, e.g. `backends.tests.BackendTestCase.test_is_usable_after_database_disconnects`:
```
  File "/django/tests/backends/tests.py", line 486, in test_is_usable_after_database_disconnects
    self.assertFalse(connection.is_usable())
AssertionError: True is not false
```

Why connection is usable after `_close()`? Maybe I'm missing something.

apollo13:
Ah that is because the connection isn't closed on `close` but only returned to the pool. I guess we really have to set `self.connection` to `None` afterwards since we no longer own it and adjust the remaining code to handle `None` there then (eg in `is_usable`)

apollo13:
So, I had some time and managed to get the testsuite passing with pooling enabled. During that I discovered one bug in the implementation that we should discuss (see below). I opened a PR against this PR, see https://github.com/sarahboyce/django/pull/10 for the details.

The code looks clean imo but there are a few things that might need some improving (note that by default the whole testsuite does not run under pooling, so the issues don't show up and even when they show up, this is basically Django testing transaction behavior. Endusers will probably never see that).

I am getting:
> rolling back returned connection: <psycopg.Connection [INTRANS] (user=florian database=test_django_test_1) at 0x7f61920c5690>

for the following two tests:
* `test_hooks_cleared_on_reconnect`
* `test_atomic_prevents_queries_in_broken_transaction_after_client_close`

Both tests explicitly close the connection (and as such return it to the pool) while a transaction is active. So the warning is expected and the only thing I can think of is using filterwarnings to filter out those pgpool warnings here. But as said above I don't think this is necessary since that is not a default configuration we run the tests against.

I had the same warning for `test_skip_class_unless_db_feature` but I was able to fix it via https://github.com/sarahboyce/django/pull/10/commits/962566c4b62ef48fbdae1cb2ff75ed69147d186c -- this wouldn't hurt in current main either, @felixxm do you want an extra cleanup PR for that one or can we just keep it as extra commit in this PR?

`test_atomic_prevents_queries_in_broken_transaction_after_client_close` exposed one bug. Because we set `connection = None` in `_close` after we returned the connection to the pool (to ensure we do not use it…), we were able to reopen a connection inside an atomic. Essentially this means that
```py
        with transaction.atomic():
            Reporter.objects.create(first_name="Archibald", last_name="Haddock")
            connection.close()
            # The connection is closed and the transaction is marked as
            # needing rollback. This will raise an InterfaceError on databases
            # that refuse to create cursors on closed connections (PostgreSQL)
            # and a TransactionManagementError on other databases.
            with self.assertRaises(Error):
                Reporter.objects.create(first_name="Cuthbert", last_name="Calculus")
```
no longer raises. Without pooling `self.connection` is not `None` and as such opening a new cursor raises. I was able to fix this in `ensure_connection` by adding:
```
        if self.in_atomic_block and self.closed_in_transaction:
            raise ProgrammingError("Cannot open a new connection in an atomic block.")
```
Basically use the existing hints to determine if we closed inside a transaction and error out. I am wondering if it would make sense to move this solely into the postgres backend. In the end though the check is not *wrong* but instead of relying on the database adapter to throw errors for closed connections we error out realy -- sounds okay?

I had to rewrite a few postgres tests to use a non-pooling connection because the pool doesn't tolerate changes to the settings dict like presented (see https://github.com/sarahboyce/django/pull/10/commits/bbd5c5db2c3523b1e0b02e3242ae5c1aa5d2c34d for details). There might be some ways to rewrite them in a pooling way, this way we can reuse the existing tests to test whether timezones & assume_role work instead of adding new ones. I don't have strong feeling either way though.

I will do some manual testing against a simple app, but so far so good.

One thing left on the todo: Incorporate health checking directly into the pool (supported as of 2.0)

apollo13:
Btw we should probably create a new ticket for this so we have something to mark as "ready to review"? :D

felixxm:
> I had the same warning for `test_skip_class_unless_db_feature` but I was able to fix it via [sarahboyce@962566c](https://github.com/sarahboyce/django/commit/962566c4b62ef48fbdae1cb2ff75ed69147d186c) -- this wouldn't hurt in current main either, @felixxm do you want an extra cleanup PR for that one or can we just keep it as extra commit in this PR?

Merged in 6e520d953773d25a3d3484db67feed446aca0bc1.

ridersofrohan:
Thanks for all the work on this! Is there any update on when it will be merged in and included in a Django Release?

apollo13:
No, we don't give any ETAs since we are doing this in our free time. If you want to get this in faster, please test it and review the code.

gerardrbentley:
I'm finding this does the trick to limit PG connections when running under ASGI (in that not every request grabs a new connection). Awesome work; much less complicated than pg_bouncer for most needs

Here's my test repo using the patch branch if anyone wants to extend: https://github.com/gerardrbentley/demo-django

django/test/utils.py
```diff
@@ -359,6 +359,9 @@ def teardown_databases(old_config, verbosity, parallel=0, keepdb=False):
     """Destroy all the non-mirror databases."""
     for connection, old_name, destroy in old_config:
         if destroy:
+            # TODO: Hack :/
```
charettes, line 362: Any reason why this can't be moved to the Postgres backend `destroy_test_db` method?

apollo13: Nothing specific no, should be doable :+1:

apollo13: Got rid of it completely, couldn't find the reason why it was there :)

apollo13: Oh I found it. When tests do fail it leaves the pool open. I wonder if we should add a hook to Django to prepare the base backend for something like `close_pool`

charettes: Does this hook already exist in `destroy_test_db`? Why can't it be moved there?

apollo13: because there are more locations where I might wanna do some closing, like here: https://github.com/django/django/blob/d1be05b3e9209fd0787841c71a95819d81061187/django/db/backends/base/creation.py#L64

But yes, for now I will probably add it inside `destroy_test_db`.

apollo13: I made it postgres specific in `_destroy_test_db` for now.

yorickr:
Hi! 
First of all, thanks for all the work that's being done on this.

I'm looking to use this thread pool in the future, in combination with `gevent`. After some local testing, I found out the current solution opens a new threadpool on every new request, which for my use case is the same as using persistent connections.

Is this something that can/will be considered in the requirements? Is this threadpool even something that can be used with gevent in the first place?

Thank you :pray:

apollo13:
As far as I know Django doesn't support `gevent` officially. So no, this is not something I would consider, I have no idea how `gevent` works. That said if the existing thread locals do work, I am kinda surprised since I am not doing anything magic (I think).

bluetech:
@yorickr

> I'm looking to use this thread pool in the future, in combination with gevent. After some local testing, I found out the current solution opens a new threadpool on every new request, which for my use case is the same as using persistent connections.

I think gevent should work without any problem (and correctly) with this implementation, can you provide a way to reproduce what you're seeing?

django/db/backends/postgresql/base.py
```diff
@@ -179,6 +197,54 @@ class DatabaseWrapper(BaseDatabaseWrapper):
     ops_class = DatabaseOperations
     # PostgreSQL backend-specific attributes.
     _named_cursor_idx = 0
+    _connection_pools = {}
+
+    @property
+    def pool(self):
+        pool_options = self.settings_dict["OPTIONS"].get("pool")
+        if not pool_options or self.alias == NO_DB_ALIAS or not is_psycopg3:
```
felixxm, line 205: Checking `is_psycopg3` seems redundant as it already raises `ImproperlyConfigured`.

felixxm:
I started working on this patch .... :construction:

django/db/backends/postgresql/base.py
```diff
@@ -256,7 +255,9 @@ def get_database_version(self):
     def get_connection_params(self):
         settings_dict = self.settings_dict
         # None may be used to connect to the default 'postgres' db
-        if settings_dict["NAME"] == "" and not settings_dict["OPTIONS"].get("service"):
+        if settings_dict["NAME"] == "" and not settings_dict.get("OPTIONS", {}).get(
```
apollo13, line 258: I explicitly checked that we always configure "OPTIONS", when might it be missing?

felixxm: Sure, maybe it's not needed. However, it has nothing to do with connection pooling, so we should merge it as a separate cleanup.

apollo13:
Fair enough:)

django/db/backends/postgresql/base.py
felixxm: #17900

django/db/backends/postgresql/base.py
felixxm: #17900

django/db/backends/base/base.py
```diff
@@ -270,6 +270,8 @@ def check_settings(self):
     @async_unsafe
     def ensure_connection(self):
         """Guarantee that a connection to the database is established."""
+        if self.in_atomic_block and self.closed_in_transaction:
+            raise ProgrammingError("Cannot open a new connection in an atomic block.")
```
felixxm, line 274: This is untested.

tests/backends/postgresql/tests.py
```diff
@@ -20,6 +20,15 @@
     is_psycopg3 = False
 
 
+def no_pool_connection(alias=None):
+    new_connection = connection.copy(alias)
+    new_connection.settings_dict = copy.deepcopy(connection.settings_dict)
+    # Ensure that the second connection circumvents the pool, this is kind
+    # of a hack, but we cannot easily change the pool connections
+    new_connection.settings_dict.setdefault("OPTIONS", {})["pool"] = None
```
felixxm, line 28:
```suggestion
    new_connection.settings_dict["OPTIONS"]["pool"] = None
```

django/db/backends/postgresql/base.py
```diff
@@ -285,38 +360,54 @@ def get_new_connection(self, conn_params):
...
-        timezone_name = self.timezone_name
-        if timezone_name and conn_timezone_name != timezone_name:
-            with self.connection.cursor() as cursor:
-                cursor.execute(self.ops.set_time_zone_sql(), [timezone_name])
-            return True
-        return False
-
-    def ensure_role(self):
-        if self.connection is None:
-            return False
```
felixxm, line 300: Moved to the #17907.

django/db/backends/postgresql/base.py
```diff
@@ -179,6 +197,53 @@ class DatabaseWrapper(BaseDatabaseWrapper):
...
+            # parallel. Since we do not open the pool during it's init above,
+            # this means that at worst during startup multiple threads generate
+            # pool objects and the first to set it wins.
+            self._connection_pools.setdefault(self.alias, pool)
+
+        return self._connection_pools[self.alias]
+
+    def close_pool(self):
+        if pool := self.pool:
+            pool.close()
```
felixxm, line 245: Do we need a temporary variable? :thinking: 
```suggestion
        if self.pool:
            self.pool.close()
```

django/db/backends/base/base.py
felixxm: As far as I'm aware, we don't need this because these three conditions:
- `in_atomic_block is True`
- `closed_in_transaction is True`
- `connection is None`

cannot be fulfilled at the same time (at least in normal flow).

sarahboyce: I couldn't figure out how to make the situation happen naturally 👍 so makes sense to me

apollo13: This resulted (at least in earlier versions) in a testsuite failure where the atomic behavior was tested. Namely psycopg by default would raise an error if you did something on a closed connection. Now that `close` sets it to `None` it transparently opened a new one :/ This was visible if the Django testsuite was ran with a settings_postgresql that had pooling enabled.

The test in question was `test_atomic_prevents_queries_in_broken_transaction_after_client_close`.

felixxm: Thanks :+1: Now I'm smarter. Let me move this to the `if connection is None` branch.

apollo13: It might make sense to copy the test in question into the postgresql pool specific tests as well and ensure that it works properly under a pool (even when the test settings themself do not run with a pool?)

felixxm:
Superseded by #17914.

PR #17597, opened 2023-12-11
Add alias for show_migrations

rca:
Whenever I want to show migrations I end up running `./manage.py show_migrations` because the other `show*` management commands contain an underscore.  This patch add the alias `show_migrations` so that the command works with and without the underscore.

Thank you!

### DIFF
--- a/django/core/management/commands/show_migrations.py
+++ b/django/core/management/commands/show_migrations.py
@@ -1,0 +1,7 @@
+"""
+An alias to the `showmigrations` command
+
+All other `show*` management commands contain an underscore except this one, which makes
+it harder for muscle memory to call this command correctly.
+"""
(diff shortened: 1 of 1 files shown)


ngnpope:
There are no other management commands provided by Django that start with `show_*`. Perhaps you're using a third-party app that is providing one with that prefix?

Additional aliases like this serve no purpose and will confuse everyone else. You can add this to your own project if you desire it.

shangxiao:
Just FYI here's one way I add aliases to my projects: https://github.com/shangxiao/stupid-django-tricks/tree/master/management_command_alias

I get tired of either typing the full `showmigrations` or pressing tab and waiting for it respond after a couple of seconds so I have an alias `sm` setup.

PR #17598, opened 2023-12-12
Flake8 updates in config.py and registry.py

alesiram:


shangxiao:
@alesiram Thanks but also don't accept linting PRs as it adds unnecessary noise to the commit history :)

PR #17599, opened 2023-12-12
feat: add shortcuts

michaelschem:
### Description

This pull request introduces keyboard shortcuts to the Django admin interface, enhancing the user experience by allowing faster navigation and actions using the keyboard. Three main shortcuts have been implemented:

1. **Save and Continue Editing**: Allows users to save changes and remain on the current page.
2. **Save**: Saves changes and returns to the admin index page.
3. **Save and Add Another**: Saves the current object and opens a new form for adding another object of the same type.

The following table summarizes the keyboard shortcuts for both Windows and Mac users:

| Action                  | Windows/Linux Shortcut        | Mac Shortcut             |
|-------------------------|-------------------------|--------------------------|
| Save and Continue Editing | Ctrl + S               | Cmd + S                  |
| Save                    | Ctrl + Shift + S        | Cmd + Shift + S          |
| Save and Add Another    | Ctrl + Alt + S          | Cmd + Option + S         |

### Implementation Details

- JavaScript code has been added to `base_site.html` to listen for specific key combinations and trigger the corresponding actions.
- The script targets the `input` elements of the admin forms based on their `name` attributes (`_continue`, `_save`, and `_addanother`).
- The implementation ensures no conflict with existing browser or system shortcuts and is designed to enhance usability without disrupting current workflows.

### Testing

- The feature has been tested across various browsers (Chrome, Firefox, Safari) to ensure compatibility and functionality.
- TODO: Add Additional tests to cover the new feature and ensure it does not interfere with existing admin functionalities.

### Documentation

- TODO: Add documentation to include the new keyboard shortcuts and instructions on their usage.

This feature aims to streamline the workflow in the Django admin interface, making it more efficient and user-friendly, especially for power users who rely heavily on keyboard navigation.

### Discussion

A discussion has been created [here](https://forum.djangoproject.com/t/django-admin-save-shortcuts/27016/2)


pauloxnet:
Please add instructions also for Linux users, they are the majority of all Django developers as per the Django Survey 2022
https://lp.jetbrains.com/django-developer-survey-2022/

michaelschem:
@pauloxnet windows and linux should be the same. Description updated.

thibaudcolas:
Hi @michaelschem 👋 I think it’d be great to have keyboard shortcuts in Django, but for a feature like this I’d personally expect some discussion in a ticket or forum post beforehand. In particular I’d like to hear people’s thoughts on:

- Which actions should have shortcuts, based on frequency of use, availability in other similar tools.
- What key combinations should be used. Based on consistency with other tools.
- How we could make those shortcuts discoverable by end users (for example, in GitHub on macOS I can press <kbd>Shift</kbd> + <kbd>?</kbd> to view a list)

michaelschem:
hey, @thibaudcolas, thanks for your feedback! I also agree a discussion about the points you brought up would be great. I was under the impression this would be the correct place for that kind of a thing, but I'll go ahead and raise the issue on the form and link it here.

michaelschem:
@thibaudcolas I added hints to the buttons on hover. I think this would be a good start for discoverability. 
image

thibaudcolas:
@michaelschem no, any use of the `title` attribute is completely unusable for keyboard users in particular, so we shouldn’t use it. Personally I’d recommend to never ever use the `title` attribute in HTML except for `iframe` elements.

michaelschem:
hey, @thibaudcolas I'm happy to make the change to add the keyboard shortcut hints, but can you merge this PR first? It seems pretty low impact as far as risk and a huge time saver for my Django admin users.

thibaudcolas:
@michaelschem please spend some time reading Django’s process for feature requests like this as defined in our docs. I’m not too up to speed with it myself but I believe we expect all proposed features to start with a forum discussion, which hasn’t happened here yet.

As far as this being low impact / a time saver – I don’t think it’s acceptable we’d add shortcuts like this without any way for users to learn them. Django doesn’t have a documentation site for the admin, so we’d have to have docs within the admin in some form.

And we’d definitely want some review of the actual shortcuts. For example, personally without spending too much time thinking about it I’d have expected the "Save" shortcut to be Cmd/Ctrl + S without modifiers.

tim-schilling:
@michaelschem as a heads up, this package was developed as a part of a GSoC project. We're still waiting on adoption before considering adopting into core. https://github.com/khanxmetu/django-admin-keyshortcuts

jacobtylerwalls:
Given the update above from Tim, I think it makes sense to opt for a PR stemming from that package. Thanks for kicking off the discussion here 👍

PR #17610, opened 2023-12-15
Improved DEFAULT_FROM_EMAIL/SERVER_EMAIL docs.

Flimm:
These settings only affect the `From` header set in sent emails. They do not affect other things, such as the SMTP username used to authenticate, and they may be different from the SMTP username. This change clarifies the role of these settings, in order to prevent confusion.

This change also documents how to set a name alongside an email address in the `From` header of these emails.

### DIFF
--- a/docs/ref/settings.txt
+++ b/docs/ref/settings.txt
@@ -1343,9 +1343,12 @@ See :ref:`Filtering error reports<filtering-error-reports>`.
 
 Default: ``'webmaster@localhost'``
 
-Default email address to use for various automated correspondence from the
-site manager(s). This doesn't include error messages sent to :setting:`ADMINS`
-and :setting:`MANAGERS`; for that, see :setting:`SERVER_EMAIL`.
+Default email address for automated correspondence from the site manager(s).
+This address is used in the ``From:`` header of outgoing emails and can take
+any format valid in the chosen email sending protocol.
+
+This doesn't affect error messages sent to :setting:`ADMINS` and
+:setting:`MANAGERS`. See :setting:`SERVER_EMAIL` for that.
 
 .. setting:: DEFAULT_INDEX_TABLESPACE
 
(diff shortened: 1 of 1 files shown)


docs/ref/settings.txt
```diff
@@ -1343,9 +1343,15 @@ See :ref:`Filtering error reports<filtering-error-reports>`.
...
 
-Default email address to use for various automated correspondence from the
-site manager(s). This doesn't include error messages sent to :setting:`ADMINS`
-and :setting:`MANAGERS`; for that, see :setting:`SERVER_EMAIL`.
+Default ``From:`` header value to use for various automated email
+correspondence from the site manager(s). This describes the email address (and
+optionally the name) of the author of the email message, to be displayed in the
+interface of email clients. If you wish to set both a name and an email
+address, use a value like the following:
+``'FirstName LastName <example@example.com>'``
```
nessita, line 1351: This adding feels a bit too verbose. I think it's enough to say that the format allowed for the setting should be valid for the email sending protocol that the service configures. Something like:

```suggestion
Default email address for automated correspondence from the site manager(s).
This address is used in the ``From:`` header of outgoing emails (except for
those reporting error messages sent to :setting:`ADMINS` and
:setting:`MANAGERS`; for that, see :setting:`SERVER_EMAIL`) and can take any
format valid in the chosen email sending protocol.
```

docs/ref/settings.txt
```diff
@@ -2534,8 +2540,12 @@ example, to define a YAML serializer, use::
...
 Default: ``'root@localhost'``
 
-The email address that error messages come from, such as those sent to
-:setting:`ADMINS` and :setting:`MANAGERS`.
+A ``From:`` header value to use for error messages sent over email, such as
+those sent to :setting:`ADMINS` and :setting:`MANAGERS`. This describes the
+email address (and optionally the name) of the author of the email message, to
+be displayed in the interface of email clients. If you wish to set both a name
+and an email address, use a value like the following:
+``'FirstName LastName <example@example.com>'``
```
nessita, line 2548: I would use a more concise phrasing as before:
```suggestion
The email address that error messages come from, such as those sent to
:setting:`ADMINS` and :setting:`MANAGERS`. This address is used in the
``From:`` header and can take any format valid in the chosen email sending
protocol.
```

nessita (commented):
Thank you for your interest in making the docs clearer, I've added some comments.

Flimm:
The suggested changes are fine by me, they are an improvement to the existing documentation, and I like the conciseness. Thank you for reviewing this.

felixxm (commented):
@Flimm Thanks :+1:

PR #17613, opened 2023-12-15
Fixed #29942 -- Used linkcode sphinx extension instead of viewcode. 

smithdc1:
Follow up to #16918. This adds resolving of wildcard imports.

### DIFF
--- a/django/forms/formsets.py
+++ b/django/forms/formsets.py
@@ -1,6 +1,6 @@
 from django.core.exceptions import ValidationError
-from django.forms import Form
 from django.forms.fields import BooleanField, IntegerField
+from django.forms.forms import Form
 from django.forms.renderers import get_default_renderer
 from django.forms.utils import ErrorList, RenderableFormMixin
 from django.forms.widgets import CheckboxInput, HiddenInput, NumberInput
--- a/docs/_ext/github_links.py
+++ b/docs/_ext/github_links.py
@@ -1,0 +1,149 @@
+import ast
+import functools
+import importlib.util
+import pathlib
+
+
+class CodeLocator(ast.NodeVisitor):
+    def __init__(self):
+        super().__init__()
+        self.current_path = []
+        self.node_line_numbers = {}
+        self.import_locations = {}
+
+    @classmethod
+    def from_code(cls, code):
+        tree = ast.parse(code)
+        locator = cls()
+        locator.visit(tree)
+        return locator
+
+    def visit_node(self, node):
+        self.current_path.append(node.name)
+        self.node_line_numbers[".".join(self.current_path)] = node.lineno
+        self.generic_visit(node)
+        self.current_path.pop()
+
+    def visit_FunctionDef(self, node):
+        self.visit_node(node)
+
+    def visit_ClassDef(self, node):
+        self.visit_node(node)
+
+    def visit_ImportFrom(self, node):
+        for alias in node.names:
+            if alias.asname:
+                # Exclude linking aliases (`import x as y`) to avoid confusion
+                # when clicking a source link to a differently named entity.
+                continue
+            if alias.name == "*":
+                # Resolve wildcard imports.
+                file = module_name_to_file_path(node.module)
+                file_contents = file.read_text(encoding="utf-8")
+                locator = CodeLocator.from_code(file_contents)
+                self.import_locations |= locator.import_locations
+                self.import_locations |= {
+                    n: node.module for n in locator.node_line_numbers if "." not in n
+                }
+            else:
+                self.import_locations[alias.name] = ("." * node.level) + (
+                    node.module or ""
+                )
+
+
+@functools.lru_cache(maxsize=1024)
+def get_locator(file):
+    file_contents = file.read_text(encoding="utf-8")
+    return CodeLocator.from_code(file_contents)
+
+
+class CodeNotFound(Exception):
+    pass
+
+
+def module_name_to_file_path(module_name):
+    # Avoid importlib machinery as locating a module involves importing its
+    # parent, which would trigger import side effects.
+
+    for suffix in [".py", "/__init__.py"]:
+        file_path = pathlib.Path(__file__).parents[2] / (
+            module_name.replace(".", "/") + suffix
+        )
+        if file_path.exists():
+            return file_path
+
+    raise CodeNotFound
+
+
+def get_path_and_line(module, fullname):
+    path = module_name_to_file_path(module_name=module)
+
+    locator = get_locator(path)
+
+    lineno = locator.node_line_numbers.get(fullname)
+
+    if lineno is not None:
+        return path, lineno
+
+    imported_object = fullname.split(".", maxsplit=1)[0]
+    try:
+        imported_path = locator.import_locations[imported_object]
+    except KeyError:
+        raise CodeNotFound
+
+    # From a statement such as:
+    # from . import y.z
+    # - either y.z might be an object in the parent module
+    # - or y might be a module, and z be an object in y
+    # also:
+    # - either the current file is x/__init__.py, and z would be in x.y
+    # - or the current file is x/a.py, and z would be in x.a.y
+    if path.name != "__init__.py":
+        # Look in parent module
+        module = module.rsplit(".", maxsplit=1)[0]
+    try:
+        imported_module = importlib.util.resolve_name(
+            name=imported_path, package=module
+        )
+    except ImportError as error:
+        raise ImportError(
+            f"Could not import '{imported_path}' in '{module}'."
+        ) from error
+    try:
+        return get_path_and_line(module=imported_module, fullname=fullname)
+    except CodeNotFound:
+        if "." not in fullname:
+            raise
+
+        first_element, remainder = fullname.rsplit(".", maxsplit=1)
+        # Retrying, assuming the first element of the fullname is a module.
+        return get_path_and_line(
+            module=f"{imported_module}.{first_element}", fullname=remainder
+        )
+
+
+def get_branch(version, next_version):
+    if version == next_version:
+        return "main"
+    else:
+        return f"stable/{version}.x"
+
+
+def github_linkcode_resolve(domain, info, *, version, next_version):
+    if domain != "py":
+        return None
+
+    if not (module := info["module"]):
+        return None
+
+    try:
+        path, lineno = get_path_and_line(module=module, fullname=info["fullname"])
+    except CodeNotFound:
+        return None
+
+    branch = get_branch(version=version, next_version=next_version)
+    relative_path = path.relative_to(pathlib.Path(__file__).parents[2])
+    # Use "/" explicitely to join the path parts since str(file), on Windows,
+    # uses the Windows path separator which is incorrect for URLs.
+    url_path = "/".join(relative_path.parts)
+    return f"https://github.com/django/django/blob/{branch}/{url_path}#L{lineno}"
--- a/tests/sphinx/test_github_links.py
+++ b/tests/sphinx/test_github_links.py
@@ -1,0 +1,211 @@
+import pathlib
+import sys
+
+from django.test import SimpleTestCase
+
+
+def last_n_parts(path, n):
+    return "/".join(path.parts[-n:])
+
+
+# The import must happen at the end of setUpClass, so it can't be imported at
+# the top of the file.
+github_links = None
+
+
+class GitHubLinkTests(SimpleTestCase):
+    @classmethod
+    def setUpClass(cls):
+        # The file implementing the code under test is in the docs folder and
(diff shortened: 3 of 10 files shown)


tests/sphinx/testdata/package/module.py
```diff
@@ -0,0 +1,28 @@
+"""
+Example docstring
+"""
+from django.utils.functional import cached_property
+
+from . import other_module
+from .other_module import MyOtherClass
```
sarahboyce, line 7: Is it worth adding an import * example and test?

docs/_ext/github_links.py
```diff
@@ -0,0 +1,148 @@
...
+
+def get_branch(version, next_version):
+    if version == next_version:
+        return "main"
+    else:
+        return f"stable/{version}.x"
+
+
+def github_linkcode_resolve(domain, info, *, version, next_version):
+    """ """
```
sarahboyce, line 125:
```suggestion
```

docs/_ext/github_links.py
```diff
@@ -0,0 +1,148 @@
...
+        module = module.rsplit(".", maxsplit=1)[0]
+    imported_module = importlib.util.resolve_name(name=imported_path, package=module)
+    try:
+        return get_path_and_line(module=imported_module, fullname=fullname)
+    except CodeNotFound:
+        if "." not in fullname:
+            raise
+
+        first_element, remainder = fullname.rsplit(".", maxsplit=1)
+        # Retrying, assuming the first element of of the fullname is a module
```
sarahboyce, line 111:
```suggestion
        # Retrying, assuming the first element of of the fullname is a module.
```

docs/_ext/github_links.py
```diff
@@ -0,0 +1,148 @@
...
+        self.visit_node(node)
+
+    def visit_ImportFrom(self, node):
+        for alias in node.names:
+            if alias.asname:
+                # Exclude linking aliases (`import x as y`) to avoid confusion when
+                # clicking a source link to a differently named entity.
+                continue
+            if alias.name == "*":
+                # Resolve wildcard imports
```
sarahboyce, line 39:
```suggestion
                # Resolve wildcard imports.
```

sarahboyce (approved):
Thank you for picking this up @smithdc1! 
This looks good to me, I think it has potential to encourage people to get started contributing as the docs become a gateway to the code 🤞 so would like to get this in

I also think this fixes [#29942](https://code.djangoproject.com/ticket/29942)? Do you want to link the PR description to that ticket?

smithdc1:
Yes. I think that's right. I'll update. 

Will this work on the docs build for the website. Does that have the whole repo available at build time? 🤔  I am nervous if there's other impacts of this which are not yet considered.

sarahboyce:
I think the viewcode extension was working by trying to do an import but then that triggered any code ran on import (which caused this to stop working). So in theory, I don't think this is dissimilar  🤔 but I don't know

tests/sphinx/testdata/package/module.py
```diff
@@ -0,0 +1,30 @@
...
+"""
+
+from django.utils.functional import cached_property
+from tests.sphinx.testdata.package.wildcard_module import *  # noqa
+
+from . import other_module
+from .other_module import MyOtherClass
+
+# This file should never get imported. If it is, then something failed already.
+raise Exception
```
nessita, line 12: Did you mean to leave this here? I'm a little lost about the reason to do so.

tests/sphinx/test_github_links.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+class GitHubLinkTests(SimpleTestCase):
+    @classmethod
+    def setUpClass(cls):
+        # The file implementing the code under test cannot be imported through
+        # standard means, so include its parent in the pythonpath for the
+        # duration of the tests.
+        cls.ext_path = str((pathlib.Path(__file__).parents[2] / "docs/_ext").resolve())
+        sys.path.insert(0, cls.ext_path)
```
nessita, line 14: Should we use a couple of `addClassCleanup` here instead of defining `tearDownClass`?

docs/_ext/github_links.py
```diff
@@ -0,0 +1,147 @@
...
+        return "main"
+    else:
+        return f"stable/{version}.x"
+
+
+def github_linkcode_resolve(domain, info, *, version, next_version):
+    if domain != "py":
+        return None
+
+    if not info.get("module", None):
```
nessita, line 128: Could we use the walrus operator here to store `module`?

docs/_ext/github_links.py
```diff
@@ -0,0 +1,147 @@
...
+        return None
+
+    module = info["module"]
+    fullname = info["fullname"]
```
nessita, line 132: Could `fullname` not be defined in some cases?

nessita (commented):
Looks great! I left some comments in order to understand a this a bit better.

Thank you!

docs/_ext/github_links.py
smithdc1: Good question. Actually both `fullname` and `module` are guaranteed to be present, see [docs](https://www.sphinx-doc.org/en/master/usage/extensions/linkcode.html#confval-linkcode_resolve). I'll update and fix our test.

tests/sphinx/testdata/package/module.py
smithdc1: Importing modules is a limitation of `viewcode` and would require a full django setup to work.  This was discussed but it is not an [acceptable solution](https://github.com/django/django/pull/12135#issuecomment-560362400).

The aim of this line is to test that we don't import the module and face the same limitations. 

Do you think there's a way we can do that without just raising an exception here? 

Maybe it is not even needed?

tests/sphinx/test_github_links.py
```diff
@@ -0,0 +1,188 @@
...
+        pass"""
+        )
+
+        self.assertEqual(locator.node_line_numbers, {"h": 5, "I": 8, "I.j": 9})
+        self.assertEqual(locator.import_locations, {"b": "a", "c": "a", "e": ".d"})
+
+    def test_module_name_to_file_path_package(self):
+        path = self.github_links.module_name_to_file_path("django")
+
+        self.assertEqual("/".join(path.parts[-2:]), "django/__init__.py")
```
nessita, line 47: Since this pattern is heavily used across many tests, what do you think if we add a helper `last_n_parts` or similar so:
```python
def last_n_parts(path, n):
    return "/".join(path.parts[-n:])
```

And then:
```python
        self.assertEqual(last_n_parts(path, 2), "django/__init__.py")
```

tests/sphinx/testdata/package/module.py
nessita: Right, I think that we should remove this, since the `raise` in the `__init__.py` file would always get in the way of importing any of the modules anyways. Do you agree?

django/forms/formsets.py
```diff
@@ -1,6 +1,6 @@
 from django.core.exceptions import ValidationError
-from django.forms import Form
 from django.forms.fields import BooleanField, IntegerField
+from django.forms.forms import Form
```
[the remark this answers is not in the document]
smithdc1: Without this change when I build the docs (on windows? 🤔 ) I get the following error message.

```
reading sources... [ 27%] ref/forms/api
Extension error (sphinx.ext.linkcode):
Handler <function doctree_read at 0x000001ED271A28E0> for event 'doctree-read' threw an exception (exception: maximum recursion depth exceeded while calling a Python object)
```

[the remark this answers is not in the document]
nessita: Thanks for the details!

tests/sphinx/testdata/package/module.py
smithdc1: Sure, I can remove it 👍

docs/_ext/github_links.py
```diff
@@ -0,0 +1,144 @@
...
+    # from . import y.z
+    # - either y.z might be an object in the parent module
+    # - or y might be a module, and z be an object in y
+    # also:
+    # - either the current file is x/__init__.py, and z would be in x.y
+    # - or the current file is x/a.py, and z would be in x.a.y
+    if path.name != "__init__.py":
+        # Look in parent module
+        module = module.rsplit(".", maxsplit=1)[0]
+    imported_module = importlib.util.resolve_name(name=imported_path, package=module)
```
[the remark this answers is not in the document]
smithdc1: What are you thinking should happen once an `ImportError` is caught? It seems to me that we would want to see these errors. An import from outside of Django's package seems surprising to me and doesn't sound like it would be desired.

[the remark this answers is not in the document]
nessita: Well I was hoping we would handle it a bit more gracefully, right now the error looks like this:

```
Extension error (sphinx.ext.linkcode):
Handler <function doctree_read at 0x7fecd096b9c0> for event 'doctree-read' threw an exception (exception: no package specified for '.config' (required for relative module names))
make: *** [Makefile:59: html] Error 2
```

and I think that may bee too cryptic for someone unfamiliar with the extension to understand what is going one. If we would print a clear and informative message about where the error is being raised from and under which state/variables values, I think that would help future contributors.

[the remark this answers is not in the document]
smithdc1: What import did you use to see this error message? I'm struggling to create that scenario.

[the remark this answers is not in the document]
nessita: What I did was changing the code to force an import error, following the [docs](https://docs.python.org/3/library/importlib.html#importlib.util.resolve_name) for `resolve_name`, specifically I forced package to be None:
```diff
diff --git a/docs/_ext/github_links.py b/docs/_ext/github_links.py
index 3cc75cd30f..b704b08532 100644
--- a/docs/_ext/github_links.py
+++ b/docs/_ext/github_links.py
@@ -103,7 +103,7 @@ def get_path_and_line(module, fullname):
         module = module.rsplit(".", maxsplit=1)[0]
     try:
         imported_module = importlib.util.resolve_name(
-            name=imported_path, package=module
+            name=imported_path, package=None  # module
         )
     except ImportError as error:
         raise ImportError(
```

With your change, this is what I see when building the docs:
```
Extension error (sphinx.ext.linkcode):
Handler <function doctree_read at 0x7fc71886b9c0> for event 'doctree-read' threw an exception (exception: Could not import '.config' in 'django.apps'.)
make: *** [Makefile:59: html] Error 2
```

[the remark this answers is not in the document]
smithdc1: Thanks for the info!

I'm not sure I can write a test for this as this is invalid syntax which causes `ast.parse()` to crash. 🤔 

```python 
import .config
```


```
  File "C:\Users\smith\projects\django\docs\_ext\github_links.py", line 16, in from_code
    tree = ast.parse(code)
           ^^^^^^^^^^^^^^^
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.11_3.11.2544.0_x64__qbz5n2kfra8p0\Lib\ast.py", line 50, in parse
    return compile(source, filename, mode, flags,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<unknown>", line 2
    import .TestImportError  # noqa
           ^
SyntaxError: invalid syntax
```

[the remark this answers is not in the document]
nessita: I'm a little lost about what you mean with invalid syntax. The line:

```python
imported_module = importlib.util.resolve_name(name=imported_path, package=None)
```

is valid syntax, in fact it can be executed in a Python interpreter:

```
In [5]: import importlib.util

In [6]: imported_path="."

In [7]: imported_module = importlib.util.resolve_name(name=imported_path, package=None)
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
Cell In[7], line 1
----> 1 imported_module = importlib.util.resolve_name(name=imported_path, package=None)

File <frozen importlib.util>:32, in resolve_name(name, package)

ImportError: no package specified for '.' (required for relative module names)
```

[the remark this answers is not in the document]
smithdc1: I added it in one of the test Python files, as that's where we'll be reading the name and package data from. 

Editing the `import_error.py` file to the following results in the crash similar to that shown above. 

```python
from .....test import Test  # noqa
import .config # This line is invalid syntax.

class MyClass:
    pass
```

I'm therefore unsure how to add a test for `package` being `None`.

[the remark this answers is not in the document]
nessita: Thank you, I think this suffices as a test!

nessita (approved):
Thank you so much @smithdc1 and @ewjoachim for this work! I'll push minimal edits (comments should be wrapped at 79 cols) and I'll merge after a green CI run.

PR #17614, opened 2023-12-15
Improved cache.tests.FileBasedCacheTests.test_touch to avoid flakiness due to slow file system access.

nessita:
I've noticed a recurrent failure in Jenkins' runs where the `test_touch` for the filesystem cache backend fails (intermittently but often enough that it bothers me). My understanding is that this is caused by the cache expiration happening faster than the test execution, specifically the code:

```python
        self.assertIs(cache.touch("expire1", timeout=4), True)
        time.sleep(2)
        self.assertIs(cache.has_key("expire1"), True)
```

is flaky in that `cache.has_key` returns False if the expiration happened faster than the 2 second sleeping plus the file opening, de-pickling and checking for its contents.

This proposal mocks `time` for this particular test and backend to manually control time and thus ensuring deterministic runs.

### DIFF
--- a/tests/cache/tests.py
+++ b/tests/cache/tests.py
@@ -1766,6 +1766,28 @@ class FileBasedCacheTests(BaseCacheTests, TestCase):
             self.assertIs(cache.has_key("key"), False)
             mocked_open.assert_called_once()
 
+    def test_touch(self):
+        """Override to manually advance time since file access can be slow."""
+
+        class ManualTickingTime:
+            def __init__(self):
+                # Freeze time, calling `sleep` will manually advance it.
+                self._time = time.time()
+
+            def time(self):
+                return self._time
+
+            def sleep(self, seconds):
+                self._time += seconds
+
+        mocked_time = ManualTickingTime()
+        with (
+            mock.patch("django.core.cache.backends.filebased.time", new=mocked_time),
+            mock.patch("django.core.cache.backends.base.time", new=mocked_time),
+            mock.patch("cache.tests.time", new=mocked_time),
+        ):
+            super().test_touch()
+
 
 @unittest.skipUnless(RedisCache_params, "Redis backend not configured")
 @override_settings(


sarahboyce:
This looks plausible. Do you have an example of a fewof the failing test runs for reference?
I'm also wondering if this might be something around test testcase classes being ran in parrallel and sharing the cache or some kinda race condition (as the line before has a smaller timeout, maybe it didn't update).

nessita:
> This looks plausible. Do you have an example of a few of the failing test runs for reference? I'm also wondering if this might be something around test testcase classes being ran in parrallel and sharing the cache or some kinda race condition (as the line before has a smaller timeout, maybe it didn't update).

Thank you @sarahboyce! I haven't stored examples of previous test failures, so I'll monitor the runs and try to gather recent examples.

Regarding whether this is an issue because of the run in parallel, I'm inclined to think is not, because I can describe a scenario where this test, using the actual time, is flaky even when run in isolation. Considering the test is as follows:

```python

    492     def test_touch(self):                                                      
    493         # cache.touch() updates the timeout.                                   
    494         cache.set("expire1", "very quickly", timeout=1)                        
    495         self.assertIs(cache.touch("expire1", timeout=4), True)                 
    496         time.sleep(2)                                                          
    497         self.assertIs(cache.has_key("expire1"), True)                          
    498         time.sleep(3)                                                          
    499         self.assertIs(cache.has_key("expire1"), False)                         
    500         # cache.touch() works without the timeout argument.                    
    501         cache.set("expire1", "very quickly", timeout=1)                        
    502         self.assertIs(cache.touch("expire1"), True)                            
    503         time.sleep(2)                                                          
    504         self.assertIs(cache.has_key("expire1"), True)                          
    505                                                                                
    506         self.assertIs(cache.touch("nonexistent"), False)                       
```
and considering that the `FileBasedCacheTests` uses the file-based cache which does file disk reading and writing, if the completion of the `cache.touch` call plus the actual sleeping for 2 seconds plus the checking of the key `"expire1"` takes more than 4 seconds, `cache.has_key("expire1")` will return False. This can be forced when using the `ManualTickingTime` and having the `sleep` implementation taking longer than the seconds provided, simulating a slow file writing in `touch` and a slow file reading in `has_key`.

I hope the above makes sense!

sarahboyce (approved):
Makes sense 👍 thank you

PR #17619, opened 2023-12-16
Add missing word in documentation.

David-Wobrock:


docs/ref/models/querysets.txt
```diff
@@ -2641,7 +2641,7 @@ limit should be used. (In particular, when prefetching across two or more
 relations, a ``chunk_size`` should be small enough that the anticipated number
 of results for each prefetched relation still falls below the limit.)
 
-So long as the QuerySet does not prefetch any related objects, providing no
+So as long as the QuerySet does not prefetch any related objects, providing no
```
ngnpope, line 2644: I'm not actually sure the change is correct. Adding in the "as" make it sound like it is missing a comma and should be "So, as long as…". But this implies some continuation of thought from the previous paragraph.

I think it is correct as it is, but it is [informal](https://dictionary.cambridge.org/grammar/british-grammar/as-long-as-and-so-long-as). Maybe you could go for "As long as…" if you think it reads better.

David-Wobrock: Thanks!

TIL: _So long as_ is actually correct to use.
I thought we had a missing word and meants _So, as long as_... 

Closing the PR :)

David-Wobrock:
Closing as per https://github.com/django/django/pull/17619#discussion_r1428751457

PR #17621, opened 2023-12-18
Fixed #35030 -- Made django.contrib.auth decorators to work with async functions.

HappyDingning:
Hi, dear reviewers,

Follow up to [ticket-35030](https://code.djangoproject.com/ticket/35030), I have achieved the `alogin_required `, the `apermission_required` and the corresponding test cases. I haven't write the documentation yet. Could you please review the code first? I will write the documentation soon.

Thank you.

### DIFF
--- a/django/contrib/auth/decorators.py
+++ b/django/contrib/auth/decorators.py
@@ -1,5 +1,8 @@
+import asyncio
 from functools import wraps
 from urllib.parse import urlparse
+
+from asgiref.sync import async_to_sync, sync_to_async
 
 from django.conf import settings
 from django.contrib.auth import REDIRECT_FIELD_NAME
@@ -17,10 +20,7 @@ def user_passes_test(
     """
 
     def decorator(view_func):
-        @wraps(view_func)
-        def _wrapper_view(request, *args, **kwargs):
-            if test_func(request.user):
-                return view_func(request, *args, **kwargs)
+        def _redirect_to_login(request):
             path = request.build_absolute_uri()
             resolved_login_url = resolve_url(login_url or settings.LOGIN_URL)
             # If the login url is the same scheme and net location then just
@@ -35,7 +35,32 @@ def user_passes_test(
 
             return redirect_to_login(path, resolved_login_url, redirect_field_name)
 
-        return _wrapper_view
+        if asyncio.iscoroutinefunction(view_func):
+
+            async def _view_wrapper(request, *args, **kwargs):
+                auser = await request.auser()
+                if asyncio.iscoroutinefunction(test_func):
+                    test_pass = await test_func(auser)
+                else:
+                    test_pass = await sync_to_async(test_func)(auser)
+
+                if test_pass:
+                    return await view_func(request, *args, **kwargs)
+                return _redirect_to_login(request)
+
+        else:
+
+            def _view_wrapper(request, *args, **kwargs):
+                if asyncio.iscoroutinefunction(test_func):
+                    test_pass = async_to_sync(test_func)(request.user)
+                else:
+                    test_pass = test_func(request.user)
+
+                if test_pass:
+                    return view_func(request, *args, **kwargs)
+                return _redirect_to_login(request)
+
+        return wraps(view_func)(_view_wrapper)
 
     return decorator
 
@@ -64,19 +89,36 @@ def permission_required(perm, login_url=None, raise_exception=False):
     If the raise_exception parameter is given the PermissionDenied exception
     is raised.
     """
+    if isinstance(perm, str):
+        perms = (perm,)
+    else:
+        perms = perm
 
-    def check_perms(user):
-        if isinstance(perm, str):
-            perms = (perm,)
+    def decorator(view_func):
+        if asyncio.iscoroutinefunction(view_func):
+
+            async def check_perms(user):
+                # First check if the user has the permission (even anon users).
+                if await sync_to_async(user.has_perms)(perms):
+                    return True
+                # In case the 403 handler should be called raise the exception.
+                if raise_exception:
+                    raise PermissionDenied
+                # As the last resort, show the login form.
+                return False
+
         else:
-            perms = perm
-        # First check if the user has the permission (even anon users)
-        if user.has_perms(perms):
-            return True
-        # In case the 403 handler should be called raise the exception
-        if raise_exception:
-            raise PermissionDenied
-        # As the last resort, show the login form
-        return False
 
-    return user_passes_test(check_perms, login_url=login_url)
+            def check_perms(user):
+                # First check if the user has the permission (even anon users).
+                if user.has_perms(perms):
+                    return True
+                # In case the 403 handler should be called raise the exception.
+                if raise_exception:
+                    raise PermissionDenied
+                # As the last resort, show the login form.
+                return False
+
+        return user_passes_test(check_perms, login_url=login_url)(view_func)
+
+    return decorator
--- a/docs/topics/auth/default.txt
+++ b/docs/topics/auth/default.txt
@@ -617,6 +617,10 @@ The ``login_required`` decorator
     :func:`django.contrib.admin.views.decorators.staff_member_required`
     decorator a useful alternative to ``login_required()``.
 
+.. versionchanged:: 5.1
+
+    Support for wrapping asynchronous view functions was added.
+
 .. currentmodule:: django.contrib.auth.mixins
 
 The ``LoginRequiredMixin`` mixin
@@ -713,6 +717,11 @@ email in the desired domain and if not, redirects to the login page::
 
         @user_passes_test(email_check, login_url="/login/")
         def my_view(request): ...
+
+    .. versionchanged:: 5.1
+
+        Support for wrapping asynchronous view functions and using asynchronous
+        test callables was added.
 
 .. currentmodule:: django.contrib.auth.mixins
 
@@ -817,6 +826,10 @@ The ``permission_required`` decorator
     This also avoids a redirect loop when :class:`.LoginView`'s
     ``redirect_authenticated_user=True`` and the logged-in user doesn't have
     all of the required permissions.
+
+.. versionchanged:: 5.1
+
+    Support for wrapping asynchronous view functions was added.
 
 .. currentmodule:: django.contrib.auth.mixins
 
--- a/docs/releases/5.1.txt
+++ b/docs/releases/5.1.txt
@@ -52,6 +52,11 @@ Minor features
   form save. This is now available in the admin when visiting the user creation
   and password change pages.
 
+* :func:`~.django.contrib.auth.decorators.login_required`,
+  :func:`~.django.contrib.auth.decorators.permission_required`, and
+  :func:`~.django.contrib.auth.decorators.user_passes_test` decorators now
+  support wrapping asynchronous view functions.
+
 :mod:`django.contrib.contenttypes`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/tests/auth_tests/test_decorators.py
+++ b/tests/auth_tests/test_decorators.py
@@ -1,3 +1,7 @@
+from asyncio import iscoroutinefunction
+
+from asgiref.sync import sync_to_async
+
 from django.conf import settings
 from django.contrib.auth import models
 from django.contrib.auth.decorators import (
@@ -18,6 +22,22 @@ class LoginRequiredTestCase(AuthViewsTestCase):
     """
     Tests the login_required decorators
     """
+
+    factory = RequestFactory()
+
+    def test_wrapped_sync_function_is_not_coroutine_function(self):
+        def sync_view(request):
+            return HttpResponse()
+
+        wrapped_view = login_required(sync_view)
+        self.assertIs(iscoroutinefunction(wrapped_view), False)
+
+    def test_wrapped_async_function_is_coroutine_function(self):
+        async def async_view(request):
+            return HttpResponse()
+
+        wrapped_view = login_required(async_view)
+        self.assertIs(iscoroutinefunction(wrapped_view), True)
 
     def test_callable(self):
         """
@@ -63,6 +83,35 @@ class LoginRequiredTestCase(AuthViewsTestCase):
             view_url="/login_required_login_url/", login_url="/somewhere/"
         )
 
+    async def test_login_required_async_view(self, login_url=None):
+        async def async_view(request):
+            return HttpResponse()
+
+        async def auser_anonymous():
+            return models.AnonymousUser()
+
+        async def auser():
+            return self.u1
+
+        if login_url is None:
+            async_view = login_required(async_view)
+            login_url = settings.LOGIN_URL
+        else:
+            async_view = login_required(async_view, login_url=login_url)
+
+        request = self.factory.get("/rand")
+        request.auser = auser_anonymous
+        response = await async_view(request)
+        self.assertEqual(response.status_code, 302)
+        self.assertIn(login_url, response.url)
+
+        request.auser = auser
+        response = await async_view(request)
+        self.assertEqual(response.status_code, 200)
+
+    async def test_login_required_next_url_async_view(self):
+        await self.test_login_required_async_view(login_url="/somewhere/")
+
 
 class PermissionsRequiredDecoratorTest(TestCase):
     """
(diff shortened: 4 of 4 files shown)


django/contrib/auth/decorators.py
```diff
@@ -21,19 +40,29 @@ def decorator(view_func):
...
-
-            return redirect_to_login(path, resolved_login_url, redirect_field_name)
+            return _redirect_to_login(request, login_url, redirect_field_name)
+
+        return _wrapper_view
+
+    return decorator
+
+
+def auser_passes_test(
```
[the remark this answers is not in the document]
HappyDingning: Hi, thank you for your comment. Here is my view.

I think what we need is an asynchronous version instead of adapting current functions to support coroutine functions. Here's why: 

1. In Django, IO-related functions usually have distinct coroutine versions, like `aget` and `acreate`. 

2. Having a dedicated async decorator allows for more controlled, simpler, and user-friendly behavior. If we modify existing sync function decorators to handle coroutine functions, it raises a question: should they return a coroutine or a sync function when decorating a sync function? My view is that they should return a sync function in a WSGI environment and a coroutine function in an ASGI environment. This approach, however, complicates the decorator's logic and makes it less intuitive. On the other hand, using a new coroutine function decorator keeps things straightforward. For instance, a sync function decorator would always return a sync function, and a coroutine function decorator would always return a coroutine function.

Do you think these are good reasons? Welcome to discuss.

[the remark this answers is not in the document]
felixxm: There is a huge difference, we don't have async decorators, we make them to handle async functions passed as parameters. So creating a separate functions with the `a` prefix is confusing and inconsistent with other decorators.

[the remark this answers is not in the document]
HappyDingning: OK, I understand what you mean.  Is there any similar case in Django?  Could you please give me an example for reference?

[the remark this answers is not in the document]
felixxm: You can check ticket-31949 and all changed decorators.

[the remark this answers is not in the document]
HappyDingning: > You can check [ticket-31949](https://code.djangoproject.com/ticket/31949) and all changed decorators.

Okay, I'll take a look at this ticket, thank you.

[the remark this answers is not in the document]
dingning-aspecta: Hi, I have read ticket-31949, I have understand what you mean. But I still have a question. For example, there is a very simple sync view, it returns HttpResponse("OK") directly. But now I decide to decorate it by using the `login_required` decorator. According to the ticket-31949, if the view is a function, the decorator will return a function too. In the end, we got a function that includes database I/O, which will run in the async context. I think it's really strange.

```
# Assuming this is an async backend.
@login_required
def foo(request):
    return HttpResponse(”OK“)
```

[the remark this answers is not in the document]
HappyDingning: Hi, I have read ticket-31949, I have understand what you mean. But I still have a question, becauce I noticed that most of the decorators referenced in the ticket-31949 do not contain any I/O, but the `login_required` contains I/O. So I think there are some different. For example, there is a very simple sync view, it returns `HttpResponse("OK")` directly. But now I decide to decorate it by using the `login_required` decorator. According to the ticket-31949, if the view is a function, the decorator will return a function too. In the end, we got a function that includes synchronous database I/O, which will run in the async context. I think it's really strange.

```
# Assuming this is an async backend.
# The decorated foo is a function that will query the database synchronously.

@login_required
def foo(request):
    return HttpResponse(”OK“)
```

Although Django has technology that allows synchronous view functions to run in async context, I still feel that this is not as good as native coroutine functions.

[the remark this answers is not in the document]
felixxm: I'm not sure what are you trying to say, TBH :thinking: Modified `@login_required` should recognized if decorated function is coroutine or not, and take different paths. This not much different from what you proposed, the main difference is that we keep one decorator that can handle both.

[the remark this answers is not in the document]
HappyDingning: Hi, could you please view my simple example that explains my view? I have created a repo here https://github.com/HappyDingning/dango_playground/tree/main

I implemented a demo `login_required`, that can decorate the view function according to its type(coroutine or not). And I also implemented a demo `alogin_required`, that can fix the issue of `login_required`.
Please read the comments in the [views.py ](https://github.com/HappyDingning/dango_playground/blob/a181bcd1e7f7798910e1e1ccac873acf1151727a/playground/views.py#L9)and [decorators.py(line 62~63)](https://github.com/HappyDingning/dango_playground/blob/a181bcd1e7f7798910e1e1ccac873acf1151727a/playground/decorators.py#L62). Thank you~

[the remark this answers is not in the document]
felixxm: I really don't understand what you're trying to achieve you've created async versions of decorators and trying to decorated sync function views with them :exploding_head: I'm puzzled.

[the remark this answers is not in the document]
felixxm: This ticket is about modifying `login_required` and `permission_required` and `user_passes_test` that already handle async functions, to be able to handle coroutines, that's it.

> Please read the comments in the [views.py ](https://github.com/HappyDingning/dango_playground/blob/89f8bd4b1bf1b61022174c015c4caedb75369bcb/playground/views.py#L9)and [decorators.py(line 62~63)](https://github.com/HappyDingning/dango_playground/blob/a181bcd1e7f7798910e1e1ccac873acf1151727a/playground/decorators.py#L62). Thank you~

I think you're looking on this wrong the wrong perspective `test_func` is not important, `view_func` is decorated and we want to handle when it's a coroutine:

```python
def user_passes_test(
    test_func, login_url=None, redirect_field_name=REDIRECT_FIELD_NAME
):
    """
    Decorator for views that checks that the user passes the given test,
    redirecting to the log-in page if necessary. The test should be a callable
    that takes the user object and returns True if the user passes.
    """

    def decorator(view_func):
        if iscoroutinefunction(view_func):
            
            async def _view_wrapper(request, *args, **kwargs):
                ...
             
        else:
            
            def _view_wrapper(request, *args, **kwargs):
                ...
                
        return wraps(view_func)(_view_wrapper)
        
    return decorator

```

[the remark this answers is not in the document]
HappyDingning: I'm really sorry for confusing you. I created a post. You are welcome to leave a message. https://forum.djangoproject.com/t/about-async-version-of-the-login-required-decorator/27077

django/contrib/auth/decorators.py
```diff
@@ -47,6 +68,7 @@ def login_required(
     Decorator for views that checks that the user is logged in, redirecting
     to the log-in page if necessary.
     """
+
```
felixxm, line 71: Remove unnecessary blank line.

docs/topics/auth/default.txt
```diff
@@ -620,6 +620,10 @@ The ``login_required`` decorator
     :func:`django.contrib.admin.views.decorators.staff_member_required`
     decorator a useful alternative to ``login_required()``.
 
+.. versionchanged:: 5.1
+
+    Added support for async views.
```
felixxm, line 625:
```suggestion
    .. versionchanged:: 5.1

        Added support for async views.
```

tests/async/test_async_auth_decorators.py
```diff
@@ -0,0 +1,145 @@
...
+from django.core.exceptions import PermissionDenied
+from django.http import HttpResponse
+from django.test import TestCase
+from django.test.client import RequestFactory
+
+
+class AsyncLoginRequiredTestCase(TestCase):
+    """
+    Tests the login_required decorators with async views.
+    """
```
felixxm, line 15: This docstring is unnecessary.

tests/async/test_async_auth_decorators.py
```diff
@@ -0,0 +1,145 @@
...
+        decorator with a login_url set.
+        """
+
+        await self.test_login_required(login_url="/somewhere/")
+
+
+class AsyncPermissionsRequiredDecoratorTest(TestCase):
+    """
+    Tests for the permission_required decorator with async views.
+    """
```
felixxm, line 63: This docstring is unnecessary, please remove it.

tests/async/test_async_auth_decorators.py
```diff
@@ -0,0 +1,145 @@
+from functools import partial
```
felixxm, line 1: Please move these tests to the `tests/decorators` as we did for other decorators.

docs/topics/auth/default.txt
```diff
@@ -719,6 +723,10 @@ email in the desired domain and if not, redirects to the login page::
         def my_view(request):
             ...
 
+    .. versionchanged:: 5.1
+
+        Added support for async ``test_func`` and async views.
```
felixxm, line 728: Does it mean that we can use async `test_func` and with sync views?

tests/async/test_async_auth_decorators.py
felixxm: Please also add two "standard" tests that we've added for all decorators with async views support: `test_wrapped_sync_function_is_not_coroutine_function`, `test_wrapped_async_function_is_coroutine_function` (e.g. 3152f9de47a317d658ebee6d5b517b7e1258aa1f).

tests/async/test_async_auth_decorators.py
```diff
@@ -0,0 +1,145 @@
...
+class AsyncLoginRequiredTestCase(TestCase):
+    """
+    Tests the login_required decorators with async views.
+    """
+
+    factory = RequestFactory()
+
+    @staticmethod
+    async def auser(user):
+        return user
```
felixxm, line 21: Can we change it to the class method and always return `cls.user`?

django/contrib/auth/decorators.py
```diff
@@ -35,6 +35,27 @@ def _wrapper_view(request, *args, **kwargs):
...
             return redirect_to_login(path, resolved_login_url, redirect_field_name)
 
+        if asyncio.iscoroutinefunction(view_func):
+
+            @wraps(view_func)
+            async def _wrapper_view(request, *args, **kwargs):
+                if asyncio.iscoroutinefunction(test_func):
+                    result = await test_func(await request.auser())
+                else:
+                    result = test_func(await request.auser())
```
felixxm, line 45: Currently it's tested only with a `lambda`, it will not work when test cannot be called from async context, e.g.
```
        def my_test(user):
            return bool(Group.objects.filter(name=user.username).exists())
```

felixxm (commented):
@HappyDingning Thanks :+1: I left more comments.

@bigfootjon Do you want to take a look?

django/contrib/auth/decorators.py
HappyDingning: Yes, you are right. it will raise an exception when `test_func` cannot be called from async context. But I think this is an expected behavior. That means, if developer provide a sync `test_func`, then it must can be called from async context. How do you think about?

tests/async/test_async_auth_decorators.py
HappyDingning: I think this could be a little difficult, because `request.auser` doesn't always return `cls.user`, it sometimes returns an `AnonymousUser`, such as https://github.com/django/django/pull/17621/commits/c73ee01b7af97106d6bd7d40f9a036f395f979c3#diff-c14ba26ed5b9575e08fdb8baf770ff7cd0d8d999db0c325345b317615e9d9784R101

But the same method in the `PermissionsRequiredDecoratorTest` can be a classmethod. How do you think about?

felixxm:
@bigfootjon Do you wan to take a look?

django/contrib/auth/decorators.py
bigfootjon: I agree with @felixxm, I think we should allow the user to pass a sync-only `test_func` to an `async` function using this decorator.

This makes the internal implementation more complicated, but it does allow users to have more flexibility in their implementations.

I think this is particularly important for a project that is "migrating" between sync and async, or using a mix of both contexts. We should not require them to implement 2 different test functions that do the same thing.

django/contrib/auth/decorators.py
```diff
@@ -35,6 +35,27 @@ def _wrapper_view(request, *args, **kwargs):
...
+                    return await view_func(request, *args, **kwargs)
+                return _redirect_to_login(request)
+
+        else:
+
+            @wraps(view_func)
+            def _wrapper_view(request, *args, **kwargs):
+                if test_func(request.user):
+                    return view_func(request, *args, **kwargs)
+                return _redirect_to_login(request)
```
bigfootjon, line 57: Same here, I think we should allow an `async` `test_func` that wraps a sync function.

bigfootjon (commented):
Looks on the right track! I think we should cover all sync/async combinations, but otherwise this looks good to me!

django/contrib/auth/decorators.py
HappyDingning: Okay, thank you for your advice. I completely understand and make sense. I'll update the code soon.

django/contrib/auth/decorators.py
```diff
@@ -35,6 +35,35 @@ def _wrapper_view(request, *args, **kwargs):
...
+        if asyncio.iscoroutinefunction(view_func):
+
+            @wraps(view_func)
+            async def _wrapper_view(request, *args, **kwargs):
+                if asyncio.iscoroutinefunction(test_func):
+                    result = await test_func(await request.auser())
+                else:
+                    try:
+                        result = test_func(await request.auser())
+                    except SynchronousOnlyOperation:
```
bigfootjon, line 47: Why do we need to catch `SynchronousOnlyOperation`? This feels wrong to me

I think the better implementation is to just use the code from the `except` block instead of calling it from an async context first. That method makes this work the same as the sync `_wrapper_view` below and would be more like what users would expect to happen.

bigfootjon:
(@HappyDingning you might want to refer to this section of the contributing guide to make sure you automatically with Django's code formatting rules: https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#pre-commit-checks. Or otherwise read that document page so you know what the CI jobs expect)

HappyDingning:
> (@HappyDingning you might want to refer to this section of the contributing guide to make sure you automatically comply with Django's code formatting rules: https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#pre-commit-checks. Or otherwise read that document page so you know what the CI jobs expect)

I'm so sorry, I'm using vscode in chrome, because I'm outside and inconvenient to use my own computer. I will rebase my commit later.

django/contrib/auth/decorators.py
bigfootjon: That may be true, there is a perf cost to making it work that way but it does make the implementation more consistent. My concern is that blindly calling a function twice is going to lead to more problems (and worse perf) than deterministic functionality.

I think we can safely trade perf for consistency here, as the perf notes are well documented: https://docs.djangoproject.com/en/5.0/topics/async/#performance

One option might be raising a warning when we encounter "mixed decoration" states (wrapper is async and view is sync, or the other way around). What do you think about that? AIUI we have a warning for when Middleware must be "adapted" to different calling environments so there is precedent for this.

bigfootjon:
> I'm so sorry, I'm using vscode in chrome, because I'm outside and inconvenient to use my own computer. I will rebase my commit later.

No problem on my end! Just looked like you were having a bad time fighting the code-style rules and thought that could help.

django/contrib/auth/decorators.py
HappyDingning: Oh, you are absolutely right, consistent is more important. I'll update the code later, thank you.

HappyDingning: I thought to raise a warning, when encounter "mixed decoration" states. But sometimes, the test_func doesn't contain any IO indeed, so it's enough to be sync (for example, the test_func within login_required).I thought to raise a warning, when encounter "mixed decoration" states. But sometimes, the test_func doesn't contain any IO indeed, so it's enough to be sync (for example, the test_func within login_required).

dingning-aspecta:
@felixxm Hello, I think we are ready for the reviewing😊.

django/contrib/auth/decorators.py
bigfootjon: > I thought to raise a warning when encounter "mixed decoration" states. But sometimes, the test_func doesn't contain any IO indeed, so sync is enough

Great point, I like the current implementation. Thanks for working through alternatives!

bigfootjon (approved):
LGTM

felixxm:
@HappyDingning Thanks :+1: I'll start working on this patch after merging some extra tests, #17943.

felixxm:
Please don't push changes to this branch. I'll work on it tomorrow.

felixxm (approved):

felixxm:
Test failure is not related with this PR.

PR #17628, opened 2023-12-20
Fixed #35048 -- Dropped support for PROJ 5.x.

smithdc1:
[Ticket #35048](https://code.djangoproject.com/ticket/35048)


felixxm (approved):
@smithdc1 Thanks :+1: 

I found one small note to remove in docs.

PR #17642, opened 2023-12-24
Fixed #35033 -- Avoid repeated header in email message

salvo-polizzi:
Fixed bug of ticket [35033](https://code.djangoproject.com/ticket/35033)

The solution that I thought is to raise a ValueError exception, as suggested in the ticket conversation, when it happens that the same header parameter is passed 2 times

### DIFF
--- a/django/core/mail/message.py
+++ b/django/core/mail/message.py
@@ -251,7 +251,6 @@ class EmailMessage:
         self.headers_with_one_occurrence = {
             "To": self.to,
             "Cc": self.cc,
-            "Bcc": self.bcc,
             "Reply-To": self.reply_to,
         }
         self.connection = connection
--- a/django/core/mail/message.py
+++ b/django/core/mail/message.py
@@ -248,6 +248,12 @@ class EmailMessage:
                 else:
                     self.attach(*attachment)
         self.extra_headers = headers or {}
+        self.headers_with_one_occurrence = {
+            "To": self.to,
+            "Cc": self.cc,
+            "Bcc": self.bcc,
+            "Reply-To": self.reply_to,
+        }
         self.connection = connection
 
     def get_connection(self, fail_silently=False):
@@ -286,6 +292,10 @@ class EmailMessage:
                 raise ValueError("Cannot insert repeated headers")
             if name.lower() != "from":  # From is already handled
                 msg[name] = value
+        for header, value_header in self.headers_with_one_occurrence.items():
+            # extra_headers takes precedence over to/cc/bcc/reply_to parameters.
+            if msg.get(header) is None and value_header:
+                msg[header] = ", ".join(str(v) for v in value_header)
         return msg
 
     def recipients(self):
--- a/django/core/mail/message.py
+++ b/django/core/mail/message.py
@@ -258,6 +258,8 @@ class EmailMessage:
         return self.connection
 
     def message(self):
+        HEADER_WITH_ONE_OCCURRENCE = ["to", "reply-to", "cc"]
+
         encoding = self.encoding or settings.DEFAULT_CHARSET
         msg = SafeMIMEText(self.body, self.content_subtype, encoding)
         msg = self._create_message(msg)
@@ -280,6 +282,8 @@ class EmailMessage:
             # Use cached DNS_NAME for performance
             msg["Message-ID"] = make_msgid(domain=DNS_NAME)
         for name, value in self.extra_headers.items():
+            if name.lower() in HEADER_WITH_ONE_OCCURRENCE and msg[name] is not None:
+                raise ValueError("Cannot insert repeated headers")
             if name.lower() != "from":  # From is already handled
                 msg[name] = value
         return msg
--- a/tests/mail/tests.py
+++ b/tests/mail/tests.py
@@ -129,6 +129,28 @@ class MailTests(HeadersCheckMixin, SimpleTestCase):
             email.recipients(), ["to@example.com", "cc@example.com", "bcc@example.com"]
         )
 
+    def test_headers_not_repeated(self):
+        tests = ["To", "Cc", "Bcc", "Reply_To"]
+        for header in tests:
+            with self.subTest(header=header):
+                header_attr = header.lower()
+                header = header.replace("_", "-")
+                email = EmailMessage(
+                    "Subject",
+                    "Content",
+                    from_email="bounce@example.com",
+                    **{
+                        header_attr: ["test@example.com"],
+                        "headers": {header: "precedence@example.com"},
+                    },
+                )
+                message = email.message()
+                self.assertEqual(
+                    message.get_all(header),
+                    ["precedence@example.com"],
+                )
+                self.assertEqual(getattr(email, header_attr), ["test@example.com"])
+
     def test_cc(self):
         """Regression test for #7722"""
         email = EmailMessage(
--- a/tests/mail/tests.py
+++ b/tests/mail/tests.py
@@ -130,7 +130,7 @@ class MailTests(HeadersCheckMixin, SimpleTestCase):
         )
 
     def test_headers_not_repeated(self):
-        tests = ["To", "Cc", "Bcc", "Reply_To"]
+        tests = ["To", "Cc", "Reply_To"]
         for header in tests:
             with self.subTest(header=header):
                 header_attr = header.lower()
--- a/tests/mail/tests.py
+++ b/tests/mail/tests.py
@@ -191,9 +191,8 @@ class MailTests(HeadersCheckMixin, SimpleTestCase):
             "bounce@example.com",
             ["to@example.com"],
             cc=["foo@example.com"],
-            headers={"Cc": "override@example.com"},
         ).message()
-        self.assertEqual(message["Cc"], "override@example.com")
+        self.assertEqual(message["Cc"], "foo@example.com")
 
     def test_cc_in_headers_only(self):
         message = EmailMessage(
@@ -359,14 +358,17 @@ class MailTests(HeadersCheckMixin, SimpleTestCase):
             "Subject",
             "Content",
             "bounce@example.com",
-            ["list-subscriber@example.com", "list-subscriber2@example.com"],
-            headers={"To": "mailing-list@example.com"},
+            headers={
+                "To": ", ".join(
+                    ["guy1@example.com", "guy2@example.com", "guy3@example.com"]
+                )
+            },
         )
         message = email.message()
-        self.assertEqual(message["To"], "mailing-list@example.com")
         self.assertEqual(
-            email.to, ["list-subscriber@example.com", "list-subscriber2@example.com"]
+            message["To"], "guy1@example.com, guy2@example.com, guy3@example.com"
         )
+        self.assertEqual(email.to, [])
 
         # If we don't set the To header manually, it should default to the `to`
         # argument to the constructor.
@@ -403,10 +405,9 @@ class MailTests(HeadersCheckMixin, SimpleTestCase):
             "bounce@example.com",
             ["to@example.com"],
             reply_to=["foo@example.com"],
-            headers={"Reply-To": "override@example.com"},
         )
         message = email.message()
-        self.assertEqual(message["Reply-To"], "override@example.com")
+        self.assertEqual(message["Reply-To"], "foo@example.com")
 
     def test_reply_to_in_headers_only(self):
         message = EmailMessage(


django/core/mail/message.py
```diff
@@ -280,6 +282,8 @@ def message(self):
             # Use cached DNS_NAME for performance
             msg["Message-ID"] = make_msgid(domain=DNS_NAME)
         for name, value in self.extra_headers.items():
+            if name.lower() in HEADER_WITH_ONE_OCCURRENCE and msg[name] is not None:
+                raise ValueError("Cannot insert repeated headers")
             if name.lower() != "from":  # From is already handled
                 msg[name] = value
```
David-Wobrock, line 288: An alternative to raising an exception, could be assigning the same way we do for `from`.

The behaviour change would then be that the headers in `extra_headers` will silently be ignored - it's probably safer to raise for users, so that they are made aware of the change.

tests/mail/tests.py
```diff
@@ -359,14 +358,17 @@ def test_to_header(self):
             "Subject",
             "Content",
             "bounce@example.com",
-            ["list-subscriber@example.com", "list-subscriber2@example.com"],
-            headers={"To": "mailing-list@example.com"},
```
David-Wobrock, line 363: Please add a test to assert that the new exception raises :)

django/core/mail/message.py
```diff
@@ -280,6 +282,8 @@ def message(self):
             # Use cached DNS_NAME for performance
             msg["Message-ID"] = make_msgid(domain=DNS_NAME)
         for name, value in self.extra_headers.items():
+            if name.lower() in HEADER_WITH_ONE_OCCURRENCE and msg[name] is not None:
```
David-Wobrock, line 285: Why the `msg[name] is not None` condition? :)

django/core/mail/message.py
```diff
@@ -258,6 +258,8 @@ def get_connection(self, fail_silently=False):
         return self.connection
 
     def message(self):
+        HEADER_WITH_ONE_OCCURRENCE = ["to", "reply-to", "cc"]
+
```
David-Wobrock, line 262: This shouldn't be defined as a variable in the method.
Could we move it and make it as a class attribute instead?

And we make probably make it a tuple :)

David-Wobrock:
Thanks for the PR @salvo-polizzi !
I left a few comments to improve the PR :)

django/core/mail/message.py
apollo13: Yeah, I would outright forbid the headers that can be set via our public API and not only if they are not set otherwise. It might also be worth to move it up to `__init__` then?

apollo13: Mhm looking at https://code.djangoproject.com/ticket/17444 we have to be very careful with this change. Looks like we explicitly want to support overriding `to` via `extra_headers` :/

tests/mail/tests.py
apollo13: I do not think it is wise to alter an existing test and change it's behavior. The test imo shows why we probably cannot raise an exception but have to prefer `extra_headers` over the init kwargs?

salvo-polizzi: Thank you all for reviews. @apollo13 Do you think that is better to handle this as suggested by @David-Wobrock above, e.g. in the same way of from?

django/core/mail/message.py
salvo-polizzi: > Why the `msg[name] is not None` condition? :)

I'm sorry for that. It is a mistake

tests/mail/tests.py
apollo13: I don't think so, to me the existing `test_to_header` explicitly tests that `extra_data` overrides the `to` argument (whereas the existing from handling simply ignores the value in `extra_headers` if I read it correctly). So we most likely have to keep this behavior.

salvo-polizzi:
thanks @David-Wobrock and @apollo13 for review. Commited a new solution

salvo-polizzi:
I'm sorry if I forbid tests in previous commits, but I'm a new contributor and so I have to take a little bit of time more to understand what to do. Now I've added tests for not having repeated headers.

django/core/mail/message.py
```diff
@@ -263,9 +268,6 @@ def message(self):
         msg = self._create_message(msg)
         msg["Subject"] = self.subject
         msg["From"] = self.extra_headers.get("From", self.from_email)
-        self._set_list_header_if_not_empty(msg, "To", self.to)
-        self._set_list_header_if_not_empty(msg, "Cc", self.cc)
-        self._set_list_header_if_not_empty(msg, "Reply-To", self.reply_to)
```
David-Wobrock, line 268: `_set_list_header_if_not_empty` is now unused in the class. Could we still use it? Or should it be deleted? (but that should maybe be documented then)

David-Wobrock (commented):
Generally, this starts looking goof to me 👍 

The PR title needs to be changed though :)

django/core/mail/message.py
salvo-polizzi: > `_set_list_header_if_not_empty` is now unused in the class. Could we still use it? Or should it be deleted? (but that should maybe be documented then)

I don't know if it can still be useful, but I could modify that function for doing the same thing instead of delete it. For example I can modify `_set_list_header_if_not_empty` in this way

```python
def _set_list_header_if_not_empty(self, header, values, msg):
  if self.extra_headers.get(header) is None:
     if values:
         msg[header] = ", ".join(str(v) for v in values)
```

and then use it 

```python
for header, values_header in self.headers_with_one_occurrence.items():
   _set_list_header_if_not_empty(self, header, values_header, msg)
```

What do you think?

David-Wobrock: Seems unnecessary indeed.

I'd suggest deleting it :) Since it's an [internal API](https://docs.djangoproject.com/en/5.0/misc/api-stability/#apis-marked-as-internal), I think there is no need to document the change.

tests/mail/tests.py
```diff
@@ -205,6 +205,22 @@ def test_cc_in_headers_only(self):
...
+            ["to@example.com"],
+            cc=["foo@example.com"],
+            headers={"Cc": "override@example.com"},
+        ).message()
+
+        count_cc = 0
+        for el in list(message.raw_items()):
+            if "Cc" in el:
+                count_cc += 1
+        self.assertEqual(count_cc, 1)
```
David-Wobrock, line 222: You can probably simplify this statement (and the ones in the other tests) using:
```suggestion
        count_cc = list(message.raw_items()).count("Cc")
        self.assertEqual(count_cc, 1)
```

David-Wobrock (approved):

tests/mail/tests.py
```diff
@@ -393,6 +426,28 @@ def test_to_in_headers_only(self):
...
+        """
+        email = EmailMessage(
+            subject="test subject",
+            body="test body",
+            from_email="from@example.com",
+            to=["guy1@example.com"],
+            headers={
+                "To": ", ".join(
+                    ["guy1@example.com", "guy2@example.com", "guy3@example.com"]
+                ),
```
felixxm, line 441: There is no need to use `join()` for static strings
```suggestion
                "To": "person1@example.com, person2@example.com",
```

tests/mail/tests.py
```diff
@@ -393,6 +426,28 @@ def test_to_in_headers_only(self):
         ).message()
         self.assertEqual(message["To"], "to@example.com")
 
+    def test_to_not_repeated(self):
+        """
+        Make sure that to is not repeated (#35033)
+        """
```
felixxm, line 432: Please remove ticket numbers from docstrings. Also, to add more explanation, docstrings should state the expected behavior and omit prefixes like "Make sure" since all tests ensure things. This guideline is from [Python coding style](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#python-style).

Personally, I'd remove docstring from new test methods as test method names are self-explanatory.

tests/mail/tests.py
```diff
@@ -384,6 +400,23 @@ def test_to_header(self):
...
+            headers={
+                "To": ", ".join(
+                    ["guy1@example.com", "guy2@example.com", "guy3@example.com"]
+                ),
+            },
+        )
+        message = email.message()
+        self.assertEqual(
+            message["To"], "guy1@example.com, guy2@example.com, guy3@example.com"
+        )
```
felixxm, line 417: This assertion could be moved to the `test_to_not_repeated` without repeating everything else.

tests/mail/tests.py
```diff
@@ -205,6 +205,22 @@ def test_cc_in_headers_only(self):
...
+            "Subject",
+            "Content",
+            "bounce@example.com",
+            ["to@example.com"],
+            cc=["foo@example.com"],
+            headers={"Cc": "override@example.com"},
+        ).message()
+
+        count_cc = 0
+        for el in list(message.raw_items()):
```
felixxm, line 219: `raw_items()` is an undocumented method. What do you think about using `get_all()` instead (e.g. `get_all("cc")`)?

salvo-polizzi:
Thanks @felixxm for review

tests/mail/tests.py
```diff
@@ -393,6 +409,23 @@ def test_to_in_headers_only(self):
...
+            subject="test subject",
+            body="test body",
+            from_email="from@example.com",
+            to=["guy1@example.com"],
+            headers={"To": "guy1@example.com, guy2@example.com, guy3@example.com"},
+        )
+        message = email.message()
+        count_to = len(message.get_all("To"))
+
+        self.assertEqual(count_to, 1)
```
felixxm, line 423: I would assert values not only the number of headers
```suggestion
        self.assertEqual(
            message.get_all("To"),
            ["guy1@example.com, guy2@example.com, guy3@example.com"],
        )     
```

tests/mail/tests.py
```diff
@@ -393,6 +409,23 @@ def test_to_in_headers_only(self):
         ).message()
         self.assertEqual(message["To"], "to@example.com")
 
+    def test_to_not_repeated(self):
+        email = EmailMessage(
+            subject="test subject",
+            body="test body",
+            from_email="from@example.com",
+            to=["guy1@example.com"],
```
felixxm, line 417: It's confusing that `to` contains a value that is also in `headers["To"]`.

tests/mail/tests.py
```diff
@@ -205,6 +205,20 @@ def test_cc_in_headers_only(self):
...
+            "Content",
+            "bounce@example.com",
+            ["to@example.com"],
+            cc=["foo@example.com"],
+            headers={"Cc": "override@example.com"},
+        )
+        message = email.message()
+
+        self.assertEqual(message.get_all("Cc"), ["override@example.com"])
+        self.assertEqual(email.cc, ["foo@example.com"])
```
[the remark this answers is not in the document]
salvo-polizzi: In this case, do i have to remove all kinds of non-repeated tests (such as `test_cc_not_repeated`, `test_to_not_repeated`, etc. )?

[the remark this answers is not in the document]
felixxm: > In this case, do i have to remove all kinds of non-repeated tests (such as `test_cc_not_repeated`, `test_to_not_repeated`, etc. )?

Yes, they'd be redundant.

django/core/mail/message.py
```diff
@@ -248,6 +248,12 @@ def __init__(
                 else:
                     self.attach(*attachment)
         self.extra_headers = headers or {}
+        self.headers_with_one_occurrence = {
+            "To": self.to,
+            "Cc": self.cc,
+            "Bcc": self.bcc,
```
felixxm, line 254: OK, `Bcc` is a different story, we should not include it in headers (check out ticket-18582 and a related ticket-28598). Please remove it from this patch.

felixxm:
TBH, now I have more questions than answers :disappointed: With this patch we will only have one `To`/`Cc`/`Reply-To` header in the message, fine, however, we still pass recipients (`self.to + self.cc + self.bcc`) to the email backends so how it will end? Who will receive a message?

Is it better to leave it as it is? and give email backends a chance to crash/handle this in a way that is preferred from their perspective. I'm afraid that we're introducing here a regression that will be really difficult to debug.

salvo-polizzi:
> TBH, now I have more questions than answers 😞 With this patch we will only have one `To`/`Cc`/`Reply-To` header in the message, fine, however, we still pass recipients (`self.to + self.cc + self.bcc`) to the email backends so how it will end? Who will receive a message?
> 
> Is it better to leave it as it is? and give email backends a chance to crash/handle this in a way that is preferred from their perspective. I'm afraid that we're introducing here a regression that will be really difficult to debug.

I'm sorry but I messed up with the rebasing, so I had to close the PR. If you think that it would be dangerous to apply these changes I will not reopen the PR.

felixxm:
Feel-free to reopen this PR. 

I hope that someone smarter than me will join a discussion :crossed_fingers: . I'd start from checking how real email servers handle duplicated headers with addresses and how they would behave with this patch.

PR #17674, opened 2024-01-03
Fixed #35033 -- Avoid repeated header in email message

salvo-polizzi:
Ref to old  [PR](https://github.com/django/django/pull/17642)
Ticket: [35033](https://code.djangoproject.com/ticket/35033)

### DIFF
--- a/django/core/mail/message.py
+++ b/django/core/mail/message.py
@@ -248,6 +248,11 @@ class EmailMessage:
                 else:
                     self.attach(*attachment)
         self.extra_headers = headers or {}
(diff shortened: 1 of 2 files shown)


salvo-polizzi:
The discussion about this PR is ongoing because it might be dangerous. As @felixxm pointed out to me, the questions are: what do we have to do with email recipients? Is it difficult to debug? Could it introduce a regression?

sarahboyce:
Thank you for your work on this @salvo-polizzi ⭐ 
I will be merging #18325 shortly and so will close this PR

PR #17676, opened 2024-01-03
Refs #33690 -- Updated tutorial for admin dark mode toggle.

adamchainz:
bc7aa2a5e91cf65fc7510edaf1776528c7ad07b4 fixed ticket-33690, and added new code to `base_site.html`. Updating this corresponding “copy-paste-modify” section from the tutorial was missed, which this PR fixes.


felixxm (approved):
@adamchainz Thanks :+1: Good catch :dart:

sabderemane (commented):
Good catch!

sabderemane (approved):

PR #17680, opened 2024-01-03
Fixed tutorial 'background.gif' reference.

adamchainz:
* ~'asgi.py' added out of order in 3930ec1bf275d17f1c36cd38b38114177b1d1565~
* 'background.gif' reference missed in 76fda7729e4cdfec715cd92b2c80d851797b05f7

### DIFF
--- a/docs/intro/reusable-apps.txt
+++ b/docs/intro/reusable-apps.txt
@@ -76,7 +76,7 @@ After the previous tutorials, our project should look like this:
             static/
                 polls/
                     images/
-                        background.gif
+                        background.png
                     style.css
             templates/
                 polls/


docs/intro/tutorial01.txt
```diff
@@ -84,9 +84,9 @@ Let's look at what :djadmin:`startproject` created:
         manage.py
         mysite/
             __init__.py
+            asgi.py
```
felixxm, line 87: Now the file descriptions in the `These files are:` are not in the same order. I wanted to add `asgi.py` next to the `wsgi.py` to make it easier to read for newcomers.

nessita: I agree that this change feels unnecessary, and that if we do merge it for whatever reason, the following section needs to be adjusted to use the same order as the filesystem list:
image

docs/intro/reusable-apps.txt
```diff
@@ -76,7 +76,7 @@ After the previous tutorials, our project should look like this:
             static/
                 polls/
                     images/
-                        background.gif
+                        background.png
```
nessita, line 79: I agree this needs fixing to match the tutorial06 section. Shal we split it in a separated PR?

docs/intro/tutorial01.txt
adamchainz: Okay, fair enough, I will revert this bit.

adamchainz:
This PR now only covers the `background.gif` change.

felixxm:
@adamchainz Thanks :+1:

PR #17696, opened 2024-01-05
Refs #35058 -- Added support for measured geometries to GDAL Point.

smithdc1:
[Ticket](https://code.djangoproject.com/ticket/35058)

This PR adds support for M dimensions to GDAL "POINT" Geometries.

### DIFF
--- a/django/contrib/gis/gdal/geometries.py
+++ b/django/contrib/gis/gdal/geometries.py
@@ -287,6 +287,22 @@ class OGRGeometry(GDALBase):
         else:
             raise ValueError(f"Input to 'set_3d' must be a boolean, got '{value!r}'.")
 
+    @property
+    def is_measured(self):
+        """Return True if the geometry has M coordinates."""
+        return capi.is_measured(self.ptr)
+
+    def set_measured(self, value):
+        """Set if this geometry has M coordinates."""
+        if value is True:
+            capi.set_measured(self.ptr, 1)
+        elif value is False:
+            capi.set_measured(self.ptr, 0)
+        else:
+            raise ValueError(
+                f"Input to 'set_measured' must be a boolean, got '{value!r}'."
+            )
+
     # #### SpatialReference-related Properties ####
 
     # The SRS property
@@ -386,14 +402,22 @@ class OGRGeometry(GDALBase):
         sz = self.wkb_size
         # Creating the unsigned character buffer, and passing it in by reference.
         buf = (c_ubyte * sz)()
-        capi.to_wkb(self.ptr, byteorder, byref(buf))
+        # For backward compatibility, export old-style 99-402 extended
+        # dimension types when geometry does not have an M dimension.
+        # https://gdal.org/api/vector_c_api.html#_CPPv417OGR_G_ExportToWkb12OGRGeometryH15OGRwkbByteOrderPh
+        to_wkb = capi.to_iso_wkb if self.is_measured else capi.to_wkb
+        to_wkb(self.ptr, byteorder, byref(buf))
         # Returning a buffer of the string at the pointer.
         return memoryview(string_at(buf, sz))
 
     @property
     def wkt(self):
         "Return the WKT representation of the Geometry."
-        return capi.to_wkt(self.ptr, byref(c_char_p()))
+        # For backward compatibility, export old-style 99-402 extended
+        # dimension types when geometry does not have an M dimension.
+        # https://gdal.org/api/vector_c_api.html#_CPPv417OGR_G_ExportToWkt12OGRGeometryHPPc
+        to_wkt = capi.to_iso_wkt if self.is_measured else capi.to_wkt
+        return to_wkt(self.ptr, byref(c_char_p()))
 
     @property
     def ewkt(self):
@@ -569,10 +593,20 @@ class Point(OGRGeometry):
             return capi.getz(self.ptr, 0)
 
     @property
+    def m(self):
+        """Return the M coordinate for this Point."""
+        if self.is_measured:
+            return capi.getm(self.ptr, 0)
+
+    @property
     def tuple(self):
         "Return the tuple of this point."
+        if self.is_3d and self.is_measured:
+            return self.x, self.y, self.z, self.m
         if self.is_3d:
-            return (self.x, self.y, self.z)
+            return self.x, self.y, self.z
+        if self.is_measured:
+            return self.x, self.y, self.m
         return self.x, self.y
 
     coords = tuple
@@ -753,7 +787,9 @@ GEO_CLASSES = {
     6: MultiPolygon,
     7: GeometryCollection,
     101: LinearRing,
-    1 + OGRGeomType.wkb25bit: Point,
+    2001: Point,  # POINT M
+    3001: Point,  # POINT ZM
+    1 + OGRGeomType.wkb25bit: Point,  # POINT Z
     2 + OGRGeomType.wkb25bit: LineString,
     3 + OGRGeomType.wkb25bit: Polygon,
     4 + OGRGeomType.wkb25bit: MultiPoint,
--- a/django/contrib/gis/gdal/prototypes/geom.py
+++ b/django/contrib/gis/gdal/prototypes/geom.py
@@ -51,6 +51,7 @@ to_kml = string_output(
 getx = pnt_func(lgdal.OGR_G_GetX)
 gety = pnt_func(lgdal.OGR_G_GetY)
 getz = pnt_func(lgdal.OGR_G_GetZ)
+getm = pnt_func(lgdal.OGR_G_GetM)
 
 # Geometry creation routines.
 if GDAL_VERSION >= (3, 3):
@@ -82,6 +83,8 @@ geom_sym_diff = geom_output(lgdal.OGR_G_SymmetricDifference, [c_void_p, c_void_p
 geom_union = geom_output(lgdal.OGR_G_Union, [c_void_p, c_void_p])
 is_3d = bool_output(lgdal.OGR_G_Is3D, [c_void_p])
 set_3d = void_output(lgdal.OGR_G_Set3D, [c_void_p, c_int], errcheck=False)
+is_measured = bool_output(lgdal.OGR_G_IsMeasured, [c_void_p])
+set_measured = void_output(lgdal.OGR_G_SetMeasured, [c_void_p, c_int], errcheck=False)
 
 # Geometry modification routines.
 add_geom = void_output(lgdal.OGR_G_AddGeometry, [c_void_p, c_void_p])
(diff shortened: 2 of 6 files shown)


django/contrib/gis/gdal/geometries.py
```diff
@@ -546,16 +590,26 @@ def y(self):
...
         "Return the tuple of this point."
         if self.coord_dim == 2:
             return (self.x, self.y)
         elif self.coord_dim == 3:
-            return (self.x, self.y, self.z)
+            if self.is_3d:
+                return (self.x, self.y, self.z)
+            return (self.x, self.y, self.m)
+        elif self.coord_dim == 4:
+            return (self.x, self.y, self.z, self.m)
```
ngnpope, line 612: I'm wondering if this can be more straightforward? Maybe:

```python
        if self.is_3d and self.is_measured:
            return self.x, self.y, self.z, self.m
        if self.is_3d:
            return self.x, self.y, self.z
        if self.is_measured:
            return self.x, self.y, self.m
        return self.x, self.y
```

This drops the `== 2` branch as [`OGR_G_CoordinateDimension`](https://gdal.org/api/vector_c_api.html#_CPPv425OGR_G_CoordinateDimension12OGRGeometryH) can only return `2`, `3`, or `4` anyway.

We could replace `if self.is_3d and self.is_measured:` with `if self.coord_dim == 4:`, but this way is perhaps clearer in meaning.

django/contrib/gis/gdal/prototypes/geom.py
```diff
@@ -112,7 +122,7 @@ def topology_func(f):
 get_area = double_output(lgdal.OGR_G_GetArea, [c_void_p])
 get_centroid = void_output(lgdal.OGR_G_Centroid, [c_void_p, c_void_p])
 get_dims = int_output(lgdal.OGR_G_GetDimension, [c_void_p])
-get_coord_dim = int_output(lgdal.OGR_G_GetCoordinateDimension, [c_void_p])
+get_coord_dim = int_output(lgdal.OGR_G_CoordinateDimension, [c_void_p])
```
ngnpope, line 125: Is this a backward compatibility concern? I think it's fine, but people could have been getting 3 returned for things that had 4 dimensions if the measured part was being ignored. That was likely incorrect anyway, and I'm not sure how likely it is that this would have been encountered -  not sure if it's worth a note somewhere.

tests/gis_tests/gdal_tests/test_geom.py
```diff
@@ -714,3 +714,56 @@ def test_geometry_types(self):
...
+
+    def test_point_measure(self):
+        p1 = OGRGeometry("POINT Z (1 2 3)")
+        self.assertEqual(p1.x, 1)
+        self.assertEqual(p1.y, 2)
+        self.assertEqual(p1.z, 3)
+        self.assertEqual(p1.m, None)
+        self.assertIs(p1.is_3d, True)
+        self.assertIs(p1.is_measured, False)
+        # "Old style" WKT for backward compatability.
```
ngnpope, line 726:
```suggestion
        # Old-style WKT for backward compatibility.
```

tests/gis_tests/gdal_tests/test_geom.py
```diff
@@ -714,3 +714,56 @@ def test_geometry_types(self):
...
+
+        p2 = OGRGeometry("POINT M (1 2 3)")
+        self.assertEqual(p2.x, 1)
+        self.assertEqual(p2.y, 2)
+        self.assertIs(p2.z, None)
+        self.assertEqual(p2.m, 3)
+        self.assertIs(p2.is_3d, False)
+        self.assertIs(p2.is_measured, True)
+        self.assertEqual(p2.wkt, "POINT M (1 2 3)")
+        # GEOSGeometry does not yet support M dimension
```
ngnpope, line 739:
```suggestion
        # GEOSGeometry does not yet support the M dimension.
```

tests/gis_tests/gdal_tests/test_geom.py
```diff
@@ -714,3 +714,56 @@ def test_geometry_types(self):
...
+        p3.is_measured = False
+        self.assertEqual(p3.m, None)
+        self.assertEqual(p3.is_measured, False)
+        self.assertEqual(p3.wkt, "POINT (1 2 3)")
+        p3.is_measured = True
+        self.assertEqual(p3.is_measured, True)
+        self.assertEqual(p3.wkt, "POINT ZM (1 2 3 4)")
+        # WKB creates geometry with the same WKT
+        self.assertEqual(OGRGeometry(p3.wkb).wkt, p3.wkt)
+        # GEOSGeometry does not yet support M dimension
```
ngnpope, line 768:
```suggestion
        # GEOSGeometry does not yet support the M dimension.
```

tests/gis_tests/gdal_tests/test_geom.py
```diff
@@ -714,3 +714,56 @@ def test_geometry_types(self):
...
+        self.assertEqual(p3.wkt, "POINT ZM (1 2 3 4)")
+        self.assertEqual(p3.tuple, (1, 2, 3, 4))
+        p3.is_measured = False
+        self.assertEqual(p3.m, None)
+        self.assertEqual(p3.is_measured, False)
+        self.assertEqual(p3.wkt, "POINT (1 2 3)")
+        p3.is_measured = True
+        self.assertEqual(p3.is_measured, True)
+        self.assertEqual(p3.wkt, "POINT ZM (1 2 3 4)")
+        # WKB creates geometry with the same WKT
```
ngnpope, line 766:
```suggestion
        # WKB creates geometry with the same WKT.
```

docs/releases/5.1.txt
```diff
@@ -53,7 +53,11 @@ Minor features
 :mod:`django.contrib.gis`
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
-* ...
+* :class:`~django.contrib.gis.gdal.OGRGeometry` added support for ``POINT M``
+  and ``POINT ZM`` geometry types. To support this ``OGRGeometry`` learnt
+  :attr:`~django.contrib.gis.gdal.OGRGeometry.is_3d` and
+  :attr:`~django.contrib.gis.gdal.OGRGeometry.is_measured`, and ``Point``
+  learnt :attr:`~django.contrib.gis.gdal.Point.m`.
```
ngnpope, line 60: Use present tense in release notes and "learnt" isn't US English. Maybe:

```suggestion
* Support for measured geometries is added via the new
  :attr:`~django.contrib.gis.gdal.OGRGeometry.is_3d` and
  :attr:`~django.contrib.gis.gdal.OGRGeometry.is_measured` properties of
  :class:`~django.contrib.gis.gdal.OGRGeometry` and the new
  :attr:`~django.contrib.gis.gdal.Point.m` property of 
  :class:`~django.contrib.gis.gdal.Point`.
```

ngnpope (changes requested):
Nice to see this addition 👍🏻

django/contrib/gis/gdal/geometries.py
```diff
@@ -268,6 +272,35 @@ def extent(self):
         "Return the envelope as a 4-tuple, instead of as an Envelope object."
         return self.envelope.tuple
 
+    @property
+    def is_3d(self):
+        "Return True if the geometry has Z coordinates."
+        return capi.is_3d(self.ptr)
+
+    @is_3d.setter
+    def is_3d(self, has_z):
```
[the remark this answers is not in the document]
claudep: At first, I found that defining a setter for a property starting with `is_` sounds a bit awkward. I'd rather see an `is_3d` read-only property on one side, and a `set_3d` method on the other side, mimicing the GDAL API. Maybe a personal preference?

django/contrib/gis/gdal/prototypes/geom.py
smithdc1: I've added some words to the backwards incompatible changes section of  `5.1.txt` for this.

django/contrib/gis/gdal/prototypes/geom.py
```diff
@@ -91,9 +97,13 @@ def topology_func(f):
 to_wkb = void_output(
     lgdal.OGR_G_ExportToWkb, None, errcheck=True
 )  # special handling for WKB.
+to_iso_wkb = void_output(lgdal.OGR_G_ExportToWkb, None, errcheck=True)
```
smithdc1, line 100:
```suggestion
to_iso_wkb = void_output(lgdal.OGR_G_ExportToIsoWkb, None, errcheck=True)
```

django/contrib/gis/gdal/geometries.py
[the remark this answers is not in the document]
smithdc1: I've attempted to implement separate `is_*` and `set_*` methods as proposed by @claudep.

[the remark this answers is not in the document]
claudep: Thanks. It was a proposal, but if others find your first version better, I will not oppose.

smithdc1:
This PR is to add support for "M" Geometries. It also adds support for "M" dimensions to "POINT" so that we have something to test.

I'd like to focus this PR on reviewing the Python bindings and the API. I propose that adding support for the "M" dimension to other `GEO_CLASSES` is a follow up PR.

django/contrib/gis/gdal/geometries.py
```diff
@@ -268,6 +275,34 @@ def extent(self):
...
+        """Return True if the geometry has Z coordinates."""
+        return capi.is_3d(self.ptr)
+
+    def set_3d(self, value):
+        if value is True:
+            capi.set_3d(self.ptr, 1)
+        elif value is False:
+            capi.set_3d(self.ptr, 0)
+        else:
+            raise ValueError(f"Input to 'set_3d' must be a boolean, got {value!r}.")
```
felixxm, line 289: What do you think about adding `is_3d` and `set_3d` in a separate PR? We could merge it first.

felixxm: I can move it if you want.

smithdc1: I'm happy to do this, the tests will need a bit of work.

Likely we can do the same with the change to use the `Iso` output for wkt/wkb.

felixxm: Sounds good :+1: Thanks.

felixxm: Merged, please rebase.

django/contrib/gis/gdal/prototypes/geom.py
smithdc1: I originally added a small release note... but on review now I'm not so sure. I don't think geometries with four dimensions could be created before this patch? 

```python
>>> OGRGeometry("POINT (1 2 3 4)")
TypeError: Unsupported geometry type: PointZM
```

felixxm: Agreed, it's not worth adding a note.

felixxm (commented):
@smithdc1 Thanks :+1:

I pushed small edits to a release note.

ngnpope:
Thanks David 👏🏻

PR #17704, opened 2024-01-06
Correct Latvian translation

googol7:


ngnpope:
Thanks for the report, but translations are handled using [Transifex](https://docs.djangoproject.com/en/dev/internals/contributing/localizing/#translations).

PR #17709, opened 2024-01-08
#34044 -- Add the admin app filter on index pages in addition to the nav sidebar

Hisham-Pak:
Ticket: https://code.djangoproject.com/ticket/34044


django/contrib/admin/static/admin/js/app_filter.js
```diff
@@ -0,0 +1,62 @@
...
+            sessionStorage.setItem('django.admin.navSidebarFilterValue', filterValue);
+        }
+
+        const nav = document.getElementById('app-filter');
+        nav.addEventListener('change', checkValue, false);
+        nav.addEventListener('input', checkValue, false);
+        nav.addEventListener('keyup', checkValue, false);
+
+        // Only retrieve the stored value if the current page is not the index page
+        if (!location.pathname.endsWith('/admin/')) {
```
sergei-maertens, line 52: This will break if you don't host the admin on `/admin/` (e.g. `/obfuscated/`) right?

django/contrib/admin/templates/admin/base.html
```diff
@@ -6,21 +6,23 @@
 <link rel="stylesheet" href="{% block stylesheet %}{% static "admin/css/base.css" %}{% endblock %}">
 {% block dark-mode-vars %}
   <link rel="stylesheet" href="{% static "admin/css/dark_mode.css" %}">
-  <script src="{% static "admin/js/theme.js" %}"></script>
+  <script src="{% static "admin/js/theme.js" %}" defer></script>
```
sergei-maertens, line 9: 1. I don't see how this change is related to the ticket?
2. The theme script applies the theme from the `localStorage`, so I suspect if you don't have the default value set, you may get brief flickering between the loading of the script and it finally executing. Or am I misunderstanding `defer`?

Hisham-Pak: Removed defer from theme.js

nessita:
I forgot to add in my previous comment that these changes also need to include an entry in the `docs/releases/5.0.1.txt` file describing the new feature.

Hisham-Pak:
@nessita, I added:

- The former author to `AUTHORS` file and made a commit message with `Co-authored-by`.
- Regarding [this](https://github.com/django/django/pull/16107#issuecomment-1262098976) comment, I could not really visualize what @knyghty really meant (wouldn't sidebar really just duplicate models on index page?). Moreover, I think a new ticket should be opened to discuss about this feature.
- Squashed commits into cleaner history.
- Also added to release notes.

nessita:
> @nessita, I added:

Thank you @Hisham-Pak for the addings! Below some questions/comments.

>     * Regarding [this](https://github.com/django/django/pull/16107#issuecomment-1262098976) comment, I could not really visualize what @knyghty really meant (wouldn't sidebar really just duplicate models on index page?). Moreover, I think a new ticket should be opened to discuss about this feature.

That's a good question, I have asked Tom in our Discord server ([thread](https://discord.com/channels/856567261900832808/1194661151464755280)) and it seems they meant that the filter should be shown also for model list at the app level (for example, under `/admin/auth/`).

I'll do a code review now and post my feedback.

django/contrib/admin/templates/admin/app_list.html
```diff
@@ -1,44 +1,46 @@
...
-        <caption>
-          <a href="{{ app.app_url }}" class="section" title="{% blocktranslate with name=app.name %}Models in the {{ name }} application{% endblocktranslate %}">{{ app.name }}</a>
-        </caption>
-        {% for model in app.models %}
-          {% with model_name=model.object_name|lower %}
-            <tr class="model-{{ model_name }}{% if model.admin_url in request.path|urlencode %} current-model{% endif %}">
-              <th scope="row" id="{{ app.app_label }}-{{ model_name }}">
-                {% if model.admin_url %}
-                  <a href="{{ model.admin_url }}"{% if model.admin_url in request.path|urlencode %} aria-current="page"{% endif %}>{{ model.name }}</a>
+<div id="content-main" class="filterable-apps-table">
```
nessita, line 3: I'm not sure this is correct. When this template is included in nav_sidebar.html, this div with `id="content-main"` gets included inside the `<nav>` element and that does not feel correct, more so knowing that there is another div with the same id further down the page.

I would suggest to completely revert changes to this file, and only add the div in the index.html file where app_list.html is included.

django/contrib/admin/templates/admin/index.html
```diff
@@ -11,10 +11,17 @@
 
 {% block nav-sidebar %}{% endblock %}
 
+{% block content_title %}
+  <div class="admin-index-title">
+    {% if title %}<h1>{{ title }}</h1>{% endif %}
```
nessita, line 16: Could you replace this with `block.super`?
```suggestion
    {{ block.super }}
```

django/contrib/admin/templates/admin/index.html
```diff
@@ -11,10 +11,17 @@
...
+    <input type="search" id="app-filter"
+           placeholder="{% translate 'Start typing to filter…' %}"
+           aria-label="{% translate 'Filter navigation items' %}">
+  </div>
+{% endblock %}
+
 {% block content %}
-<div id="content-main">
```
nessita, line 15: I think this needs to be restored.

django/contrib/admin/templates/admin/index.html
```diff
@@ -11,10 +11,17 @@
 
 {% block nav-sidebar %}{% endblock %}
 
+{% block content_title %}
+  <div class="admin-index-title">
+    {% if title %}<h1>{{ title }}</h1>{% endif %}
+    <input type="search" id="app-filter"
```
nessita, line 17: Could we isolate this into a separated template to be included in both nav_sidebar.html and index.html?

nessita (commented):
Added an initial set of comments.

Since we are doing a rename from sidebar to something more appropriate, I would suggest that:

1. We split the changes into two commits, one doing only the renames and moving to isolated files, and other commit that makes use of that in the admin index template.
2.  The new name is something closer to "model filter" and not "app filter", since the behavior is really about filtering model names and not apps, right?

What do you think @knyghty?

nessita:
This work may change pending resolution of https://forum.djangoproject.com/t/adding-the-admin-nav-sidebar-to-almost-every-page/26921

knyghty:
My opinion on this at the moment is that based on this [forum thread](https://forum.djangoproject.com/t/adding-the-admin-nav-sidebar-to-almost-every-page/26921/8) we should instead try to implement it via adding the nav sidebar across all pages. If that turns out to be more complex than I think it is, we should go back to this as an interim solution.

PR #17724, opened 2024-01-11
Fixed GIS tests with Oracle 23c.

smithdc1:
My attempt at a patch to allow GIS tests to pass with Oracle 23c. 

The change to `LayerMapping` is due to Oracle 23c changing "MULTI" geometries containing one value to their single type. 

<details>

```sql
create table LAYERMAP_COUNTY_TEST
(
    ID       NUMBER(11),
    NAME     NVARCHAR2(25),
    STATE_ID NUMBER(11),
    MPOLY    MDSYS.SDO_GEOMETRY
);

INSERT INTO "LAYERMAP_COUNTY_TEST" ("NAME",
                               "STATE_ID",
                               "MPOLY")
VALUES ('Galveston', 3, SDO_GEOMETRY('MULTIPOLYGON (((30 20, 45 40, 10 40, 30 20)))',4269));

INSERT INTO "LAYERMAP_COUNTY_TEST" ("NAME",
                               "STATE_ID",
                               "MPOLY")
VALUES ('Hawaii', 4, SDO_GEOMETRY('MULTIPOLYGON (((30 20, 45 40, 10 40, 30 20)),
((15 5, 40 10, 10 20, 5 10, 15 5)))',4269));

select SDO_UTIL.TO_WKTGEOMETRY("LAYERMAP_COUNTY_TEST"."MPOLY")
 from LAYERMAP_COUNTY_TEST;

"POLYGON ((30.0 20.0, 45.0 40.0, 10.0 40.0, 30.0 20.0))"
"MULTIPOLYGON (((30.0 20.0, 45.0 40.0, 10.0 40.0, 30.0 20.0)), ((15.0 5.0, 40.0 10.0, 10.0 20.0, 5.0 10.0, 15.0 5.0)))"
```

</details>


tests/gis_tests/geoapp/tests.py
```diff
@@ -711,10 +711,16 @@ def test_unionagg_tolerance(self):
         tx = Country.objects.get(name="Texas").mpoly
         # Tolerance is greater than distance between Forney and Dallas, that's
         # why Dallas is ignored.
-        forney_houston = GEOSGeometry(
-            "MULTIPOINT(-95.363151 29.763374, -96.467222 32.751389)",
-            srid=4326,
-        )
+        if connection.ops.oracle and connection.oracle_version >= (23,):
+            forney_houston = GEOSGeometry(
+                "MULTIPOINT (-95.363151 29.763374, -96.801611 32.782057)",
```
smithdc1, line 716: Marking this PR as draft. 

This is the edit to make the test pass, but it seems wrong to me. This new point is Dallas (entering the point at https://www.geometrymapper.com/), before it was Forney. Note above would suggest this change is wrong. 

Thoughts welcome 😄

felixxm: Maybe it's related with undefined ordering :thinking:

django/contrib/gis/utils/layermapping.py
```diff
@@ -646,6 +646,14 @@ def _save(feat_range=default_range, num_feat=0, num_saved=0):
                             else:
                                 geom = geom_value.ogr
                                 new = OGRGeometry(kwargs[self.geom_field])
+                                if (
+                                    geom_typeid := geom.geos.geom_typeid
+                                ) in self.MULTI_TYPES:
+                                    # Database returned non-MULI type, reconvert.
```
felixxm, line 652:
```suggestion
                                    # Database returned non-MULTI type, reconvert.
```

tests/gis_tests/geoapp/tests.py
smithdc1: Taking another look at this morning. Just leading a small note here for now to say that I've just realised the `supports_tolerance_parameter` feature (and therefore this test) is only enabled on Oracle.

tests/gis_tests/relatedapp/tests.py
```diff
@@ -100,9 +100,22 @@ def test_related_union_aggregate(self):
...
         # implementation-dependent (DB backend, GEOS version)
-        self.assertEqual({p.ewkt for p in ref_u1}, {p.ewkt for p in u1})
-        self.assertEqual({p.ewkt for p in ref_u2}, {p.ewkt for p in u2})
-        self.assertEqual({p.ewkt for p in ref_u1}, {p.ewkt for p in u3})
+        for point in [*u1, *u2, *u3]:
+            self.assertEqual(point.srid, 4326)
+
+        wkt_w = WKTWriter(trim=True, precision=6)
+        self.assertEqual(
+            {wkt_w.write(p).decode("utf-8") for p in ref_u1},
```
smithdc1, line 108: I'm suspecting these test failures are due to version differences in GDAL/GEOS. 

There's [`equals_exact()`](https://docs.djangoproject.com/en/5.0/ref/contrib/gis/geos/#django.contrib.gis.geos.GEOSGeometry.equals_exact) which accepts a tolerance. That would be useful here, but what's tricky here is that we're comparing two sets. I'm not sure the Points can be ordered in a deterministic(?) way first.

felixxm: > I'm suspecting these test failures are due to version differences in GDAL/GEOS.
> 
> There's [`equals_exact()`](https://docs.djangoproject.com/en/5.0/ref/contrib/gis/geos/#django.contrib.gis.geos.GEOSGeometry.equals_exact) which accepts a tolerance. That would be useful here, but what's tricky here is that we're comparing two sets. I'm not sure the Points can be ordered in a deterministic(?) way first.

Yes, now we have the same failure on Oracle 19c. GEOS is responsible for the `wkt` format,  so it's rather a failure on Oracle and GEOS 3.12+. Can we move it to the separate PR?

felixxm:
I think we can close it for now.

PR #17726, opened 2024-01-12
Fixed #27225 -- Added "Age" header when fetching cached responses.

e11bits:
This addresses Ticket 27225. I took the PR #7246 from @renskiy and rebased it to the current main branch.

Tests pass locally and I added just a minor change to `hello_world_view` .  I haven't yet added any missing changes recommended in #7246 by @timgraham , so I create this as a draft PR first.

This is my fourth PR for django, so please advise as you see fit.

### DIFF
--- a/django/middleware/cache.py
+++ b/django/middleware/cache.py
@@ -43,6 +43,8 @@ More details about how the caching works:
 
 """
 
+import time
+
 from django.conf import settings
 from django.core.cache import DEFAULT_CACHE_ALIAS, caches
 from django.utils.cache import (
@@ -53,6 +55,7 @@ from django.utils.cache import (
     patch_response_headers,
 )
 from django.utils.deprecation import MiddlewareMixin
+from django.utils.http import parse_http_date_safe
 
 
 class UpdateCacheMiddleware(MiddlewareMixin):
@@ -170,6 +173,15 @@ class FetchFromCacheMiddleware(MiddlewareMixin):
         if response is None:
             request._cache_update_cache = True
             return None  # No cache information available, need to rebuild.
+
+        # Derive the age estimation of the cached response.
+        if (max_age_seconds := get_max_age(response)) is not None and (
+            expires_timestamp := parse_http_date_safe(response["Expires"])
+        ) is not None:
+            now_timestamp = int(time.time())
+            remaining_seconds = expires_timestamp - now_timestamp
+            # Use Age: 0 if local clock got turned back.
+            response["Age"] = max(0, max_age_seconds - remaining_seconds)
 
         # hit, return cached response
         request._cache_update_cache = False
--- a/tests/cache/tests.py
+++ b/tests/cache/tests.py
@@ -2752,6 +2752,37 @@ class CacheMiddlewareTest(SimpleTestCase):
 
         self.assertIsNot(thread_caches[0], thread_caches[1])
 
+    def test_cache_control_max_age(self):
+        view = cache_page(2)(hello_world_view)
+        request = self.factory.get("/view/")
+
+        # First request. Freshly created response gets returned with no Age
+        # header.
+        with mock.patch.object(
+            time, "time", return_value=1468749600
+        ):  # Sun, 17 Jul 2016 10:00:00 GMT
+            response = view(request, 1)
+            response.close()
+            self.assertIn("Expires", response)
+            self.assertEqual(response["Expires"], "Sun, 17 Jul 2016 10:00:02 GMT")
+            self.assertIn("Cache-Control", response)
+            self.assertEqual(response["Cache-Control"], "max-age=2")
+            self.assertNotIn("Age", response)
+
+        # Second request one second later. Response from the cache gets
+        # returned with an Age header set to 1 (second).
+        with mock.patch.object(
+            time, "time", return_value=1468749601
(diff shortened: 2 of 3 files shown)


e11bits:
In addition to the original PR I did the following:

- squashed to commits
- Removed the change in `django/utils/cache.py` . I also couldn't see why this change would be necessary. see original comment https://github.com/django/django/pull/7246#discussion_r81975841
- Added a comment and renamed some of the variables in `django/middleware/cache.py` to make it more obvious how the calculation is done. see original comment https://github.com/django/django/pull/7246#discussion_r81975935
- Changed the comment in the test case to explain the value of the `Age: 1` header entry. see original comment https://github.com/django/django/pull/7246#discussion_r81976035

e11bits:
Ok, now I know, why the former code in `django/utils/cache.py` was introduced by the original PR. This is due to another bug for which probably have to open a new ticket first.

e11bits:
This is currently blocked by #17776

MiirzaBaig:
There is an error in cache middleware system.

e11bits:
Not blocked by https://github.com/django/django/pull/17776 anymore

tests/cache/tests.py
```diff
@@ -2477,8 +2477,8 @@ class PrefixedCacheI18nTest(CacheI18nTest):
     pass
 
 
-def hello_world_view(request, value):
-    return HttpResponse("Hello World %s" % value)
+def hello_world_view(request, value=None):
+    return HttpResponse("Hello World %s" % (value or ""))
```
felixxm, line 2481: I'd revert this change and pass `value` to the `view()` calls:
```diff
diff --git a/tests/cache/tests.py b/tests/cache/tests.py
index ce02872d4e..40d07fd45e 100644
--- a/tests/cache/tests.py
+++ b/tests/cache/tests.py
@@ -2477,8 +2477,8 @@ class PrefixedCacheI18nTest(CacheI18nTest):
     pass
 
 
-def hello_world_view(request, value=None):
-    return HttpResponse("Hello World %s" % (value or ""))
+def hello_world_view(request, value):
+    return HttpResponse("Hello World %s" % value)
 
 
 def csrf_view(request):
@@ -2760,7 +2760,7 @@ class CacheMiddlewareTest(SimpleTestCase):
         with mock.patch.object(
             time, "time", return_value=1468749600
         ):  # Sun, 17 Jul 2016 10:00:00 GMT
-            response = view(request)
+            response = view(request, 1)
             response.close()
             self.assertIn("Expires", response)
             self.assertEqual(response["Expires"], "Sun, 17 Jul 2016 10:00:02 GMT")
@@ -2773,7 +2773,7 @@ class CacheMiddlewareTest(SimpleTestCase):
         with mock.patch.object(
             time, "time", return_value=1468749601
         ):  # Sun, 17 Jul 2016 10:00:01 GMT
-            response = view(request)
+            response = view(request, 1)
             response.close()
             self.assertIn("Expires", response)
             self.assertEqual(response["Expires"], "Sun, 17 Jul 2016 10:00:02 GMT")

```

felixxm (commented):
@e11bits Thanks :+1: Can you correct the commit message? This PR is not about `max-age`.

django/middleware/cache.py
```diff
@@ -171,6 +174,17 @@ def process_request(self, request):
             request._cache_update_cache = True
             return None  # No cache information available, need to rebuild.
 
+        # Derive the age estimation of the cached entry using max_age and the Expires
+        # timestamp. max_age_seconds is None if CACHE_MIDDLEWARE_SECONDS is
+        # misconfigured to be a float.
+        max_age_seconds = get_max_age(response)
+        if max_age_seconds is not None:
+            expires_timestamp = parse_http_date(response["Expires"])
```
kezabelle, line 182: 🤔 I'm wondering out loud, mostly. Is this a scenario where we care/need to handle the possibility of an exception being raised? There's other places which act similarly upon "response" headers[^1] and do so by leveraging `parse_http_date_safe`[^2] to avoid crashing, but it's decidedly unclear to me as to whether that's relevant here...

Just flagging it for consideration, as such.

[^1]: "last modified" & "if modified since"
[^2]: which just catches and returns `None`

e11bits: I thought about that too, but came to the conclusion, that extra caution is not required here, since the `Expires` response header is set by a django cache. 

`response.headers["Expires"] = http_date(time.time() + cache_timeout)`

Two cases you mentioned in `cache.py` are dealing with request headers `If-Modified-Since` and `If-Unodified-Since` that could likely contain badly formatted date strings and need `parse_http_date_safe`

`Last-Modified` from a request will be saved in the cache as is and could be badly formatted as well and is used for a conditional response.

felixxm: > ...since the `Expires` response header is set by a django cache.

Yes, but only if it's not already set.

e11bits: > > ...since the `Expires` response header is set by a django cache.
> 
> Yes, but only if it's not already set.

True. I missed `Expires` being part of `_not_modified`.

e11bits: > What do you think about the following change?

Looks good. Made the change.

felixxm:
@e11bits Thanks :+1:

PR #17765, opened 2024-01-22
Added 'python' to the console command in contributing.txt for clarity

iammrpresident:
Hey Django team,

I've made a small but, I believe, valuable update to the **contributing.txt** file. Specifically, I added the '**python**' prefix to the console command in the **runtests.py** section. This adjustment aims to enhance clarity and ensure a more seamless experience for contributors running tests using this command.

While I was trying to run my first test the '**runtests.py**' command kept opening the test file without running the test. The addition of '**python**' makes the command more explicit and user-friendly.

This helps contributors, especially those new to the project, to better understand and execute the testing process.

I have locally tested these changes to ensure they don't introduce any issues and that the modified command works as expected.

### DIFF
--- a/docs/intro/contributing.txt
+++ b/docs/intro/contributing.txt
@@ -222,7 +222,7 @@ some other flavor of Unix, run:
 
 .. console::
 
-    $ ./runtests.py
+    $ python ./runtests.py
 
 Now sit back and relax. Django's entire test suite has thousands of tests, and
 it takes at least a few minutes to run, depending on the speed of your
@@ -361,7 +361,7 @@ that's really what happens. ``cd`` to the Django ``tests/`` directory and run:
 
 .. console::
 
-    $ ./runtests.py shortcuts
+    $ python ./runtests.py shortcuts
 
 If the tests ran correctly, you should see one failure corresponding to the test
 method we added, with this error:
@@ -390,7 +390,7 @@ whether the code we added is working correctly. Again, navigate to the Django
 
 .. console::
 
-    $ ./runtests.py shortcuts
+    $ python ./runtests.py shortcuts
 
 Everything should pass. If it doesn't, make sure you correctly added the
 function to the correct file.
@@ -409,7 +409,7 @@ directory and run:
 
 .. console::
 
-    $ ./runtests.py
+    $ python ./runtests.py
 
 Writing Documentation
 =====================


smithdc1 (approved):
On Windows (for me) without `python`, this command will result in `runtests.py` being opened in notepad. 

Looks good 👍

nessita:
Thank you @iammrpresident for this proposal.

While I encourage being inclusive of the different development environments, this change only applies to Windows users, and also there are many other places in the docs where `./manage.py` or `./runtests.py` is used. Because of that, I think a better alternative would be to evaluate the adding of a single yet general clarification note for Windows users somewhere at the beginning of the docs.

What do you think? Would you like to make such evaluation and propose a change?

smithdc1:
I'm not so familiar with non windows systems. Is it standard to run python scripts without `python`.

The tutorial seems to use `python` throughout. 

https://github.com/django/django/blob/e412d85b4626bc56eb25206a40c3529162ce5dfc/docs/intro/tutorial01.txt#L131

iammrpresident:
Thank you @nessita for your feedback and suggestion. 

I appreciate your perspective on the broader use of `./manage.py` or `./runtests.py` in the documentation.

I agree that a more general clarification note for Windows users, placed at the beginning of the docs, could provide a comprehensive solution. I'll evaluate this approach and work on proposing the necessary changes to ensure inclusivity for users across different development environments.


Feel free to share any additional thoughts or considerations.

iammrpresident:
Great question! @smithdc1 

On non-Windows systems, it's common to run Python scripts by directly invoking the Python interpreter without explicitly using the "python" command. This practice assumes that the system has the appropriate Python version available in the environment variable PATH.

The tutorial predominantly uses the python command for consistency and clarity, especially for users who may be new to Python or different operating systems. However, in many environments, using python is interchangeable with invoking the script directly, like `./manage.py` or `./runtests.py`.

In the specific case of the proposed change, the addition of "python" to the command was aimed at ensuring clarity and consistency, especially for Windows users.

Feel free to ask if you have any more questions or if there's anything else you'd like clarification on.

docs/intro/contributing.txt
```diff
@@ -222,7 +222,7 @@ some other flavor of Unix, run:
 
 .. console::
 
-    $ ./runtests.py
+    $ python ./runtests.py
```
sarahboyce, line 225: Just above this it says "If you're using GNU/Linux, macOS, or some other flavor of Unix, run:"
This should be updated if we're making changes for Windows 🙂

iammrpresident: Well noted 👍
I would include that when making the changes for Windows.

iammrpresident:
I have updated the docs accordingly, @nessita kindly review.

PR #17773, opened 2024-01-23
Added test for the sendtestemail command when no recipients are given.

bmispelon:
I was reading the code for the `sendtestemail` command (as one does) and was confused as to how it made sure that either some recipients were passed on the command line, or the `--admins` (or `--managers`) flag was set.
Turns out that the presence of the `missing_args_message` argument on the `Command` class trigger some magic validation, but that behavior was not tested for that particular command (commenting out the argument doesn't trigger any failures in the test suite).

So I added a test for it. This seems like a small enough change that it doesn't require a ticket, but let me know if that's not the case.

Thanks!

### DIFF
--- a/tests/mail/test_sendtestemail.py
+++ b/tests/mail/test_sendtestemail.py
@@ -1,5 +1,5 @@
 from django.core import mail
-from django.core.management import call_command
+from django.core.management import CommandError, call_command
 from django.test import SimpleTestCase, override_settings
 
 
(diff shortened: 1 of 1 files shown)


nessita (approved):
Looks great, thank you!

I pushed a minor edit and a fix to docstring line wrapping. I will merge when CI passes.

bmispelon:
Excellent, thanks! ✨

PR #17784, opened 2024-01-27
Fixed#35137 - Interactive Collapsible error container

abhijeet-singh800:
This PR addresses - Ticket [#35137](https://code.djangoproject.com/ticket/35137) 

The change mainly include 
- Adding an Interactive button on the front of the Error Container (accessible using the keyboard)
- Changes in the Toggle Function in change the classes for error container and the button 
- approprite CSS for Hover messages and the button on based of the toggle function 

Note: This is my first PR so any guidance will be appriciated.


MHLut:
Somewhat related; this PR also deals with collapsible widgets, only in the Django admin instead of error reporting: https://github.com/django/django/pull/17910

knyghty:
Hmm this also seems like a good fit for `<datails>`/`<summary>` rather than extending the JS.

sarahboyce:
Hi @abhijeet-singh800 thank you for being patient and for raising a PR, welcome 👋 

I agree with @knyghty that this should be refactored to use the `<details>/<summary>` approach (see https://github.com/django/django/pull/17910) and you can probably add an assert to `tests.view_tests.tests.test_debug.DebugViewTests.test_technical_500` to cover this change.

It looks like all of the JavaScript in `django/views/templates/technical_500.html` is untested and ideally we'd have some selenium tests to cover this. This can be achieved in a separate PR, but will need to be done in this PR if we update the JavaScript.

I'm going to mark that `Patch needs improvement` on the ticket, please unset this when you're ready for another review after refactoring to use  `<details>/<summary>` and updated tests.

abhijeet-singh800:
Understood, will start working on this soon 🤠

sarahboyce:
Thank you for your work here @abhijeet-singh800 
Closing as this is being progressed here: https://github.com/django/django/pull/18198 :+1:

PR #17786, opened 2024-01-28
Refs #35058 -- Made centroid available on all geometry types.

smithdc1:
[Ticket #35058](https://code.djangoproject.com/ticket/35058)

 Centroid is available on all geometry types since GDAL 1.8.0. Previously it was restricted to Polygon.

 https://gdal.org/doxygen/classOGRGeometry.html#a91787f669b2a148169667e270e7e40df


felixxm (commented):
@smithdc1 Thanks :+1:

PR #17787, opened 2024-01-28
Fixed #35143 -- Improved accessibility of 404/500 debug pages.

MHLut:
Fixes [ticket #35143](https://code.djangoproject.com/ticket/35143).

## Notes

The HTTP 400 error page uses the same template as HTTP 500.

### Requested changes

**The page should have a header, main, and footer.**

Done. I did not fix the indentation within the `<main>` element in the HTTP 500 template because it would cause a shift in most lines. The new CSS is hence somewhat hacky.

The indentation is currently inconsistent and is best fixed in a separate ticket.

**Each table should have scope attributes on th elements.**

Done.

**Monospaced text in the patterns list should use a code element.**

Done for all the pattern items in the HTTP 404 template. I kept the CSS line that sets the list font to monospace, to prevent a visual change to the list marker font.

I have also wrapped the `Python Path` value in the HTTP 500 page in a code block, but couldn't get a working unit test for this scenario. (there is a test draft in the related commit message).

I left all other `<pre>` elements without `<code>` for future tickets.

**The small text within the h1 should use a small element instead of a span.**

Done. Since `<small>` looked different from the original styled span, I reused the former's CSS.

### DIFF
--- a/django/views/templates/technical_404.html
+++ b/django/views/templates/technical_404.html
@@ -9,9 +9,9 @@
     body * { padding:10px 20px; }
     body * * { padding:0; }
     body { font:small sans-serif; background:#eee; color:#000; }
-    body>div { border-bottom:1px solid #ddd; }
+    body > :where(header, main, footer) { border-bottom:1px solid #ddd; }
     h1 { font-weight:normal; margin-bottom:.4em; }
-    h1 span { font-size:60%; color:#666; font-weight:normal; }
+    h1 small { font-size:60%; color:#666; font-weight:normal; }
     table { border:none; border-collapse: collapse; width:100%; }
     td, th { vertical-align:top; padding:2px 3px; }
     th { width:12em; text-align:right; color:#666; padding-right:.5em; }
@@ -24,27 +24,28 @@
   </style>
 </head>
 <body>
-  <div id="summary">
-    <h1>Page not found <span>(404)</span></h1>
+  <header id="summary">
+    <h1>Page not found <small>(404)</small></h1>
     {% if reason and resolved %}<pre class="exception_value">{{ reason }}</pre>{% endif %}
     <table class="meta">
       <tr>
-        <th>Request Method:</th>
+        <th scope="row">Request Method:</th>
         <td>{{ request.META.REQUEST_METHOD }}</td>
       </tr>
       <tr>
-        <th>Request URL:</th>
+        <th scope="row">Request URL:</th>
         <td>{{ request.build_absolute_uri }}</td>
       </tr>
       {% if raising_view_name %}
       <tr>
-        <th>Raised by:</th>
+        <th scope="row">Raised by:</th>
         <td>{{ raising_view_name }}</td>
       </tr>
       {% endif %}
     </table>
-  </div>
-  <div id="info">
+  </header>
+
+  <main id="info">
     {% if urlpatterns %}
       <p>
       Using the URLconf defined in <code>{{ urlconf }}</code>,
@@ -54,8 +55,10 @@
         {% for pattern in urlpatterns %}
           <li>
             {% for pat in pattern %}
+              <code>
                 {{ pat.pattern }}
                 {% if forloop.last and pat.name %}[name='{{ pat.name }}']{% endif %}
+              </code>
             {% endfor %}
           </li>
         {% endfor %}
@@ -69,14 +72,14 @@
         {% if resolved %}matched the last one.{% else %}didn’t match any of these.{% endif %}
       </p>
     {% endif %}
-  </div>
+  </main>
 
-  <div id="explanation">
+  <footer id="explanation">
     <p>
       You’re seeing this error because you have <code>DEBUG = True</code> in
       your Django settings file. Change that to <code>False</code>, and Django
       will display a standard 404 page.
     </p>
-  </div>
+  </footer>
 </body>
 </html>
--- a/django/views/templates/technical_500.html
+++ b/django/views/templates/technical_500.html
@@ -10,7 +10,7 @@
     body * { padding:10px 20px; }
     body * * { padding:0; }
     body { font:small sans-serif; background-color:#fff; color:#000; }
-    body>div { border-bottom:1px solid #ddd; }
+    body > :where(header, main, footer) { border-bottom:1px solid #ddd; }
     h1 { font-weight:normal; }
     h2 { margin-bottom:.8em; }
     h3 { margin:1em 0 .5em 0; }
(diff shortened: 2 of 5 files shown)


MHLut:
Apologies for the multiple failed checks; I have the pre-commit running locally now.

django/views/templates/technical_500.html
```diff
@@ -195,11 +199,11 @@ <h3>{{ template_info.message|force_escape }}</h3>
       {% if template_info.bottom != template_info.total %} cut-bottom{% endif %}">
    {% for source_line in template_info.source_lines %}
    {% if source_line.0 == template_info.line %}
-   <tr class="error"><th>{{ source_line.0 }}</th>
+   <tr class="error"><th scope="row">{{ source_line.0 }}</th>
      <td>{{ template_info.before }}<span class="specific">{{ template_info.during }}</span>{{ template_info.after }}</td>
    </tr>
    {% else %}
-      <tr><th>{{ source_line.0 }}</th>
+      <tr><th scope="row">{{ source_line.0 }}</th>
```
felixxm, line 206: I don't think that the line number in templates can be treated as a header. IMO, we should change it to `<td>`.

django/views/templates/technical_500.html
```diff
@@ -195,11 +199,11 @@ <h3>{{ template_info.message|force_escape }}</h3>
       {% if template_info.bottom != template_info.total %} cut-bottom{% endif %}">
    {% for source_line in template_info.source_lines %}
    {% if source_line.0 == template_info.line %}
-   <tr class="error"><th>{{ source_line.0 }}</th>
+   <tr class="error"><th scope="row">{{ source_line.0 }}</th>
```
felixxm, line 202: I don't think that the line number in templates can be treated as a header. IMO, we should change it to `<td>`.

felixxm (commented):
@MHLut Thanks :+1: I left comments.

django/views/templates/technical_500.html
MHLut: That's a good point, and I agree, but that would leave the table in a heading-less state. Neither situation is preferable.

Changing anything more significant about this code would be out of the scope of this PR. What would you recommend we do? Maybe @thibaudcolas can chime in on this one?

django/views/templates/technical_500.html
MHLut: See https://github.com/django/django/pull/17787#discussion_r1469336676

tests/view_tests/tests/test_debug.py
```diff
@@ -238,7 +259,22 @@ def test_technical_500(self):
...
+        self.assertContains(
+            response,
+            '<main id="info">',
+            status_code=500,
+        )
+        self.assertContains(
+            response,
+            '<footer id="explanation">',
+            status_code=500,
+        )
```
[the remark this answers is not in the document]
MHLut: Done!

django/views/templates/technical_500.html
felixxm: As far as I'm aware, tables without headings are better than tables with incorrect ones.

felixxm: We could also display it in a single cell:
```html
<tr><td>{{ source_line.0 }}: {{ source_line.1 }}</td></tr>
```
or use `<ol start="..."><li>...</li></ol>` instead (like for tracebacks).

MHLut: I'm thinking about this but don't have a good answer yet.

A single-cell approach would turn this table into a list, which is not semantically correct.

(Solution 1) We could add a header row to this table with a heading cell for "Path" and "Error".

(Solution 2) If not, switching to an ordered list would be good, too.
However, the traceback feature is inaccessible, making me reluctant to introduce it now. (Fixing that would be a separate ticket)

MHLut: Screenshot 2024-01-29 at 11 29 07
Quick mock-up of solution 1. 

I'll leave it with this screenshot for now. It does not feel good to me to replace the `<th>` with a `<td>`, but I also don't know enough to decide the best course of action.

Since Thibaud proposed this change and is part of the accessibility team, I would also like to hear his opinion.

If you rather go ahead: say the word, and I'll make the change so you can merge it.

thibaudcolas (changes requested):

django/views/templates/technical_500.html
felixxm: Convinced.

MHLut: Alright, I'm leaving it as it is now, then.

django/views/templates/technical_500.html
MHLut: Leaving it as-is per the discussion on the other, similar table above.

django/views/templates/technical_500.html
felixxm: Yes, please do.

django/views/templates/technical_404.html
```diff
@@ -54,8 +55,8 @@ <h1>Page not found <span>(404)</span></h1>
         {% for pattern in urlpatterns %}
           <li>
             {% for pat in pattern %}
-                {{ pat.pattern }}
-                {% if forloop.last and pat.name %}[name='{{ pat.name }}']{% endif %}
+              <code>{{ pat.pattern }}</code>
+              {% if forloop.last and pat.name %}<code>[name='{{ pat.name }}']</code>{% endif %}
             {% endfor %}
```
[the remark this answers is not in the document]
MHLut: I wrapped the content in a single code element.

MHLut:
As far as I can see, I've resolved all the issues we've found in the last two days.

django/views/templates/technical_404.html
```diff
@@ -9,9 +9,9 @@
     body * { padding:10px 20px; }
     body * * { padding:0; }
     body { font:small sans-serif; background:#eee; color:#000; }
-    body>div { border-bottom:1px solid #ddd; }
+    body > :where(header, main, footer) { border-bottom:1px solid #ddd; }
```
felixxm, line 12: We should do the same in `technical_500.html`.

MHLut: Thanks, done!

felixxm:
@MHLut Thanks for updates :+1:

MHLut:
Note: The Selenium test that failed is `test_inline_add_another_widgets (admin_views.test_autocomplete_view.SeleniumTests.test_inline_add_another_widgets)`.

As far as I know, my changes should not touch the admin, and I'm not sure where this is coming from.

PR #17792, opened 2024-01-29
Fixed #31405 -- Added LoginRequiredAuthenticationMiddleware force all views to require authentication by default.

Hisham-Pak:
Ticket: https://code.djangoproject.com/ticket/31405
Mailing list: https://groups.google.com/forum/#!topic/django-developers/PUQQUHIxEXQ
Follows: #12632

Any suggestions are welcome.

### DIFF
--- a/django/contrib/admin/sites.py
+++ b/django/contrib/admin/sites.py
@@ -7,11 +7,12 @@ from django.contrib.admin import ModelAdmin, actions
 from django.contrib.admin.exceptions import AlreadyRegistered, NotRegistered
 from django.contrib.admin.views.autocomplete import AutocompleteJsonView
 from django.contrib.auth import REDIRECT_FIELD_NAME
+from django.contrib.auth.decorators import login_not_required
 from django.core.exceptions import ImproperlyConfigured
 from django.db.models.base import ModelBase
 from django.http import Http404, HttpResponsePermanentRedirect, HttpResponseRedirect
 from django.template.response import TemplateResponse
-from django.urls import NoReverseMatch, Resolver404, resolve, reverse
+from django.urls import NoReverseMatch, Resolver404, resolve, reverse, reverse_lazy
 from django.utils.decorators import method_decorator
 from django.utils.functional import LazyObject
 from django.utils.module_loading import import_string
@@ -259,6 +260,8 @@ class AdminSite:
                 return self.admin_view(view, cacheable)(*args, **kwargs)
 
             wrapper.admin_site = self
+            # Used by LoginRequiredMiddleware.
+            wrapper.login_url = reverse_lazy("admin:login", current_app=self.name)
             return update_wrapper(wrapper, view)
 
         # Admin-site-wide views.
@@ -402,6 +405,7 @@ class AdminSite:
         return LogoutView.as_view(**defaults)(request)
 
     @method_decorator(never_cache)
+    @login_not_required
     def login(self, request, extra_context=None):
         """
         Display the login form for the given HttpRequest.
--- a/django/contrib/auth/checks.py
+++ b/django/contrib/auth/checks.py
@@ -4,8 +4,25 @@ from types import MethodType
 from django.apps import apps
 from django.conf import settings
 from django.core import checks
+from django.utils.module_loading import import_string
 
 from .management import _get_builtin_permissions
+
+
+def _subclass_index(class_path, candidate_paths):
+    """
+    Return the index of dotted class path (or a subclass of that class) in a
+    list of candidate paths. If it does not exist, return -1.
+    """
+    cls = import_string(class_path)
+    for index, path in enumerate(candidate_paths):
+        try:
+            candidate_cls = import_string(path)
+            if issubclass(candidate_cls, cls):
+                return index
+        except (ImportError, TypeError):
+            continue
+    return -1
 
 
 def check_user_model(app_configs=None, **kwargs):
@@ -217,4 +234,29 @@ def check_models_permissions(app_configs=None, **kwargs):
                 )
             codenames.add(codename)
 
+    return errors
+
+
+def check_middleware(app_configs, **kwargs):
+    errors = []
+
+    login_required_index = _subclass_index(
+        "django.contrib.auth.middleware.LoginRequiredMiddleware",
+        settings.MIDDLEWARE,
+    )
+
+    if login_required_index != -1:
+        auth_index = _subclass_index(
+            "django.contrib.auth.middleware.AuthenticationMiddleware",
+            settings.MIDDLEWARE,
+        )
+        if auth_index == -1 or auth_index > login_required_index:
+            errors.append(
+                checks.Error(
+                    "In order to use django.contrib.auth.middleware."
+                    "LoginRequiredMiddleware, django.contrib.auth.middleware."
+                    "AuthenticationMiddleware must be defined before it in MIDDLEWARE.",
+                    id="auth.E013",
+                )
+            )
     return errors
--- a/django/contrib/auth/decorators.py
+++ b/django/contrib/auth/decorators.py
@@ -60,6 +60,10 @@ def user_passes_test(
                     return view_func(request, *args, **kwargs)
                 return _redirect_to_login(request)
 
+        # Attributes used by LoginRequiredMiddleware.
+        _view_wrapper.login_url = login_url
+        _view_wrapper.redirect_field_name = redirect_field_name
+
         return wraps(view_func)(_view_wrapper)
 
     return decorator
@@ -80,6 +84,14 @@ def login_required(
     if function:
         return actual_decorator(function)
     return actual_decorator
+
+
+def login_not_required(view_func):
+    """
+    Decorator for views that allows access to unauthenticated requests.
+    """
+    view_func.login_required = False
+    return view_func
 
 
 def permission_required(perm, login_url=None, raise_exception=False):
--- a/django/contrib/auth/middleware.py
+++ b/django/contrib/auth/middleware.py
@@ -1,9 +1,13 @@
 from functools import partial
+from urllib.parse import urlparse
 
+from django.conf import settings
 from django.contrib import auth
-from django.contrib.auth import load_backend
+from django.contrib.auth import REDIRECT_FIELD_NAME, load_backend
 from django.contrib.auth.backends import RemoteUserBackend
+from django.contrib.auth.views import redirect_to_login
 from django.core.exceptions import ImproperlyConfigured
+from django.shortcuts import resolve_url
 from django.utils.deprecation import MiddlewareMixin
 from django.utils.functional import SimpleLazyObject
 
@@ -32,6 +36,56 @@ class AuthenticationMiddleware(MiddlewareMixin):
             )
         request.user = SimpleLazyObject(lambda: get_user(request))
         request.auser = partial(auser, request)
+
+
+class LoginRequiredMiddleware(MiddlewareMixin):
+    """
+    Middleware that redirects all unauthenticated requests to a login page.
+
+    Views using the login_not_required decorator will not be redirected.
+    """
+
+    redirect_field_name = REDIRECT_FIELD_NAME
+
+    def process_view(self, request, view_func, view_args, view_kwargs):
+        if request.user.is_authenticated:
+            return None
+
+        if not getattr(view_func, "login_required", True):
+            return None
+
+        return self.handle_no_permission(request, view_func)
+
+    def get_login_url(self, view_func):
+        login_url = getattr(view_func, "login_url", None) or settings.LOGIN_URL
+        if not login_url:
+            raise ImproperlyConfigured(
+                "No login URL to redirect to. Define settings.LOGIN_URL or "
+                "provide a login_url via the 'django.contrib.auth.decorators."
+                "login_required' decorator."
+            )
+        return str(login_url)
+
+    def get_redirect_field_name(self, view_func):
+        return getattr(view_func, "redirect_field_name", self.redirect_field_name)
+
+    def handle_no_permission(self, request, view_func):
+        path = request.build_absolute_uri()
+        resolved_login_url = resolve_url(self.get_login_url(view_func))
+        # If the login url is the same scheme and net location then use the
+        # path as the "next" url.
+        login_scheme, login_netloc = urlparse(resolved_login_url)[:2]
+        current_scheme, current_netloc = urlparse(path)[:2]
+        if (not login_scheme or login_scheme == current_scheme) and (
+            not login_netloc or login_netloc == current_netloc
+        ):
+            path = request.get_full_path()
+
+        return redirect_to_login(
+            path,
+            resolved_login_url,
+            self.get_redirect_field_name(view_func),
+        )
 
 
 class RemoteUserMiddleware(MiddlewareMixin):
--- a/docs/ref/checks.txt
+++ b/docs/ref/checks.txt
@@ -868,6 +868,10 @@ The following checks are performed on the default
   for its builtin permission names to be at most 100 characters.
 * **auth.E012**: The permission codenamed ``<codename>`` of model ``<model>``
   is longer than 100 characters.
+* **auth.E013**: In order to use
+  :class:`django.contrib.auth.middleware.LoginRequiredMiddleware`,
+  :class:`django.contrib.auth.middleware.AuthenticationMiddleware` must be
+  defined before it in MIDDLEWARE.
 
 ``contenttypes``
 ----------------
--- a/docs/ref/middleware.txt
+++ b/docs/ref/middleware.txt
@@ -495,6 +495,58 @@ Adds the ``user`` attribute, representing the currently-logged-in user, to
 every incoming ``HttpRequest`` object. See :ref:`Authentication in web requests
 <auth-web-requests>`.
 
+.. class:: LoginRequiredMiddleware
+
+.. versionadded:: 5.1
+
+Redirects all unauthenticated requests to a login page. For admin views, this
+redirects to the admin login. For all other views, this will redirect to
+:setting:`settings.LOGIN_URL <LOGIN_URL>`. This can be customized by using the
+:func:`~.django.contrib.auth.decorators.login_required` decorator and setting
+``login_url`` or ``redirect_field_name`` for the view. For example::
+
+    @method_decorator(
+        login_required(login_url="/login/", redirect_field_name="redirect_to"),
+        name="dispatch",
+    )
+    class MyView(View):
+        pass
+
+
+    @login_required(login_url="/login/", redirect_field_name="redirect_to")
+    def my_view(request): ...
+
+Views using the :func:`~django.contrib.auth.decorators.login_not_required`
+decorator are exempt from this requirement.
+
+.. admonition:: Ensure that your login view does not require a login.
+
+    To prevent infinite redirects, ensure you have
+    :ref:`enabled unauthenticated requests
+    <disable-login-required-middleware-for-views>` to your login view.
+
+**Methods and Attributes**
+
+.. attribute:: redirect_field_name
+
+    Defaults to ``"next"``.
+
+.. method:: get_login_url()
+
+    Returns the URL that unauthenticated requests will be redirected to. If
+    defined, this returns the ``login_url`` set on the
+    :func:`~.django.contrib.auth.decorators.login_required` decorator. Defaults
+    to :setting:`settings.LOGIN_URL <LOGIN_URL>`.
+
+.. method:: get_redirect_field_name()
+
+    Returns the name of the query parameter that contains the URL the user
+    should be redirected to after a successful login. If defined, this returns
+    the ``redirect_field_name`` set on the
+    :func:`~.django.contrib.auth.decorators.login_required` decorator. Defaults
+    to :attr:`redirect_field_name`. If ``None`` is returned, a query parameter
+    won't be added.
+
 .. class:: RemoteUserMiddleware
 
 Middleware for utilizing web server provided authentication. See
@@ -596,6 +648,12 @@ Here are some hints about the ordering of various Django middleware classes:
 #. :class:`~django.contrib.auth.middleware.AuthenticationMiddleware`
 
    After ``SessionMiddleware``: uses session storage.
+
+#. :class:`~django.contrib.auth.middleware.LoginRequiredMiddleware`
+
+   .. versionadded:: 5.1
+
+   After ``AuthenticationMiddleware``: uses user object.
 
 #. :class:`~django.contrib.messages.middleware.MessageMiddleware`
 
--- a/docs/ref/settings.txt
+++ b/docs/ref/settings.txt
@@ -3060,8 +3060,9 @@ Default: ``'/accounts/login/'``
 The URL or :ref:`named URL pattern <naming-url-patterns>` where requests are
 redirected for login when using the
 :func:`~django.contrib.auth.decorators.login_required` decorator,
-:class:`~django.contrib.auth.mixins.LoginRequiredMixin`, or
-:class:`~django.contrib.auth.mixins.AccessMixin`.
+:class:`~django.contrib.auth.mixins.LoginRequiredMixin`,
+:class:`~django.contrib.auth.mixins.AccessMixin`, or when
+:class:`~django.contrib.auth.middleware.LoginRequiredMiddleware` is installed.
 
 .. setting:: LOGOUT_REDIRECT_URL
 
--- a/docs/releases/5.1.txt
+++ b/docs/releases/5.1.txt
@@ -26,6 +26,20 @@ only officially support the latest release of each series.
 What's new in Django 5.1
 ========================
 
+Middleware to require authentication by default
+-----------------------------------------------
+
+The new :class:`~django.contrib.auth.middleware.LoginRequiredMiddleware`
+redirects all unauthenticated requests to a login page. Views can allow
+unauthenticated requests by using the new
+:func:`~django.contrib.auth.decorators.login_not_required` decorator.
+
+The :class:`~django.contrib.auth.middleware.LoginRequiredMiddleware` respects
+the ``login_url`` and ``redirect_field_name`` values set via the
+:func:`~.django.contrib.auth.decorators.login_required` decorator, but does not
+support setting ``login_url`` or ``redirect_field_name`` via the
+:class:`~django.contrib.auth.mixins.LoginRequiredMixin`.
+
 Minor features
 --------------
 
--- a/docs/topics/auth/default.txt
+++ b/docs/topics/auth/default.txt
@@ -656,7 +656,24 @@ inheritance list.
     ``is_active`` flag on a user, but the default
     :setting:`AUTHENTICATION_BACKENDS` reject inactive users.
 
+.. _disable-login-required-middleware-for-views:
+
 .. currentmodule:: django.contrib.auth.decorators
+
+The ``login_not_required`` decorator
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 5.1
+
+When :class:`~django.contrib.auth.middleware.LoginRequiredMiddleware` is
+installed, all views require authentication by default. Some views, such as the
+login view, may need to disable this behavior.
+
+.. function:: login_not_required()
+
+    Allows unauthenticated requests without redirecting to the login page when
+    :class:`~django.contrib.auth.middleware.LoginRequiredMiddleware` is
+    installed.
 
 Limiting access to logged-in users that pass a test
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/tests/auth_tests/test_middleware.py
+++ b/tests/auth_tests/test_middleware.py
@@ -1,8 +1,14 @@
-from django.contrib.auth.middleware import AuthenticationMiddleware
+from django.conf import settings
+from django.contrib.auth import REDIRECT_FIELD_NAME
+from django.contrib.auth.middleware import (
+    AuthenticationMiddleware,
+    LoginRequiredMiddleware,
+)
 from django.contrib.auth.models import User
 from django.core.exceptions import ImproperlyConfigured
 from django.http import HttpRequest, HttpResponse
-from django.test import TestCase
+from django.test import TestCase, modify_settings, override_settings
+from django.urls import reverse
 
 
 class TestAuthenticationMiddleware(TestCase):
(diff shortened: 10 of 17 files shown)


django/contrib/auth/middleware.py
```diff
@@ -34,6 +38,41 @@ def process_request(self, request):
         request.auser = partial(auser, request)
 
 
+class LoginRequiredAuthenticationMiddleware(AuthenticationMiddleware):
```
Hisham-Pak, line 41: Should this even inherit from AuthenticationMiddleware?

Hisham-Pak:
why does `make black` expects (at all code blocks):
```
class MyView(TestMixin1, TestMixin2, View): ...
```
instead of:
```
class MyView(TestMixin1, TestMixin2, View):
    ...
```

nessita:
> why does `make black` expects (at all code blocks):
> 
> ```
> class MyView(TestMixin1, TestMixin2, View): ...
> ```
> 
> instead of accepting:
> 
> ```
> class MyView(TestMixin1, TestMixin2, View):
>     ...
> ```

I believe the cause of this is that the formatting tool Black has been recently updated to its 2024 style. The Django code has been updated as well, so if you rebase your branch onto main (`git rebase main` with no uncommitted changes) you will get the necessary fixes.

Hisham-Pak:
My bad, thanks for the fix!

mdisec:
Amazing work @Hisham-Pak !! Couple of weeks ago friend of mine remind me this.. I was literally thinking about coming back and getting the work done on this PR.

Hisham-Pak:
Thank you for the kind words. Appreciate the effort you put in!

django/contrib/auth/middleware.py
Hisham-Pak: Now inherits MiddlewareMixin not AuthenticationMiddleware

sarahboyce (commented):
Thank you for this @Hisham-Pak 👍 
This is looking good, I added a few comments.

As a side note, we should keep an eye on the async work that @bigfootjon is doing (PR: https://github.com/django/django/pull/18036/ discussion: https://forum.djangoproject.com/t/asyncifying-django-contrib-auth-and-signals-and-maybe-sessions/18770/26). This might require some updates here 👍

tests/auth_tests/test_middleware.py
```diff
@@ -50,3 +57,79 @@ async def test_auser(self):
         self.assertEqual(auser, self.user)
         auser_second = await self.request.auser()
         self.assertIs(auser, auser_second)
+
+
+class TestLoginRequiredMiddleware(TestCase):
+    class EmptyResponseBaseView(View):
+        def get(self, request, *args, **kwargs):
+            return HttpResponse()
```
bigfootjon, line 65: I think it would be worthwhile to have a few `async` test cases as well

django/contrib/auth/middleware.py
```diff
@@ -34,6 +38,49 @@ def process_request(self, request):
         request.auser = partial(auser, request)
 
 
+class LoginRequiredMiddleware(MiddlewareMixin):
```
bigfootjon, line 41: Using `MiddlewareMixin` requires a context-switch in async code. Considering that `contrib.auth` has an async interface (as applied here: you can use `await request.auser()` to get the current user, instead of `request.user`). can this middleware be rewritten to not require a context switch in async code?

Ref: https://github.com/django/django/pull/18036/commits/2c59d74e56ea6364274a5780d6d4ed32f25ce6e3

(note that that code is untested, but hopefully that demonstrates what I'm talking about)

bigfootjon (changes requested):
@sarahboyce thanks for tagging me! I added a few comments inline. The test cases are essential, but the `Middleware` reworking is nice-to-have.

hermann-wenninger:
a

django/contrib/auth/middleware.py
bigfootjon: I've updated the above commit reference after I added testing and cleaned up the implementation.

django/contrib/auth/mixins.py
```diff
@@ -57,6 +56,8 @@ def handle_no_permission(self):
             not login_netloc or login_netloc == current_netloc
         ):
             path = self.request.get_full_path()
+        from django.contrib.auth.views import redirect_to_login
```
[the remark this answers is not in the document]
Hisham-Pak: We already do a local import to avoid circular import [here](https://github.com/django/django/blob/549320946dfdaac6ccb6c0d3150c46db92637187/django/contrib/auth/decorators.py#L34). We can also move `redirect_to_login` to `utils.py` and follow a [similar](https://github.com/django/django/commit/98e05ccde440cc9b768952cc10bc8285f4924e1f#diff-76917634c6c088f56f8dec9493294c657953e61b01e38e33b02d876d5e96dd3aR7-R8) approach for backwards compatibility. What do you prefer?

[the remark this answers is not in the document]
sarahboyce: Another instance of avoiding a circular import sounds like more reason to move `redirect_to_login` but this change can come in a separate PR.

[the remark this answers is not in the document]
Hisham-Pak: Moved it [here](https://github.com/django/django/pull/18072)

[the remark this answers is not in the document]
sarahboyce: Commented on the PR, but I checked with other contributors is they thought this was the right approach. As this is a change to a documented stable API, decided the benefit is not strong enough to justify the depreciation 
But thank you again for your efforts here, very appreciated

django/contrib/auth/middleware.py
```diff
@@ -34,6 +39,68 @@ def process_request(self, request):
...
+
+        try:
+            if request.path.startswith(reverse("admin:index")):
+                return redirect_to_login(
+                    request.get_full_path(), reverse("admin:login")
+                )
+        except NoReverseMatch:
+            pass
+
+        return redirect_to_login(request.get_full_path())
```
bigfootjon, line 101: Defining `process_view` will still result in context switches :(

Ref this code:

https://github.com/django/django/blob/main/django/core/handlers/base.py#L77-L81

I think you need to rename this function to something else (so it doesn't get caught by that auto-rewrite code) and then call it from `__call__` / `__acall__`

bigfootjon (changes requested):
(just a friendly reminder, when your PR is ready for review uncheck the "patch needs improvement" checkbox on the ticket)

django/contrib/auth/middleware.py
Hisham-Pak: Yes, I thought nothing in `process_view` was io bound so it didn't really matter much if it got wrapped in `sync_to_async`? 

The way I see to achieve no context switch here is to have our local [resolve_request](https://github.com/django/django/blob/32d70b2f55b1f74736fd11bc8efce890ad5fa2f0/django/core/handlers/base.py#L300). But I don't know if its worth it? What do you think?

bigfootjon: > Yes, I thought nothing in process_view was io bound so it didn't really matter much if it got wrapped in sync_to_async?

Wrapping in sync_to_async still causes a context switch / performance hit :(

> Yes, I thought nothing in process_view was io bound so it didn't really matter much if it got wrapped in sync_to_async?

The way I see to achieve no context switch here is to have our local [resolve_request](https://github.com/django/django/blob/32d70b2f55b1f74736fd11bc8efce890ad5fa2f0/django/core/handlers/base.py#L300). But I don't know if its worth it? What do you think?

I don't think that's necessary, I think you can just change the name to *not* hit any of the names that will be rewritten and then call this function from `__call__` / `__acall__`.

Hisham-Pak: Yes, but the thing is `__call__`/`__acall__` is not provided a `view_func`/`callback` it is only passed to `process_view` [here](https://github.com/django/django/blob/32d70b2f55b1f74736fd11bc8efce890ad5fa2f0/django/core/handlers/base.py#L186) only after [resolving the request](https://github.com/django/django/blob/32d70b2f55b1f74736fd11bc8efce890ad5fa2f0/django/core/handlers/base.py#L181).

So the way forward I see is:
1) Using a private resolve_request (this gets called twice here and in handler)
2) Changing how handler works to support native async (not using same `process_view` name for both sync and async)
3) Finding a way to not use `view_func`/`callback` at all (I didn't look much into it yet)

Am I missing something?

bigfootjon: > Changing how handler works to support native async (not using same process_view name for both sync and async)

I'm kinda in favor of this idea, but I can understand your hesitancy towards doing it. I think for now we can leave this as you have implemented in this PR and I'll think about how we can solve it better in the general case.

bigfootjon (approved):
Async code LGTM, and thanks for adding the test cases. I leave the rest of the review to @sarahboyce et al

tests/auth_tests/urls.py
```diff
@@ -198,7 +248,18 @@ def get_default_redirect_url(self):
...
+    path("protected_view/", ProtectedView.as_view(), name="protected_view"),
+    path("protected_function_view/", protected_view, name="protected_function_view"),
+    path("protected_async_view/", ProtectedAsyncView.as_view(), name="protected_view"),
+    path(
+        "protected_async_function_view/",
+        protected_async_view,
+        name="protected_function_view",
+    ),
     path("setlang/", set_language, name="set_language"),
-    # This line is only required to render the password reset with is_admin=True
```
sarahboyce, line 202: Please revert this line change.

django/contrib/auth/middleware.py
```diff
@@ -34,6 +38,49 @@ def process_request(self, request):
...
+        try:
+            admin_url = reverse("admin:index")
+            if request.path.startswith(admin_url):
+                return HttpResponseRedirect(
+                    reverse("admin:login") + "?next=" + request.get_full_path()
+                )
+        except NoReverseMatch:
+            pass
+
+        return redirect_to_login(request.get_full_path())
```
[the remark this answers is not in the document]
sarahboyce: @Hisham-Pak thank you for adding the tests here, I still don't think the redirect behviour is quite what we want.

I think we should have that this middleware has the exact same behaviour as using the `LoginRequiredMixin`, this redirect behaviour differs currently (I think we should reuse the `AccessMixin.handle_no_permission`)
I being able to update the `login_url` on a particular view is also nice and maybe expected if this middleware gives the same behaviour of having `LoginRequiredMixin` everywhere.

django/contrib/auth/mixins.py
```diff
@@ -73,6 +74,15 @@ def dispatch(self, request, *args, **kwargs):
         return super().dispatch(request, *args, **kwargs)
 
 
+class LoginNotRequiredMixin:
+    """
+    Mixin for CBV that marks that the view is accessible by unauthenticated
+    users.
+    """
```
adamchainz, line 81: Follow the pattern of the other mixins and use a simple sentence like so:

```suggestion
    """
    Mark the view as accessible by unauthenticated users.
    """
```

django/contrib/auth/middleware.py
adamchainz: +1 for changing the handler, but in a separate ticket and PR.

django/contrib/auth/middleware.py
```diff
@@ -34,6 +39,68 @@ def process_request(self, request):
...
+                "before "
+                "'django.contrib.auth.middleware.LoginRequiredMiddleware'."
+            )
+        if self.async_mode:
+            return self.__acall__(request)
+        self.user_is_authenticated = request.user.is_authenticated
+        return self.get_response(request)
+
+    async def __acall__(self, request):
+        self.user_is_authenticated = (await request.auser()).is_authenticated
```
adamchainz, line 79: The current approach is insecure. Storing values on `self` allows those values to be shared across threads and async tasks. This leads to the possibility that two concurrent requests to a threaded/async server will mix up the “is authenticated” values and incorrectly show a private page.

Please change the middleware to inherit from `MiddlewareMixin`, drop the custom `__init__`, `__call__` and `__acall__`, and only implement `process_view`.

django/contrib/auth/middleware.py
```diff
@@ -34,6 +39,68 @@ def process_request(self, request):
...
+    def __call__(self, request):
+        if not hasattr(request, "user"):
+            raise ImproperlyConfigured(
+                "The Django login required middleware requires authentication "
+                "middleware to be installed. Edit your MIDDLEWARE setting to "
+                "insert "
+                "'django.contrib.auth.middleware.AuthenticationMiddleware' "
+                "before "
+                "'django.contrib.auth.middleware.LoginRequiredMiddleware'."
+            )
```
adamchainz, line 72: I don't think we need this error. We don’t do equivalent checks in `LoginRequiredMixin`, for example.

To defend against the case where the auth middleware is missing or out of order, it would be better done in a system check. That way there won’t be any extra work on the per-request path. The admin system checks have a small example of checking the middleware setting.

adamchainz (changes requested):
Thanks for taking this on. This is a small feature I’ve wanted to see for a while.

The current approach is insecure for some server configurations - please see my comment on `__call__`.

django/contrib/auth/mixins.py
```diff
@@ -73,6 +74,15 @@ def dispatch(self, request, *args, **kwargs):
         return super().dispatch(request, *args, **kwargs)
 
 
+class LoginNotRequiredMixin:
```
angusholder, line 77: I might be missing something, but what's the reason for also having a mixin class? It seems like the `@login_not_required` decorator will work perfectly well for class-based views too no?

django/contrib/auth/middleware.py
Hisham-Pak: Thanks for confirming! Reverted async stuff in https://github.com/django/django/pull/17792/commits/3dd63ed05bef5eabf016375af9c2be53a200f62d

tests/auth_tests/urls.py
Hisham-Pak: This statement is not true any more because we are testing admin redirect now. Should we remove that? Or can we go something along?
```
# This line is required to render the password reset with is_admin=True and for testing admin redirects under login required middleware
```

django/contrib/auth/middleware.py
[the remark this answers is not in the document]
Hisham-Pak: I made middleware behave like `LoginRequiredMixin` in https://github.com/django/django/commit/3dd63ed05bef5eabf016375af9c2be53a200f62d. I also made decorators set `login_url` value (which is what mixin does on view) so they can be easily changed on FBVs. What do you think?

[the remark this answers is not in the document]
sarahboyce: Thank you!

tests/auth_tests/urls.py
sarahboyce: Understood 👍

django/contrib/admin/sites.py
```diff
@@ -259,6 +260,9 @@ def wrapper(*args, **kwargs):
                 return self.admin_view(view, cacheable)(*args, **kwargs)
 
             wrapper.admin_site = self
+            # login_url attribute is used by LoginRequiredMiddleware so it
+            # redirects to admin login.
+            wrapper.login_url = reverse_lazy("admin:login")
```
sarahboyce, line 265:
```suggestion
            wrapper.login_url = reverse("admin:login", current_app=self.name)
```
(also wondering if we should reference `login_url` above in the `redirect_to_login` call 🤔

docs/ref/middleware.txt
```diff
@@ -495,6 +495,19 @@ Adds the ``user`` attribute, representing the currently-logged-in user, to
...
+.. versionadded:: 5.1
+
+By default, this middleware mandates login for all views. However, views
+employing the ``login_not_required`` decorator or the
+``LoginNotRequiredMixin`` mixin are exempt from this requirement. In the
+absence of these, the middleware will guide the user to the login page or to
+a custom login page if the view has defined ``login_url`` or
+``redirect_field_name`` attributes. This is typically seen when using the
+``login_required`` decorator or the ``LoginRequiredMixin`` mixin and
+modifying the default behavior.
```
sarahboyce, line 509: Maybe something like:
```
Redirects all requests by non-authenticated users to a login page. For admin 
views, this will redirect to the admin login. For all other views, this will 
redirect to :setting:`settings.LOGIN_URL <LOGIN_URL>`. This can be customized 
by setting ``login_url`` or ``redirect_field_name`` attributes on your view.
```
And then we can insert an example of doing this for a class based and function based view. Something like

```python
class MyView(View):
     login_url = "/login/"
     redirect_field_name = "redirect_to"
```
and....? Not sure if this is right but
```python
@login_required(login_url="/accounts/login/")
def my_view(request): ...
```
and then finish with
```
Views using the :func:`~django.contrib.auth.decorators.login_not_required` 
decorator or the :class:`~django.contrib.auth.mixins.LoginNotRequiredMixin` 
mixin are exempt from this requirement.
```

docs/releases/5.1.txt
```diff
@@ -45,6 +45,11 @@ Minor features
 * The default iteration count for the PBKDF2 password hasher is increased from
   720,000 to 870,000.
 
+* The :class:`~django.contrib.auth.middleware.LoginRequiredMiddleware` is now
+  available. It forces all views to require login by default. The
+  ``LoginNotRequiredMixin`` mixin and ``login_not_required`` decorator have
+  also been added.
```
sarahboyce, line 51:
```suggestion
  :class:`~django.contrib.auth.mixins.LoginNotRequiredMixin` mixin and 
  :func:`~django.contrib.auth.decorators.login_not_required` decorator have 
  also been added.
```

docs/topics/auth/default.txt
```diff
@@ -763,6 +763,48 @@ email in the desired domain and if not, redirects to the login page::
 
 .. currentmodule:: django.contrib.auth.decorators
 
+The ``login_not_required`` decorator
```
sarahboyce, line 766: Let's move both of these to be just under `The ``LoginRequiredMixin`` mixin` section

sarahboyce (commented):
Thank you for the updates! 
I have a few more comments, can you also rebase and squash the commits? ⭐

django/contrib/admin/sites.py
Hisham-Pak: We need to lazily reverse because url configuration is not loaded yet. I added the missing argument though.

django/contrib/auth/mixins.py
Hisham-Pak: I believe it largely comes down to personal preferences.

django/contrib/auth/middleware.py
```diff
@@ -34,6 +38,100 @@ def process_request(self, request):
...
+        if (not login_scheme or login_scheme == current_scheme) and (
+            not login_netloc or login_netloc == current_netloc
+        ):
+            path = request.get_full_path()
+
+        return redirect_to_login(
+            path,
+            resolved_login_url,
+            self.get_redirect_field_name(view_func),
+        )
```
Hisham-Pak, line 132: test these

django/contrib/auth/middleware.py
Hisham-Pak: Added it to system check.

docs/ref/middleware.txt
```diff
@@ -495,6 +495,30 @@ Adds the ``user`` attribute, representing the currently-logged-in user, to
...
+        redirect_field_name = "redirect_to"
+
+
+    @login_required(redirect_field_name="redirect_to", login_url="/login/")
+    def my_view(request): ...
+
+Views using the :func:`~django.contrib.auth.decorators.login_not_required` 
+decorator or the :class:`~django.contrib.auth.mixins.LoginNotRequiredMixin` 
+mixin are exempt from this requirement.
+
```
sarahboyce, line 521: We potentially want a warning here that the login url (either set on the middleware or the settings) must not require a login itself, otherwise there is an infinite redirect loop
It _should_ be self explanatory but maybe we should call it out in a docs warning 🤔 what do we think?

docs/releases/5.1.txt
```diff
@@ -63,6 +63,12 @@ Minor features
   accessibility of the
   :class:`~django.contrib.auth.forms.UserChangeForm`.
 
+* The :class:`~django.contrib.auth.middleware.LoginRequiredMiddleware` is now
+  available. It forces all views to require login by default. The
+  :class:`~django.contrib.auth.mixins.LoginNotRequiredMixin` mixin and
+  :func:`~django.contrib.auth.decorators.login_not_required` decorator have
+  also been added.
```
sarahboyce, line 70: @nessita do you think this should be a headline feature?

sarahboyce (commented):
Thank you so much for this @Hisham-Pak
I _think_ we are basically there 🥳💃

I have pushed very small updates and I have a couple small remaining thoughts/questions about the docs.

docs/ref/middleware.txt
Hisham-Pak: I agree, in addition to middleware or the settings it can also be set on views. But I think a general warning to not require a login on login views is okay?

docs/ref/checks.txt
```diff
@@ -868,6 +868,8 @@ The following checks are performed on the default
   for its builtin permission names to be at most 100 characters.
 * **auth.E012**: The permission codenamed ``<codename>`` of model ``<model>``
   is longer than 100 characters.
+* **auth.E013**: The login required middleware requires authentication
+  middleware to be installed, and come before the login required middleware.
```
nessita, line 872: This feels a tiny odd, what about:
```suggestion
  middleware to be installed, and should be defined before the login required middleware.
```

docs/ref/middleware.txt
```diff
@@ -495,6 +495,30 @@ Adds the ``user`` attribute, representing the currently-logged-in user, to
 every incoming ``HttpRequest`` object. See :ref:`Authentication in web requests
 <auth-web-requests>`.
 
+.. class:: LoginRequiredMiddleware
+
+.. versionadded:: 5.1
+
+Redirects all requests by unauthenticated users to a login page. For admin
```
nessita, line 502: I would suggest:
```suggestion
Redirects all unauthenticated requests to a login page. For admin
```

docs/releases/5.1.txt
nessita: Yes, I think this would be a nice adding to the `What's new in Django 5.1` section, good thinking Sarah! Thank you for proposing it.

docs/ref/checks.txt
sarahboyce: Just so you know where this is coming from, this is model-ed off https://github.com/django/django/blob/f030236a86a64a4befd3cc8093e2bbeceef52a31/django/contrib/messages/storage/session.py#L17

Shall I still update the wording or keep it to be the same format as this other error

docs/ref/middleware.txt
sarahboyce: Added something here 👍

docs/topics/auth/default.txt
```diff
@@ -633,7 +633,7 @@ inheritance list.
 
 .. class:: LoginRequiredMixin
 
-    If a view is using this mixin, all requests by non-authenticated users will
+    If a view is using this mixin, all requests by unauthenticated users will
```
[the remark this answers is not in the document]
sarahboyce: Yes, reverting for now 👍

docs/ref/checks.txt
nessita: My proposal would be to update the error message in this PR (both in the exception raising code and in the docs), since it reads awkwardly, and follow up with a separated PR to fix phrasing for the session middleware. What do you think?

adamchainz:
I suggest we do **not** support class-based attributes on views, as currently defined in `LoginNotRequiredMixin`, and as @angusholder [previously queried](https://github.com/django/django/pull/17792/files#r1581904420).

Supporting the class-level attributes adds significant complexity to the middleware implementation, as it has to look in two places on views for all of the attributes (`login_required`, `login_url`, redirect_field_name`). This makes it harder to understand the middleware, as well as any class-based views that might combine both options.

Instead, I think we should only support view-function-level attributes. Class-based views can use `@method_decorator`, as covered in the [“Decorating the class” docs](https://docs.djangoproject.com/en/5.0/topics/class-based-views/intro/#decorating-the-class).

This would introduce the oddity that if you subclass a `View` with `@method_decorator(login_not_required, name='dispatch')` applied, and override `dispatch()`, that new `dispatch()` method will not have the `login_required = False` attribute. But we consider this okay as it “fails closed”, the view simply isn’t public until the decorator is reapplied in the subclass.

Alternatively, we could make the mixin still work with an implementation like:

```python
class LoginNotRequiredMixin(View):
    def __init_subclass__(cls):
        method_decorator(login_not_required, name='dispatch')(cls)
```

But IMO better to go with only `@method_decorator` here, to keep the API minimal for this critical security feature. We can always add a mixin later if there is demand.

django/contrib/auth/middleware.py
```diff
@@ -34,6 +38,89 @@ def process_request(self, request):
...
+                f"{self.__class__.__name__}.login_url, settings.LOGIN_URL, or override "
+                f"{self.__class__.__name__}.get_login_url()."
+            )
+        return str(login_url)
+
+    def get_permission_denied_message(self):
+        """
+        Override this method to override the permission_denied_message attribute.
+        """
+        return self.permission_denied_message
```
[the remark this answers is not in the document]
sarahboyce: This is mostly following the [`AccessMixin` api](https://docs.djangoproject.com/en/5.0/topics/auth/default/#django.contrib.auth.mixins.AccessMixin), I have dropped `get_permission_denied_message` and have expanded the docs on the `LoginRequiredMiddleware`.
I can just remove `raise_exception` and `permission_denied_message` and we wait to see if this is requested?

django/contrib/auth/middleware.py
```diff
@@ -34,6 +38,89 @@ def process_request(self, request):
...
+        if not getattr(view_func, "login_required", True):
+            return None
+
+        return self.handle_no_permission(request, view_func)
+
+    def get_login_url(self, view_func):
+        """
+        Override this method to override the login_url attribute.
+        """
+        login_url = self.login_url or getattr(view_func, "login_url", None)
```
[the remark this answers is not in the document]
sarahboyce: Good catch, updated 👍

docs/ref/checks.txt
sarahboyce: Error message is updated 👍

django/contrib/auth/mixins.py
```diff
@@ -2,7 +2,6 @@
 
 from django.conf import settings
 from django.contrib.auth import REDIRECT_FIELD_NAME
-from django.contrib.auth.views import redirect_to_login
```
Hisham-Pak, line 5: @sarahboyce I guess we no longer need an inner import :D

sarahboyce: Star thank you!

django/contrib/auth/middleware.py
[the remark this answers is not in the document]
adamchainz: Oh, yeah, I think we can drop this feature altogether. It’s just a few lines to override `handle_no_permission()` and add that feature if needed.

Also, it seems somewhat unintentinoal that `LoginRequiredMixin` has the feature. `@login_required` does not, but `@permission_required` does. `LoginRequiredMixin` gets it from `AccessMixin`, perhaps without much thought. At least the inconsistency here points to it not being popular, since otherwise we could expect it to have been added to `@login_required`...

django/contrib/auth/middleware.py
```diff
@@ -34,6 +38,75 @@ def process_request(self, request):
...
+        if not getattr(view_func, "login_required", True):
+            return None
+
+        return self.handle_no_permission(request, view_func)
+
+    def get_login_url(self, view_func):
+        login_url = getattr(view_func, "login_url", None)
+        if hasattr(view_func, "view_class"):
+            login_url = getattr(view_func.view_class, "login_url", login_url)
+        login_url = login_url or self.login_url or settings.LOGIN_URL
```
adamchainz, line 67: The `view_class` block can be removed not that we’re no longer using class-level attributes.

```suggestion
        login_url = (
        getattr(view_func, "login_url", None)
        or self.login_url
        or settings.LOGIN_URL
    )
```

django/contrib/auth/middleware.py
```diff
@@ -34,6 +38,75 @@ def process_request(self, request):
...
+        """
+        Override this method to override the redirect_field_name attribute.
+        """
+        redirect_field_name = getattr(view_func, "redirect_field_name", None)
+        view_class = getattr(view_func, "view_class", None)
+        if view_class:
+            redirect_field_name = getattr(
+                view_class, "redirect_field_name", redirect_field_name
+            )
+        return redirect_field_name or self.redirect_field_name
```
adamchainz, line 86: Ditto.

```suggestion
        return (
            getattr(view_func, "redirect_field_name", None)
            or self.redirect_field_name
        )
```

adamchainz (changes requested):
Thanks Sarah, that has really tidied up the implementation. I spotted a couple places where `view_class` is still being looked at. Can you check the test coverage of the middleware?

I also agree on removing the `raise_exception` feature.

django/contrib/auth/middleware.py
sarahboyce: If you were using `LoginRequiredMixin` before, set the `login_url` this would now be ignored. I don't think we should ignore it or we will have to state that the `LoginRequiredMixin` is incompatible?

sarahboyce: I don't intuitive that we respect the `login_required` decorator attributes but not the `LoginRequiredMixin` attributes

django/contrib/auth/middleware.py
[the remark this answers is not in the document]
sarahboyce: Removed 👍

django/contrib/auth/middleware.py
sarahboyce: Updated, I also removed the `login_url` attribute on the middleware as couldn't find the extra value compared to setting the `LOGIN_URL` setting

django/contrib/auth/middleware.py
knyghty: Maybe I'm shaving microseconds for no reason but could you then use `return getattr(view_func, "redirect_field_name", self.redirect_field_name)`?

docs/ref/middleware.txt
```diff
@@ -495,6 +495,30 @@ Adds the ``user`` attribute, representing the currently-logged-in user, to
...
+views, this will redirect to the admin login. For all other views, this will
+redirect to :setting:`settings.LOGIN_URL <LOGIN_URL>`. This can be customized
+by setting ``login_url`` or ``redirect_field_name`` attributes on your view.
+
+Here are examples of how you can apply this in both class based and function
+based views::
+
+    class MyView(View):
```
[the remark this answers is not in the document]
sarahboyce: Turns out it does this by default

django/contrib/auth/middleware.py
```diff
@@ -34,6 +38,57 @@ def process_request(self, request):
...
+        if request.user.is_authenticated:
+            return None
+
+        if not getattr(view_func, "login_required", True):
+            return None
+
+        return self.handle_no_permission(request, view_func)
+
+    def get_login_url(self, view_func):
+        login_url = getattr(view_func, "login_url", None) or settings.LOGIN_URL
```
knyghty, line 61: Then I'd have the same suggestion here:

```suggestion
        login_url = getattr(view_func, "login_url", settings.LOGIN_URL) 
```

django/contrib/auth/middleware.py
adamchainz: Oh, right. Yeah, it’s not intuitive that `LoginRequiredMixin` wouldn’t be respected. I guess we can add back class-level support for `login_url` only..?

Hisham-Pak: `redirect_field_name` as well otherwise redirect could be broken after login if it is not the default "next"?

django/contrib/auth/middleware.py
sarahboyce: I tried to change this but `redirect_field_name` can be an attribute that is `None` which then makes this return `None`

django/contrib/auth/middleware.py
sarahboyce: (sorry about the previous comment) I tried to update this at the same time, however the `login_url` has a default on `None` when the decorator is used which means this update returns `None` rather than `settings.LOGIN_URL`

django/contrib/auth/middleware.py
sarahboyce: Added to the release notes that `LoginRequiredMixin` is not respected. If users which to use rather than the `login_required` decorator, this can be added in a feature release 👍

docs/ref/middleware.txt
```diff
@@ -495,6 +495,59 @@ Adds the ``user`` attribute, representing the currently-logged-in user, to
 every incoming ``HttpRequest`` object. See :ref:`Authentication in web requests
 <auth-web-requests>`.
 
+.. class:: LoginRequiredMiddleware
+
+    .. versionadded:: 5.1
+
+    Redirects all unauthenticated requests to a login page. For admin views,
```
nessita, line 502: I know that we said in our sync today that classes have indented content in other docs, but this file does not seem to comply with that, since classes before this one does not indent their description:
image

I *think* that we have to unindent everything up to the end to be consistent with the rest of the file, and later on consider a separated branch updating this docs to match other files, like everything under `class-based-views` which seems to indent all class definitions.

docs/ref/settings.txt
```diff
@@ -3060,8 +3060,9 @@ Default: ``'/accounts/login/'``
 The URL or :ref:`named URL pattern <naming-url-patterns>` where requests are
 redirected for login when using the
 :func:`~django.contrib.auth.decorators.login_required` decorator,
-:class:`~django.contrib.auth.mixins.LoginRequiredMixin`, or
-:class:`~django.contrib.auth.mixins.AccessMixin`.
+:class:`~django.contrib.auth.mixins.LoginRequiredMixin`,
+:class:`~django.contrib.auth.mixins.AccessMixin`, or
+:class:`~django.contrib.auth.middleware.LoginRequiredMiddleware` is installed.
```
nessita, line 3065: Perhaps slightly clearer?

```suggestion
:class:`~django.contrib.auth.mixins.LoginRequiredMixin` or
:class:`~django.contrib.auth.mixins.AccessMixin` views, or the
:class:`~django.contrib.auth.middleware.LoginRequiredMiddleware`
is installed.
```

docs/releases/5.1.txt
```diff
@@ -26,6 +26,20 @@ only officially support the latest release of each series.
 What's new in Django 5.1
 ========================
 
+Require authentication for all views by default
```
nessita, line 29: I would prepend "Middleware" otherwise others may panic for a second like I did when I read this, my initial interpretation was that Django would require authentication everywhere by default! :-)
```suggestion
Middleware to require authentication by default
```

docs/releases/5.1.txt
```diff
@@ -26,6 +26,20 @@ only officially support the latest release of each series.
...
+-----------------------------------------------
+
+The new :class:`~django.contrib.auth.middleware.LoginRequiredMiddleware`
+redirects all unauthenticated requests to a login page. Views can allow
+unauthenticated requests by using the
```
nessita, line 34:
```suggestion
unauthenticated requests by using the also new
```

docs/releases/5.1.txt
```diff
@@ -26,6 +26,20 @@ only officially support the latest release of each series.
...
+:func:`~django.contrib.auth.decorators.login_not_required` decorator.
+
+The :class:`~django.contrib.auth.middleware.LoginRequiredMiddleware` respects
+``login_url`` and ``redirect_field_name`` set via the
```
nessita, line 38:
```suggestion
the ``login_url`` and ``redirect_field_name`` values set via the
```

docs/releases/5.1.txt
```diff
@@ -26,6 +26,20 @@ only officially support the latest release of each series.
...
+:func:`~.django.contrib.auth.decorators.login_required` decorator, but does not
+support setting ``login_url`` or ``redirect_field_name`` via the
+:class:`~django.contrib.auth.mixins.LoginRequiredMixin`.
+
```
nessita, line 42: Very optional, but would there be a way to grow this section with a small yet (ideally) complete example like we have done in the past for choices allowing a dict or db defaults?

docs/topics/auth/default.txt
```diff
@@ -656,8 +656,25 @@ inheritance list.
...
 .. currentmodule:: django.contrib.auth.decorators
 
+The ``login_not_required`` decorator
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 5.1
+
+When :class:`~django.contrib.auth.middleware.LoginRequiredMiddleware` is
+installed, all views require authentication by default. Some views, such as the
+login view, may want to disable this behavior.
```
nessita, line 670: Maybe?
```suggestion
login view, may need to disable this behavior.
```

django/contrib/auth/checks.py
```diff
@@ -4,10 +4,27 @@
...
+    Return the index of dotted class path (or a subclass of that class) in a
+    list of candidate paths. If it does not exist, return -1.
+    """
+    cls = import_string(class_path)
+    for index, path in enumerate(candidate_paths):
+        try:
+            candidate_cls = import_string(path)
+            if issubclass(candidate_cls, cls):
+                return index
+        except (ImportError, TypeError):
```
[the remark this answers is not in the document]
sarahboyce: Updated 👍

docs/releases/5.1.txt
sarahboyce: Not that I can think of?

django/contrib/auth/middleware.py
```diff
@@ -34,6 +38,57 @@ def process_request(self, request):
         request.auser = partial(auser, request)
 
 
+class LoginRequiredMiddleware(MiddlewareMixin):
+    """
+    Middleware that forces all views to require login by default.
+
+    Views that have login_not_required decorator will allow unauthenticated
+    requests. Otherwise, it will redirect to a login page.
```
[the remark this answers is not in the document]
sarahboyce: I rewrote the docstring

django/contrib/auth/decorators.py
```diff
@@ -82,6 +86,14 @@ def login_required(
     return actual_decorator
 
 
+def login_not_required(view_func):
+    """
+    Decorator for views that allows access to unauthenticated users.
```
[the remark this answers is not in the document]
sarahboyce: This format is currently consistent with the other docstrings. I can delete it entirely

django/contrib/auth/checks.py
```diff
@@ -218,3 +235,29 @@ def check_models_permissions(app_configs=None, **kwargs):
...
+    )
+
+    if login_required_index != -1 and (
+        auth_index == -1 or auth_index > login_required_index
+    ):
+        errors.append(
+            checks.Error(
+                "django.contrib.auth.middleware.LoginRequiredMiddleware requires"
+                " django.contrib.auth.middleware.AuthenticationMiddleware and must "
+                "follow it.",
```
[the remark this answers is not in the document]
sarahboyce: I have rewritten

docs/releases/5.1.txt
nessita: Fair enough, thank you! I really appreciate all the effort you've put into this.

django/contrib/auth/decorators.py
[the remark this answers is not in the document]
sarahboyce: Yes! Thank you ⭐

PR #17806, opened 2024-02-01
Migrated setuptools configuration to pyproject.toml.

claudep:


pauloxnet:
From a quick review, it seems fine to me 👍🏻

nessita:
I will check these new instructions during the incoming release (early April).

/remind me to test this PR on April 2nd

claudep:
Thanks Nick! All suggestions applied.

claudep:
Sorry, I messed up at some point. Now it should be better.

claudep:
Mmmh, looks like deleting the `find` line is disturbing the tests. I'll try reintroduce it.

ngnpope:
Yeah. I'll have to have a look next week at the details of how [auto-discovery](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html) behaves.

nessita:
I haven't tested this yet because the previous release was the "pair with Sarah" release and I prioritized showing the current flow.

/remind me to recheck this work on May 6th

LilyFoote (approved):

adamchainz:
Can we retitle this “Migrated setuptools configuration to pyproject.toml.”? That’s the major change here, using `build` is orthogonal (and could even be a second commit).

ngnpope:
I was having a detailed look at this last night and there are some things that need tweaking. Will try to update with something soon.

pyproject.toml
```diff
@@ -1,12 +1,68 @@
...
-known_first_party = 'django'
+profile = "black"
+default_section = "THIRDPARTY"
+known_first_party = "django"
+
+[tool.setuptools.dynamic]
+version = {attr = "django.__version__"}
+
+[tool.setuptools.packages]
+find = {}
```
ngnpope, line 68: So I found that the wheel produced contained `docs/`, `tests/`, etc. that it shouldn't.

It seems to be due to the fact that `setuptools` will consider implicit namespaces by default when using a `pyproject.toml` file according to the note under [this section](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#finding-simple-packages).

We could do the following instead to disable this:

```suggestion
[tool.setuptools.packages.find]
namespaces = false
```

But then I get a large number of warnings in the build output like this:

```
.../lib/python3.12/site-packages/setuptools/command/build_py.py:207: _Warning: Package 'django.conf.app_template' is absent from the `packages` configuration.
!!                                                                                                                      
                                                                                                                        
        ********************************************************************************                                
        ############################                                                                                    
        # Package would be ignored #                                                                                    
        ############################                                                                                    
        Python recognizes 'django.conf.app_template' as an importable package[^1],                                      
        but it is absent from setuptools' `packages` configuration.                                                     
                                                                                                                        
        This leads to an ambiguous overall configuration. If you want to distribute this                                
        package, please make sure that 'django.conf.app_template' is explicitly added                                   
        to the `packages` configuration field.                                                                          
                                                                                                                        
        Alternatively, you can also rely on setuptools' discovery methods                                               
        (for example by using `find_namespace_packages(...)`/`find_namespace:`                                          
        instead of `find_packages(...)`/`find:`).                                                                       
                                                                                                                        
        You can read more about "package discovery" on setuptools documentation page:                                   
                                                                                                                        
        - https://setuptools.pypa.io/en/latest/userguide/package_discovery.html                                         
                                                                                                                        
        If you don't want 'django.conf.app_template' to be distributed and are                                          
        already explicitly excluding 'django.conf.app_template' via                                                     
        `find_namespace_packages(...)/find_namespace` or `find_packages(...)/find`,                                     
        you can try to use `exclude_package_data`, or `include-package-data=False` in                                   
        combination with a more fine grained `package-data` configuration.                                              
                                                                                                                        
        You can read more about "package data files" on setuptools documentation page:                                  
                                                                                                                        
        - https://setuptools.pypa.io/en/latest/userguide/datafiles.html                                                 
                                                                                                                        
                                                                                                                        
        [^1]: For Python, any directory (with suitable naming) can be imported,                                         
              even if it does not contain any `.py` files.                                                              
              On the other hand, currently there is no concept of package data                                          
              directory, all directories are treated like packages.                                                     
        ********************************************************************************                                
                                                                                                                        
!!                                                                                                                      
  check.warn(importable)                                                                                                
```

It turns out we had that before when using `packages = find:` in `setup.cfg` as implicit namespaces weren't considered, but it sounds like something that ought to be fixed. So instead I attempted to explicitly exclude the stuff we don't want in the wheel:

```suggestion
[tool.setuptools.packages.find]
exclude = ["docs*", "extras*", "js_tests*", "scripts*", "tests*"]
```

This works, but I think for safety in case something else is added we should just include exactly what we want:

```suggestion
[tool.setuptools.packages.find]
include = ["django*"]
```

In theory, according to the [flat-layout docs](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#flat-layout) this shouldn't be necessary, as many of these are in the default exclude, but it seems that the `namespaces` thing takes precedence. And we likely want that due to the spewing of warnings that we get.

extras/Makefile
```diff
@@ -1,9 +1,9 @@
...
 
 sdist:
-	python setup.py sdist
+	python -m build --sdist
 
 bdist_wheel:
-	python setup.py bdist_wheel
+	python -m build --wheel
 
 .PHONY : sdist bdist_wheel
```
ngnpope, line 9: According to `python -m build --help`, passing `--wheel` means that it won't create the wheel from the sdist, but from the source folder. The advantage of creating the wheel from the sdist is that it can verify that the sdist can be used to build wheels:

> By default, a source distribution (sdist) is built from `{srcdir}` and a binary distribution (wheel) is built from the sdist. This is recommended as it will ensure the sdist can be used to build wheels.
> 
> Pass `-s`/`--sdist` and/or `-w`/`--wheel` to build a specific distribution. If you do this, the default behavior will be disabled, and all artifacts will be built from `{srcdir}` (even if you combine `-w`/`--wheel` with `-s`/`--sdist`, the wheel will be built from `{srcdir}`).

In light of this, I suggest that we ditch this `extras/Makefile` file and replace `make -f extras/Makefile` in `docs/internals/howto-release-django.txt` with `python -m build` which is one simple command - the makefile is superfluous. It also makes those instructions flow better, as they contain `python -m pip install build twine` already, so using `make` is just obscuring what we're doing.

pyproject.toml
```diff
@@ -1,12 +1,68 @@
 [build-system]
-requires = ['setuptools>=40.8.0']
-build-backend = 'setuptools.build_meta'
+requires = ["setuptools>=61.0.0"]
```
ngnpope, line 2: So it turns out that [v69.3.0](https://setuptools.pypa.io/en/latest/history.html#v69-3-0) supports PEP 625 and so canonicalizes package names. I think they really mean "normalizes". What this means is that we now get the following files:

```
Django-X.Y-py3-none-any.whl
django-X.Y.tar.gz
```

Note that the sdist filename is now lower-cased. It seems that this is going to be the way forward and that the wheel also ought to be the same, but that remains to be seen.

setup.cfg
```diff
@@ -1,55 +1,3 @@
...
-    Release notes = https://docs.djangoproject.com/en/stable/releases/
-    Funding = https://www.djangoproject.com/fundraising/
-    Source = https://github.com/django/django
-    Tracker = https://code.djangoproject.com/
-
-[options]
-python_requires = >=3.10
-packages = find:
-include_package_data = true
-zip_safe = false
```
ngnpope, line 39: I checked the contents of the sdist archives before and after and we are missing a single file with these changes: `Django.egg-info/`[`not-zip-safe`](https://setuptools.pypa.io/en/latest/deprecated/python_eggs.html#zip-safe-and-not-zip-safe). This is missing because of not migrating over the obsolete [`zip-safe`](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration) option. This is fine 👍🏻

ngnpope (changes requested):
Other than the things below I also saw the following warning in the output both before and after:

```
warning: no previously-included files matching '__pycache__' found anywhere in distribution
```
It seems that we should be able to remove `global-exclude __pycache__` from `MANIFEST.in`.

Also, as we now depend on `setuptools>=61` and given that `setup.py install` has been deprecated since [v58.3.0](https://setuptools.pypa.io/en/stable/history.html#v58-3-0) in October 2021, I think we can consider removing `setup.py` in its entirety. _(It contains a bunch of workarounds to avoid overlaying an existing install which is no longer really applicable.)_

pyproject.toml
nessita: This! I got this during the Django 5.1 alpha release and it was a nightmare (and I was not using this PR).
I started the alpha release on a system that only had Python 3.12, and since `distutils` is gone in 3.12, I had to manually install `setuptools` which brought the lastest version and that resulted in lowercase-named `django` tarball.

Because the lowercase-name `django` generated complains after the release that Sarah did at the beginning of May, and also the Jenkins CI job fails for lowercase `django`, I resorted to using a Python 3.11 docker image to build the release files, but we'd need a plan for this.

extras/Makefile
claudep: Thanks, suggestion applied.

pyproject.toml
claudep: @ngnpope Do you suggest we require setuptools>=69.3.0?

@nessita Did you identify what's the main issues with lowercased sdist filename?

ngnpope: > Do you suggest we require setuptools>=69.3.0?

I guess we could do that to ensure we always get the normalised name as that looks to be the way it'll be going forward and we won't get a day in it.

Alternatively, we can pin to `setuptools>=61.0.0,<69.3.0` here for now to give time to communicate the change if necessary as, with this PR, we get build isolation. But we'll need to unpin at some point.

> …and also the Jenkins CI job fails for lowercase `django`…

@nessita This sounds suspect. I don't understand - which job is this and why should it care about the name of the sdist? Are we not always installing a wheel via `pip`? That wouldn't care as it will normalise the name anyway.

ngnpope:
@claudep Did you have any thoughts on [these comments](https://github.com/django/django/pull/17806#pullrequestreview-2070796314)?

claudep:
Oh sorry, missed it. I just pushed commits with your suggestion.

pyproject.toml
nessita: @claudep @ngnpope The issue with the "lowercased-name" django tarball I mentioned above was related to the *confirm release* CI job (sorry I wasn't explicit yesterday, it was a tough day).

At the beginning of May, when @sarahboyce did her first release, she used Python 3.12 and pip-installed the latest `setuptools`. The resulting django tarball was lowercased and that led to "issues" (there was some complaining in Mastodon and a failed CI run for the "confirm-release" job), so I *thought* that the CI job failed because of the "wrong casing" (see [run 1](https://djangoci.com/job/confirm-release/196/console) and [run 2](https://djangoci.com/job/confirm-release/197/console)). **But**, after a chat with Sarah today, she reminded me that the failures were caused by a manual rename of the tarball from `django-...` to `Django-..` (capitalized) so that was the cause of the CI issue.

Long story short, the CI confirm-release job should be OK with either casing.

Then, and as mentioned yesterday, I was also using python 3.12 and latest setuptools to build the binaries, which resulted in the "django-..." named tarball, so I panicked and used a docker image with Python 3.11 (which includes distutils) to complete the release. Having looked at this again with fresher eyes, I could've used Python 3.12 with `pip install --force "setuptools<67"` and that would have let me produce the usual casing "Django-..." tarball and wheel.

In summary, the lowercase django may be fine but unclear if "the community" may complain for reasons I can't think of today.

adamchainz: I think you might encounter some complaining from repackagers, whose scripts might need adjusting for the different casing, but I wouldn’t expect that to be a big issue considering it’s a standard and happening to all packages.

nessita:
@ngnpope Hey there, we have a new release in a week (5.1 beta release), would you have some time to re-review this PR before then to consider merging this work and using it in that release? Thank you.

setup.cfg
```diff
@@ -1,55 +1,3 @@
...
-
-[options.entry_points]
-console_scripts =
-    django-admin = django.core.management:execute_from_command_line
-
-[options.extras_require]
-argon2 = argon2-cffi >= 19.1.0
-bcrypt = bcrypt
-
 [flake8]
```
ngnpope, line 1: Optional: Wondering if we should rename `setup.cfg` to `.flake8`?

Pretty much every other tool now supports `pyproject.toml`, with `flake8` being an exception, and `setup.cfg` really belongs to `setuptools`.

ngnpope:
> What's more confusing is why the wheel filename is still uppercased as I thought the whole point of PEP 625 was to align the naming to match the wheel filename convention... 🤔

It seems that there is a bug and the wheel filename should be lowercased according to this:

https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode

More reading:

- https://github.com/pypa/setuptools/issues/4300#issuecomment-2058930005
- https://github.com/pypa/packaging.python.org/commit/6d4efbc12ea8220a2b78136a0ab13ca2857bf38a
- https://github.com/pypa/setuptools/issues/3777

Urgh. What a mess.

TL;DR: This is good to go as the filename problem existed before switching to `pyproject.toml` anyway. You have two options - pin to `setuptools>=61.0.0,<69.3.0` until `setuptools` is fixed to produce correct wheel filenames or switch to `name = "django"` to pre-normalize the name.

setup.cfg
adamchainz: +1 to switching it. It may help avoid confusing other tools that look for data in `setup.cfg`.

nessita:
> TL;DR: This is good to go as the filename problem existed before switching to `pyproject.toml` anyway. You have two options - pin to `setuptools>=61.0.0,<69.3.0` until `setuptools` is fixed to produce correct wheel filenames or switch to `name = "django"` to pre-normalize the name.

Thank you so much @ngnpope for the complete summary. My preference is to pin `setuptools`, so we can merge this and progress this modernization without disrupting the project name and the binary casing. One question though: from my experiments, we need `setuptools<67` to avoid any name case change. But you are proposing `<69.3.0` so I wanted to double check with you about the proper upper threshold.

setup.cfg
claudep: Another option would be to define the flake8 config in `tox.ini`, what do you think?

ngnpope: We could do, but I don't really like one tool using another tool's configuration file. This is why `pyproject.toml` is nice as it is intended to be tool-agnostic.

Also `tox` does have rudimentary support for `pyproject.toml`, but it's basically one setting with a large string blob of everything - so not worth merging that yet.

We could also keep `setup.cfg` and move `tox.ini` in there, which also works, but - as I said - `setup.cfg` really belongs to (legacy) `setuptools`.

ngnpope:
> One question though: from my experiments, we need `setuptools<67` to avoid any name case change. But you are proposing `<69.3.0` so I wanted to double check with you about the proper upper threshold.

v63.9.0 introduced the changes that affected the filename, so I think the bounds I gave were correct? (It's what worked for me.)

nessita:
> > One question though: from my experiments, we need `setuptools<67` to avoid any name case change. But you are proposing `<69.3.0` so I wanted to double check with you about the proper upper threshold.
> 
> v63.9.0 introduced the changes that affected the filename, so I think the bounds I gave were correct? (It's what worked for me.)

I'm assuming you meant `v69.3.0` :-)
I tested again with a clean venv and you are correct, `pip install "setuptools>=61.0.0,<69.3.0"` allowed me to build these binaries:
```
-rw-r--r-- 1 nessita nessita  8200954 jun 19 10:22 Django-5.2.dev20240618152543-py3-none-any.whl
-rw-r--r-- 1 nessita nessita 10742249 jun 19 10:22 Django-5.2.dev20240618152543.tar.gz
```
So I think we are good to go!

setup.cfg
claudep: Done.

nessita (approved):
Thank you everyone, this looks good. I have tested this in a clean environment and I have updated our "ad-hoc" release script. I will push a cleanup for what I think is a missed replace and then I'll merge:
```
docs/internals/contributing/writing-code/coding-style.txt:49:  ``setup.cfg`` file contains...
```

nessita:
@claudep @ngnpope I had an action item to evaluate backporting these changes to 4.2, since we need to maintain 4.2 for another year (5.0 was EOLd in April so no changes there). I have done so in https://github.com/django/django/pull/19436, I would appreciate your reviews. Thanks!

PR #17826, opened 2024-02-06
[gsoc] django_auto_import shell feature

Ammar-Munirr:
Adam requested me to create a draft PR [on the forum](https://forum.djangoproject.com/t/have-solved-80-of-auto-importing-shell-feature/27782/6). This feature is one of the ideas for [GSOC 2024](https://code.djangoproject.com/wiki/SummerOfCode2024#Auto-importingshell).


timgraham:
Hi, this is implemented in [shell_plus](https://django-extensions.readthedocs.io/en/latest/shell_plus.html#shell-plus) command of the django-extensions package.

shangxiao:
Hey Tim, just FYI Adam requested that Ammar-Munirr create a draft PR [on the forum](https://forum.djangoproject.com/t/have-solved-80-of-auto-importing-shell-feature/27782/6) 😅

(This feature is one of the ideas for [GSOC 2024](https://code.djangoproject.com/wiki/SummerOfCode2024#Auto-importingshell))

Ammar-Munirr:
Hey @timgraham , I'm aware this functionality is part of the well-known Django extension called shell_plus. However, referencing this [forum](https://forum.djangoproject.com/t/have-solved-80-of-auto-importing-shell-feature/27782/6), @adamchainz  asked for a pull request to be submitted.

They want to review my code and offer suggestions for improvement. For further details, you can check out the project's page mentioned in [GSOC 2024](https://code.djangoproject.com/wiki/SummerOfCode2024#Auto-importingshell).

django/core/management/helper.py
```diff
@@ -0,0 +1,66 @@
...
+    {"Max": "django.db.models.Max"},
+    {"Min": "django.db.models.Min"},
+    {"Prefetch": "django.db.models.Prefetch"},
+    {"Q": "django.db.models.Q"},
+    {"Sum": "django.db.models.Sum"},
+    {"When": "django.db.models.When"},
+    {"timezone": "django.utils.timezone"},
+    {"reverse": "django.urls.reverse"},
+    {"Exists": "django.db.models.Exists"},
+]
```
adamchainz, line 34: I am not sure that we should add any non-model default imports. Yes, it’s tempting to add a few oft-used things, but preferences will vary between users and projects. I would prefer that we document how users can extend the `shell` function to add extra objects.

django/core/management/helper.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def get_objects(style):
+    import_objects = {}
+    for x in CHANGED_IMPORTS:
+        for key, val in x.items():
+            import_objects[key] = import_string(val)
+    print(style.MIGRATE_HEADING("# Importing Necessary modules"))
+    for x in DJANGO_DEFAULT_IMPORTS:
+        print(style.SUCCESS(x))
```
adamchainz, line 44: As discussed on the forum, I’d prefer we only write a single message like: “Imported N default objects.”. Although there’s probably better wording that doesn’t sound like a data import.

django/core/management/helper.py
```diff
@@ -0,0 +1,66 @@
+from django import setup
```
adamchainz, line 1: This extra module is unnecessary, and has an especially unclear name.

adamchainz (changes requested):
Well done on getting this started ahead of GSoC. Note there’s still no guarantee of this project or you being accepted, but initial exploration is useful.

I proposed you create a “draft PR”, which avoids notifying those following the repository such as fellows (or ex-fellows like Tim!). You can convert it to a draft still: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft

Second, I think you’ve complicated things a bit too much. I propose that we add a method to `shell.Command` called `get_namespace` or similar. This should perform the imports and return a dictionary mapping names to objects. YOu could fold your existing code in there.

Further comments on the PR.

django/core/management/helper.py
Ammar-Munirr: How about this approach ?
Screenshot from 2024-02-07 11-45-46

django/core/management/helper.py
Ammar-Munirr: Yeah I am planing to add functionality that user set there default imports in settings.py:
SHELL_DEFAULT_IMPORT = [
'from django.contrib.auth.models import get_user_model'
]
it they don't set any default imports the shell will only import models of installed applications.

Ammar-Munirr:
I apologize for the delay in responding. Right now, Pakistan is having elections, and the government has cut off internet access nationwide to ensure election security. That's why I haven't been able to update or respond to pull requests promptly.
anyway I have fix the code, now it will work for both plain and Ipython shell for confirmation I am attaching the SS.
Screenshot from 2024-02-09 23-41-27

felixxm:
Superseded by #18158.

PR #17834, opened 2024-02-07
Fixed #7732 -- Added support for connection pooling in oracle backend.

suraj-ora-2020:
Added support for session pool in Django for oracle backend. Supported with [oracledb](https://github.com/oracle/python-oracledb) driver.
[Ticket- 7732](https://code.djangoproject.com/ticket/7732)


felixxm (commented):
Please start from fixing doc and test failures.

suraj-ora-2020:
@felixxm Updated.

django/db/backends/oracle/base.py
```diff
@@ -238,6 +240,38 @@ def __init__(self, *args, **kwargs):
...
+
+        # Verify that we are not running with persistent connections
+        if self.settings_dict.get("CONN_MAX_AGE", 0) != 0:
+            raise ImproperlyConfigured(
+                """Pooling doesn't support persistent connections, unset
+                   conn_max_age"""
+            )
+        # Pooling feature is only supported for oracledb
+        if not is_oracledb:
+            raise ImproperlyConfigured("Not supported by cx_Oracle install oracledb")
```
felixxm, line 260: This should raise an exception much earlier when connection params are checked.

tests/backends/oracle/tests.py
```diff
@@ -69,6 +69,11 @@ def test_check_database_version_supported(self, mocked_get_database_version):
             connection.check_database_version_supported()
         self.assertTrue(mocked_get_database_version.called)
 
+    @unittest.skipUnless(connection.is_pool, "Pool specific tests")
+    def test_connect_pool_set_to_true(self):
+        test_pool = connection.settings_dict["OPTIONS"]["pool"]
+        self.assertTrue(test_pool)
```
felixxm, line 75: This hardly test any of added code.

docs/releases/5.1.txt
```diff
@@ -146,6 +146,12 @@ CSRF
 
 * ...
 
+oracle backends
+~~~~~~~~~~~~~~~
+
+* The new ``"pool"`` option is now supported in :setting:`OPTIONS` on
+  oracle to allow using connection pool.
```
felixxm, line 153: Move release notes to the `Database backends` sessions.

django/db/backends/oracle/base.py
```diff
@@ -238,6 +240,38 @@ def __init__(self, *args, **kwargs):
...
+            del connect_kwargs["pool"]
+            pool = Database.create_pool(
+                user=self.settings_dict["USER"],
+                password=self.settings_dict["PASSWORD"],
+                dsn=dsn(self.settings_dict),
+                **connect_kwargs,
+            )
+            self._connection_pools.setdefault(self.alias, pool)
+
+        return self._connection_pools[self.alias]
```
felixxm, line 273: New functionality should be tested, you can check new tests in #17914 for an inspiration.

docs/ref/databases.txt
```diff
@@ -1082,6 +1082,46 @@ Example of a full DSN string::
         "(CONNECT_DATA=(SERVICE_NAME=orclpdb1)))"
     )
 
+.. _oracle-pool:
+
+Connection Pool
```
felixxm, line 1087:
```suggestion
Connection pool
```

docs/ref/databases.txt
```diff
@@ -1082,6 +1082,46 @@ Example of a full DSN string::
...
+part of your database configuration :setting:`DATABASES`::
+
+    DATABASES = {
+        "default": {
+            "ENGINE": "django.db.backends.oracle",
+            # ...
+            "OPTIONS": {
+                "pool": True,
+                "min": 1,
+                "max": 10,
```
felixxm, line 1117: I'd recommend to use the same pattern as in #17914.

felixxm (commented):
@suraj-ora-2020 Thanks :+1: I left initial comments.

I've tried to run our test suite with `"pool": True` and encountered dozens or errors/failures, please check them before you will mark this ready for another reviews.

django/db/backends/oracle/base.py
```diff
@@ -238,17 +240,59 @@ def __init__(self, *args, **kwargs):
         )
         self.features.can_return_columns_from_insert = use_returning_into
 
+    @property
+    def is_pool(self):
+        return self.settings_dict["OPTIONS"].get("pool")
+
+    @property
+    def pool(self):
+        if not self.settings_dict["OPTIONS"].get("pool"):
```
felixxm, line 249: Why we don't reuse `is_pool` here?

django/db/backends/oracle/base.py
```diff
@@ -238,17 +240,59 @@ def __init__(self, *args, **kwargs):
...
+            return None
+
+        # Verify that we are not running with persistent connections
+        if self.settings_dict.get("CONN_MAX_AGE", 0) != 0:
+            raise ImproperlyConfigured(
+                """Pooling doesn't support persistent connections, unset
+                conn_max_age"""
```
felixxm, line 256:
```suggestion
                "Pooling doesn't support persistent connections."
```

docs/ref/databases.txt
```diff
@@ -1107,6 +1107,46 @@ Example of a full DSN string::
         "(CONNECT_DATA=(SERVICE_NAME=orclpdb1)))"
     )
 
+.. _oracle-pool:
+
+Connection pool
```
felixxm, line 1112: As far as I'm aware connection and session pooling are not the same thing. I think you copy too many things from PostgreSQL connection pooling PR. Please correct wording.

docs/ref/databases.txt
felixxm: This is not fixed, it should be:
```
"pool": {
    "min": 1,
    "max": 10,
}
```
Also, I cannot find any place where pool's parameters are used.

docs/releases/5.1.txt
```diff
@@ -170,6 +170,9 @@ Database backends
 * ``"pool"`` option is now supported in :setting:`OPTIONS` on PostgreSQL to
   allow using :ref:`connection pools <postgresql-pool>`.
 
+* ``"pool"`` option is now supported in :setting:`OPTIONS` on Oracle to
+  allow using :ref:`connection pools <oracle-pool>`.
```
felixxm, line 174: Again, "session" not connection pools.

tests/backends/oracle/tests.py
```diff
@@ -69,6 +70,63 @@ def test_check_database_version_supported(self, mocked_get_database_version):
...
+        new_connection.closed_in_transaction = True
+        with self.assertRaisesMessage(ProgrammingError, msg):
+            new_connection.ensure_connection()
+
+    @unittest.skipUnless(connection.is_pool, "Pool specific tests")
+    def test_pooling_not_support_persistent_connections(self):
+        new_connection = connection.copy()
+        new_connection.settings_dict["CONN_MAX_AGE"] = 10
+        msg = """Pooling doesn't support persistent connections, unset
+                conn_max_age"""
```
felixxm, line 126:
```suggestion
        msg = "Pooling doesn't support persistent connections."
```

tests/backends/oracle/tests.py
```diff
@@ -69,6 +70,63 @@ def test_check_database_version_supported(self, mocked_get_database_version):
             connection.check_database_version_supported()
         self.assertTrue(mocked_get_database_version.called)
 
+    @unittest.skipUnless(connection.is_pool, "Pool specific tests")
+    def test_connect_pool_set_to_true(self):
+        test_pool = connection.settings_dict["OPTIONS"]["pool"]
+        self.assertTrue(test_pool)
+
+    @unittest.skipUnless(connection.is_pool, "Pool specific tests")
+    def test_connect_pool(self):
```
felixxm, line 79: This test hangs.

suraj-ora-2020: The test was hanging because the default value for min and max connections in oracledb is 1. As the test was trying to hold two connections but max was not specified in the options dictionary of the settings.py file, it caused the program to hang. To resolve this issue, I modified the code to set max as 4 by default if it is not specified in the options dictionary.

docs/ref/databases.txt
suraj-ora-2020: @felixxm  oracledb documentation also mentions it as  connection pooling. For reference https://python-oracledb.readthedocs.io/en/latest/user_guide/connection_handling.html#connection-pooling . Should I replace connection with sessions?

docs/ref/databases.txt
suraj-ora-2020: You are suggesting "pool" can have 2 values .
1 - True/False (boolean)
2 - { "min":1, "max":2 } (Dictionary)

django/db/backends/oracle/base.py
suraj-ora-2020: Yes we can use. I have modified.

docs/ref/databases.txt
sarahboyce: @suraj-ora-2020 I believe Mariusz is right, if you look into the docs you linked it says:

>In python-oracledb Thick mode, the pool implementation uses Oracle’s [**session pool** technology](https://www.oracle.com/pls/topic/lookup?ctx=dblatest&id=GUID-F9662FFB-EAEF-495C-96FC-49C6D1D9625C)

And then if you go into the linked Oracle docs, session pooling and connection pooling are different.

suraj-ora-2020: @sarahboyce  drivers like the python driver (**oracledb**) or node.js driver refer to a connection+session combination as a connection. And OCI's session pooling is nothing but the same, a session from a session pool actually means  session+connection.

cjbj: I agree with @suraj-ora-2020 : drivers like python-oracledb and node-oracledb deliberately use 'connection pool', so that terminology should also be used in Django and frameworks that use the Oracle driver.

docs/releases/5.1.txt
cjbj: See above: the driver terminology is 'connection pool'.

django/db/backends/oracle/base.py
```diff
@@ -238,17 +240,69 @@ def __init__(self, *args, **kwargs):
...
+            del self._connection_pools[self.alias]
+
     def get_database_version(self):
         return self.oracle_version
 
     def get_connection_params(self):
+        # Pooling feature is only supported for oracledb
+        if self.is_pool and not is_oracledb:
+            raise ImproperlyConfigured(
+                "Pooling Not supported by cx_Oracle install oracledb"
```
cjbj, line 294: This is awkward. How about 'Pooling is not supported by cx_Oracle.  Use python-oracledb instead' ?

django/db/backends/oracle/features.py
```diff
@@ -137,6 +137,25 @@ def django_test_skips(self):
                     },
                 }
             )
+        if self.connection.settings_dict["OPTIONS"].get("pool"):
+            skips.update(
+                {
+                    "Pooling does not supports persistent connection": {
```
cjbj, line 143: Typo: move the 's' from "supports" to "connection", i.e "Pooling does not support persistent connections"

docs/ref/databases.txt
```diff
@@ -1107,6 +1107,49 @@ Example of a full DSN string::
...
+---------------
+
+.. versionadded:: 5.1
+
+To use a connection pool with `oracledb`_, you can set the ``"pool"`` in the
```
cjbj, line 1117: I don't understand this sentence: it is complex.  Perhaps split it up?  Use bullet points?  Cut out text and let the examples show what is needed?

docs/releases/5.1.txt
```diff
@@ -179,6 +179,9 @@ Database backends
 * ``"pool"`` option is now supported in :setting:`OPTIONS` on PostgreSQL to
   allow using :ref:`connection pools <postgresql-pool>`.
 
+* ``"pool"`` option is now supported in :setting:`OPTIONS` on Oracle to
```
cjbj, line 182: Change to 'Oracle Database'

tests/backends/oracle/tests.py
```diff
@@ -69,6 +70,57 @@ def test_check_database_version_supported(self, mocked_get_database_version):
...
+            # copy() reuses the existing alias and as such the same pool.
+            conn = new_connection.copy()
+            conn.connect()
+            connections.append(conn)
+            return conn
+
+        try:
+            connection_1 = get_connection()  # First connection.
+            get_connection()  # Get the second connection.
+            sql = "select sys_context('userenv', 'sid') from dual"
```
cjbj, line 95: I think this should check the SID & serial # combination is reused: something like `select sid || ' ' || serial# from v$session where sid = sys_context('USERENV','SID')`

cjbj (changes requested):

django/db/backends/oracle/base.py
sarahboyce: This is also missing test coverage 👍

tests/backends/oracle/tests.py
suraj-ora-2020: To query v$session one requires a sysdba connection . Also to verify connection reuse in oracledb similar query is used.
https://github.com/oracle/python-oracledb/blob/main/tests/test_2400_pool.py#L760

django/db/backends/oracle/base.py
```diff
@@ -238,17 +240,69 @@ def __init__(self, *args, **kwargs):
...
+            del self._connection_pools[self.alias]
+
     def get_database_version(self):
         return self.oracle_version
 
     def get_connection_params(self):
+        # Pooling feature is only supported for oracledb
+        if self.is_pool and not is_oracledb:
+            raise ImproperlyConfigured(
+                "Pooling Not supported by cx_Oracle. Use python-oracledb instead"
```
cjbj, line 294: Change "not" to lowercase and while you're editing it, make it "isn't" to match the "doesn't" from line 255.

Or spell both out as "is not" and "does not" to match the "does not" used in django/db/backends/oracle/features.py

suraj-ora-2020:
@sarahboyce Could we move forward with the next steps for this pull request?

sarahboyce (commented):
@suraj-ora-2020 I've added a few comments

Can you squash to a single commit with a message something like "Fixed #7732 -- Added support for connection pooling in oracle backend."?

django/db/backends/base/creation.py
```diff
@@ -62,6 +62,10 @@ def create_test_db(
         self._create_test_db(verbosity, autoclobber, keepdb)
 
         self.connection.close()
+
+        if self.connection.is_pool:
+            self.connection.close_pool()
+
```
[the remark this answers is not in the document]
suraj-ora-2020: Moved to oracle layer.

django/db/backends/oracle/base.py
```diff
@@ -239,17 +241,69 @@ def __init__(self, *args, **kwargs):
...
+            connect_kwargs = self.get_connection_params()
+            pool_options = connect_kwargs["pool"]
+
+            if pool_options is True:
+                connect_kwargs["max"] = 4
+            else:
+                for key, val in pool_options.items():
+                    connect_kwargs[key] = val
+                if not connect_kwargs.get("max"):
+                    connect_kwargs["max"] = 4
```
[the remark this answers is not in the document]
suraj-ora-2020: Yes, the python-oracledb driver have max = 2 as default value. However, I've set it as 4 because the psycopg_pool has the default minimum and maximum values set as 4. Additionally, the test **test_pool_reuse** requires 3 connections, so setting the maximum to 2 would cause it to fail. Therefore, 4 seems to be a more suitable value, especially considering the psycopg_pool.

docs/releases/5.1.txt
```diff
@@ -210,6 +210,16 @@ Database backends
...
+Decorators
+~~~~~~~~~~
+
+* ...
+
+Email
+~~~~~
+
+* ...
+
```
sarahboyce, line 222:
```suggestion
```
There should be no changes to the 5.1 release notes

django/db/backends/oracle/base.py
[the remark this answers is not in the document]
sarahboyce: Ah then I think `test_pool_reuse` should be setting the connection parameters similar to `tests.backends.postgresql.tests.Tests.test_connect_pool`. I don't think we need to have the max size defaults align to postgres and it means the code and docs is simpler

tests/test_utils/tests.py
```diff
@@ -2186,6 +2186,8 @@ def test_allowed_database_copy_queries(self):
         finally:
             new_connection.validate_thread_sharing()
             new_connection._close()
+            if new_connection.is_pool:
+                new_connection.close_pool()
```
sarahboyce, line 2190: I'm not really sure why this needs a `close_pool` but `test_allowed_threaded_database_queries` does not 
What happens when you remove this?

django/db/backends/base/base.py
```diff
@@ -47,6 +47,7 @@ class BaseDatabaseWrapper:
     introspection_class = None
     ops_class = None
     validation_class = BaseDatabaseValidation
+    is_pool = False
```
sarahboyce, line 50: From what I can tell, this is mostly used for skipping tests
Have you looked into whether those tests can have the connection settings updated in the test (similar to `tests.backends.postgresql.tests.Tests.test_connect_pool_set_to_true` for example) so that we have some test coverage in our current ci and we don't impact the base backend?

sarahboyce (commented):
Thank you for the updates @suraj-ora-2020

tests/test_utils/tests.py
suraj-ora-2020: In the `test_allowed_database_copy_queries` test, it's necessary to explicitly close the pool because it copies a connection with the `dynamic_connection` alias. Due to this new alias, a new pool is created with the key `(dynamic_connection, user)` , and this pool is never closed because `new_connection._close()` only releases the connection back to the pool, but never closes the connection. As a result, when dropping the test user at the end of the test suite run, an error is raised because the test user is already connected due to this test.

In the case of `test_allowed_threaded_database_queries`, the alias is `default`, so it doesn't create a new pool; instead, it reuses the existing pool `(default, user)` and gets closed in the `_destroy_test_db()` call.

So if you delete this line, at the end of the test suite run, it will raise **cannot drop a user which is already connected** during the `_destroy_test_user()` call.

sarahboyce: Thank you for the context! Using a connection pool config for postgres also has an error when destroying the test db for this test

@apollo13 I feel like some of the test errors when running a connection pool config for postgres where ignored as "expected errors". Do you think it's worth adding a `close_pool` call to `test_allowed_database_copy_queries`?

If we don't want to add a `is_pool` property, we could do something like
```python
    if hasattr(new_connection, "close_pool"):
        new_connection.close_pool()
```

@suraj-ora-2020 note that `is_pool` is currently `False` for postgres connection pools - but I'm still not 100% convinced we want to add this property

suraj-ora-2020: @sarahboyce , I agree that `is_pool` is currently `False` for postgres connection pools and other database backends. However, I think the pooling functionality will be extended to other backends in the future, including sqlite and mysql. In the base layer, we can set `is_pool` as False, and in the backend-specific layer, we can add the following code:

```python
@property
def is_pool(self):
    return self.settings_dict["OPTIONS"].get("pool", False)
```

This code can be added to postgres and the other backends when they extend this functionality.
If you have any other suggestions or if you're not convinced, feel free to let me know.

django/db/backends/base/base.py
suraj-ora-2020: I have removed `is_pool` from the base backend and made the test similar to `tests.backends.postgresql.tests.Tests.test_connect_pool_set_to_true`.

tests/test_utils/tests.py
suraj-ora-2020: I have updated the code according to the suggestion. This will work for both Oracle and PostgreSQL and will resolve the issue we are facing when dropping the user in the case of connection pooling for both backends.
`if hasattr(new_connection, "close_pool"):
        new_connection.close_pool()
`

django/db/backends/oracle/base.py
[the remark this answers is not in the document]
suraj-ora-2020: I have removed the default value of "max" to be 4. Now, in case of default configuration, it will use whatever defaults are defined in oracledb.

sarahboyce:
buildbot, test on oracle.

sarahboyce:
Thank you for the updates @suraj-ora-2020 :star: 

> If all the changes seem okay, I can create a new PR with a single commit as per the requirement.

I'm happy with the updates, can you squash to a single commit and rebase on main?

suraj-ora-2020:
Sure.

django/db/backends/oracle/base.py
```diff
@@ -239,17 +241,66 @@ def __init__(self, *args, **kwargs):
...
+            del self._connection_pools[pool_key]
+
     def get_database_version(self):
         return self.oracle_version
 
     def get_connection_params(self):
+        # Pooling feature is only supported for oracledb.
+        if self.is_pool and not is_oracledb:
+            raise ImproperlyConfigured(
+                "Pooling isn't supported by cx_Oracle. Use python-oracledb instead"
```
felixxm, line 292:
```suggestion
                "Pooling isn't supported by cx_Oracle. Use python-oracledb instead."
```

django/db/backends/oracle/base.py
```diff
@@ -239,17 +241,66 @@ def __init__(self, *args, **kwargs):
...
+            )
+            self._connection_pools.setdefault(pool_key, pool)
+
+        return self._connection_pools[pool_key]
+
+    def close_pool(self):
+        pool_key = (self.alias, self.settings_dict["USER"])
+        if self.pool:
+            self.pool.close(force=True)
+            del self._connection_pools[pool_key]
```
felixxm, line 283: There is no need to create `pool_key` when it's not used:
```suggestion
        if self.pool:
            self.pool.close(force=True)
            pool_key = (self.alias, self.settings_dict["USER"])
            del self._connection_pools[pool_key]
```

django/db/backends/oracle/features.py
```diff
@@ -139,6 +139,25 @@ def django_test_skips(self):
                     },
                 }
             )
+        if self.connection.settings_dict["OPTIONS"].get("pool"):
```
felixxm, line 142: Should not we use `is_pool` instead?

tests/test_utils/tests.py
```diff
@@ -2186,6 +2186,8 @@ def test_allowed_database_copy_queries(self):
         finally:
             new_connection.validate_thread_sharing()
             new_connection._close()
+            if hasattr(new_connection, "close_pool"):
+                new_connection.close_pool()
```
felixxm, line 2190: I think this should be in `_close()` for the Oracle backend instead :thinking:

django/db/backends/oracle/base.py
```diff
@@ -334,13 +385,21 @@ def check_constraints(self, table_names=None):
             cursor.execute("SET CONSTRAINTS ALL DEFERRED")
 
     def is_usable(self):
+        if self.connection is None:
+            return False
```
felixxm, line 389: This change looks unrelated :thinking:

tests/backends/oracle/tests.py
```diff
@@ -1,12 +1,31 @@
...
 
 from ..models import Square, VeryLongModelNameZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZ
 
+if connection.vendor == "oracle":
+    try:
+        from django.db.backends.oracle.oracledb_any import is_oracledb
+    except ImportError:
+        is_oracledb = False
+else:
+    is_oracledb = False
```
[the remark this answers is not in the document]
suraj-ora-2020: @felixxm I had a similar change in the commit [80c88f0](https://github.com/django/django/pull/17834/commits/80c88f077b3ce1d310e4dc39dd97f959a29613ad)
It was failing the CI tests.

django/db/backends/oracle/base.py
suraj-ora-2020: Done.

tests/test_utils/tests.py
suraj-ora-2020: The `_close()` function only releases the connection back to the pool, it doesn't close the pool itself. This means that the connection remains valid and connected to the database unless we close the pool itself. As a result, dropping the user at the end of the test suite completion was causing the error "cannot drop a user who is already connected". This issue is also present in Postgres, as the `_close` function in Postgres also only releases the connection back to the pool. Therefore, this change fixes the issue for both backends.

tests/backends/oracle/tests.py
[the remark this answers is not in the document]
felixxm: IMO, this behavior should be changed, #18557.

django/db/backends/oracle/base.py
suraj-ora-2020: Done.

django/db/backends/oracle/features.py
suraj-ora-2020: Updated.

django/db/backends/oracle/base.py
suraj-ora-2020: You are right it seems not useful with the current changes. I have removed this.

tests/backends/oracle/tests.py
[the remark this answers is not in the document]
suraj-ora-2020: I am satisfied with this simplification, but I had to step back earlier due to CI errors. @sarahboyce once https://github.com/django/django/pull/18557  get merged i will add this change.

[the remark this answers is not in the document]
sarahboyce: Merged in :+1: thank you both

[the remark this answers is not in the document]
suraj-ora-2020: Updated the code in my PR. Thanks.

suraj-ora-2020:
@sarahboyce @felixxm I have updated the code as per the review comments. Let me know if the changes seems okay. I will  superseded this PR by a new PR with the single  commit message `Fixed #7732 -- Added support for connection pooling in oracle backend`.

sarahboyce:
> I will supersed this PR with a new PR with the single commit message `Fixed #7732 -- Added support for connection pooling in oracle backend`.

Go ahead :+1:

felixxm:
> Let me know if the changes seems okay

I didn't have time to check this, but is it work with the minimal support version of `oracledb`, i.e. `1.3.2`, or do we need to bump that requirement.

suraj-ora-2020:
> > Let me know if the changes seems okay
> 
> I didn't have time to check this, but is it work with the minimal support version of `oracledb`, i.e. `1.3.2`, or do we need to bump that requirement.

Sure I will test once and update you.

suraj-ora-2020:
Superseded by #18568

PR #17847, opened 2024-02-10
Fixed #34917 . Links in the Django admin should be underlined.

Myrausman:
Fixed ticket [#34917](https://code.djangoproject.com/ticket/34917) - Links in the Django admin are now underlined.

Added CSS rule for underlining links in Django admin. 
for links i have added ` text-decoration: underline;`

### DIFF
--- a/django/contrib/admin/static/admin/css/base.css
+++ b/django/contrib/admin/static/admin/css/base.css
@@ -84,6 +84,8 @@ html[data-theme="light"],
         "Segoe UI Emoji",
         "Segoe UI Symbol",
         "Noto Color Emoji";
+
+    color-scheme: light;
 }
 
 html, body {
--- a/django/contrib/admin/static/admin/css/base.css
+++ b/django/contrib/admin/static/admin/css/base.css
@@ -105,15 +105,17 @@ body {
 
 a:link, a:visited {
     color: var(--link-fg);
-    text-decoration: none;
+    text-decoration: underline;
     transition: color 0.15s, background 0.15s;
 }
 
 a:focus, a:hover {
     color: var(--link-hover-color);
+    text-decoration: underline;
 }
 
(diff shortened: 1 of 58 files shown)


django/contrib/admin/static/admin/css/base.css
```diff
@@ -105,16 +105,18 @@ body {
...
 }
 
 a:focus, a:hover {
     color: var(--link-hover-color);
+    text-decoration: underline;
 }
 
 a:focus {
     text-decoration: underline;
+    text-decoration: underline;
```
knyghty, line 119: This is the same as the line above.

knyghty (commented):
Hi @Myrausman,

Thanks for this. In the ticket, I suggested that we should only underline the links in the main part of the page (not the sidebar or header, other than the links in the header that already underlined). Could you update your work here?

Myrausman:
```

/* LINKS */

a:link, a:visited {
    color: var(--link-fg);
    transition: color 0.15s, background 0.15s;
}

a:focus, a:hover {
    color: var(--link-hover-color);
}

a:focus {
    text-decoration: underline;
}
a img {
	@@ -123,7 +125,7 @@ a img {

a.section:link, a.section:visited {
    color: var(--header-link-color);
    text-decoration: none;
}
a.section:focus, a.section:hover {
    text-decoration: underline;
}

#content-main a:link,
#content-main a:visited {
    text-decoration: underline;
}


```
My apologies for not understanding clearly before. Can you confirm if this is okay, or should I make any other changes?

knyghty:
I suggest you try some things out and check locally if the right links are underlined and then update your code here.

Myrausman:
I believe there are some errors in this branch, including commits attributed to me that I did not make. Therefore, I've decided to create a new pull request.

PR #17848, opened 2024-02-10
Fixed #35179 -- Made admindocs detect positional/keyword-only arguments.

salvo-polizzi:
Fix [ticket 35179](https://code.djangoproject.com/ticket/35179)

### DIFF
--- a/django/utils/inspect.py
+++ b/django/utils/inspect.py
@@ -68,9 +68,7 @@ def func_accepts_var_args(func):
 
 def method_has_no_args(meth):
     """Return True if a method only accepts 'self'."""
-    count = len(
-        [p for p in _get_callable_parameters(meth) if p.kind == p.POSITIONAL_OR_KEYWORD]
-    )
+    count = len([p for p in _get_callable_parameters(meth) if p.kind in ARG_KINDS])
     return count == 0 if inspect.ismethod(meth) else count == 1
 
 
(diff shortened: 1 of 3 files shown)


django/utils/inspect.py
```diff
@@ -56,14 +56,22 @@ def get_func_full_args(func):
...
 def func_accepts_var_args(func):
     """
     Return True if function 'func' accepts positional arguments *args.
     """
-    return any(p for p in _get_callable_parameters(func) if p.kind == p.VAR_POSITIONAL)
+    return any(
+        p
+        for p in _get_callable_parameters(func)
+        if p.kind == p.VAR_POSITIONAL or p.kind == p.POSITIONAL_ONLY
+    )
```
shangxiao, line 74: I'm sorry but it appears this patch breaks these 2 functions.

`func_accepts_var_args()` should only return `True` for `VAR_POSITIONAL` - the name of the function and docstrings are clear on this 😊

as for `func_accepts_kwargs()` - the name is a misnomer but the docstrings state the intended purpose. If you look at its usage throughout the codebase though the intention that it only return `True` for `**kwargs`.

shangxiao (changes requested):
Hi thanks for the patch though I doubt this is the correct approach 🤷

django/utils/inspect.py
salvo-polizzi: Thanks. I had to read better before

shangxiao:
@salvo-polizzi thanks 🏆

PR #17876, opened 2024-02-19
Refs #35234 -- Deprecated CheckConstraint.check in favor of .condition.

charettes:
ticket-35234

The first three commits could likely be combined into a single one, decided to push them separated to ease review.

---

I think there is an argument to be made to move constraint related checks to their own namespace (e.g. `models.E041` -> `constraints.E001`) like we do with `fields` and also ultimately make the error message mention _which_ constraint (by name) is failing the check but that's another can of worms.

### DIFF
--- a/django/db/models/constraints.py
+++ b/django/db/models/constraints.py
@@ -134,19 +134,54 @@ class BaseConstraint:
 
 
 class CheckConstraint(BaseConstraint):
+    # RemovedInDjango60Warning: when the deprecation ends, replace with
+    # def __init__(
+    #  self, *, condition, name, violation_error_code=None, violation_error_message=None
+    # )
     def __init__(
-        self, *, check, name, violation_error_code=None, violation_error_message=None
+        self,
+        *,
+        name,
+        condition=None,
+        check=None,
+        violation_error_code=None,
+        violation_error_message=None,
     ):
-        self.check = check
-        if not getattr(check, "conditional", False):
+        if check is not None:
+            warnings.warn(
+                "CheckConstraint.check is deprecated in favor of `.condition`.",
+                RemovedInDjango60Warning,
+                stacklevel=2,
+            )
+            condition = check
+        self.condition = condition
+        if not getattr(condition, "conditional", False):
             raise TypeError(
-                "CheckConstraint.check must be a Q instance or boolean expression."
+                "CheckConstraint.condition must be a Q instance or boolean expression."
             )
         super().__init__(
             name=name,
             violation_error_code=violation_error_code,
             violation_error_message=violation_error_message,
         )
+
+    def _get_check(self):
+        warnings.warn(
+            "CheckConstraint.check is deprecated in favor of `.condition`.",
+            RemovedInDjango60Warning,
+            stacklevel=2,
+        )
+        return self.condition
+
+    def _set_check(self, value):
+        warnings.warn(
+            "CheckConstraint.check is deprecated in favor of `.condition`.",
+            RemovedInDjango60Warning,
+            stacklevel=2,
+        )
+        self.condition = value
+
+    check = property(_get_check, _set_check)
 
     def _check(self, model, connection):
         errors = []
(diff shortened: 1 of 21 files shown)


tests/invalid_models_tests/test_models.py
```diff
@@ -436,7 +436,7 @@ class Model(models.Model):
             fk_2 = models.ForeignKey(Target, models.CASCADE, related_name="target_2")
 
             class Meta:
-                constraints = [
+                indexes = [
```
charettes, line 439: This one made me wonder if we should have a check to ensure members of `Meta.constraints` are instances of `BaseConstraint`.

felixxm: I think it will fail anyway during migrations.

django/db/models/constraints.py
```diff
@@ -94,7 +94,7 @@ class CheckConstraint(BaseConstraint):
     def __init__(
         self, *, check, name, violation_error_code=None, violation_error_message=None
     ):
-        self.check = check
```
felixxm, line 97: Unfortunately, this is a [documented attribute](https://docs.djangoproject.com/en/5.0/ref/models/constraints/#check) so we cannot freely change its name.

felixxm:
I merged the 1st commit in abe6c1f93eb112066691b590052704948789c60f.

django/db/models/constraints.py
charettes: It's a shame really given it's the only abstraction that doesn't conform to the `.check` convention but I flipped the problem around and used `_check` instead for the method meant to be used to perform the system checks.

tests/invalid_models_tests/test_models.py
charettes: True but that's the case for any other problem we added a system check for.

felixxm:
Merged the first three commits in 0fb104dda287431f5ab74532e45e8471e22b58c8 & rebased.

django/contrib/postgres/constraints.py
```diff
@@ -77,6 +78,15 @@ def _get_expressions(self, schema_editor, query):
             expressions.append(expression)
         return ExpressionList(*expressions).resolve_expression(query)
 
+    def _check(self, model, connection):
+        references = set()
+        for expr, _ in self.expressions:
+            if isinstance(expr, str):
+                references.add(expr.split(LOOKUP_SEP))
```
felixxm, line 85: I would wrap in `F()` and use `_get_expr_references()` to get the same references for `F(lookup_string)` and `lookup_string`. Now, we have `list` vs. `tuple`.

felixxm:
I pushed the main change with extra tests to #17921.

django/db/models/constraints.py
```diff
@@ -135,19 +135,50 @@ def clone(self):
...
+            RemovedInDjango60Warning,
+            stacklevel=1,
+        )
+        return self.condition
+
+    def _set_check(self, value):
+        warnings.warn(
+            "`check` property of `CheckConstraint` is replaced by `condition`.",
+            RemovedInDjango60Warning,
+            stacklevel=1,
```
felixxm, line 176:
```suggestion
            stacklevel=2,
```

felixxm (commented):
@charettes Thanks :+1: I rebased and left small comments.

charettes:
All of these changes make sense to me @felixxm. Would you like me to push them myself?

felixxm:
> All of these changes make sense to me @felixxm. Would you like me to push them myself?

Pushed :) Let me double-check and merge.

django/db/models/constraints.py
```diff
@@ -135,19 +135,50 @@ def clone(self):
...
 class CheckConstraint(BaseConstraint):
     def __init__(
-        self, *, check, name, violation_error_code=None, violation_error_message=None
+        self,
+        *,
+        name,
+        condition=None,
+        check=None,
+        violation_error_code=None,
+        violation_error_message=None,
```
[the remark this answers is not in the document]
charettes: Should we also add a comment in `BaseConstraint._check` to rename to `check` once the deprecation period ends?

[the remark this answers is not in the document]
felixxm: We should have a separate ticket to rename `_check()` methods, use them, and [document](https://docs.djangoproject.com/en/5.0/topics/checks/#field-model-manager-and-database-checks) when the deprecation ends.

felixxm (approved):

django/db/models/constraints.py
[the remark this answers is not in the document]
charettes: Created ticket-35358 for this purpose.

PR #17883, opened 2024-02-19
Fixed #34917 -- Underlined Links in Django Admin.

Myrausman:
Fixed ticket [#34917](https://code.djangoproject.com/ticket/34917) - Links in the Django admin are now underlined.

### DIFF
--- a/django/contrib/admin/static/admin/css/changelists.css
+++ b/django/contrib/admin/static/admin/css/changelists.css
@@ -11,6 +11,14 @@
     min-width: 0;
 }
 
+#changelist .changelist-form-container a {
+    text-decoration: underline;
+}
+
+#changelist .changelist-form-container table thead a {
+    text-decoration: none;
+}
+
 #changelist table {
     width: 100%;
 }
@@ -48,6 +56,10 @@
     border-bottom: 1px solid var(--hairline-color);
     background: var(--body-bg);
     overflow: hidden;
+}
+
+#changelist .paginator a{
+    text-decoration: none;
 }
 
 /* CHANGELIST TABLES */


cheng-liven:
Can we have an option to turn this off easily? Most of our admin lists have links in every column and it will look aggressively.

(Editable fields in the list as shown in the example are probably not something very popular due to the risk of data lost from concurrent editing.)

knyghty:
@cheng-liven you can already easily turn it off by overriding the CSS, nothing more is needed here.

nessita:
Hello! Thanks for your contribution. Please remember that when this work is ready for review, you need to ensure that the title references the corresponding ticket and to adjust the ticket flags in the Trac system so this PR gets added to the "branches needing review" section of the [Django Developer Dahsboard](https://dashboard.djangoproject.com/). More information in [these docs](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/submitting-patches/#patch-style).

sarahboyce (commented):
Thank you for this @Myrausman this looks a great start ⭐ 

The scope of the ticket describes that all links in the "main content area" of the admin should be underlined. "main content area" is defined as "not the header or nav sidebar".

This currently only covers links for the admin change list. I beleive the scope is a bit larger, as other admin pages are not included here. 

Some things that come to mind and need investigating include:
- [rendering FileField links](https://github.com/django/django/blob/9a27c76021f934201cccf12215514a3091325ec8/django/contrib/admin/utils.py#L456)
- [rendering deleted objects links](https://github.com/django/django/blob/9a27c76021f934201cccf12215514a3091325ec8/django/contrib/admin/utils.py#L165)

We need a decision on more custom things like if a form added a link in a help text.

Can you investigate some of the other pages of the Django admin and confirm if we need to do more css updates to cover these?

Myrausman:
Hi, I apologize for not being able to work on this PR due to time constraints. I am closing it now. Thank you.

PR #17885, opened 2024-02-19
Fixed #26001 -- Fixed non-string field exact lookups in ModelAdmin.search_fields.

0saurabh0:
ticket : https://code.djangoproject.com/ticket/26001

 [reference](https://code.djangoproject.com/ticket/26001#comment:5)

### DIFF
--- a/django/contrib/admin/options.py
+++ b/django/contrib/admin/options.py
@@ -41,6 +41,7 @@ from django.core.exceptions import (
 from django.core.paginator import Paginator
 from django.db import models, router, transaction
 from django.db.models.constants import LOOKUP_SEP
+from django.db.models.functions import Cast
 from django.forms.formsets import DELETION_FIELD_NAME, all_valid
 from django.forms.models import (
     BaseInlineFormSet,
@@ -1207,9 +1208,33 @@ class ModelAdmin(BaseModelAdmin):
         may_have_duplicates = False
         search_fields = self.get_search_fields(request)
         if search_fields and search_term:
-            orm_lookups = [
-                construct_search(str(search_field)) for search_field in search_fields
-            ]
+            str_annotations = {}
+            orm_lookups = []
+            for field in search_fields:
+                if field.endswith("__exact"):
+                    field_name = field.rsplit("__exact", 1)[0]
+                    try:
+                        field_obj = queryset.model._meta.get_field(field_name)
+                    except FieldDoesNotExist:
+                        lookup = construct_search(field)
+                        orm_lookups.append(lookup)
+                        continue
+                    # Add string cast annotations for non-string exact lookups.
+                    if not isinstance(field_obj, (models.CharField, models.TextField)):
+                        str_annotations[f"{field_name}_str"] = Cast(
+                            field_name, output_field=models.CharField()
+                        )
+                        orm_lookups.append(f"{field_name}_str__exact")
+                    else:
+                        lookup = construct_search(field)
+                        orm_lookups.append(lookup)
+                else:
+                    lookup = construct_search(str(field))
+                    orm_lookups.append(lookup)
+
+            if str_annotations:
+                queryset = queryset.annotate(**str_annotations)
+
             term_queries = []
             for bit in smart_split(search_term):
                 if bit.startswith(('"', "'")) and bit[0] == bit[-1]:
--- a/tests/admin_changelist/tests.py
+++ b/tests/admin_changelist/tests.py
@@ -860,6 +860,25 @@ class ChangeListTests(TestCase):
         cl = m.get_changelist_instance(request)
         self.assertCountEqual(cl.queryset, [abcd])
 
+    def test_search_with_exact_lookup_for_non_string_field(self):
+        child = Child.objects.create(name="Asher", age=11)
+        model_admin = ChildAdmin(Child, custom_site)
+
+        for search_term, expected_result in [
+            ("11", [child]),
+            ("Asher", [child]),
+            ("1", []),
+            ("A", []),
+            ("random", []),
+        ]:
+            request = self.factory.get("/", data={SEARCH_VAR: search_term})
+            request.user = self.superuser
+            with self.subTest(search_term=search_term):
+                # 1 query for filtered result, 1 for filtered count, 1 for total count.
+                with self.assertNumQueries(3):
+                    cl = model_admin.get_changelist_instance(request)
+                self.assertCountEqual(cl.queryset, expected_result)
+
     def test_no_distinct_for_m2m_in_list_filter_without_params(self):
         """
         If a ManyToManyField is in list_filter but isn't in any lookup params,
--- a/tests/admin_changelist/admin.py
+++ b/tests/admin_changelist/admin.py
@@ -48,6 +48,7 @@ class ChildAdmin(admin.ModelAdmin):
     list_display = ["name", "parent"]
     list_per_page = 10
     list_filter = ["parent", "age"]
+    search_fields = ["age__exact", "name__exact"]
 
     def get_queryset(self, request):
         return super().get_queryset(request).select_related("parent")


django/contrib/admin/options.py
```diff
@@ -1213,7 +1213,14 @@ def construct_search(field_name):
...
                 )
                 term_queries.append(or_queries)
-            queryset = queryset.filter(models.Q.create(term_queries))
+            try:
+                queryset = queryset.filter(models.Q.create(term_queries))
+            except (ValueError, ValidationError):
+                # If a ValueError occurs
+                # (e.g. when a non-integer term is used with __exact)
+                # return an empty queryset.
+                return queryset.none(), may_have_duplicates
```
sarahboyce, line 1222: I think returning no results is not the desired behavior
Let's say I have a `Person` model with an `age` IntegerField and `name` CharField
I have a `Person(name="Alice", age=10)`
If I have `search_fields = ["age", "name__exact"]`, I search:
- `Alice`, and I get results for alice
- `10`, and I get results for alice
- `1`, and I get results for alice (becuase it isn't `exact`)
I update to `search_fields = ["age__exact", "name__exact"]`, I search:
- `Alice`, and I get no results (:x: this is incorrect, I should get alice, but the queryset is noned out)
- `10`, and I get results for alice
- `1`, and I get no results (correct)
This means you effectively cannot search the name and can only search for the age when we really want an "OR" behavior
Ideally I'd want us to check whether the `search_term` is valid for any of the filters and only filter against the valid ones

0saurabh0: Hi Sarah thanks for the review.
I get what you're saying here and will push the according changes soon.

django/contrib/admin/options.py
```diff
@@ -1214,12 +1215,31 @@ def construct_search(field_name):
...
-            queryset = queryset.filter(models.Q.create(term_queries))
+
+                field_queries = []
+                for orm_lookup in orm_lookups:
+                    try:
+                        # Attempt to create a query for each field
+                        field_query = models.Q(**{orm_lookup: bit})
+                        queryset.filter(
+                            field_query
+                        ).exists()  # Test if the query is valid
```
sarahboyce, line 1226: This is better but this adds extra queries
I think we might be able to annotate values which are cast to strings and search these when the lookup is "exact" and it is not a Char or TextField

0saurabh0: Ah, good 👀.
Will fix soon.

0saurabh0:
@sarahboyce have a look

tests/admin_changelist/tests.py
```diff
@@ -860,6 +860,36 @@ def test_custom_lookup_with_pk_shortcut(self):
...
+
+        request = self.factory.get("/", data={SEARCH_VAR: "fsfsl"})
+        request.user = self.superuser
+        cl = model_admin.get_changelist_instance(request)
+        self.assertCountEqual(cl.queryset, [])
+
+        request = self.factory.get("/", data={SEARCH_VAR: "3.14"})
+        request.user = self.superuser
+        cl = model_admin.get_changelist_instance(request)
+        self.assertCountEqual(cl.queryset, [])
```
sarahboyce, line 891:
```suggestion
    def test_search_with_exact_lookup_for_non_string_field(self):
        child = Child.objects.create(name="Asher", age=11)
        model_admin = ChildAdmin(Child, custom_site)

        for search_term, expected_result in [
            ("11", [child]),
            ("Asher", [child]),
            ("1", []),
            ("A", []),
            ("random", []),
        ]:
            request = self.factory.get("/", data={SEARCH_VAR: search_term})
            request.user = self.superuser
            with self.subTest(search_term=search_term):
                # 1 query for filtered result, 1 for filtered count, 1 for total count.
                with self.assertNumQueries(3):
                    cl = model_admin.get_changelist_instance(request)
                self.assertCountEqual(cl.queryset, expected_result)
```
I think using `subTest` and `assertNumQueries` would be nice here

django/contrib/admin/options.py
```diff
@@ -1201,12 +1202,49 @@ def construct_search(field_name):
                     if hasattr(field, "path_infos"):
                         # Update opts to follow the relation.
                         opts = field.path_infos[-1].to_opts
+            # Return the field name as-is if it already includes __exact
```
sarahboyce, line 1205:
```suggestion
            # Return the field name as-is if it already includes __exact.
```
(nit)

django/contrib/admin/options.py
```diff
@@ -1219,7 +1257,10 @@ def construct_search(field_name):
                     connector=models.Q.OR,
                 )
                 term_queries.append(or_queries)
-            queryset = queryset.filter(models.Q.create(term_queries))
+            try:
+                queryset = queryset.filter(models.Q.create(term_queries))
+            except (ValueError, ValidationError):
+                return queryset.none(), may_have_duplicates
```
sarahboyce, line 1263:
```suggestion
            queryset = queryset.filter(models.Q.create(term_queries))
```
No longer needed

django/contrib/admin/options.py
```diff
@@ -1201,12 +1202,49 @@ def construct_search(field_name):
...
+                            (models.CharField, models.TextField),
+                        )
+                        else field
+                    )
+                    for field in search_fields
+                ]
+
             orm_lookups = [
                 construct_search(str(search_field)) for search_field in search_fields
             ]
```
sarahboyce, line 1250: We should be able to build `orm_lookups` as we loop through the `search_fields` to avoid multiple loops through  `search_fields`
```suggestion
            # Add string cast annotations for non-string exact lookups.
            str_annotations = {}
            orm_lookups = []
            for field in search_fields:
                if field.endswith("__exact"):
                    field_name = field.rsplit("__exact", 1)[0]
                    try:
                        field_obj = queryset.model._meta.get_field(field_name)
                    except FieldDoesNotExist:
                        lookup = construct_search(field)
                        orm_lookups.append(lookup)
                        continue
                    if not isinstance(field_obj, (models.CharField, models.TextField)):
                        str_annotations[f"{field_name}_str"] = Cast(
                            field_name, output_field=models.CharField()
                        )
                        lookup = construct_search(f"{field_name}_str__exact")
                        orm_lookups.append(lookup)
                    else:
                        lookup = construct_search(field)
                        orm_lookups.append(lookup)
                else:
                    lookup = construct_search(str(field))
                    orm_lookups.append(lookup)
```

sarahboyce (commented):
Thank you for the updates @0saurabh0 this looks really strong! 

I have a few more comments, when you're done with them can you squash this down to a single commit?

django/contrib/admin/options.py
0saurabh0: Nice!  
Slightly more complex logic within the loop, but definitely reduces the number of iterations
let me just quickly run it locally?

sarahboyce: Yeah sure, feel free to simplify it also 👍

django/contrib/admin/options.py
```diff
@@ -1201,15 +1202,39 @@ def construct_search(field_name):
...
+                            field_name, output_field=models.CharField()
+                        )
+                        lookup = construct_search(f"{field_name}_str__exact")
+                        orm_lookups.append(lookup)
+                    else:
+                        lookup = construct_search(field)
+                        orm_lookups.append(lookup)
+                else:
+                    lookup = construct_search(str(field))
+                    orm_lookups.append(lookup)
```
sarahboyce, line 1237:
```suggestion
                    orm_lookups.append(lookup)

            if str_annotations:
                queryset = queryset.annotate(**str_annotations)
```
(sorry I forgot this)

django/contrib/admin/options.py
```diff
@@ -1201,15 +1202,39 @@ def construct_search(field_name):
...
+                    except FieldDoesNotExist:
+                        lookup = construct_search(field)
+                        orm_lookups.append(lookup)
+                        continue
+                    if not isinstance(field_obj, (models.CharField, models.TextField)):
+                        str_annotations[f"{field_name}_str"] = Cast(
+                            field_name, output_field=models.CharField()
+                        )
+                        lookup = construct_search(f"{field_name}_str__exact")
+                        orm_lookups.append(lookup)
```
sarahboyce, line 1231: We can get rid of above
```python
            # Return the field name as-is if it already includes __exact.
            if field_name.endswith("__exact"):
                return field_name
```
If we update this to:
```suggestion
                        orm_lookups.append(f"{field_name}_str__exact")
```

0saurabh0:
> can you squash this down to a single commit?

I can definitely squash the commits but when I attempt to rebase/reset, I see several other commits authored by different contributors, probably due to merging `main` into this branch. 
If everything looks good, you can go with squash and merge.

sarahboyce (approved):
Thank you @0saurabh0 - this looks good to me :+1:

PR #17886, opened 2024-02-19
Refs #35224 -- Removed special-case GenericForeignKey system check.

adamchainz:
ticket-35224

Profiling checks after 6002df06713cb0a7050432263527a25754190c27 was merged, I found that `check_generic_foreign_keys()` was taking ~2% of the total time. It seems that iterating over all fields of all models is fairly expensive.

The special check function can be eliminated now that `GenericForeignKey` if we make `Model._check_fields` cover private fields as well. That seems like a change we’d want anyway.


tests/model_fields/models.py
```diff
@@ -411,6 +411,12 @@ class Meta:
         required_db_features = {"supports_json_field"}
 
 
+class GfkModel(models.Model):
+    object_id = models.PositiveIntegerField()
+    content_type = models.ForeignKey(ContentType, models.CASCADE)
+    gfk = GenericForeignKey()
```
adamchainz, line 417: Needed for the below `GenericRelation` to have something to refer to, otherwise the `contenttypes.E004` check fails.

tests/model_meta/models.py
```diff
@@ -5,7 +5,9 @@
 
 
 class Relation(models.Model):
-    pass
+    content_type = models.ForeignKey(ContentType, models.CASCADE, related_name="+")
+    object_id = models.PositiveIntegerField()
+    content_object = GenericForeignKey()
```
adamchainz, line 10: As above, fields required to fill in for `GenericRelation`s below, preventing `contenttypes.E004` errors.

django/db/models/base.py
```diff
@@ -1789,6 +1789,8 @@ def _check_fields(cls, **kwargs):
         errors = []
         for field in cls._meta.local_fields:
             errors.extend(field.check(**kwargs))
+        for field in cls._meta.private_fields:
+            errors.extend(field.check(**kwargs))
```
adamchainz, line 1793: One outstanding wrinkle: `private_fields` are [copied into subclasses](https://github.com/django/django/blob/b6f07e2dc69b9a55606f26ac8f13769531148c73/django/db/models/base.py#L347-L350), so checks for one field get repeated in subclasses.

I don’t think there’s a way with the current attributes to tell if a field is local *and* private right now. Maybe more attributes or reorganization could work.

django/contrib/contenttypes/fields.py
```diff
@@ -357,7 +357,7 @@ def __init__(
 
     def check(self, **kwargs):
         return [
-            *super().check(**kwargs),
+            *self._check_field_name(),
```
adamchainz, line 360: After the fix to make the `GenericRelation` checks run, the test suite showed many errors: fields.E304, fields.E305, and fields.E311. For example:

```
aggregation_regress.HardbackBook.tags: (fields.E311) 'ItemTag.book_ptr' must be unique because it is referenced by a foreign key.
    HINT: Add unique=True to this field or add a UniqueConstraint (without condition) in the model Meta.constraints.

generic_relations_regress.Cafe.links: (fields.E304) Reverse accessor 'Link.places' for 'generic_relations_regress.Cafe.links' clashes with reverse accessor for 'generic_relations_regress.Place.links'.
    HINT: Add or change a related_name argument to the definition for 'generic_relations_regress.Cafe.links' or 'generic_relations_regress.Place.links'.
```

I think these are false positives. The easiest way to prevent them was to modify `GenericRelation.check()` to not call `super()`, like `GenericForeignKey.check()`. I tried to look through all existing checks for what’s relevant, and the field name check is the only one that really stood out.

nessita:
@adamchainz I'm reviewing code around ticket-35224 to do a full review of ticket-36151. Is this PR still relevant? Could we close it if not?

PR #17893, opened 2024-02-20
Fixed #35241 -- Cached model's full parent list.

adamchainz:
ticket-35241

### DIFF
--- a/django/db/models/deletion.py
+++ b/django/db/models/deletion.py
@@ -305,13 +305,11 @@ class Collector:
         if not collect_related:
             return
 
-        if keep_parents:
-            parents = set(model._meta.get_parent_list())
         model_fast_deletes = defaultdict(list)
         protected_objects = defaultdict(list)
         for related in get_candidate_relations_to_delete(model._meta):
             # Preserve parent reverse relationships if keep_parents=True.
-            if keep_parents and related.model in parents:
+            if keep_parents and related.model in model._meta.all_parents:
                 continue
             field = related.field
             on_delete = field.remote_field.on_delete
--- a/django/db/models/options.py
+++ b/django/db/models/options.py
@@ -692,16 +692,24 @@ class Options:
                 return res
         return []
 
-    def get_parent_list(self):
+    @cached_property
+    def all_parents(self):
         """
-        Return all the ancestors of this model as a list ordered by MRO.
+        Return all the ancestors of this model as a tuple ordered by MRO.
         Useful for determining if something is an ancestor, regardless of lineage.
         """
         result = OrderedSet(self.parents)
         for parent in self.parents:
-            for ancestor in parent._meta.get_parent_list():
+            for ancestor in parent._meta.all_parents:
                 result.add(ancestor)
-        return list(result)
+        return tuple(result)
+
+    def get_parent_list(self):
+        """
+        Return all the ancestors of this model as a list ordered by MRO.
+        Backward compatibility method.
+        """
+        return list(self.all_parents)
 
     def get_ancestor_link(self, ancestor):
         """
--- a/django/forms/models.py
+++ b/django/forms/models.py
@@ -1214,20 +1214,19 @@ def _get_foreign_key(parent_model, model, fk_name=None, can_fail=False):
         fks_to_parent = [f for f in opts.fields if f.name == fk_name]
         if len(fks_to_parent) == 1:
             fk = fks_to_parent[0]
-            parent_list = parent_model._meta.get_parent_list()
-            parent_list.append(parent_model)
+            all_parents = (*parent_model._meta.all_parents, parent_model)
             if (
                 not isinstance(fk, ForeignKey)
                 or (
                     # ForeignKey to proxy models.
                     fk.remote_field.model._meta.proxy
-                    and fk.remote_field.model._meta.proxy_for_model not in parent_list
+                    and fk.remote_field.model._meta.proxy_for_model not in all_parents
                 )
                 or (
                     # ForeignKey to concrete models.
                     not fk.remote_field.model._meta.proxy
                     and fk.remote_field.model != parent_model
-                    and fk.remote_field.model not in parent_list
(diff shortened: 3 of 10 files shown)


adamchainz:
Testing this fix on the project, I found that Wagtail, at least, calls `get_parent_list()`. The below patch adds a backwards compatibility method. Should we do that, or is it better that projects using this internal method update?

```diff
diff --git django/db/models/options.py django/db/models/options.py
index d11ca85154..3cea5a0d0a 100644
--- django/db/models/options.py
+++ django/db/models/options.py
@@ -704,6 +704,10 @@ def parent_list(self):
                 result.add(ancestor)
         return list(result)
 
+    def get_parent_list(self):
+        # Backwards compatibility method
+        return self.parent_list
+
     def get_ancestor_link(self, ancestor):
         """
         Return the field on the current model which points to the given
diff --git tests/model_meta/tests.py tests/model_meta/tests.py
index bdcc991aee..f2fa32deaa 100644
--- tests/model_meta/tests.py
+++ tests/model_meta/tests.py
@@ -334,6 +334,9 @@ def test_parent_list(self):
             Child._meta.parent_list, [FirstParent, SecondParent, CommonAncestor]
         )
 
+    def test_get_parent_list(self):
+        self.assertEqual(Child._meta.get_parent_list(), Child._meta.parent_list)
+
 
 class PropertyNamesTests(SimpleTestCase):
     def test_person(self):
```

adamchainz:
There were some non-deterministic checks failures due to my missing that `django.forms.models._get_foreign_key()` *mutates* the return list. I fixed that and have tried changing the function to return a tuple instead, which will prevent any other such errors. Maybe we should rename the attribute though - to e.g. `parents_mro` or something? `parents` already contains the direct parents.

django/forms/models.py
```diff
@@ -1214,8 +1214,7 @@ def _get_foreign_key(parent_model, model, fk_name=None, can_fail=False):
         fks_to_parent = [f for f in opts.fields if f.name == fk_name]
         if len(fks_to_parent) == 1:
             fk = fks_to_parent[0]
-            parent_list = parent_model._meta.get_parent_list()
-            parent_list.append(parent_model)
+            parent_list = [*parent_model._meta.parent_list, parent_model]
```
kezabelle, line 1217: Where you mentioned that this method:
> [...] _mutates_ the return list [...]

Did you mean the `.append` that was here, or does additional mutation occur somewhere in here and I'm just being blind to it? If it's the former, we could also do:
```python
parent_list = (*parent_model._meta.parent_list, parent_model)
```
to illustrate that, maybe? Doesn't make a blind lick of difference in the scheme of things, and I only bring it up precisely because you called out a _mutation_ occurring...

nessita:
> Testing this fix on the project, I found that Wagtail, at least, calls `get_parent_list()`. The below patch adds a backwards compatibility method. Should we do that, or is it better that projects using this internal method update?

While `get_parent_list` is not publicly documented, I'm not comfortable with removing it, considering is "cheap" for us to maintain some backward compatibility. Furthermore, I'm also not comfortable with changing its type from list to tuple, for similar reasons, so I think that the compatibility shim should be:

```python
   def get_parent_list(self):
       # Backwards compatibility method
        return list(self.parent_list)
```

nessita:
Further thoughts: with my last comment in mind, we could certainly rename it to something more accurate/meaningful. +1 to that!

adamchainz:
I went with `all_parents`, how does that sound?

django/db/models/options.py
```diff
@@ -692,16 +692,21 @@ def get_base_chain(self, model):
...
         result = OrderedSet(self.parents)
         for parent in self.parents:
-            for ancestor in parent._meta.get_parent_list():
+            for ancestor in parent._meta.all_parents:
                 result.add(ancestor)
-        return list(result)
+        return tuple(result)
+
+    def get_parent_list(self):
+        # Backwards compatibility method
```
smithdc1, line 708: For `en-us`:

```suggestion
        # Backward compatibility method.
```

django/db/models/options.py
```diff
@@ -692,16 +692,21 @@ def get_base_chain(self, model):
                 return res
         return []
 
-    def get_parent_list(self):
+    @cached_property
+    def all_parents(self):
         """
         Return all the ancestors of this model as a list ordered by MRO.
```
pauloxnet, line 698:
```suggestion
        Return all the ancestors of this model as a tuple ordered by MRO.
```

django/db/models/options.py
pauloxnet, line 708: Can we maintain here the docstring and add the backward note?

django/db/models/options.py
adamchainz: Done.

django/db/models/options.py
adamchainz: Done.

django/db/models/deletion.py
```diff
@@ -306,7 +306,7 @@ def collect(
             return
 
         if keep_parents:
-            parents = set(model._meta.get_parent_list())
+            parents = set(model._meta.all_parents)
```
nessita, line 309: Now that `all_parents` is a cached property, I think this part of the code can be removed and later just do:

```python
            # Preserve parent reverse relationships if keep_parents=True.
            if keep_parents and related.model in model._meta.all_parents:
                continue
```
More so because the highlighted `if` branch is defining a `parents` variable that will produce a `NameError` if `keep_parents` is False (and that is hard to spot in the code, and also a code smell IMHO).

nessita (commented):
Looks good! One final suggestion and this is ready to be merged :partying_face:

django/db/models/deletion.py
adamchainz: Nice spot. The suggestion *will* change the lookup behaviour from O(1) to O(n), but I think that is likely okay given the typical shallowness of model inheritance where 10 layers would be excessive.

PR #17902, opened 2024-02-23
Fixed #20960 - moved DEFAULT_TABLESPACE to DATABASES

bcail:


tests/model_options/test_tablespaces.py
```diff
@@ -133,3 +136,47 @@ def test_tablespace_for_many_to_many_field(self):
...
+
+        # 1 for the table + 1 for the index on the primary key
+        self.assertNumContains(sql, tablespace, 2)
+
+    @skipUnlessDBFeature("supports_tablespaces")
+    @isolate_apps("model_options")
+    def test_default_index_tablespace(self):
+        index_tablespace = "default_index_tablespace"
+        databases = copy.deepcopy(settings.DATABASES)
+        databases["default"]["OPTIONS"]["DEFAULT_INDEX_TABLESPACE"] = index_tablespace
```
charettes, line 173: I could be wrong here but I _think_ these options should live at the same level as `OPTIONS` and not be nested within `OPTIONS`?

In other words, I think the setting should be defined as 

```
DATABASES = {
    "default": {
        "ENGINE": "...",
        ...,
        "DEFAULT_TABLESPACE": "..."
        "DEFAULT_INDEX_TABLESPACE": "...",
     }
}
```

As `OPTIONS` is meant to be per-backend flags that are usually passed directly to the underlying client bindings.

bcail: Thanks, @charettes. The [ticket comment](https://code.djangoproject.com/ticket/20960#comment:2) seems to suggest they should be inside "OPTIONS", but I'm fine with whatever.  @aaugustin do you have any comments?

aaugustin: Sorry, I haven't been involved in developing Django for several years now and I'm unable to help there.

bcail: No problem, thanks.

I'll update the code to move these options up to the same level as `OPTIONS`.

bcail: @charettes @felixxm @shangxiao I've updated the code to move those options up.

django/db/backends/base/schema.py
```diff
@@ -276,10 +276,15 @@ def table_sql(self, model):
...
-            tablespace_sql = self.connection.ops.tablespace_sql(
-                model._meta.db_tablespace
-            )
+        db_tablespace = None
+        if settings.DATABASES[self.connection.alias].get("DEFAULT_TABLESPACE"):
+            db_tablespace = settings.DATABASES[self.connection.alias][
+                "DEFAULT_TABLESPACE"
+            ]
+        elif model._meta.db_tablespace:
+            db_tablespace = model._meta.db_tablespace
```
hartungstenio, line 285: This way the DEFAULT_TABLESPACE has a higher priority than the model's Meta tablespace.

I think it's because of the DEFAULT_TABLESPACE [assignment](https://github.com/django/django/blob/main/django/db/models/options.py#L125) in options, but it might lead to unexpected behaviour if someone is not using the default.

bcail: Thanks, @hartungstenio - I adjusted it.

Alok-work23:
I am planning to continue this PR since it appears to be stalled. I will base updated work on current main, resolve conflicts, and address the deprecation-warning related test failures.
Credit for the original work is fully retained.

PR #17909, opened 2024-02-26
Refs #32114 -- Fixed test crash on non-picklable objects in subtests when PickleError is raised.

David-Wobrock:
Refs https://github.com/django/django/pull/17650 and [ticket-32114](https://code.djangoproject.com/ticket/32114).

Probably related to https://github.com/python/cpython/issues/73373.
Reproducing the same behaviour between CPython 3.10 and PyPy 3.10


felixxm (approved):
@David-Wobrock Thanks :+1:

PR #17910, opened 2024-02-26
Fixed #35189 -- Render admin collapsible fieldsets with <details>.

MHLut:
TL;DR: Improve the accessibility of admin fieldsets and collapse them with a native HTML implementation.

- Ticket: [#35189](https://code.djangoproject.com/ticket/35189)
- Proof of concept: [Example 2: Details inside (no JS)](https://codepen.io/MHLut/pen/LYaqYrz)
- [Initial discussion on Discord, accessibility channel ](https://discord.com/channels/856567261900832808/1207376857800900618)

## Related tickets and PRs

Related items:

- [Ticket #35382](https://code.djangoproject.com/ticket/35382); this ticket removed unused CSS temporarily unearthed by this PR.
- [Ticket #35380](https://code.djangoproject.com/ticket/35380); documentation screenshots aren't being automatically generated.
- https://github.com/django/django/pull/18080; add more forced-colors screenshots.

## Important facts

- Works with the main model fieldset and stacked inline fieldsets.
- This change removes all JavaScript for collapsible fieldsets.
- I have recycled as much CSS as possible.
- JavaScript used to check if a fieldset could be collapsible. With JS gone, I moved that check to the Fieldset helper class in Python.
- The translations for "Show" and "Hide" are now redundant.

## Known issues

- No existing CSS bugs were fixed. 
  - Example: The font size of the fieldset is below 16px/1rem.
  - Example: The fieldset heading is hard to distinguish in `forced-colors` mode.
  - Context: Style for `.collapse. h3` in `forms.css` was copied from `.module h2, .module caption, .inline-group h2` in `base.css`.

Note: Several CSS fixes that we could make would cause scope creep beyond the collapsible fieldsets, so I recommend fixing them in separate PRs.


carltongibson:
buildbot, test on selenium.

MHLut:
I have removed the draft status of the PR. I expect that people will have feedback on the look and feel of the feature and that the code will require some changes. As such, I have not fixed every Selenium test yet.

MHLut:
I didn't realize you can make inline admins collapsible as well. I will work on adding that feature.

tests/admin_inlines/tests.py
```diff
@@ -1885,17 +1883,15 @@ def test_added_stacked_inline_with_collapsed_fields(self):
         )
         self.selenium.find_element(By.LINK_TEXT, "Add another Child").click()
         test_fields = ["#id_child_set-0-name", "#id_child_set-1-name"]
-        show_links = self.selenium.find_elements(By.LINK_TEXT, "SHOW")
-        self.assertEqual(len(show_links), 2)
+        summaries = self.selenium.find_elements(By.TAG_NAME, "summary")
+        self.assertEqual(len(summaries), 3)
```
MHLut, line 1887: I increased the count from 2 to 3 here on purpose.

There are two visible fieldsets, but a third one is hidden in the code that is used to copy a new fieldset when clicking "Add another Child".

You can see that `test_collapsed_inlines()` already looked for three elements.

sarahboyce:
Hi @MHLut I think this looks really promising thank you ⭐ 
I've found something I'd like you to look into. I had some screenshots generated from this PR. You can see these [here when you download the **Artifacts**](https://github.com/django/django/actions/runs/8663609768).
These screenshots are taken under a few scenarios we try to support include dark mode and right to left languages (side note we should think about adding forced-colors mode to these).

I can see on your PR that in RTL we have this:
test_selectbox_height_collapsible_fieldset_rtl-selectbox-collapsible

Collapisble is on the left here where as before it was on the right
test_selectbox_height_collapsible_fieldset_rtl-selectbox-collapsible

I am also not sure if the collapse icon (> triangle thingy) should be on the left, so before Fields in the right to left case (maybe @sakhawy might be able to tell us 🙏 ) 

You might need to look into `django/contrib/admin/static/admin/css/rtl.css` 👍

sakhawy:
@sarahboyce I guess it "feels off" because we're used to seeing dropdowns with the triangle after the text and not before it :D 
But I think the triangle here is part of the `<details>` tag's default style, so nothing to worry about!

One reason for checking the `django/contrib/admin/static/admin/css/rtl.css` file would be this line: 
https://github.com/django/django/blob/88274f6997b73326453695db8607822b0922d89f/django/contrib/admin/static/admin/css/forms.css#L225

I think it's the reason 'Collapsible' is on the left when `dir="rtl"`.

MHLut:
@sarahboyce @sakhawy Thanks for the heads-up. I can't remember why I put the `text-align: left` in `forms.py`, but it's gone now. The heading aligns to the right side again in RTL mode.

The triangle is—as Moe mentioned—part of `<details>` native styling; I'd rather not mess with that. Since Moe seems OK with it, I'll keep that as-is.

FYI: I'm currently doing some free-form CSS experiments for forced-colors mode.

knyghty (commented):
I didn't look much at the code, or look at it on a screen.

From an accessibility standpoint I think we're good with this approach. The outline seems a little odd still (that there are multiple h3s per link, rather than one h3 and then going to h4 or something similar) but I think this is better than what we currently have and shouldn't block this merge - to be honest I'm not even sure if it's a problem at all, I could be totally wrong.

If I find time before someone else I will also review the code.

MHLut:
I've added a border around the collapsible fieldsets so that their fields look more grouped, just like the original collapsible widget.

The collapsible-fieldset-within-an-inline looks OK enough. It also still has the heading level bug (H3 instead of H4). We can argue it's still better than before and not a blocker, as @knyghty mentioned above.

Screenshot 2024-04-1
![Screenshot 2024-04-15 at 20 57 12
Screenshot 2024-04-15 at 20 57 35
5 at 20 57 02](https://github.com/django/django/assets/2210890/975b30f9-ae7f-4e1e-83e7-cfb900a12879)
Screenshot 2024-04-15 at 20 57 23

tests/admin_inlines/tests.py
```diff
@@ -1853,7 +1853,7 @@ def test_delete_inlines(self):
     def test_collapsed_inlines(self):
```
sarahboyce, line 1853: I think it's worth us adding in some screenshots to these tests. Something like:
```diff
diff --git a/tests/admin_inlines/tests.py b/tests/admin_inlines/tests.py
index 91a8affba5..40db1dd040 100644
--- a/tests/admin_inlines/tests.py
+++ b/tests/admin_inlines/tests.py
@@ -4,7 +4,9 @@ from django.contrib.admin.tests import AdminSeleniumTestCase
 from django.contrib.auth.models import Permission, User
 from django.contrib.contenttypes.models import ContentType
 from django.test import RequestFactory, TestCase, override_settings
+from django.test.selenium import screenshot_cases
 from django.urls import reverse
+from django.utils.translation import gettext

 from .admin import InnerInline
 from .admin import site as admin_site
@@ -1850,6 +1852,7 @@ class SeleniumTests(AdminSeleniumTestCase):
             "form#profilecollection_form tr.dynamic-profile_set#profile_set-2", 1
         )

+    @screenshot_cases(["desktop_size", "mobile_size", "rtl", "dark"])
     def test_collapsed_inlines(self):
         from selenium.webdriver.common.by import By

@@ -1865,15 +1868,19 @@ class SeleniumTests(AdminSeleniumTestCase):
         ]
         summaries = self.selenium.find_elements(By.TAG_NAME, "summary")
         self.assertEqual(len(summaries), 3)
+        self.take_screenshot("loaded")
         for show_index, field_name in enumerate(test_fields, 0):
             self.wait_until_invisible(field_name)
             summaries[show_index].click()
             self.wait_until_visible(field_name)
+        self.take_screenshot("expanded")
         for hide_index, field_name in enumerate(test_fields, 0):
             self.wait_until_visible(field_name)
             summaries[hide_index].click()
             self.wait_until_invisible(field_name)
+        self.take_screenshot("collapsed")

+    @screenshot_cases(["desktop_size", "mobile_size", "rtl", "dark"])
     def test_added_stacked_inline_with_collapsed_fields(self):
         from selenium.webdriver.common.by import By

@@ -1881,18 +1888,22 @@ class SeleniumTests(AdminSeleniumTestCase):
         self.selenium.get(
             self.live_server_url + reverse("admin:admin_inlines_teacher_add")
         )
-        self.selenium.find_element(By.LINK_TEXT, "Add another Child").click()
+        add_text = gettext("Add another %(verbose_name)s") % {"verbose_name": "Child"}
+        self.selenium.find_element(By.LINK_TEXT, add_text).click()
         test_fields = ["#id_child_set-0-name", "#id_child_set-1-name"]
         summaries = self.selenium.find_elements(By.TAG_NAME, "summary")
         self.assertEqual(len(summaries), 3)
+        self.take_screenshot("loaded")
         for show_index, field_name in enumerate(test_fields, 0):
             self.wait_until_invisible(field_name)
             summaries[show_index].click()
             self.wait_until_visible(field_name)
+        self.take_screenshot("expanded")
         for hide_index, field_name in enumerate(test_fields, 0):
             self.wait_until_visible(field_name)
             summaries[hide_index].click()
             self.wait_until_invisible(field_name)
+        self.take_screenshot("collapsed")

     def assertBorder(self, element, border):
         width, style, color = border.split(" ")
```

This gives me a couple more screenshots that I think we might need to do some css tweaking for, specifically...
"Name" is on the left for RTL and generally the contrast of this looks not strong enough to me
image

I haven't investigated if these are existing issues yet 🤔 might well be

django/contrib/admin/helpers.py
```diff
@@ -438,6 +443,12 @@ def inline_formset_data(self):
     def forms(self):
         return self.formset.forms
 
+    @cached_property
+    def is_collapsible(self):
+        if any(self.formset.errors):
+            return False
+        return "collapse" in self.classes
+
```
sarahboyce, line 451: Is this and the changes to `django/contrib/admin/templates/admin/edit_inline/tabular.html` and `django/contrib/admin/templates/admin/edit_inline/stacked.html` required?
The way I read the existing feature it looks like it is applied only for fieldsets rather than formsets.

sarahboyce (commented):
We also need to tweak the docs: https://docs.djangoproject.com/en/5.0/ref/contrib/admin/
It says things like: 
> with a collapse class will be initially collapsed and their header will have a small “show” link.
And an existing wrong sounding:
> Fieldsets with the collapse style will be initially collapsed in the admin and replaced with a small “click to expand” link.

We probably need a "versionchanged" note and release note also 👍

django/contrib/admin/helpers.py
MHLut: Yes it is. Formsets can also be collapsible; you can put a collapsible fieldset inside a collapsible stacked or tabular inline.

The original logic for fieldsets is that they should not be collapsible when one of its fields has an error. Originally, JavaScript performed that check.

It makes sense to restrict complete inlines from collapsing in case of an error, since that is in line with the fieldset logic. This way we ensure that error'ed fields stay visible.

MHLut: @sarahboyce Forgot to tag you.

tests/admin_inlines/tests.py
MHLut: @sarahboyce I've added these lines of code, see the artifacts for screenshots.

django/contrib/admin/helpers.py
sarahboyce: I see, I missed that the inline html are using `<fieldset></fieldset>` 👍

MHLut:
> We also need to tweak the docs: https://docs.djangoproject.com/en/5.0/ref/contrib/admin/ It says things like:
> 
> > with a collapse class will be initially collapsed and their header will have a small “show” link.
> 
> And an existing wrong sounding:
> 
> > Fieldsets with the collapse style will be initially collapsed in the admin and replaced with a small “click to expand” link.
> 
> We probably need a "versionchanged" note and release note also 👍

@sarahboyce I have updated the texts and release notes that you mentioned. I am not entirely pleased with the wording yet, feel free to update these! (I won't be around for the rest of the day)

sarahboyce (approved):
Thank you for this @MHLut 🎉 I think this is a great change

@django/accessibility I've approved as LGTM and tested the best I knew how 😁 
If someone on your team would like to do a code review or some testing, there's still a month before the alpha feature freeze and happy to wait before merging.
Let me know if you explicitly want to look at it, if I haven't heard anything by Monday I will start the steps to merge this in

knyghty (approved):
@sarahboyce I'm very happy as is. I think it would be good to create a ticket for the contrast issue, any remaining RTL issues, and maybe also the header outline, but IMO this is good to go.

MHLut:
Tagging @thibaudcolas since he is the author of the original ticket.

docs/ref/contrib/admin/index.txt
```diff
@@ -2306,8 +2311,7 @@ The ``InlineModelAdmin`` class adds or customizes:
     A list or tuple containing extra CSS classes to apply to the fieldset that
     is rendered for the inlines. Defaults to ``None``. As with classes
     configured in :attr:`~ModelAdmin.fieldsets`, inlines with a ``collapse``
-    class will be initially collapsed and their header will have a small "show"
-    link.
+    class will be initially collapsed using an expandable widget.
```
sarahboyce, line 2314: @django/triage-review `InlineModelAdmin.classes` references `ModelAdmin.fieldsets` which has a `versionchanged` note, do you think that same version changed note is also required here for `InlineModelAdmin.classes`?

Also suggestions are welcome around tweaking the release notes/version changed note/docs updates. I've done my best but not quite mastered the art of good docs. ❤️

django/contrib/admin/helpers.py
```diff
@@ -116,10 +117,14 @@ def __init__(
...
     def media(self):
-        if "collapse" in self.classes:
-            return forms.Media(js=["admin/js/collapse.js"])
         return forms.Media()
 
+    @cached_property
+    def is_collapsible(self):
+        if any([error in self.fields for error in self.form.errors]):
+            return False
+        return "collapse" in self.classes
```
ngnpope, line 126: Is it worth making `"collapse"` a constant somewhere? 🤔

Also, would it be better to switch to `"collapsible"`? I know this is probably kept for backward compatibility, but IIUC we'll be requiring everyone to switch to using `<details>`, so it is a breaking change anyway?

django/contrib/admin/static/admin/css/forms.css
```diff
@@ -76,6 +76,24 @@ form ul.inline li {
...
 
+/* FIELDSETS */
+
+fieldset legend {
+    inline-size: 100%;
+    padding: 0;
+}
+
+fieldset h3 {
+    /* The border exists for line continuity + forced-colors mode */
```
ngnpope, line 87: I think we really want to refer to "high-constrast"? IIUC "forced-colors" is just the mechanism for forcing a particular color mode in a browser? See the description [here](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors).

```suggestion
    /* The border exists for line continuity and high-contrast mode. */
```

django/contrib/admin/helpers.py
knyghty: I think this change is only really "breaking" if the user is overriding some JS or template around this. For most users nothing needs to change, so IMO not worth making everyone using this update their code.

django/contrib/admin/static/admin/css/forms.css
sarahboyce: I removed the comment as I think we don't need it but I understand your point 👍

django/contrib/admin/helpers.py
```diff
@@ -116,10 +117,14 @@ def __init__(
 
     @property
     def media(self):
-        if "collapse" in self.classes:
-            return forms.Media(js=["admin/js/collapse.js"])
         return forms.Media()
 
+    @cached_property
+    def is_collapsible(self):
+        if any([error in self.fields for error in self.form.errors]):
```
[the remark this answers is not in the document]
sarahboyce: Updated

docs/ref/contrib/admin/index.txt
MHLut: I'm also unsure about this. If one source references the other, then it's probably OK to list a versionchanged with the latter only as a single source of truth. But I am unaware of Django's practices here.

nessita:
Hello @MHLut! Thank you so much for your work on this, for all the investigation and sharing your thoughts in the description and in Discord.

I read thru the Discord thread (I may have missed something), and while I acknowledge that you said the following:

> So far, I like Example 2 most, Example 1 is OK, and don't like Example 3.
> [...] the third example just feels meh to me. The look and feel is clunky, with a duplicate mention of "Address", where the <legend> content feels useless.
> Code-wise though; the second example is the cleanest, and allows us to reuse the existing structure of the fieldsets.
> The first example has the same perks, but needs JS to get it to work.
> The first example also requires CSS to add in the triangle icons which have no specific meaning.

I honestly dislike the "Fields" summary. It feels (to me, I know, very objectively) very unpolished in terms of UI, mainly because the fieldset is supposed to be collapsed and yet Django is "stubbornly" showing just a peak, a little more of it. I think is, to some extent, important to keep the previous "implicit agreement" that a collapsed fieldset would only show the fieldset name and nothing else.

With that in mind, do you think there is a way to have Example 3 working (purely in terms of UI) without duplicating the fieldset name? Basically have Example 3 without the words highlighted in red:
image

Happy to coordinate a chat or meet if you feel that discussing this IRL(ish) would be easier.

MHLut:
@nessita Thank you for your feedback. As far as I have seen, there is no good way to do the third option whilst hiding information. Here is why.

The HTML requires the following items:

- A <details> element.
- A <summary> inside the details.
- A <fieldset> element.
- A <legend> inside a fieldset.
- A heading (<h3> in this case) inside the fieldset or summary.

We can place the heading inside the legend.
However, we cannot place the summary inside anything other than the top level of the details element.
This means that both the heading and the summary need to have some form of descriptive text.

I picked the solution that looks the most streamlined (in my opinion) and has the least replication. 

By ensuring that the heading comes first in the structure, we don't have to add the heading contents to any of its child elements. (Since the fieldset name has already been announced)
If we choose the third solution, we would have to repeat the fieldset name in the summary.

I prefer solution 2 to the old UI, so let's agree to disagree there.
I also think that showing just a hint of the fieldset contents alongside the details triangle icon gives the user a visual hint of an expandable area.

Nb. If desired, the text "Fields" could be replaced with another text.

nessita:
> The HTML requires the following items:
>     * A `<details>` element.
>     * A `<summary>` inside the details.
>     * A `<fieldset>` element.
>     * A `<legend>` inside the fieldset.
>     * A heading (`<h3>` in this case) inside the fieldset or summary.
> 
> We can place the heading inside the legend. However, we cannot place the summary inside anything other than the top level of the details element. This means that both the heading and the summary need to have some form of descriptive text.

Thanks for these details. To be honest, I'm struggling to understand why all these elements are necessary. Based on the Discord messages I read, it seems there are two primary reasons: firstly, to enable the use of CSS for selectors and to handle the hide/show behavior without JavaScript, and secondly, to ensure proper accessibility and assist screen readers. However, I'm still not entirely clear on why each element is essential. I'm unsure how to experiment with this and comprehend the consequences of removing each item individually.

To try to understand a bit more, I copied the HTML from your codepen for Example 3, and below these I put some questions:

```html
<section>
  <h2>Example 3: Details outside (no JS)</h2>
  <div id="details-example-outside" class="example example--details-outside">
    <details open="">
      <summary>Address fields</summary>
      <fieldset class="module aligned collapse">
        <h3>Address</h3>

        <div class="description">Optional description.</div>
        <div class="form-row field-address_line_1">...</div>
        <div class="form-row field-address_line_2">...</div>
        <div class="form-row field-city">...</div>
        <div class="form-row field-state">...</div>
        <div class="form-row field-zip_code">...</div>
        <div class="form-row field-country">...</div>

      </fieldset>
    </details>
  </div>
</section>
```

* Could we simply remove the `h3`? If not, could you please explain why not? It would be easier for me to understand if you could outline what happens if the element is not present.
* If we absolutely cannot remove the `h3`, could we hide it instead? That way, we could retain the necessary information for screen readers while making it visually invisible.

With the above, and changing "Address fields" -> "Address", I think we would have a nice solution.

> By ensuring that the heading comes first in the structure, we don't have to add the heading contents to any of its child elements. (Since the fieldset name has already been announced) If we choose the third solution, we must repeat the fieldset name in the summary.

This I'm not sure what it means, sorry. Could you please help me understand? An example would certainly help me!

> I prefer solution 2 to the old UI, so let's agree to disagree there.

Sure.

> I also think that showing just a hint of the fieldset contents alongside the details triangle icon gives the user a visual hint of an expandable area.

I agree that the hint about the expandable area is necessary, but I think the triangle in the "fieldset title row" (the blue one with white letters) accomplish that perfectly. There is no need for an extra row IMHO.

Having yet another row (white background in example 2) seems unnecessary (from an UI POV) and doesn't provide any useful information (I know we can change the text, I replied about this further down). I consider this a regression, in that Example 2 makes the interface feel cluttered and can be confusing for users who expect collapsible sections to minimize the amount of visible content, especially when vertical space is limited.

I believe it's unnecessary to append or mention "fields" anywhere, as it seems redundant. This is similar to how the admin doesn't append "Model" everywhere because it's implied.

> Nb. If desired, "Fields" could be replaced with another text.

Yes, absolutely, though I can't think of a text to put in there that would justify the extra row. Do you have any ideas?

tests/admin_inlines/tests.py
sarahboyce: @MHLut I have some slightly bad news, the color contrast is currently an issue on this patch because the css was previously redundant and never applied. With the new html, it has resurrected this dodgy css 

Previously the header would look like this:
image

There is a ticket to delete this css here: https://code.djangoproject.com/ticket/35382
This needs to happen before this PR can be merged

MHLut: @sarahboyce Well, I dislike this light blue heading since it doesn't match the modern admin style. I'm on board with making it look more like the current style on `main` again.

See also the comment I placed on that ticket.

Either someone removes the redundant CSS in a separate PR, or I'll remove it in this PR; it doesn't matter because I will have to update the styles anyway.

knyghty:
@MHLut, @nessita,

I had a (very quick!) look into this.

The `<legend>` element is not (strictly speaking) necessary. It is useful, but in our case I think the `<summary>` element achieves the same goal. We do need to make sure we keep it for `<fieldset>`s that do not use `collapse` however (not sure how much the admin uses fieldsets otherwise).

As for the `<h3>`, I am not sure. We could move it into the `<summary>` element, this is allowed. I think that we should just do whatever makes the outline more correct here, whether that's having a heading or not.

With removing the `<legend>` element we should probably test that it makes sense when using a screen reader (I'd guess so, and I'd guess possibly it's even less confusing without the repeated text.

MHLut:
Notes regarding UI changes to the third proof of concept (link in top description):

I tried to combine the fieldset and details elements, but the Firefox accessibility inspector threw warnings.

The problem is that a fieldset needs a legend to describe its contents to assistive technology, and the legend should be its first child.

The same goes for details: the summary describes its contents and should be the first child. 
(If you omit a summary, the collapsible widget's label defaults to "Details" in English.)

When putting the legend inside the summary, the accessibility inspector did not recognize it as a label.

I also tried to attach the fieldset name to the fieldset through `aria-label`, but the accessibility inspector indicated that this isn't good enough (and once again hinted at using a legend).

I don't see how we can use the third UI solution without sacrificing accessibility.
Unless we could somehow use `aria-label` or `aria-describedby`.
(My third proof of concept also lacks a legend, meaning it is an incomplete solution.)

NB. Regarding `aria-describedby`:
If we'd use that, we need to add an ID to the fieldset's heading element and use the heading as a descriptor.
However, fieldset names are not mandatory, and sometimes there is no heading.
We also lack a property guaranteeing a unique ID output.

nessita:
@MHLut could we use a "complete" solution for the third option without sacrificing accessibility by hiding the `<h3>` with a custom CSS class that allows to screen readers to still mention it?

MHLut:
@nessita The `<h3>` isn't the problem. 

We cannot intertwine the `<legend>` and the `<summary>` without sacrificing accessibility.
We would need to do that to achieve the look and feel of the third solution without JavaScript or hacks.

I also would prefer to keep the `<fieldset>` as the outer element so that most of the outline will be identical between collapsible and regular fieldsets.
Solution 2 provides that, but solution 3 doesn't.

If we sacrifice the `<legend>`, solution 3 would probably be achievable, even retaining `<fieldset>` as the outer element.
However, adding a legend is one of the original assignments of this ticket (point 3 under "Accessibility bugs").

I have tried several things and don't know what else would be good.
Accessibility-wise, the current solution is probably the best we will get.
Thibaud hasn't looked at this yet, though, and I hope he has ideas for an improvement.

nessita:
@MHLut Thank you. Could you please, when you have some time, push changes in another branch with what you have for solution 3? Ideally this temporary branch would have a complete solution in terms of accessibility (ie all the required HTML elements for assistive technologies would be present). I'd like to see how it looks and use it a bit, if possible.

MHLut:
Per @nessita's request, I have updated the PR (in a separate commit for now) with a new UI style.

See these notes I added to [the PR](https://github.com/MHLut/django/pull/1):

```
## Accessibility notes

I discussed this solution with a blind developer. He suggested we replace `<fieldset>` with `<div role="group">` if the lack of `<legend>` bothers the accessibility checkers.

Due to styling issues, I had to change the heading level of inline fieldsets to `<h4>`. This also fixes a heading bug I mentioned in the original PR.

## Other notes

This solution hinges on adding a unique ID to a fieldset heading so that we can use `aria-labelledby` on the `<fieldset>.` Due to the lack of a clean way of generating such an ID in the template, I added another property to the `Fieldset` class.
```

MHLut:
Tagging @jscholes as my aforementioned advisor.

MHLut:
Apologies for the failed tests; I've added some classes to existing headings, but the tests do not like it. I will fix these later, but you can test this new UI design already!

tests/admin_inlines/tests.py
```diff
@@ -1850,10 +1852,11 @@ def test_delete_inlines(self):
             "form#profilecollection_form tr.dynamic-profile_set#profile_set-2", 1
         )
 
+    @screenshot_cases(["desktop_size", "mobile_size", "rtl", "dark"])
```
sarahboyce, line 1855: Now that https://github.com/django/django/pull/18080 is merged, please rebase and add "high_contrast" to all added screenshot_cases ⭐

sarahboyce: Thank you

MHLut:
@nessita Could you check if you approve of this look and feel? If so, I will fix the unit tests. Otherwise, I will revert the commit.

MHLut:
@nessita I'm glad you like it! Per your feedback:

1. Changing the cursor to a pointer on elements that don't naturally have one is typically considered bad practice. I am reluctant to make this change and would like a second opinion from the accessibility team.

2. The header is in line with the current admin design, and changing it would also affect the non-collapsible UI.

I don't have much time until the feature freeze, and fixing the tests alone will be a task. James (@jscholes) uses a screen reader and clarified that whatever we push will already be a significant improvement compared to the current UI. As such, I would rather see us commit these changes to Django 5.1 and make overarching UI improvements in future tickets.

claudep:
>     1. Changing the cursor to a pointer on elements that don't naturally have one is typically considered bad practice. I am reluctant to make this change and would like a second opinion from the accessibility team.

I find myself changing the CSS to set the cursor as a pointer in all my own usages of details/summary. I admit I didn't search for the rationales of not making this behavior the default in the specs.

nessita:
> @nessita I'm glad you like it! Per your feedback:
> 
>     1. Changing the cursor to a pointer on elements that don't naturally have one is typically considered bad practice.

Would you have any reference link for this affirmation? It was a surprise to me to read this since I would generally think that the pointer mouse cursor would be an additional (and welcomed) "clickability" indicator. I tried searching for articles or reasons explaining why it'd be a bad practice and I couldn't find any, so I would definitely appreciate if you could provide some.

I did find a very interesting link supporting the pointer mouse cursor for anything that can be "interacted with" (but this should be in addition to other indicators like the "arrow pointing right" that you are proposing, which I this is a positive adding), see https://github.com/w3c/csswg-drafts/issues/1936 (very long conversation) with a [resolution from the CSS Working Group to add](https://github.com/w3c/csswg-drafts/issues/1936#issuecomment-420722704):

> Authors should use pointer on links and may use on other interactive elements.

I know it says "MAY" use but the whole thread and [many](https://kizu.dev/cursor-pointer/) [other](https://12daysofweb.dev/2021/details-summary/#styling-considerations-for-details-and-summary) [related](https://css-tricks.com/two-issues-styling-the-details-element-and-how-to-solve-them/#aa-issue-1-the-summary-cursor) [issues](https://github.com/webhintio/webhintio.github.io/issues/453) focus on how this can enhance user experience and make the interface more intuitive, especially for users who are accustomed to the pointer cursor indicating interactivity on the web. Furthermore, the [MDN Web docs change the mouse cursor to `pointer`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#customizing_the_appearance) in the examples for customizing the appearance of the `details` element.

Because of the above, I think we should change it.

>     2. The header is in line with the current admin design, and changing it would also affect the non-collapsible UI.

Fair enough.

> I don't have much time until the feature freeze, and fixing the tests alone will be a task. James (@jscholes) uses a screen reader and clarified that whatever we push will already be a significant improvement compared to the current UI. As such, I would rather see us commit these changes to Django 5.1 and make overarching UI improvements in future tickets.

Thanks for the heads up!

knyghty:
FWIW I agree with adding `cursor: pointer` here for Natalia's reasons and more. It's a handy way to know something is interactive (there are times you can see non-interactive arrows, e.g. in lists (should they? maybe not, but we're here).

Another notable site to have a pointer on these elements is this one.

<details>
  <summary>For example</summary>
  etc.
</details>

ngnpope:
Also +1 for `cursor: pointer`. As Tom says, GitHub does it too - they have the following rule to apply it globally which seems sensible:

```css
details summary {
    cursor: pointer;
}
```

We have also already done the same for changelist filters:

```css
#changelist-filter h3,                                                                                                  
#changelist-filter details summary {                                                                                    
    font-weight: 400;                                                                                                   
    padding: 0 15px;                                                                                                    
    margin-bottom: 10px;                                                                                                
    cursor: pointer;                                                                                                                                                     
}                                                                                                                       
```

If we apply it globally, we can remove the duplication from that rule.

MHLut:
I'll add the pointer, although I would like to add that GitHub is not necessarily an example of good usability. Agreed on the global styling for this one, since we're also introducing more details and summaries in other PRs.

MHLut:
@nessita The tests finally seem to pass, so that you can return to testing and reviewing!

I don't have much time for follow-ups, but I would like to see this merged before the feature freeze, so feel free to update the branch as you see fit.

@sakhawy might also be able to assist with issues that don't require knowing the entire PR history.

nessita:
> @nessita The tests finally seem to pass, so that you can return to testing and reviewing!
> 
> I don't have much time for follow-ups, but I would like to see this merged before the feature freeze, so feel free to update the branch as you see fit.
> 
> @sakhawy might also be able to assist with issues that don't require knowing the entire PR history.

Thank you so much @MHLut, I'll work in this PR tomorrow or Thu at the latest.

django/contrib/admin/helpers.py
nessita: I appreciate this conversation because it addressed some questions that I had, but there is one issue with "automatically expanding a fieldset that has errors" in that, for example, a fieldset inside a `StackedInline` would be rendered as an `h4`, but when there are errors, it's rendered as an `h2`. This is also an issue in main (though it uses `h3` instead of `h4`) so I created this ticket-35456 to track this.

docs/ref/contrib/admin/index.txt
nessita: I will tweak these docs (if necessary) as part of my review process.

nessita:
@MHLut Hey there! I have pushed some changes removing the definition and usage of `heading_id` because it felt inconsistent with the existing code, and also because it's usually a bad practice to rely on random numbers to generate unique IDs (random module does not guarantee uniqueness, even in large ranges). We could use UUID instead (which does guarantee uniqueness) but that feels a bit overkill for the branch's need, so I built the fieldset heading ID by using information available in the template (or its parent template), similar to what formsets already do.

@sarahboyce Would you fancy to re-review that commit (it's isolated in a second commit to ease review)? (Since I changed some of the already approved code.)

FYI I still need to review in detail tests and docs, but overall I think the branch is in very good shape.

django/contrib/admin/templates/admin/includes/fieldset.html
```diff
@@ -1,5 +1,11 @@
-<fieldset class="module aligned {{ fieldset.classes }}">
-    {% if fieldset.name %}<h2>{{ fieldset.name }}</h2>{% endif %}
+{% with prefix=fieldset.formset.prefix|default:"fieldset" id_prefix=id_prefix|default:"0" id_suffix=id_suffix|default:"0" name=fieldset.name|default:""|slugify %}
+<fieldset class="module aligned {{ fieldset.classes }}" {% if name %}aria-labelledby="{{ prefix }}-{{ id_prefix}}-{{ name }}-{{ id_suffix }}"{% endif %}>
+    {% if name %}
+        {% if fieldset.is_collapsible %}<details><summary>{% endif %}
+        <h{{ heading_level|default:2 }} id="{{ prefix }}-{{ id_prefix}}-{{ name }}-{{ id_suffix }}" class="fieldset-heading">{{ fieldset.name }}</h{{ heading_level|default:2 }}>
```
sarahboyce, line 5: I think either this id is too complex or the previously formed id `{{ inline_admin_formset.formset.prefix }}-heading` is too simple. Maybe `{{ fieldset.formset.prefix }}-fieldset-header-{{ id_suffix }}` would be more consistent.

This logic is also missing tests such as
- checking for id uniqueness i.e. `self.assertContains(response, `id="this-id-thing"`, html=True, count=1)`
- checking for `aria-labelledby`

django/contrib/admin/templates/admin/edit_inline/stacked.html
```diff
@@ -19,11 +23,16 @@ <h3><b>{{ inline_admin_formset.opts.verbose_name|capfirst }}:</b> <span class="i
     {% if inline_admin_formset.formset.can_delete and inline_admin_formset.has_delete_permission and inline_admin_form.original %}<span class="delete">{{ inline_admin_form.deletion_field.field }} {{ inline_admin_form.deletion_field.label_tag }}</span>{% endif %}
   </h3>
   {% if inline_admin_form.form.non_field_errors %}{{ inline_admin_form.form.non_field_errors }}{% endif %}
-  {% for fieldset in inline_admin_form %}
-    {% include "admin/includes/fieldset.html" %}
-  {% endfor %}
+
+  {% with parent_counter=forloop.counter0 %}
+    {% for fieldset in inline_admin_form %}
+      {% include "admin/includes/fieldset.html" with heading_level=4 id_prefix=parent_counter id_suffix=forloop.counter0 %}
```
sarahboyce, line 29: we should add an assert checking for the header level 4 (can also check that it comes after a h3/2/1 but that's more a nice to have)

docs/ref/contrib/admin/_images/fieldsets.png
sarahboyce, line 1: We should update this picture again 👍

sarahboyce (commented):
I didn't do a full review since the change of approach. 👍 
This generally looks good, we're missing tests checking the logic that exists in the templates

django/contrib/admin/templates/admin/includes/fieldset.html
nessita: Thank you! I agree that tests are missing, I will add some, which will definitely help validate what I expose down below. What follows may be confusing though I tried to be as clear as possible, but the summary is that I'll build a test case with the scenario below and proper asserts.

> I think either this id is too complex or the previously formed id `{{ inline_admin_formset.formset.prefix }}-heading` is too simple.

Regarding the complexity of the proposed `id`, and given that the `id` attr needs to have a unique value, we need to consider the following scenarios where the `fieldset.html` is included in other templates:
1. via "toplevel" inclusion in the `change_form.html` (https://github.com/django/django/blob/main/django/contrib/admin/templates/admin/change_form.html#L48)
2. via a nested for loop via `StackedInline` in `stacked.html` (https://github.com/django/django/blob/main/django/contrib/admin/templates/admin/edit_inline/stacked.html#L22)

>  Maybe `{{ fieldset.formset.prefix }}-fieldset-header-{{ id_suffix }}` would be more consistent.

Sadly, `{{ fieldset.formset.prefix }}-fieldset-header-{{ id_suffix }}` is not truly unique, and as far as my research goes, the "only" way to built a truly unique `id` for the fieldset heading is using all of the following elements (in all cases, the fieldset name is accessory in that it does not provide uniqueness but it helps building a nicer `id`):

1. For item 1, where there is no formset associated with the fieldset, the fieldset name and the for loop counter suffices
2. For item 2, we need to consider:
    * the formset prefix (which is also used as part of most `id`s in `stacked.html`), since in an admin page there can be X instances of inlines, so that would mean there are X different formset prefixes
    * the for loop counter that iterates over each formset (since there can be Y `inline_admin_form` inside each formset), this is what the proposal defines as `parent_counter` in `stacked.html` and is used as `id_prefix` in `fieldset.html`)
    * the for loop counter that iterates over each fieldset in each form in each formset (since there can be Z fieldsets in each inline_admin_form), this is what the proposal uses as `id_suffix` in `fieldset.html`
    * As with item 1, the fieldset name is used to make the `id` a little nicer, not because it adds uniqueness

So for models and admins like the ones defined at the bottom of this comment, we get fieldset headers like this (the ids are of the form `cover_set-X-Y-details-Z`, where `coverset-X` is the formset prefix, Y and X are as described above, and `details` is the fieldset name):
```html
        <!-- these two are toplevel fieldsets, defined at the admin model level -->
        <h2 id="fieldset-0-advanced-options-1" class="fieldset-heading">Advanced options</h2>
        <h2 id="fieldset-0-advanced-options-2" class="fieldset-heading">Advanced options</h2>

  <h2 id="cover_set-heading" class="inline-heading"> <!-- this one `CoverTabularInline` instance -->
  <h2 id="cover_set-2-heading" class="inline-heading"> <!-- this the second `CoverTabularInline` instance -->

  <h2 id="cover_set-3-heading" class="inline-heading"> <!-- this is the `CoverStackedExtra2Inline` instance -->
        <h4 id="cover_set-3-0-details-1" class="fieldset-heading">Details</h4> <!-- first fielset of the first form -->
        <h4 id="cover_set-3-0-details-2" class="fieldset-heading">Details</h4> <!-- second fielset of the first form -->
        <h4 id="cover_set-3-1-details-1" class="fieldset-heading">Details</h4> <!-- first fielset of the second form -->
        <h4 id="cover_set-3-1-details-2" class="fieldset-heading">Details</h4> <!-- etc. -->
        <h4 id="cover_set-3-2-details-1" class="fieldset-heading">Details</h4>
        <h4 id="cover_set-3-2-details-2" class="fieldset-heading">Details</h4>

  <h2 id="cover_set-4-heading" class="inline-heading"> <!-- this is the `CoverStackedExtra3Inline` instance -->
        <h4 id="cover_set-4-0-details-1" class="fieldset-heading">Details</h4>
        <h4 id="cover_set-4-0-details-2" class="fieldset-heading">Details</h4>
        <h4 id="cover_set-4-1-details-1" class="fieldset-heading">Details</h4>
        <h4 id="cover_set-4-1-details-2" class="fieldset-heading">Details</h4>
        <h4 id="cover_set-4-2-details-1" class="fieldset-heading">Details</h4>
        <h4 id="cover_set-4-2-details-2" class="fieldset-heading">Details</h4>
        <h4 id="cover_set-4-3-details-1" class="fieldset-heading">Details</h4>
        <h4 id="cover_set-4-3-details-2" class="fieldset-heading">Details</h4>
```
* models.py
```python
from django.db import models
from django.utils.timezone import now


class Book(models.Model):
    title = models.CharField(max_length=100)
    author = models.CharField(max_length=100)
    publisher = models.CharField(max_length=100)
    creation_date = models.DateField(default=now)
    update_date = models.DateField(default=now)
    publication_date = models.DateField(default=now)

    def __str__(self):
        return self.title


class Cover(models.Model):
    book = models.ForeignKey(Book, on_delete=models.CASCADE)
    image = models.CharField(max_length=100)
    title = models.CharField(max_length=100)
    description = models.TextField()
    creation_date = models.DateField(default=now)
    update_date = models.DateField(default=now)
    updated_by = models.CharField(max_length=100)
```
* admin.py
```python
from django.db import models

from .models import Book, Cover


class CoverInlineMixin:
    model = Cover
    extra = 2
    fieldsets = (
        (None, {"fields": ("image", "title")}),
        ("Details", {
            "fields": ("description", "creation_date"),
        }),
        ("Details", {  # Name intentionally duplicated
            "fields": ("update_date", "updated_by"),
        }),
    )


class CoverTabularInline(CoverInlineMixin, admin.TabularInline):
    pass


class CoverStackedExtra2Inline(CoverInlineMixin, admin.StackedInline):
    pass


class CoverStackedExtra3Inline(CoverInlineMixin, admin.StackedInline):
    extra = 3


class BookAdmin(admin.ModelAdmin):
    list_display = ("title", "author", "publisher","publication_date")
    search_fields = ("title", "author")
    list_filter = ("author", "publication_date")
    fieldsets = (
        (None, {
            "fields": ("title", "author")
        }),
        ("Advanced options", {
            "fields": ("publisher", "publication_date",)
        }),
        ("Advanced options", {
            "fields": ("creation_date", "update_date",)
        }),
    )
    inlines = [
        CoverTabularInline,
        CoverTabularInline,
        CoverStackedExtra2Inline,
        CoverStackedExtra3Inline,
    ]


admin.site.register(Book, BookAdmin)
```

django/contrib/admin/templates/admin/edit_inline/stacked.html
nessita: Yes, I agree, will add that.

docs/ref/contrib/admin/_images/fieldsets.png
nessita: Yes, it's updated in my local copy, will push it shortly.

docs/ref/contrib/admin/index.txt
```diff
@@ -425,8 +425,13 @@ subclass::
             }
 
         Fieldsets with the ``collapse`` style will be initially collapsed in
-        the admin and replaced with a small "click to expand" link. Fieldsets
-        with the ``wide`` style will be given extra horizontal space.
+        the admin and replaced with an expandable widget. Fieldsets with the
```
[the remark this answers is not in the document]
nessita: Good point! Will include that.

django/contrib/admin/helpers.py
```diff
@@ -116,10 +117,14 @@ def __init__(
...
     def media(self):
-        if "collapse" in self.classes:
-            return forms.Media(js=["admin/js/collapse.js"])
         return forms.Media()
 
+    @cached_property
+    def is_collapsible(self):
+        if any([field in self.fields for field in self.form.errors]):
+            return False
+        return "collapse" in self.classes
```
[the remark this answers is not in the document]
MHLut: @sarahboyce @nessita I have considered this change several times and ultimately decided not to put this check in Python. I felt a UI-specific check would be better made in the template.

If you look at `fieldset.html`, you'll see on lines 3 and 39 that there is a check for the fieldset name, a prerequisite for the is_collapsible check.

If we move this check to Python, we can simplify the aforementioned if-statements in the template. I'm all for it if you are also OK with this since the nested if-statement made the template slightly more complicated to understand for those unaware of this bug.

MHLut:
@nessita @sarahboyce Thank you for the heading ID update! I cannot review the ID changes in detail right now, but the only thing important to me is that they are predictable and unique since `aria-labelledby` relies on them for accessibility reasons.

A nice to have for IDs would be for them to be short and readable so that they are understandable when read from a browser URL.

django/contrib/admin/templates/admin/includes/fieldset.html
nessita: @sarahboyce I have pushed a complete test for the headings, including "h-level" and id. I realized the heading needed a "-heading" suffix to be consistent with other ids, but otherwise the ID is as described above. Resolving but let me know if you would like something changed! Thank you.

docs/ref/contrib/admin/index.txt
nessita: I tweaked docs and duplicated the `versionadded` note for `InlineModelAdmin.classes`. Explicit is better than implicit.

nessita (approved):

sarahboyce (commented):
Looks good! 
If we want to split commits I would have:
- setting the header levels and adding the `aria-labelledby` as "Associated fieldsets with their headers."
- updating collapse to use details/summary

docs/releases/5.1.txt
```diff
@@ -414,6 +414,15 @@ Miscellaneous
...
   ``<div id="main">`` element.
 
+* In order to improve accessibility, the expandable widget used for
+  :attr:`ModelAdmin.fieldsets <django.contrib.admin.ModelAdmin.fieldsets>` and
+  :attr:`InlineModelAdmin.fieldsets <django.contrib.admin.InlineModelAdmin>`,
+  when the fieldset has a name and use the ``collapse`` class, now includes
+  ``<details>`` and ``<summary>`` elements.
+
+* The JavaScript file ``collapse.js`` is removed since it is no longer needed
+  in the Django administrative interface.
```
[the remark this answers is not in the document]
claudep: I would let that, as you never know if some people are linking to this file in their projects.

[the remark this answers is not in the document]
ngnpope: I'd also change `administrative interface` to `admin site` which is more commonly used in the docs.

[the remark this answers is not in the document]
sarahboyce: Okay dokey 👍

nessita:
> Looks good! If we want to split commits I would have:
> 
>     * setting the header levels and adding the `aria-labelledby` as "Associated fieldsets with their headers."
> 
>     * updating collapse to use details/summary

This is a good proposal, thank you! I will try to work this out after our weekly sync.

docs/releases/5.1.txt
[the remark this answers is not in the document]
nessita: I'm happy to change the wording to "admin site", I was using what I found in existing release notes (but you are right Nick, "admin site" is a bit more common that "administrative interface").

tests/admin_inlines/tests.py
```diff
@@ -1466,30 +1683,143 @@ class BothVerboseNameProfileInline(TabularInline):
...
+
+        # Headings for the toplevel named fieldsets.
+        for x, fieldset in enumerate(response.context["adminform"]):
+            if not fieldset.name:
+                continue
+            fieldset_heading = (
+                f'<h2 id="fieldset-0-advanced-options-{x}-heading"'
+                ' class="fieldset-heading">Advanced options</h2>'
+            )
+            self.assertContains(response, fieldset_heading, html=True, count=1)
```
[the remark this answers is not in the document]
nessita: Sounds good, will change following this advice.

tests/admin_inlines/tests.py
```diff
@@ -1466,30 +1683,143 @@ class BothVerboseNameProfileInline(TabularInline):
...
+
+        # Headings and subheadings for all the inlines.
+        for inline_admin_formset in response.context["inline_admin_formsets"]:
+            prefix = inline_admin_formset.formset.prefix
+            formset_heading = (
+                f'<h2 id="{prefix}-heading" class="inline-heading">Photos</h2>'
+            )
+            self.assertContains(response, formset_heading, html=True, count=1)
```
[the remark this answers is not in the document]
nessita: Agreed, good thinking!

tests/admin_inlines/tests.py
```diff
@@ -1466,30 +1683,143 @@ class BothVerboseNameProfileInline(TabularInline):
...
+                        if "collapse" in fieldset.classes:
+                            fieldset_heading = (
+                                f"<details><summary>{fieldset_heading}</summary>"
+                            )
+                        fieldset_html = (
+                            f'<fieldset class="module aligned {fieldset.classes}" '
+                            f'aria-labelledby="{heading_id}">\n    \n        '
+                            f"{fieldset_heading}"
+                        )
+                        self.assertContains(response, fieldset_html, count=1)
```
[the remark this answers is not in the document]
nessita: I like your proposal, thanks!

nessita:
I split the first commit into two logical commits.

PR #17912, opened 2024-02-27
More readable and efficient conditions. Removed explicit return.

rrupy:
Removed explicit return None in middleware/csrf.py. More readable and efficient conditions in if statements in middleware module.

### DIFF
--- a/django/middleware/csrf.py
+++ b/django/middleware/csrf.py
@@ -204,7 +204,6 @@ class CsrfViewMiddleware(MiddlewareMixin):
         # request.  This will be relevant when both decorator and middleware
         # are used.
         request.csrf_processing_done = True
-        return None
 
     def _reject(self, request, reason):
         response = _get_failure_view()(request, reason=reason)
--- a/django/middleware/locale.py
+++ b/django/middleware/locale.py
@@ -27,9 +27,8 @@ class LocaleMiddleware(MiddlewareMixin):
         )
         language_from_path = translation.get_language_from_path(request.path_info)
         if (
-            not language_from_path
-            and i18n_patterns_used
-            and not prefixed_default_language
+                not (language_from_path and prefixed_default_language)
+                and i18n_patterns_used
         ):
             language = settings.LANGUAGE_CODE
         translation.activate(language)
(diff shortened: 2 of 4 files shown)


django/middleware/security.py
```diff
@@ -20,10 +20,9 @@ def __init__(self, get_response):
 
     def process_request(self, request):
         path = request.path.lstrip("/")
-        if (
-            self.redirect
-            and not request.is_secure()
-            and not any(pattern.search(path) for pattern in self.redirect_exempt)
+        if self.redirect and not (
+            request.is_secure()
+            and any(pattern.search(path) for pattern in self.redirect_exempt)
```
shangxiao, line 25: (!x && !y) is not the same thing as !(x && y)

Please see De Morgan's laws: https://en.wikipedia.org/wiki/De_Morgan%27s_laws

django/middleware/csrf.py
```diff
@@ -204,7 +204,6 @@ def _accept(self, request):
         # request.  This will be relevant when both decorator and middleware
         # are used.
         request.csrf_processing_done = True
-        return None
```
shangxiao, line 207: "Explicit is better than implicit" … although… there are many cases in Django where implicit None is returned.

In any case it's not worth the change imho 🤷‍♂️

django/middleware/locale.py
```diff
@@ -27,9 +27,8 @@ def process_request(self, request):
         )
         language_from_path = translation.get_language_from_path(request.path_info)
         if (
-            not language_from_path
-            and i18n_patterns_used
-            and not prefixed_default_language
+                not (language_from_path and prefixed_default_language)
+                and i18n_patterns_used
```
shangxiao, line 31: Not sure this is equivalent?

shangxiao:
Sorry but we can't accept changes that randomly change conditional logic because it simply appears to be faster or more readable.

Any change based on performance needs to be accompanied with some statistics showing significant improvements.

django/middleware/security.py
rrupy: Thanks, i didn't pay attention to it, you are right

PR #17923, opened 2024-02-29
Refs #22997: stopped adding default for added AutoFields

bcail:


tests/migrations/test_autodetector.py
```diff
@@ -1350,6 +1350,29 @@ def test_add_date_fields_with_auto_now_add_not_asking_for_null_addition(
...
+            "Author",
+            [
+                ("pkfield", models.IntegerField(primary_key=True)),
+            ],
+        )
+        updated_state = ModelState(
+            "testapp",
+            "Author",
+            [
+                ("id", models.BigAutoField()),
```
DevilsAutumn, line 1365: This should be a primary key.
```suggestion
                ("id", models.BigAutoField(primary_key=True)),
```

sarahboyce:
Hi @bcail 👋 Thank you for this

I created a Django project with a SQLite backend and followed the instructions from the ticket i.e.

> 1. create new model, set some field as primary key
> 2. set up migrations with initial migration
> 3. modify model so field is not primary key
> 4. generate migration

Previously it would ask for default value for auto-generated ID field and will try to create id field as primary key, before modifying existing key (as also described in the ticket). 
Then I use the patch, the migration has been updated as we would hope and it works 🥳 

I have found a very similar existing issue that is still broken with this patch, if I
1. create new model, with one field as a primary key
2. set up migrations with initial migration
3. remove that field
4. generate migration

Then when I am using this patch, I get no prompt and the migration is generated, however when I try to apply that migration - it crashes with `sqlite3.OperationalError: near ")": syntax error`
Note that this is also broken on main but I also get a prompt

<details><summary>Full trace</summary>

```
  Applying app3.0002_remove_modelwithprimarykey_some_pk_and_more...Traceback (most recent call last):
  File "django-project\venv\Lib\site-packages\django\db\backends\utils.py", line 105, in _execute
  File "django-project\venv\Lib\site-packages\django\db\backends\sqlite3\base.py", line 354, in execute
sqlite3.OperationalError: near ")": syntax error

  File "django-project\mysite\manage.py", line 22, in <module>
    main()
  File "django-project\mysite\manage.py", line 18, in main
  File "django-project\venv\Lib\site-packages\django\core\management\__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "django-projectvenv\Lib\site-packages\django\core\management\__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "django-project\venv\Lib\site-packages\django\core\management\base.py", line 413, in run_from_argv
    self.execute(*args, **cmd_options)
  File "django-project\venv\Lib\site-packages\django\core\management\base.py", line 459, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "django-project\venv\Lib\site-packages\django\core\management\base.py", line 107, in wrapper
    res = handle_func(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "django-project\venv\Lib\site-packages\django\core\management\commands\migrate.py", line 356, in handle
    post_migrate_state = executor.migrate(
                         ^^^^^^^^^^^^^^^^^
  File "django-project\venv\Lib\site-packages\django\db\migrations\executor.py", line 135, in migrate
    state = self._migrate_all_forwards(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "django-project\venv\Lib\site-packages\django\db\migrations\executor.py", line 167, in _migrate_all_forwards
    state = self.apply_migration(
            ^^^^^^^^^^^^^^^^^^^^^
  File "django-project\venv\Lib\site-packages\django\db\migrations\executor.py", line 252, in apply_migration
    state = migration.apply(state, schema_editor)
  File "django-project\venv\Lib\site-packages\django\db\migrations\migration.py", line 132, in apply
    operation.database_forwards(
  File "django-projectvenv\Lib\site-packages\django\db\migrations\operations\fields.py", line 174, in database_forwards
    schema_editor.remove_field(
  File "django-project\venv\Lib\site-packages\django\db\backends\sqlite3\schema.py", line 336, in remove_field
    self._remake_table(model, delete_field=field)
  File "django-project\venv\Lib\site-packages\django\db\backends\sqlite3\schema.py", line 236, in _remake_table
    self.execute(
  File "django-project\venv\Lib\site-packages\django\db\backends\base\schema.py", line 201, in execute
    cursor.execute(sql, params)
  File "django-project\venv\Lib\site-packages\django\db\backends\utils.py", line 122, in execute
    return super().execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "django-project\venv\Lib\site-packages\django\db\backends\utils.py", line 79, in execute
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "django-project\venv\Lib\site-packages\django\db\backends\utils.py", line 92, in _execute_with_wrappers
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "django-project\venv\Lib\site-packages\django\db\backends\utils.py", line 100, in _execute
  File "django-project\venv\Lib\site-packages\django\db\utils.py", line 91, in __exit__
  File "django-project\venv\Lib\site-packages\django\db\backends\utils.py", line 105, in _execute
  File "django-project\venv\Lib\site-packages\django\db\backends\sqlite3\base.py", line 354, in execute
django.db.utils.OperationalError: near ")": syntax error

```
</details>

<details><summary>Generated migration</summary>

```python
# Generated by Django 5.1 on 2024-04-30 10:46

from django.db import migrations, models


class Migration(migrations.Migration):

    dependencies = [
        ('app3', '0001_initial'),
    ]

    operations = [
        migrations.RemoveField(
            model_name='modelwithprimarykey',
            name='some_pk',
        ),
        migrations.AddField(
            model_name='modelwithprimarykey',
            name='id',
            field=models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID'),
        ),
    ]
```
</details>

If I was to switch the order so that the field is added before the other one is removed, then the migration applies cleanly.

[I will update this comment with next steps shortly, either this needs to be resolved in this PR as they're very similar/related issues or I create a new ticket for this]

bcail:
Sure, I'll plan to work on this.

@sarahboyce (or anyone) - since this issue has to do with generating and applying migrations to the database, is there a good place to add an integration test that handles both generating and applying migrations?

bcail:
@sarahboyce I've updated the PR, with a test for fixing that sqlite syntax error you mentioned. However, that scenario (removing a primary key from a table so the table is empty) is failing on mysql, saying we have to use `DROP TABLE` instead of removing the field.

Maybe this scenario of dropping the primary key field so the model is empty isn't something we need to try to support?

sarahboyce:
> However, that scenario (removing a primary key from a table so the table is empty) is failing on mysql, saying we have to use DROP TABLE instead of removing the field.

The table won't be empty in the end as we add a field. Can we not add the new id before dropping the old primary key field?

bcail:
@sarahboyce You're talking about the ticket [35416](https://code.djangoproject.com/ticket/35416) case right? Where the model ends up empty and it's just the implicit ID field that's added in the migration? If so, I think we would at least need to modify the old field, to make it not a primary key, then add the new implicit ID field, and then drop the old field. (I don't think Postgres likes to have multiple primary key fields at once.)

bcail:
OK, I'll look more into the case where field removals would result in a table with no columns.

bcail:
@sarahboyce It looks like in [Schema.remove_field](https://github.com/django/django/blob/main/django/db/backends/base/schema.py#L788) I could check to see if the model has no fields left, and if so, I could switch it to a `delete_model` call? Should I put the check in the MySQL schema, or in the base schema? Seems like maybe we would want to keep this just for MySQL?

sarahboyce:
> @sarahboyce It looks like in [Schema.remove_field](https://github.com/django/django/blob/main/django/db/backends/base/schema.py#L788) I could check to see if the model has no fields left, and if so, I could switch it to a `delete_model` call? Should I put the check in the MySQL schema, or in the base schema? Seems like maybe we would want to keep this just for MySQL?

This is not quite what I would expect. If I have a model:
```python
class SimpleModel(Model):
    pass
```
I do have a table, with one implicit id field. 
If I add a primary key field (which removes the implicit id), migrate, change my mind, put it back again, migrate - I would expect that I'm back where I started. The table being deleted in this case would be unexpected to me. 
At no point am I expecting to have a table with no columns but I might be missing something, does that make sense to you?

bcail:
Sorry, right - so at least on mysql, we need to add a step to alter the old field removing the pk attribute, move the implicit pk field add up, and then drop the old pk field. I'll need to look at it more, and see if that needs to be in the migrations-generation step or if it could somehow be done in the `schema.remove_field` step.

PR #17927, opened 2024-03-01
Refs #21286 -- Removed invalid PK data comments in serializer data tests

adamzap:
Removed commented-out test data that fails the `fields.E201` system check. `FileField`s (and thus `ImageField`s) cannot be primary keys, so serialization support for this case should not be implemented.


felixxm (commented):
@adamzap Thanks :+1: 

I also removed `XMLPKData`.

PR #17944, opened 2024-03-06
ngettext returns undefined if there is either invalid plural or catalog with empty array provided.

pkawula:
# Branch description
We've managed to situation when in catalog we had:
```js
const newcatalog = {
  '%s something': [],
  ...
}
```
When we invoked `window.ngettext('%s something', '%s somethings', 1);` the `ngettext` function returned undefined which caused the `window.interpolate` to be called with undefined and the page crashed.
The proposed solution is to always return string with a fallback to original string provided.

# Checklist

### DIFF
--- a/django/views/templates/i18n_catalog.js
+++ b/django/views/templates/i18n_catalog.js
@@ -42,7 +42,15 @@
       if (typeof value === 'undefined') {
         return (count == 1) ? singular : plural;
       } else {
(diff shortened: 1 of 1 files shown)


pkawula:
Oh, I am sorry! I did it in a bit rush.
Closing this now, and will make this in a right way

nessita:
> Oh, I am sorry! I did it in a bit rush. Closing this now, and will make this in a right way

Thank you, we appreciate this! :100:

PR #17977, opened 2024-03-14
Docs: Updated content of "sites" framework page.

sdarwin:
Hi,
When learning the sites framework, a few questions came to mind and needed to be researched. The idea with this pull request is to address those topics. Feedback is welcome.

- When considering adding "sites" to our Django site in order to enable a feature of Allauth, it was not clear if the typical implementation of "sites" could be done with one application server (ideally) or must be scaled across multiple servers, each with their own SITE_ID (which would be problematic).  It seems the original design would require that.  Thus, the sentences about "multiple app servers" to explain the point.  I think it's interesting information.  

- added the word "optional" to the directive that you must add a SITE_ID in the settings.  

- added a mention about what is "recommended". If you believe that is going too far, and it's not recommended, it could be edited.


mahiuddin-dev:
Yes. I like your point

nessita:
As mentioned in the ticket, I think the changes should focus only on updating the "Enabling the sites framework" to reiterate the ability to use `get_current_site`.

sdarwin:
Hi Nessita,

I see that you are taking issue with word "server".  

To revisit that particular point, in the original implementation of "sites", where you would hard-code the SITE_ID in the settings file, then in that case... imagine if you would like to have varying behavior for multiple so-called "sites".  One nginx vhost, in a docker container, isn't enough.   Horizontally scaled replicas of that same container with the same files isn't enough.   You really have to do something "extra". And for that I am calling them "servers", to emphasis the point.  Most features of Django will "just work" on your one server, when you update the python code.    But not this.  Not a static SITE_ID.     So perhaps the wording could be altered, and I am open to suggestions. "server" conveyed the idea.  Perhaps "instances" or "vhosts" or "deployment" instead.  In reality, in the world of k8s, you'd probably have multiple deployments. Or multiple VMs. 

By the way, pytest is failing on the Boost website if SITE_ID is missing.  :-) But per this very issue here, it should be ok , to not set that variable.

PR #17981, opened 2024-03-15
Fixed #35233 -- Moved template engine system checks to backend methods.

g-nie:
[ticket-35233](https://code.djangoproject.com/ticket/35233)

I used the [staticfiles finders checks](https://code.djangoproject.com/ticket/35233#comment:3) as the template for refactoring this.
The E001 is removed because it's covered by the [ImproperlyConfigured in Engine](https://github.com/django/django/blob/593067a8ee43e2167c5ffc92e3cc3c5e40ec4aa4/django/template/engine.py#L43), and we now force backends to all be instantiated during checks.


django/core/checks/templates.py
```diff
@@ -1,75 +1,16 @@
-import copy
-from collections import defaultdict
+from django.template import engines
```
adamchainz, line 1: Can we push this to an inner import in `check_templates()`? It will help avoid loading the whole `django.template` package when checks *aren’t* run, such as when running some commands in production.

Yes, This wasn’t the case before, but IMO the change is worth it since we’re touching this module and it’s now only one check function.

django/template/backends/base.py
```diff
@@ -23,6 +23,12 @@ def __init__(self, params):
                 "Unknown parameters: {}".format(", ".join(params))
             )
 
+    def check(self, **kwargs):
+        raise NotImplementedError(
+            "subclasses may provide a check() method to verify the template engine "
+            "is configured correctly."
+        )
```
adamchainz, line 30: I know this is copied from the finders, but now I think about it, `NotImplementedError` is a bit weird. It forces the caller to catch it, complicating the implementation. We don’t do that for most other base class methods, like `Model.check()` or whatever.

Returning an empty list makes more sense to me:

```suggestion
        return []
```

django/template/backends/django.py
```diff
@@ -25,6 +37,54 @@ def __init__(self, params):
         super().__init__(params)
         self.engine = Engine(self.dirs, self.app_dirs, **options)
 
+    def check(self, **kwargs):
+        errors = []
+        errors.extend(self._check_string_if_invalid_is_string())
+        errors.extend(self._check_for_template_tags_with_the_same_name())
+        return errors
```
adamchainz, line 44: Shorter:
```suggestion
        return [
            *self._check_string_if_invalid_is_string(),
            *self._check_for_template_tags_with_the_same_name(),
        ]
```

django/template/backends/django.py
```diff
@@ -25,6 +37,54 @@ def __init__(self, params):
...
+
+    @staticmethod
+    def _check_string_if_invalid_is_string():
```
adamchainz, line 47: Okay, you’ve missed part of the motivation behind the ticket. The `check()` method should check *this backend only*. It should be an instance method, not a static one:

```suggestion
    def _check_string_if_invalid_is_string(self):
```

Then the method should check the configuration of only *this* backend via e.g. `self.engine`. Looping over `settings.TEMPLATES` checks *all* backends.

Both check methods need changing this way.

adamchainz (changes requested):
Thanks for getting this started. Good work navigating the changes and working on the tests.

There’s one key oversight, that you left the check functions checking *all* backends, rather than specific backends. I left a detailed comment on this. I didn’t review the test changes because I think you’ll want to make more in order to match that change.

django/template/backends/django.py
g-nie: Ah that's right, leaving the `settings.TEMPLATES` in would mean checking all engines, for all engines. So, updating `_check_string_if_invalid_is_string` to behave as an instance method should be straightforward. The `_check_for_template_tags_with_the_same_name` is an interesting case though, because as far as I can tell, the logic is not `Engine` instance specific. It needs to check for name duplicates across backends so it probably requires some kind of access to `settings.TEMPLATES` (?). The way I updated this check, it runs only  for `DjangoTemplates` but it also may do duplicate work for the >1 `DjangoTemplates` backends case, which is not great. The other alternative is to leave that as a _global_ check in `checks/templates.py` but I don't like this much either since it kind of goes against our plan here.

adamchainz: It’s fine to repeat `_check_for_template_tags_with_the_same_name` per backend. It still depends on the per-backend `libraries` option, so one backend may have a clash whilst another doesn’t.

g-nie: > But there’s no problem with two backends using the same name for different libraries.

I see. This confused me a bit because there's an explicit test for the 2 DjangoTemplates backends case, which expects to get the warning: https://github.com/django/django/blob/b07e2d57a000d98c73492e5242fed91d502a780a/tests/check_framework/test_templates.py#L145
So probably the test is wrong/needs to be updated?
(I should have mentioned this test already, my bad)

g-nie: Giving this another thought it's obvious that the test needs updating, so I'm pushing changes to match that behavior. Thank you for the guidance Adam.

django/template/backends/base.py
g-nie: Yeah good point, this makes more sense to me too,

django/core/checks/templates.py
g-nie: Sure, sounds good to me.

django/template/backends/django.py
```diff
@@ -25,6 +37,46 @@ def __init__(self, params):
...
+
+    def _check_string_if_invalid_is_string(self):
+        string_if_invalid = self.engine.string_if_invalid
+        if not isinstance(string_if_invalid, str):
+            error = copy.copy(E002)
+            error.msg = error.msg.format(
+                string_if_invalid, type(string_if_invalid).__name__
+            )
+            return [error]
+        return []
```
adamchainz, line 54: Whilst we’re touching this, let’s drop the weird `copy()` pattern. Most checks just instantiate the message objects directly, we can do that here too and it makes the code clearer:

```suggestion
        value = self.engine.string_if_invalid
        if not isinstance(value, str):
            return[
                Error(
                    "'string_if_invalid' in TEMPLATES OPTIONS must be a "
                    f"string but got: {value!r} "
                    f"({type(value)}).",
                    obj=self,
                    id="templates.E002",
                )
        return []
```

I swapped for an f-string and used `!r` to repr the value and dropped `__name__` to show the full type path, like:

```
In [1]: class X:
   ...:     pass
   ...:

In [2]: type(X())
Out[2]: __main__.X

In [3]: type(X()).__name__
Out[3]: 'X'
```

I also added the `obj` value for the current backend for identification of the error, now that we are checking the separate the backends.

This will need some repetition of the expected messages in the tests, but that’s in line with other checks and IMO a good thing as it avoids the tests “peeking” into the implementation.

django/template/backends/django.py
```diff
@@ -1,15 +1,27 @@
...
 
+E002 = Error(
+    "'string_if_invalid' in TEMPLATES OPTIONS must be a string but got: {} ({}).",
+    id="templates.E002",
+)
+W003 = Warning(
+    "{} is used for multiple template tag modules: {}",
+    id="templates.E003",
+)
+
```
adamchainz, line 24: per below comment
```suggestion
```

django/template/backends/django.py
```diff
@@ -1,15 +1,27 @@
+import copy
```
adamchainz, line 1: per below comment
```suggestion
```

django/template/backends/django.py
```diff
@@ -25,6 +37,46 @@ def __init__(self, params):
...
+                    Warning(
+                        W003.msg.format(
+                            repr(library_name),
+                            ", ".join(repr(item) for item in sorted(items)),
+                        ),
+                        id=W003.id,
+                    )
+                )
+
+        return errors
```
adamchainz, line 78: Ditto, inline the creation of the `Warning`, use an f-string, and provide `obj`:

```suggestion
        errors = []
        libraries = defaultdict(set)

        for module_name, module_path in get_template_tag_modules():
            libraries[module_name].add(module_path)

        for module_name, module_path in self.engine.libraries.items():
            libraries[module_name].add(module_path)

        for library_name, items in libraries.items():
            if len(items) > 1:
               items = ", ".join(repr(item) for item in sorted(items))
                errors.append(
                    Warning(
                        f"{library_name!r} is used for multiple template tag "
                        f"modules: {items}",
                        obj=self,
                        id="templates.E003",
                    )
                )

        return errors
```

docs/ref/checks.txt
```diff
@@ -573,9 +573,6 @@ Templates
 The following checks verify that your :setting:`TEMPLATES` setting is correctly
 configured:
 
-* **templates.E001**: You have ``'APP_DIRS': True`` in your
```
adamchainz, line 576: I think we should leave the docs in but add a note, like we already have for e.g. `security.W007`. An extra line like:

```
*This check is removed in Django 5.1 as system checks may now raise `ImproperlyConfigured` instead.*
```

tests/check_framework/test_templates.py
```diff
@@ -1,59 +1,74 @@
...
+            mock.patch.object(
+                django_engine, "_check_string_if_invalid_is_string", lambda: [error1]
+            ),
+            mock.patch.object(
+                django_engine,
+                "_check_for_template_tags_with_the_same_name",
+                lambda: [error2],
+            ),
+        ):
+            self.assertListEqual(django_engine.check(), [error1, error2])
```
adamchainz, line 57: I really appreciate that you’ve tested the top-level function, but I think the tests are too complicated, especially given how simple the function is. Also, mocking should be avoided -  `override_settings` can instead be used to swap the engine classes, if they are defined outside of the test method.

Below is an untested example test that I think would suffice for everything:

```
class ErrorEngine(BaseEngine):
    def check(self, **kwargs):
        return [Error("Example")]


class CheckTemplatesTests(SimpleTestCase):
    @override_settings(
        TEMPLATES=[
            {"BACKEND": f"{__name__}.{ErrorEngine.__qualname__}"},
            {"BACKEND": f"{__name__}.{ErrorEngine.__qualname__}"},
        ]
    )
    def test_errors_aggregated(self):
        errors = check_templates(None)
        self.assertEqual(
            errors,
            [Error("Example")] * 2,
        ]
```

Note you don’t normally need `assertListEqual` or other typed functions, unittest maps `assertEqual` to type-specific functions: https://docs.python.org/3.12/library/unittest.html#type-specific-methods

tests/check_framework/test_templates.py
```diff
@@ -78,30 +93,36 @@ def setUpClass(cls):
             string_if_invalid2, type(string_if_invalid2).__name__
         )
 
+    @staticmethod
+    def _check_all_engines():
```
adamchainz, line 97: `@staticmethod` should [generally be avoided](https://www.seporaitis.net/posts/2020/05/05/python-staticmethod-usage/):

```
    def _check_all_engines(self):
```

tests/check_framework/test_templates.py
```diff
@@ -120,43 +141,51 @@ def setUpClass(cls):
...
+            "NAME": name,
             "OPTIONS": {
                 "libraries": {
                     module_name: f"check_framework.template_test_apps.{module_path}",
                 },
             },
         }
 
+    @staticmethod
+    def _check_all_engines():
```
adamchainz, line 156:
```suggestion
    def _check_all_engines(self):
```

adamchainz (changes requested):
Nice, looking better, another round of refinement..

tests/check_framework/test_templates.py
g-nie: Noted, I was not aware of that.

django/template/backends/django.py
g-nie: Nice that's clearer for sure 🙂

django/template/backends/django.py
```diff
@@ -25,6 +27,50 @@ def __init__(self, params):
...
+        errors = []
+        libraries = defaultdict(set)
+
+        for module_name, module_path in get_template_tag_modules():
+            libraries[module_name].add(module_path)
+
+        for module_name, module_path in self.engine.libraries.items():
+            libraries[module_name].add(module_path)
+
+        for library_name, items in libraries.items():
```
adamchainz, line 60: reordering for clarity

```suggestion
        libraries = defaultdict(set)

        for module_name, module_path in get_template_tag_modules():
            libraries[module_name].add(module_path)

        for module_name, module_path in self.engine.libraries.items():
            libraries[module_name].add(module_path)

        errors = []
        for library_name, items in libraries.items():
```

tests/check_framework/test_templates.py
```diff
@@ -1,162 +1,161 @@
...
         )
-        cls.error2.msg = cls.error2.msg.format(
-            string_if_invalid2, type(string_if_invalid2).__name__
+
+    def _check_all_engines(self, all_engines):
+        return list(
+            chain.from_iterable(
+                e._check_string_if_invalid_is_string() for e in all_engines
+            )
         )
```
adamchainz, line 65: This no longer checks *all* engines, but the passed in engines. Can you change the naming accordingly? And in tests below.
```suggestion
    def _check_engines(self, engines):
        return list(
            chain.from_iterable(
                e._check_string_if_invalid_is_string() for e in engines
            )
        )
```

tests/check_framework/test_templates.py
```diff
@@ -1,162 +1,161 @@
...
-                "templatetags.same_tags', "
-                "'check_framework.template_test_apps.same_tags_app_2."
-                "templatetags.same_tags'",
-            ),
-            id=W003.id,
-        )
-
-    @staticmethod
-    def get_settings(module_name, module_path):
+    def get_settings(self, module_name, module_path, name="django"):
```
adamchainz, line 101: Nice!

tests/check_framework/test_templates.py
```diff
@@ -1,162 +1,161 @@
...
+            obj=engine,
+            id="templates.E003",
+        )
+
+    def _check_all_engines(self, all_engines):
+        return list(
+            chain.from_iterable(
+                e._check_for_template_tags_with_the_same_name() for e in all_engines
+            )
+        )
```
adamchainz, line 124: ditto
```suggestion
    def _check_engines(self, engines):
        return list(
            chain.from_iterable(
                e._check_for_template_tags_with_the_same_name() for e in engines
            )
        )
```

adamchainz (changes requested):
Some small changes but looking good!

adamchainz (approved):
Looking good. Maybe fellows will spot a few more things to improve.

This may just be worthy of a release note, to help any third-party package template backends discover the new method.

django/template/backends/django.py
```diff
@@ -25,6 +27,51 @@ def __init__(self, params):
...
+        ]
+
+    def _check_string_if_invalid_is_string(self):
+        value = self.engine.string_if_invalid
+        if not isinstance(value, str):
+            return [
+                Error(
+                    "'string_if_invalid' in TEMPLATES OPTIONS must be a "
+                    f"string but got: {value!r} "
+                    f"({type(value)}).",
```
felixxm: We don't allow function/method calls in f-strings. This guideline is from [Python coding style](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#python-style).

felixxm (commented):
@g-nie Thanks :+1: 

I will push final edits tomorrow morning.

django/template/backends/django.py
adamchainz: Ah, sorry, that was my suggestion.

felixxm (commented):
@g-nie Thanks :+1: I added a release note, documentation changes, and pushed small edits.

django/template/backends/django.py
```diff
@@ -25,6 +27,50 @@ def __init__(self, params):
...
+
+        for library_name, items in libraries.items():
+            if len(items) > 1:
+                items = ", ".join(repr(item) for item in sorted(items))
+                errors.append(
+                    Warning(
+                        f"{library_name!r} is used for multiple template tag "
+                        f"modules: {items}",
+                        obj=self,
+                        id="templates.E003",
```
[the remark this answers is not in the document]
giannis-bugshell: Ah yes, that was a leftover.

Thank you for the changes @felixxm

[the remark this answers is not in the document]
g-nie: Oh, that was a leftover.

Thank you for the changes @felixxm

PR #17984, opened 2024-03-16
Fixed #35309 -- Made prefetch clear ordering for single-valued relationships.

LLyaudet:
To avoid useless order by when prefetching by id
for ForwardManyToOneDescriptor (ForeignKey and OneToOneField forward) and ReverseOneToOneDescriptor (OneToOneField backward).

# Trac ticket number

ticket-35309

# Branch description
For big prefetches with prefetch_related() (hundreds or thousands of ids),
I noticed a speedup of 10 to 15 % when removing the useless order by
coming from class Meta: ordering =...

# Checklist

### DIFF
--- a/django/db/models/fields/related_descriptors.py
+++ b/django/db/models/fields/related_descriptors.py
@@ -195,6 +195,9 @@ class ForwardManyToOneDescriptor:
         else:
             query = {"%s__in" % self.field.related_query_name(): instances}
         queryset = queryset.filter(**query)
+        # There can be only one object prefetched for each instance so clear
+        # ordering if the query allows it without side effects.
+        queryset.query.clear_ordering()
 
         # Since we're going to assign directly in the cache,
         # we must manage the reverse relation cache manually.
@@ -469,6 +472,9 @@ class ReverseOneToOneDescriptor:
         instances_dict = {instance_attr(inst): inst for inst in instances}
         query = {"%s__in" % self.related.field.name: instances}
         queryset = queryset.filter(**query)
+        # There can be only one object prefetched for each instance so clear
+        # ordering if the query allows it without side effects.
+        queryset.query.clear_ordering()
 
         # Since we're going to assign directly in the cache,
         # we must manage the reverse relation cache manually.
--- a/tests/prefetch_related/test_prefetch_related_objects.py
+++ b/tests/prefetch_related/test_prefetch_related_objects.py
@@ -1,7 +1,7 @@
 from django.db.models import Prefetch, prefetch_related_objects
 from django.test import TestCase
 
-from .models import Author, Book, Reader
+from .models import Author, Book, House, Reader, Room
 
 
 class PrefetchRelatedObjectsTests(TestCase):
@@ -33,6 +33,17 @@ class PrefetchRelatedObjectsTests(TestCase):
         cls.reader1.books_read.add(cls.book1, cls.book4)
         cls.reader2.books_read.add(cls.book2, cls.book4)
 
+        cls.house1 = House.objects.create(name="b1", address="1")
+        cls.house2 = House.objects.create(name="b2", address="2")
+
+        cls.room1 = Room.objects.create(name="a1", house=cls.house1)
+        cls.room2 = Room.objects.create(name="a2", house=cls.house2)
+
+        cls.house1.main_room = cls.room1
+        cls.house1.save()
+        cls.house2.main_room = cls.room2
+        cls.house2.save()
+
     def test_unknown(self):
         book1 = Book.objects.get(id=self.book1.id)
         with self.assertRaises(AttributeError):
@@ -58,19 +69,74 @@ class PrefetchRelatedObjectsTests(TestCase):
 
     def test_foreignkey_forward(self):
         authors = list(Author.objects.all())
-        with self.assertNumQueries(1):
+        with self.assertNumQueries(1) as ctx:
             prefetch_related_objects(authors, "first_book")
+        self.assertNotIn("ORDER BY", ctx.captured_queries[0]["sql"])
 
         with self.assertNumQueries(0):
             [author.first_book for author in authors]
 
+        authors = list(Author.objects.all())
+        with self.assertNumQueries(1) as ctx:
+            prefetch_related_objects(
+                authors,
+                Prefetch("first_book", queryset=Book.objects.order_by("-title")),
+            )
+        self.assertNotIn("ORDER BY", ctx.captured_queries[0]["sql"])
+
     def test_foreignkey_reverse(self):
         books = list(Book.objects.all())
-        with self.assertNumQueries(1):
+        with self.assertNumQueries(1) as ctx:
             prefetch_related_objects(books, "first_time_authors")
+        self.assertIn("ORDER BY", ctx.captured_queries[0]["sql"])
 
         with self.assertNumQueries(0):
             [list(book.first_time_authors.all()) for book in books]
+
+        books = list(Book.objects.all())
+        with self.assertNumQueries(1) as ctx:
+            prefetch_related_objects(
+                books,
+                Prefetch(
+                    "first_time_authors",
+                    queryset=Author.objects.order_by("-name"),
+                ),
+            )
+        self.assertIn("ORDER BY", ctx.captured_queries[0]["sql"])
+
+    def test_one_to_one_forward(self):
+        houses = list(House.objects.all())
+        with self.assertNumQueries(1) as ctx:
+            prefetch_related_objects(houses, "main_room")
+        self.assertNotIn("ORDER BY", ctx.captured_queries[0]["sql"])
+
+        with self.assertNumQueries(0):
+            [house.main_room for house in houses]
+
+        houses = list(House.objects.all())
+        with self.assertNumQueries(1) as ctx:
+            prefetch_related_objects(
+                houses,
+                Prefetch("main_room", queryset=Room.objects.order_by("-name")),
+            )
+        self.assertNotIn("ORDER BY", ctx.captured_queries[0]["sql"])
+
+    def test_one_to_one_reverse(self):
+        rooms = list(Room.objects.all())
+        with self.assertNumQueries(1) as ctx:
+            prefetch_related_objects(rooms, "main_room_of")
+        self.assertNotIn("ORDER BY", ctx.captured_queries[0]["sql"])
+
+        with self.assertNumQueries(0):
+            [room.main_room_of for room in rooms]
+
+        rooms = list(Room.objects.all())
+        with self.assertNumQueries(1) as ctx:
+            prefetch_related_objects(
+                rooms,
+                Prefetch("main_room_of", queryset=House.objects.order_by("-name")),
+            )
+        self.assertNotIn("ORDER BY", ctx.captured_queries[0]["sql"])
 
     def test_m2m_then_m2m(self):
         """A m2m can be followed through another m2m."""


LLyaudet:
I have a problem with the checks for tests.
It seems that connection.queries is empty in the test env.
I will try to see what I can do, to make it work.

LLyaudet:
I have a problem for installing pre-commit in the venv I created for Django source code repository.
Nevertheless I passed black 24.3.0 on all the repository and the following files would be modified:
tests/asgi/tests.py
tests/cache/failing_cache.py
diff --git a/tests/asgi/tests.py b/tests/asgi/tests.py
index 963f45f798..3689c7354f 100644
--- a/tests/asgi/tests.py
+++ b/tests/asgi/tests.py
@@ -217,7 +217,6 @@ class ASGITest(SimpleTestCase):
 
         # Request class that always fails creation with RequestDataTooBig.
         class TestASGIRequest(ASGIRequest):
-
             def __init__(self, scope, body_file):
                 super().__init__(scope, body_file)
                 raise RequestDataTooBig()
diff --git a/tests/cache/failing_cache.py b/tests/cache/failing_cache.py
index 1c9b5996d6..dc6fda3aa6 100644
--- a/tests/cache/failing_cache.py
+++ b/tests/cache/failing_cache.py
@@ -2,7 +2,6 @@ from django.core.cache.backends.locmem import LocMemCache
 
 
 class CacheClass(LocMemCache):
-
     def set(self, *args, **kwargs):
         raise Exception("Faked exception saving to cache")
 
I haven't commited those changes since it is not related to the PR.

felixxm:
Closing per ticket.

felixxm:
Reopen per ticket :)

django/db/models/fields/related_descriptors.py
```diff
@@ -195,6 +195,13 @@ def get_prefetch_querysets(self, instances, querysets=None):
         else:
             query = {"%s__in" % self.field.related_query_name(): instances}
         queryset = queryset.filter(**query)
+        # Since there can be only one object prefetched for each instance
+        # order by is insignificant and slows down the query.
+        # queryset = queryset.order_by()
+        # queryset was cloned by filter() just above, no need to re-clone it.
+        queryset._in_place_empty_order_by()
+        # No need to handle the case where the query is_sliced (LIMIT)
+        # because no filter can be added to a sliced query.
```
charettes, line 204: I'm pretty sure you can simply do [the following](https://github.com/django/django/blob/b07e2d57a000d98c73492e5242fed91d502a780a/django/db/models/sql/query.py#L2272C9-L2286) to avoid cloning and work around the possible limit issue.

```suggestion
        queryset = queryset.filter(**query)
        # There can be only one object prefetched for each instance so
        # clear ordering if the query allows it without side effects.
        queryset.query.clear_ordering(force=False)
```

django/db/models/fields/related_descriptors.py
```diff
@@ -469,6 +476,13 @@ def get_prefetch_querysets(self, instances, querysets=None):
         instances_dict = {instance_attr(inst): inst for inst in instances}
         query = {"%s__in" % self.related.field.name: instances}
         queryset = queryset.filter(**query)
+        # Since there can be only one object prefetched for each instance
+        # order by is insignificant and slows down the query.
+        # queryset = queryset.order_by()
+        # queryset was cloned by filter() just above, no need to re-clone it.
+        queryset._in_place_empty_order_by()
+        # No need to handle the case where the query is_sliced (LIMIT)
+        # because no filter can be added to a sliced query.
```
charettes, line 485: Ditto about `clear_ordering`

django/db/models/query.py
```diff
@@ -1701,6 +1701,9 @@ def order_by(self, *field_names):
         obj.query.add_ordering(*field_names)
         return obj
 
+    def _in_place_empty_order_by(self):
+        self.query.clear_ordering(force=True, clear_default=True)
+
```
charettes, line 1706: I don't think this warrants a `QuerySet` level method. Calling `clear_ordering` directly should do. We also want to avoid force clearing if possible as users might compose a complex query through `Prefetch(queryset)` where distinct and / or slicing require `ORDER BY` to remain present to preserve the semantic of the query.

charettes (commented):
Left some comments, in summary

1. Call `sql.Query.clear_ordering` directly
2. Avoid passing `force=True`
3. Reuse existing test scaffolding

tests/prefetch_related/models.py
```diff
@@ -313,3 +313,26 @@ class Flea(models.Model):
...
+    class Meta:
+        ordering = ["name"]
+
+
+class C35309(models.Model):
+    name = models.CharField(max_length=200)
+    a = models.OneToOneField(A35309, related_name="c", on_delete=models.CASCADE)
+
+    class Meta:
+        ordering = ["name"]
```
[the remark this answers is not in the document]
LLyaudet: Yes I suggested models that would fit for that in the Ticket one hour ago. I will do.

tests/prefetch_related/tests.py
```diff
@@ -2052,3 +2055,81 @@ def __getattribute__(self, name):
...
+                )
+            )
+            self.assertNotIn("ORDER BY", ctx.captured_queries[-1]["sql"])
+        with self.assertNumQueries(2) as ctx:
+            list(
+                A35309.objects.all().prefetch_related(
+                    Prefetch("c", queryset=C35309.objects.order_by("-name")),
+                )
+            )
+            self.assertNotIn("ORDER BY", ctx.captured_queries[-1]["sql"])
```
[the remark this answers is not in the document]
LLyaudet: I will check also the existing tests.

django/db/models/fields/related_descriptors.py
LLyaudet: Yes that's what I wrote in the ticket one hour ago. I will take this solution.

LLyaudet: Note that you need `clear_default=True` as I did. Otherwise it will not work.

django/db/models/query.py
LLyaudet: Ok that was one the options I suggested. Will do.

django/db/models/fields/related_descriptors.py
charettes: No need to explicit pass it; `clear_ordering(clear_default)` [default value is already `True`](https://github.com/django/django/blob/b07e2d57a000d98c73492e5242fed91d502a780a/django/db/models/sql/query.py#L2272C9-L2286).

LLyaudet: Indeed, sorry about that, it is here

tests/prefetch_related/tests.py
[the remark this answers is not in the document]
LLyaudet: The tests for one to one are splitted between multiple files and there is no such tests in test_prefetch_related_objects. I will have to add at least one test case:
image

LLyaudet:
Hello @charettes :)
I did the improvements.
I think it should be ok now.

django/db/models/fields/related_descriptors.py
```diff
@@ -195,6 +195,13 @@ def get_prefetch_querysets(self, instances, querysets=None):
         else:
             query = {"%s__in" % self.field.related_query_name(): instances}
         queryset = queryset.filter(**query)
+        # Since there can be only one object prefetched for each instance
+        # order by is insignificant and slows down the query.
+        # queryset = queryset.order_by()
+        # queryset was cloned by filter() just above, no need to re-clone it.
+        queryset.query.clear_ordering()
+        # No need to handle the case where the query is_sliced (LIMIT)
+        # because no filter can be added to a sliced query.
```
charettes, line 204: Again, we want to `force=False` as users might compose a complex query through `Prefetch(queryset)` where distinct and / or slicing requires `ORDER BY` to remain present to preserve the semantic of the query.

```suggestion
        # There can be only one object prefetched for each instance so
        # clear ordering if the query allows it without side effects.
        queryset.query.clear_ordering(force=False)
```

There is no need to be overly aggressive in the optimization given `clear_ordering(force=False)` will handle
1. `Meta.ordering` which is the reported issue in this ticket
2. An explicitly provided `order_by` assuming neither `distinct` or `slicing` is specified (which can be handy when windows functions are involved)

tests/prefetch_related/test_prefetch_related_objects.py
```diff
@@ -58,20 +69,84 @@ def test_m2m_reverse(self):
...
         with self.assertNumQueries(0):
             [list(book.first_time_authors.all()) for book in books]
 
+        books = list(Book.objects.all())
+        with self.assertNumQueries(1) as ctx:
+            prefetch_related_objects(
+                books,
+                Prefetch(
+                    "first_time_authors",
+                    queryset=Author.objects.order_by("-name"),
```
charettes, line 105: I'm not convinced we need to test explicit usage of `order_by` but I'll let mergers decide.

tests/prefetch_related/test_prefetch_related_objects.py
```diff
@@ -58,20 +69,84 @@ def test_m2m_reverse(self):
...
+        with self.assertNumQueries(0):
+            [house.main_room for house in houses]
+
+        houses = list(House.objects.all())
+        with self.assertNumQueries(1) as ctx:
+            prefetch_related_objects(
+                houses,
+                Prefetch(
+                    "main_room",
+                    queryset=Room.objects.order_by("-name"),
```
charettes, line 125: Ditto

tests/prefetch_related/test_prefetch_related_objects.py
```diff
@@ -58,20 +69,84 @@ def test_m2m_reverse(self):
...
+        with self.assertNumQueries(0):
+            [room.main_room_of for room in rooms]
+
+        rooms = list(Room.objects.all())
+        with self.assertNumQueries(1) as ctx:
+            prefetch_related_objects(
+                rooms,
+                Prefetch(
+                    "main_room_of",
+                    queryset=House.objects.order_by("-name"),
```
charettes, line 145: Ditto

charettes (commented):
Tests are looking good, I'm not convinced explicitly testing for `Prefetch(qs.order_by)` is necessary but I'll mergers decide.

The last blocker is the usage of `clear_ordering` without specifying `force=False`. I believe forcing the clear might be risky as it's possible users are passing complex queries through `Prefetch` that warrant the use of `order_by` so I think it's preferable not to risk it.

`force=False` should handle the reported use case and only make a different in marginal use cases of `Prefetch` anyway so I don't think it removes any value from the patch.

django/db/models/fields/related_descriptors.py
LLyaudet: `def clear_ordering(self, force=False, clear_default=True):`
`force=False` is the default.
I don't understand your reasoning if you want to explicit `force=False` and not `clear_default=True`.

tests/prefetch_related/test_prefetch_related_objects.py
LLyaudet: From my point of view this is necessary since it corresponds to the case where `queryset.query.order_by` is filled, instead of the case where it is filled later because `queryset.query.default_ordering = True`

LLyaudet:
Hello @charettes 
I think there is no more any blocker.
We'll wait what the mergers say about testing the explicit orders.
The rest should be ok.
Best regards,
    Laurent Lyaudet

django/db/models/fields/related_descriptors.py
felixxm, line 204: Such a detailed comment is unnecessary. I'd prefer Simon's suggestion:
```suggestion
        # There can be only one object prefetched for each instance so
        # clear ordering if the query allows it without side effects.
        queryset.query.clear_ordering()
```

django/db/models/fields/related_descriptors.py
```diff
@@ -469,6 +476,13 @@ def get_prefetch_querysets(self, instances, querysets=None):
         instances_dict = {instance_attr(inst): inst for inst in instances}
         query = {"%s__in" % self.related.field.name: instances}
         queryset = queryset.filter(**query)
+        # Since there can be only one object prefetched for each instance
+        # order by is insignificant and slows down the query.
+        # queryset = queryset.order_by()
+        # queryset was cloned by filter() just above, no need to re-clone it.
+        queryset.query.clear_ordering()
+        # No need to handle the case where the query is_sliced (LIMIT)
+        # because no filter can be added to a sliced query.
```
felixxm, line 485:
```suggestion
        # There can be only one object prefetched for each instance so
        # clear ordering if the query allows it without side effects.
        queryset.query.clear_ordering()
```

felixxm:
@LLyaudet Thanks :+1: 

Can I ask why you checked the following point from the checklist?

_"The commit message is written in past tense, mentions the ticket number, and ends with a period."_

I'm trying to understand why folks blindly check all points without trying to actually do it first.

django/db/models/fields/related_descriptors.py
LLyaudet: Done

django/db/models/fields/related_descriptors.py
LLyaudet: Done

LLyaudet:
You're welcome @felixxm :)

> Can I ask why you checked the following point from the checklist?
> "The commit message is written in past tense, mentions the ticket number, and ends with a period."
> I'm trying to understand why folks blindly check all points without trying to actually do it first.

Because all commits messages mentions the ticket number, and ends with a period.
And for the past tense, I wasn't sure of how to apply it and wanted to create my PR fast because I spent a night on it :)
Sorry about that. What do you suggest to improve the current PR before merging ?
Some squashing/force pushing ?

felixxm (commented):
@LLyaudet Thanks :+1: Welcome aboard :boat:

LLyaudet:
Thanks @felixxm :)

PR #17991, opened 2024-03-18
 Refs #17235 -- Made MultiPartParser leave request.FILES immutable.

bcail:
# Trac ticket number

ticket-17235

# Branch description
Made request.FILES immutable. Continued the work from PR #10110 by @vinayinvicible.

# Checklist


django/utils/datastructures.py
```diff
@@ -170,8 +182,21 @@ def setlistdefault(self, key, default_list=None):
...
+        self._assert_mutable()
+        return super().pop(key, *args)
+
+    def popitem(self):
+        self._assert_mutable()
+        return super().popitem()
+
+    def clear(self):
+        self._assert_mutable()
+        super().clear()
```
sarahboyce, line 195: As `QueryDict` inherits `MultiValueDict`, you might be able to remove/refactor some of the code there. 
Might be more than just these

sarahboyce (commented):
Thank you @bcail ⭐ I have initial comments

sarahboyce:
> I wonder about backward compatability here. It's been this way _forever_ so, while it's not recommended, I'm sure someone will be relying on it being mutable.

Oh you make a good point 🤔
I guess this needs a release note at the very least

> One suggestion would be to add this change in behaviour as part of the new content type parsing (#17546). The next step there is for someone to write a DEP so this behaviour change could form part of that.

@bcail what do you think? Being part of the push for #17546 would be amazing. I really appreciate the work you've been doing and I think having a look at this and the DEP process could be a valuable experience if you're up for it?

bcail:
I'll plan to look into this, but if anyone wants to work on it first, feel free.

PR #18007, opened 2024-03-23
Refs #34406 -- Added support for GDAL curved geometries. 

smithdc1:
# Trac ticket number

ticket-34406

# Branch description
Adds curved geometry support for GDAL geometries. 

As part of this I tried to add the `ForceTo*` bindings (e.g. [ForceToPolygon](https://gdal.org/api/vector_c_api.html#_CPPv420OGR_G_ForceToPolygon12OGRGeometryH) ). However, I was unable to achieve this. I think it's likely due to me not understanding the return type. 

> Returns:  the converted geometry (ownership to caller).

While I've been unable to solve that problem I think it's likely reviewing the other work I've done here which adds recognition of these types and allows conversion between linear and curved geometries.

# Checklist

### DIFF
--- a/django/contrib/gis/gdal/prototypes/geom.py
+++ b/django/contrib/gis/gdal/prototypes/geom.py
@@ -85,6 +85,13 @@ is_3d = bool_output(lgdal.OGR_G_Is3D, [c_void_p])
 set_3d = void_output(lgdal.OGR_G_Set3D, [c_void_p, c_int], errcheck=False)
 is_measured = bool_output(lgdal.OGR_G_IsMeasured, [c_void_p])
 set_measured = void_output(lgdal.OGR_G_SetMeasured, [c_void_p, c_int], errcheck=False)
+has_curve_geom = bool_output(lgdal.OGR_G_HasCurveGeometry, [c_void_p, c_int])
+get_linear_geom = geom_output(
+    lgdal.OGR_G_GetLinearGeometry, [c_void_p, c_double, POINTER(c_char_p)]
+)
+get_curve_geom = geom_output(
+    lgdal.OGR_G_GetCurveGeometry, [c_void_p, POINTER(c_char_p)]
+)
 
 # Geometry modification routines.
 add_geom = void_output(lgdal.OGR_G_AddGeometry, [c_void_p, c_void_p])
--- a/docs/ref/contrib/gis/gdal.txt
+++ b/docs/ref/contrib/gis/gdal.txt
@@ -611,6 +611,26 @@ coordinate transformation:
         >>> polygon.geom_count
         1
 
+    .. attribute:: has_curve
+
+    .. versionadded:: 5.2
+
+    A boolean indicating if this geometry is or contains a curve geometry.
+
+    .. method:: get_linear_geometry
+
+    .. versionadded:: 5.2
+
+    Returns a linear version of the geometry. If no conversion can be made, the
+    original geometry is returned.
+
+    .. method:: get_curve_geometry
+
+    .. versionadded:: 5.2
+
+    Returns a curved version of the geometry. If no conversion can be made, the
+    original geometry is returned.
+
     .. attribute:: point_count
 
     Returns the number of points used to describe this geometry:
--- a/tests/gis_tests/gdal_tests/test_geom.py
+++ b/tests/gis_tests/gdal_tests/test_geom.py
@@ -8,6 +8,8 @@ from django.contrib.gis.gdal import (
     OGRGeomType,
     SpatialReference,
 )
+from django.contrib.gis.gdal.geometries import CircularString, CurvePolygon
+from django.contrib.gis.geos import GEOSException
 from django.template import Context
 from django.template.engine import Engine
 from django.test import SimpleTestCase
@@ -646,11 +648,11 @@ class OGRGeomTest(SimpleTestCase, TestDataMixin):
             ("Multilinestring", 5, True),
             ("MultiPolygon", 6, True),
             ("GeometryCollection", 7, True),
-            ("CircularString", 8, False),
-            ("CompoundCurve", 9, False),
-            ("CurvePolygon", 10, False),
-            ("MultiCurve", 11, False),
-            ("MultiSurface", 12, False),
+            ("CircularString", 8, True),
+            ("CompoundCurve", 9, True),
+            ("CurvePolygon", 10, True),
+            ("MultiCurve", 11, True),
+            ("MultiSurface", 12, True),
             # 13 (Curve) and 14 (Surface) are abstract types.
             ("PolyhedralSurface", 15, False),
             ("TIN", 16, False),
@@ -664,11 +666,11 @@ class OGRGeomTest(SimpleTestCase, TestDataMixin):
             ("Multilinestring Z", -2147483643, True),  # 1005
             ("MultiPolygon Z", -2147483642, True),  # 1006
             ("GeometryCollection Z", -2147483641, True),  # 1007
-            ("CircularString Z", 1008, False),
-            ("CompoundCurve Z", 1009, False),
-            ("CurvePolygon Z", 1010, False),
-            ("MultiCurve Z", 1011, False),
-            ("MultiSurface Z", 1012, False),
+            ("CircularString Z", 1008, True),
+            ("CompoundCurve Z", 1009, True),
+            ("CurvePolygon Z", 1010, True),
+            ("MultiCurve Z", 1011, True),
+            ("MultiSurface Z", 1012, True),
             ("PolyhedralSurface Z", 1015, False),
             ("TIN Z", 1016, False),
             ("Triangle Z", 1017, False),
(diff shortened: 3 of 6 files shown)


nessita:
@claudep Hi! Would you have any time to take a look to this PR?

claudep (approved):
I had a look, and couldn't find anything to object to. Thanks for the work! The only improvement that could be made (maybe), it's some docs complements at the top of the GDAL docs mentioning curve geometries. But that could be done later.

docs/ref/contrib/gis/gdal.txt
```diff
@@ -611,6 +611,26 @@ coordinate transformation:
         >>> polygon.geom_count
         1
 
+    .. attribute:: has_curve
+
+    .. versionadded:: 5.1
+
+    A boolean indicating if this geometry is or has curve geometry.
```
nessita, line 618: Is the "is or has" distinction really necessary? Could we just say "has"?

docs/ref/contrib/gis/gdal.txt
```diff
@@ -611,6 +611,26 @@ coordinate transformation:
...
+
+    .. method:: get_linear_geometry
+
+    .. versionadded:: 5.1
+
+    Returns a linear version of the geometry. If no conversion can be made the
+    original geometry is returned.
```
nessita, line 625: I'm a little worried about this statement, could we, in any way, signal to the caller whether the conversion could be made? Could return a boolean with the result?

docs/ref/contrib/gis/gdal.txt
```diff
@@ -611,6 +611,26 @@ coordinate transformation:
...
+
+    .. method:: get_curve_geometry
+
+    .. versionadded:: 5.1
+
+    Returns a curved version of the geometry. If no conversion can be made the
+    original geometry is returned.
```
nessita, line 632: Same concern as above.

docs/releases/5.1.txt
```diff
@@ -112,6 +112,12 @@ Minor features
   now support the optional ``srid`` argument (except for Oracle where it is
   ignored).
 
+* Added GDAL support curved geometries ``CurvePolygon``, ``CompoundCurve``,
```
nessita, line 115:
```suggestion
* Added GDAL support for curved geometries ``CurvePolygon``, ``CompoundCurve``,
```

tests/gis_tests/gdal_tests/test_geom.py
```diff
@@ -967,6 +969,60 @@ def test_multi_geometries_m_dimension(self):
...
+                    geom.geos
+        geom = OGRGeometry("POINT (0 1)")
+        self.assertIs(geom.has_curve, False)
+
+    def test_get_linear_and_curve_geometry(self):
+        circular_string = "CIRCULARSTRING (-0.797 0.466,-0.481 0.62,-0.419 0.473)"
+        geom = OGRGeometry(circular_string)
+        linear = geom.get_linear_geometry()
+        self.assertEqual(linear.geom_name, "LINESTRING")
+        self.assertIs(linear.has_curve, False)
```
nessita, line 988: I'm keen to separate these two in individual tests to be a little more aligned with unit testing primitives. What do you think?

nessita (commented):
Hello @smithdc1, thank you for this branch!

I added some comments but please let me know if those make sense. I would also like to know your thoughts on the docs suggestion from Claude :star2:

docs/ref/contrib/gis/gdal.txt
smithdc1: How about?
```suggestion
    A boolean indicating if this geometry is or contains a curve geometry.
```

docs/ref/contrib/gis/gdal.txt
smithdc1: Looking at this again, I think curved -> linear is likely always going to work, but linear -> curved may not (for example if it's just a straight line!). 

Maybe I should have stopped after the first sentence or could add something like this (from GDAL docs)

> Returns a geometry that has no CIRCULARSTRING, COMPOUNDCURVE, CURVEPOLYGON, MULTICURVE or MULTISURFACE in it, by approximating curve geometries.

But we should consider the linear -> curve case.

Currently the API here is similar to the GDAL API, albeit with 'pythonic' names. As far as I'm aware we've generally stayed close to mirroring the GDAL API. Checking if two geometries are the same is also very straight forward if users wanted to do that. I'm therefore rather unsure about this. 

@claudep would welcome your opinion on this.

tests/gis_tests/gdal_tests/test_geom.py
smithdc1: Sure! Let me have a look at that. Appreciate your feedback 🙏

docs/ref/contrib/gis/gdal.txt
claudep: Did you try adding a linear -> curved transform with a straight line in this patches tests?

smithdc1: Not specifically. Maybe I can add some more tests. 

``` python
>>> geom = OGRGeometry("LINESTRING (0 0, 1 0, 2 0)")
>>> geom2 = geom.get_curve_geometry()
>>> geom2.wkt
'LINESTRING (0 0,1 0,2 0)'
```

claudep: Yes, please!

claudep: Now about the original issue, I think it could be tricky to make any report about the conversion result. I would let to the user any control about the result. Agreed with @smithdc1 that at this level we are essentially mimicking the GDAL API and its return values.

nessita:
@smithdc1 I'm going thru pending PRs that were on my "maybe 5.1" list, I wanted to double check that this is not yet ready for a re-review (given the ticket flags). Is that accurate?

smithdc1:
Hi @nessita it is quite close. However, I'd not worry about getting this in ahead of the feature freeze. 

Let me update the versions and have a final look over. Maybe we can then get it in early in the next cycle?

nessita:
> Hi @nessita it is quite close. However, I'd not worry about getting this in ahead of the feature freeze.
> 
> Let me update the versions and have a final look over. Maybe we can then get it in early in the next cycle?

Yes, absolutely!

/remind me to check on this in one month and a half

claudep:
@nessita Do you think we could merge that work now?

nessita:
> @nessita Do you think we could merge that work now?

Thank you Claude for the nudge, I'll re-review later this week! (currently dealing with release blockers and security reports).

claudep:
No rush, thanks for your amazing work! :heart_eyes:

django/contrib/gis/gdal/prototypes/geom.py
```diff
@@ -85,6 +85,14 @@ def topology_func(f):
 set_3d = void_output(lgdal.OGR_G_Set3D, [c_void_p, c_int], errcheck=False)
 is_measured = bool_output(lgdal.OGR_G_IsMeasured, [c_void_p])
 set_measured = void_output(lgdal.OGR_G_SetMeasured, [c_void_p, c_int], errcheck=False)
+has_curve = bool_output(lgdal.OGR_G_HasCurveGeometry, [c_void_p, c_int])
+get_linear_geometry = geom_output(
+    lgdal.OGR_G_GetLinearGeometry, [c_void_p, c_double, POINTER(c_char_p)]
+)
+get_curve_geometry = geom_output(
+    lgdal.OGR_G_GetCurveGeometry, [c_void_p, POINTER(c_char_p)]
+)
```
nessita, line 94: Looking at the previous patterns, it seems that the trend is to use `geom` for `Geometry`, so perhaps:
```suggestion
get_linear_geom = geom_output(
    lgdal.OGR_G_GetLinearGeometry, [c_void_p, c_double, POINTER(c_char_p)]
)
get_curve_geo = geom_output(lgdal.OGR_G_GetCurveGeometry, [c_void_p, POINTER(c_char_p)])
```

docs/ref/contrib/gis/gdal.txt
```diff
@@ -611,6 +611,26 @@ coordinate transformation:
...
+
+    .. versionadded:: 5.2
+
+    A boolean indicating if this geometry is or contains a curve geometry.
+
+    .. method:: get_linear_geometry
+
+    .. versionadded:: 5.2
+
+    Returns a linear version of the geometry. If no conversion can be made the
```
nessita, line 624:
```suggestion
    Returns a linear version of the geometry. If no conversion can be made, the
```

docs/ref/contrib/gis/gdal.txt
```diff
@@ -611,6 +611,26 @@ coordinate transformation:
...
+    original geometry is returned.
+
+    .. method:: get_curve_geometry
+
+    .. versionadded:: 5.2
+
+    Returns a curved version of the geometry. If no conversion can be made the
```
nessita, line 631:
```suggestion
    Returns a curved version of the geometry. If no conversion can be made, the
```

nessita (commented):
Thank you @smithdc1, I invested some time learning and testing around this code. Great work! :star2: 

I added some optional comments. Let me know if you agree, I'm happy to merge soon.

nessita (approved):
Amazing work, thanks for the final polishing bits @smithdc1! :trophy:

PR #18013, opened 2024-03-24
ticket-[35321]

Jkhall81:
ticket-[35321]

# Branch description
Added CSS rule for label element making font-size 1rem to address automatic zoom issue in mobile Safari browsers.

### DIFF
--- a/django/contrib/admin/static/admin/css/base.css
+++ b/django/contrib/admin/static/admin/css/base.css
@@ -101,6 +101,10 @@ body {
     background: var(--body-bg);
 }
 
+label {
+    font-size: 1rem;
+}
+
 /* LINKS */
 
 a:link, a:visited {


felixxm:
@simonw Does it work for you?

sarahboyce:
Hi @Jkhall81 👋 thank you for this

This is not quite what I'm expecting to see here. I would expect these two lines to be updated with `font-size: 1rem`:
- https://github.com/django/django/blob/6f7c0a4d66f36c59ae9eafa168b455e462d81901/django/contrib/admin/static/admin/css/responsive.css#L174
- https://github.com/django/django/blob/6f7c0a4d66f36c59ae9eafa168b455e462d81901/django/contrib/admin/static/admin/css/responsive.css#L195

This css file is what has mobile overrides.
It's also the input field and textarea etc that have the zoom action that is undesirable, but increasing the label font size is to make the label not be smaller than the input (if that makes sense)

Jkhall81:
Hi Sarah!

Thanks for explaining that.  I am still learning this, but now I know what the responsive.css file does.  Makes sense based on the name I guess.  

I updated the two font-sizes, and took out my previous change in the base.css file.  

Do I need to document the change?  Or is it minor enough that it's not necessary to document?

sarahboyce:
Don't worry, I'm still learning and I've been doing this a while

> Do I need to document the change? Or is it minor enough that it's not necessary to document?

No docs changes needed 👍

Jkhall81:
Cool. Thanks for the help!  I really appreciate it!

PR #18014, opened 2024-03-25
Fixed #35328 - Improved debug messaging behind proxies.

ryanhiebert:
# Trac ticket number

ticket-35328

# Branch description
When in debug mode, we can  make some better error messaging when the `Origin` header checking fails. In common scenarios, we can notice that they are likely missing headers from proxies. In any case, we can remove the verbose messaging around CSRF Tokens, which are *not* the way they have failed and are unhelpful to solving the error they are viewing.

# Checklist

### DIFF
--- a/django/middleware/csrf.py
+++ b/django/middleware/csrf.py
@@ -26,7 +26,10 @@ logger = logging.getLogger("django.security.csrf")
 # This matches if any character is not in CSRF_ALLOWED_CHARS.
 invalid_token_chars_re = _lazy_re_compile("[^a-zA-Z0-9]")
 
-REASON_BAD_ORIGIN = "Origin checking failed - %s does not match any trusted origins."
+REASON_DISALLOWED_HOST = "Host is not allowed."
+REASON_BAD_ORIGIN = (
+    "Origin checking failed - %r does not match %r or any other trusted origins."
+)
 REASON_NO_REFERER = "Referer checking failed - no Referer."
 REASON_BAD_REFERER = "Referer checking failed - %s does not match any trusted origins."
 REASON_NO_CSRF_COOKIE = "CSRF cookie not set."
@@ -268,19 +271,10 @@ class CsrfViewMiddleware(MiddlewareMixin):
             # Set the Vary header since content varies with the CSRF cookie.
             patch_vary_headers(response, ("Cookie",))
 
-    def _origin_verified(self, request):
+    def _origin_verified(self, request, host_origin):
         request_origin = request.META["HTTP_ORIGIN"]
-        try:
-            good_host = request.get_host()
-        except DisallowedHost:
-            pass
-        else:
-            good_origin = "%s://%s" % (
-                "https" if request.is_secure() else "http",
-                good_host,
-            )
-            if request_origin == good_origin:
-                return True
+        if request_origin == host_origin:
+            return True
         if request_origin in self.allowed_origins_exact:
             return True
         try:
@@ -434,9 +428,17 @@ class CsrfViewMiddleware(MiddlewareMixin):
         # Reject the request if the Origin header doesn't match an allowed
         # value.
         if "HTTP_ORIGIN" in request.META:
-            if not self._origin_verified(request):
+            try:
+                host_origin = "%s://%s" % (
+                    "https" if request.is_secure() else "http",
+                    request.get_host(),
+                )
+            except DisallowedHost:
+                return self._reject(request, REASON_DISALLOWED_HOST)
+            if not self._origin_verified(request, host_origin):
                 return self._reject(
-                    request, REASON_BAD_ORIGIN % request.META["HTTP_ORIGIN"]
+                    request,
+                    REASON_BAD_ORIGIN % (request.META["HTTP_ORIGIN"], host_origin),
                 )
         elif request.is_secure():
             # If the Origin header wasn't provided, reject HTTPS requests if
--- a/django/views/csrf.py
+++ b/django/views/csrf.py
@@ -61,6 +61,16 @@ def csrf_failure(request, reason="", template_name=CSRF_FAILURE_TEMPLATE_NAME):
             "re-enable them, at least for this site, or for “same-origin” "
             "requests."
         ),
+        "bad_origin": reason.startswith("Origin checking failed"),
+        "forwarded_may_fix": (
+            request.headers.get("X-Forwarded-Proto", "") == "https"
+            and not request.is_secure()
+            or request.headers.get("Origin", "").endswith(
+                f'://{request.headers.get("X-Forwarded-Host", "")}'
+            )
+        ),
+        "x_forwarded_proto": request.headers.get("X-Forwarded-Proto"),
+        "x_forwarded_host": request.headers.get("X-Forwarded-Host"),
         "DEBUG": settings.DEBUG,
         "docs_version": get_docs_version(),
         "more": _("More information is available with DEBUG=True."),
--- a/django/views/templates/csrf_403.html
+++ b/django/views/templates/csrf_403.html
@@ -43,6 +43,34 @@
     </pre>
     {% endif %}
 
+  {% if bad_origin %}
+  {% if forwarded_may_fix %}
+  <p>The <code>Origin</code> header does not match
+  the expected server origin,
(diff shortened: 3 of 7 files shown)


django/middleware/csrf.py
```diff
@@ -26,7 +26,9 @@
 # This matches if any character is not in CSRF_ALLOWED_CHARS.
 invalid_token_chars_re = _lazy_re_compile("[^a-zA-Z0-9]")
 
-REASON_BAD_ORIGIN = "Origin checking failed - %s does not match any trusted origins."
+REASON_BAD_ORIGIN = (
+    "Origin checking failed - %s does not match any trusted origins to server %s."
+)
```
ryanhiebert, line 31: This ended up requiring a number of test changes. I think that this doesn't run afoul of our backward compatibility policy because I don't think that this is a documented property, but this seems like a likely place for a reviewer to have concerns.

django/middleware/csrf.py
```diff
@@ -436,7 +444,12 @@ def process_view(self, request, callback, callback_args, callback_kwargs):
         if "HTTP_ORIGIN" in request.META:
             if not self._origin_verified(request):
                 return self._reject(
-                    request, REASON_BAD_ORIGIN % request.META["HTTP_ORIGIN"]
+                    request,
+                    REASON_BAD_ORIGIN
+                    % (
+                        request.META["HTTP_ORIGIN"],
+                        self._host_origin(request) or "<disallowed host>",
```
ryanhiebert, line 451: It would probably be even better to handle this with a dedicated reason, but I didn't want to delay submitting since even with the messaging like this since even as it is it seems like an improvement over the status quo.

django/views/templates/csrf_403.html
```diff
@@ -53,7 +79,7 @@ <h2>Help</h2>
     <li>Your browser is accepting cookies.</li>
 
     <li>The view function passes a <code>request</code> to the template’s <a
-    href="https://docs.djangoproject.com/en/dev/topics/templates/#django.template.backends.base.Template.render"><code>render</code></a>
+    href="https://docs.djangoproject.com/en/{{ docs_version }}/topics/templates/#django.template.backends.base.Template.render"><code>render</code></a>
```
ryanhiebert, line 82: This is unrelated to my changes, except that I'm in here and noticed that it wasn't using the same version format for these doc urls.

sarahboyce (commented):
Thank you Ryan ⭐ I have initial comments

django/middleware/csrf.py
```diff
@@ -436,7 +444,12 @@ def process_view(self, request, callback, callback_args, callback_kwargs):
         if "HTTP_ORIGIN" in request.META:
             if not self._origin_verified(request):
```
adamchainz, line 445: Could we restructure the code to avoid calling `_host_origin` twice (within `_origin_verified()` and then on line 451). So here could look like:

```suggestion
            host_origin = self._host_origin(request)
            if not self._origin_verified(request, host_origin):
```

And the error message could use `host_origin or "<disallowed host>"`.

adamchainz (commented):
Thank you for your work here @ryanhiebert ! I have some comments with small improvements.

django/views/templates/csrf_403.html
ryanhiebert: Can do. Would the process be to create a separate ticket in Trac as well, or is that overkill for this?

django/middleware/csrf.py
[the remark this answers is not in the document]
ryanhiebert: Yeah, that's definitely a clearer message.

adamchainz:
Lint error:
```
Error: ./django/views/csrf.py:26:5: F401 'django.middleware.csrf.REASON_BAD_ORIGIN' imported but unused
```

django/middleware/csrf.py
adamchainz: You still need to change line 451 to:

```
host_origin or "<disallowed host>",
```

tests/csrf_tests/tests.py
```diff
@@ -907,7 +907,10 @@ def test_bad_origin_bad_domain(self):
         with self.assertLogs("django.security.csrf", "WARNING") as cm:
             response = mw.process_view(req, post_form_view, (), {})
         self.assertEqual(response.status_code, 403)
-        msg = REASON_BAD_ORIGIN % req.META["HTTP_ORIGIN"]
+        msg = REASON_BAD_ORIGIN % (
+            req.META["HTTP_ORIGIN"],
+            "http://" + req.META["HTTP_HOST"],
+        )
```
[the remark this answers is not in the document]
adamchainz: Still pending

adamchainz (changes requested):

tests/csrf_tests/tests.py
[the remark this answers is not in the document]
ryanhiebert: yes, thank you. WIP on my local machine

django/middleware/csrf.py
ryanhiebert: With @sarahboyce 's agreement that it should be a separate error code, it'll be fixed up a different way, but i have WIP code that does as you suggest. Working on it now.

django/views/templates/csrf_403.html
ryanhiebert: Awesome. Here's the new PR: https://github.com/django/django/pull/18180

django/views/csrf.py
```diff
@@ -61,6 +61,16 @@ def csrf_failure(request, reason="", template_name=CSRF_FAILURE_TEMPLATE_NAME):
...
             "requests."
         ),
+        "bad_origin": reason.startswith("Origin checking failed"),
+        "forwarded_may_fix": (
+            request.headers.get("X-Forwarded-Proto", "") == "https"
+            and not request.is_secure()
+            or request.headers.get("Origin", "").endswith(
+                f'://{request.headers.get("X-Forwarded-Host", "")}'
+            )
+        ),
```
sarahboyce, line 71: No tests fail if I was to update to:
```suggestion
        "forwarded_may_fix": True,
```
Can you write a test for the `False` case (probably asserts that `If the expected server origin looks correct, you may wish to add...` in the template).

django/middleware/csrf.py
```diff
@@ -268,19 +271,24 @@ def _set_csrf_cookie(self, request, response):
...
             good_host = request.get_host()
         except DisallowedHost:
-            pass
+            return None
         else:
-            good_origin = "%s://%s" % (
+            return "%s://%s" % (
                 "https" if request.is_secure() else "http",
                 good_host,
             )
```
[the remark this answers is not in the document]
ryanhiebert: Unfortunately, this suggestion broke the tests, likely related to `_is_secure_override`. I kept the refactor, but reverted this particular suggestion.

docs/releases/5.1.txt
```diff
@@ -190,7 +190,8 @@ Cache
 CSRF
 ~~~~
 
-* ...
+* The error messaging is improved when ``Origin`` checking fails,
+  including better hints when likely proxy headers are detected.
```
sarahboyce, line 194:
```suggestion
* The ``'403_csrf.html'`` template now has improved messaging when ``Origin`` 
  checking fails, including better hints when likely proxy headers are detected.
```
[Docs are wrapped at 80 characters](https://docs.djangoproject.com/en/5.0/internals/contributing/writing-documentation/#guidelines-for-restructuredtext-files) and we will now need to rebase and move this to the 5.2 release notes.

tim-schilling:
Hey @ryanhiebert, is this something you are intending to come back to? Otherwise I was going to add it to the pile of DjangoCon US Sprint tickets. I think this may have gotten lost because Patch Needs Improvement was never unset after your last push. Now it'll need a rebase and handling any conflicts.

ryanhiebert:
Hi @tim-schilling ! I have lots of good intentions, but as is apparent they haven't materialized into action at this point. The DjangoCon sprints seem like a potentially perfect place to have the right people in the room to iron out details and get this done, and I would hate to see that opportunity wasted because I'm not there or active enough.

I may yet do a little more work on this before then, and I'll do my best to be available on Discord during the sprints. Please help me be in the right channel(s) if I can be helpful.

PR #18020, opened 2024-03-26
Fixed #35326 -- added `allow_overwrite` parameter to FileSystemStorage.

bcail:
# Trac ticket number
ticket-35326

# Branch description
This patch handled temporary uploaded files when the filesystem storage allows overwriting files.

# Checklist

### DIFF
--- a/django/core/files/storage/filesystem.py
+++ b/django/core/files/storage/filesystem.py
@@ -1,13 +1,16 @@
 import os
+import warnings
 from datetime import datetime, timezone
 from urllib.parse import urljoin
 
 from django.conf import settings
+from django.core.exceptions import SuspiciousFileOperation
 from django.core.files import File, locks
 from django.core.files.move import file_move_safe
 from django.core.signals import setting_changed
 from django.utils._os import safe_join
 from django.utils.deconstruct import deconstructible
+from django.utils.deprecation import RemovedInDjango60Warning
 from django.utils.encoding import filepath_to_uri
 from django.utils.functional import cached_property
 
@@ -21,8 +24,7 @@ class FileSystemStorage(Storage, StorageSettingsMixin):
     Standard filesystem storage
     """
 
-    # The combination of O_CREAT and O_EXCL makes os.open() raise OSError if
-    # the file already exists before it's opened.
+    # RemovedInDjango60Warning: remove OS_OPEN_FLAGS.
     OS_OPEN_FLAGS = os.O_WRONLY | os.O_CREAT | os.O_EXCL | getattr(os, "O_BINARY", 0)
 
     def __init__(
@@ -31,12 +33,23 @@ class FileSystemStorage(Storage, StorageSettingsMixin):
         base_url=None,
         file_permissions_mode=None,
         directory_permissions_mode=None,
+        allow_overwrite=False,
     ):
         self._location = location
         self._base_url = base_url
         self._file_permissions_mode = file_permissions_mode
         self._directory_permissions_mode = directory_permissions_mode
+        self._allow_overwrite = allow_overwrite
         setting_changed.connect(self._clear_cached_properties)
+        # RemovedInDjango60Warning: remove this warning.
+        if self.OS_OPEN_FLAGS != os.O_WRONLY | os.O_CREAT | os.O_EXCL | getattr(
+            os, "O_BINARY", 0
+        ):
+            warnings.warn(
+                "Overriding OS_OPEN_FLAGS is deprecated. Use "
+                "the allow_overwrite parameter instead.",
+                RemovedInDjango60Warning,
+            )
 
     @cached_property
     def base_location(self):
@@ -98,12 +111,30 @@ class FileSystemStorage(Storage, StorageSettingsMixin):
             try:
                 # This file has a file path that we can move.
                 if hasattr(content, "temporary_file_path"):
-                    file_move_safe(content.temporary_file_path(), full_path)
+                    file_move_safe(
+                        content.temporary_file_path(),
+                        full_path,
+                        allow_overwrite=self._allow_overwrite,
+                    )
 
                 # This is a normal uploadedfile that we can stream.
                 else:
-                    # The current umask value is masked out by os.open!
-                    fd = os.open(full_path, self.OS_OPEN_FLAGS, 0o666)
+                    # The combination of O_CREAT and O_EXCL makes os.open() raises an
+                    # OSError if the file already exists before it's opened.
+                    open_flags = (
+                        os.O_WRONLY
+                        | os.O_CREAT
+                        | os.O_EXCL
+                        | getattr(os, "O_BINARY", 0)
+                    )
+                    # RemovedInDjango60Warning: when the deprecation ends, replace with:
+                    # if self._allow_overwrite:
+                    #     open_flags = open_flags & ~os.O_EXCL
+                    if self.OS_OPEN_FLAGS != open_flags:
+                        open_flags = self.OS_OPEN_FLAGS
+                    elif self._allow_overwrite:
+                        open_flags = open_flags & ~os.O_EXCL
+                    fd = os.open(full_path, open_flags, 0o666)
                     _file = None
                     try:
                         locks.lock(fd, locks.LOCK_EX)
@@ -162,7 +193,13 @@ class FileSystemStorage(Storage, StorageSettingsMixin):
             pass
 
     def exists(self, name):
-        return os.path.lexists(self.path(name))
+        try:
+            exists = os.path.lexists(self.path(name))
+        except SuspiciousFileOperation:
+            raise
+        if self._allow_overwrite:
+            return False
+        return exists
 
     def listdir(self, path):
         path = self.path(path)
--- a/docs/ref/files/storage.txt
+++ b/docs/ref/files/storage.txt
@@ -28,7 +28,7 @@ Django provides convenient ways to access the default storage class:
 The ``FileSystemStorage`` class
 ===============================
 
-.. class:: FileSystemStorage(location=None, base_url=None, file_permissions_mode=None, directory_permissions_mode=None)
+.. class:: FileSystemStorage(location=None, base_url=None, file_permissions_mode=None, directory_permissions_mode=None, allow_overwrite=False)
 
     The :class:`~django.core.files.storage.FileSystemStorage` class implements
     basic file storage on a local filesystem. It inherits from
@@ -59,6 +59,13 @@ The ``FileSystemStorage`` class
 
         The file system permissions that the directory will receive when it is
         saved. Defaults to :setting:`FILE_UPLOAD_DIRECTORY_PERMISSIONS`.
+
+    .. attribute:: allow_overwrite
+
+        .. versionadded:: 5.1
+
+        Flag to control allowing saving a new file over an existing one.
+        Defaults to ``False``.
 
     .. method:: get_created_time(name)
 
--- a/docs/releases/5.1.txt
+++ b/docs/releases/5.1.txt
@@ -207,7 +207,10 @@ Error Reporting
 File Storage
 ~~~~~~~~~~~~
 
-* ...
+* The :attr:`~django.core.files.storage.FileSystemStorage.allow_overwrite`
+  parameter has been added to
+  :class:`~django.core.files.storage.FileSystemStorage`, to allow saving new
+  files over existing ones.
 
 File Uploads
 ~~~~~~~~~~~~
@@ -463,6 +466,12 @@ Miscellaneous
 
 * The ``check`` keyword argument of ``CheckConstraint`` is deprecated in favor
   of ``condition``.
+
+* The undocumented ``OS_OPEN_FLAGS`` property of
+  :class:`~django.core.files.storage.FileSystemStorage` has been deprecated.
+  To allow overwriting files in storage, set the new
+  :attr:`~django.core.files.storage.FileSystemStorage.allow_overwrite` option
+  to ``True`` instead.
 
 Features removed in 5.1
 =======================
--- a/docs/internals/deprecation.txt
+++ b/docs/internals/deprecation.txt
@@ -79,6 +79,9 @@ details on these changes.
 
 * The ``check`` keyword argument of ``CheckConstraint`` will be removed.
 
+* The ``OS_OPEN_FLAGS`` attribute of
+  :class:`~django.core.files.storage.FileSystemStorage` will be removed.
+
 .. _deprecation-removed-in-5.1:
 
 5.1
--- a/tests/file_storage/tests.py
+++ b/tests/file_storage/tests.py
@@ -25,11 +25,18 @@ from django.core.files.uploadedfile import (
 )
 from django.db.models import FileField
 from django.db.models.fields.files import FileDescriptor
-from django.test import LiveServerTestCase, SimpleTestCase, TestCase, override_settings
+from django.test import (
+    LiveServerTestCase,
+    SimpleTestCase,
+    TestCase,
+    ignore_warnings,
+    override_settings,
+)
 from django.test.utils import requires_tz_support
 from django.urls import NoReverseMatch, reverse_lazy
 from django.utils import timezone
 from django.utils._os import symlinks_supported
+from django.utils.deprecation import RemovedInDjango60Warning
 
 from .models import (
     Storage,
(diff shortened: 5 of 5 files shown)


django/core/files/storage/filesystem.py
```diff
@@ -98,12 +96,29 @@ def _save(self, name, content):
...
+                        os.O_WRONLY
+                        | os.O_CREAT
+                        | os.O_EXCL
+                        | getattr(os, "O_BINARY", 0)
+                    )
+
+                    if self.ALLOW_OVERWRITE:
+                        open_flags = OS_OPEN_FLAGS & ~os.O_EXCL
+                    else:
+                        open_flags = OS_OPEN_FLAGS
```
adamchainz, line 119: I haven’t understood the problem hard enough to have an opinion about this change, but this pattern would be simpler:

```suggestion
                    open_flags = (
                        os.O_WRONLY
                        | os.O_CREAT
                        | getattr(os, "O_BINARY", 0)
                    )

                    if not self.ALLOW_OVERWRITE:
                        open_flags |= os.O_EXCL
```

bcail: Thanks, I updated the PR.

jschneier:
What about using the `OPTIONS` dict to implement this feature more generally?

django/core/files/storage/filesystem.py
```diff
@@ -21,21 +24,19 @@ class FileSystemStorage(Storage, StorageSettingsMixin):
...
-    # the file already exists before it's opened.
-    OS_OPEN_FLAGS = os.O_WRONLY | os.O_CREAT | os.O_EXCL | getattr(os, "O_BINARY", 0)
-
     def __init__(
         self,
         location=None,
         base_url=None,
         file_permissions_mode=None,
         directory_permissions_mode=None,
+        allow_overwrite=False,
```
sarahboyce, line 33: We need an update to the docs for [`FileSystemStorage`](https://docs.djangoproject.com/en/5.0/ref/files/storage/#the-filesystemstorage-class) adding `allow_overwite` and a `.. versionchanged::` note.
We also need a release note for `allow_overwite` and a deprecation/removed note for `OS_OPEN_FLAGS`.

django/core/files/storage/filesystem.py
```diff
@@ -64,6 +65,19 @@ def directory_permissions_mode(self):
...
+            name = str(name).replace("\\", "/")
+            dir_name, file_name = os.path.split(name)
+            if ".." in pathlib.PurePath(dir_name).parts:
+                raise SuspiciousFileOperation(
+                    "Detected path traversal attempt in '%s'" % dir_name
+                )
+            validate_file_name(file_name)
+            return name
+        else:
+            return super().get_available_name(name, max_length)
```
sarahboyce, line 79: Can you explain why this change is required? I also believe it is untested.

sarahboyce (commented):
Thank you for this @bcail! I have initial comments 👍

django/core/files/storage/filesystem.py
bcail: This change is needed so that the name is allowed even if there's already a file by that name. If I remove this change, the two `file_storage.tests.OverwritingStorageTests` tests fail.

bcail:
@sarahboyce Thanks for your review. I've added `OS_OPEN_FLAGS` back in, for deprecation, and I added some documentation. Do we need to add a deprecation warning in the code if `OS_OPEN_FLAGS` is used? If so, I'm not sure how to do that since we're always setting `OS_OPEN_FLAGS` here in the django code.

django/core/files/storage/filesystem.py
sarahboyce: I think we need to add `file_storage.test_generate_filename.GenerateFilenameStorageTests.test_storage_dangerous_paths_dir_name` or create a very similar test for when to `FileSystemStorage` has `allow_overwrite=True`

sarahboyce (commented):
> I added a commit with a warning if the default OS_OPEN_FLAGS is overridden.

Thank you for doing this @bcail, I think this approach looks good to me.

I have a few minor comments, this is getting really close 👍

django/core/files/storage/filesystem.py
```diff
@@ -24,19 +29,30 @@ class FileSystemStorage(Storage, StorageSettingsMixin):
     # The combination of O_CREAT and O_EXCL makes os.open() raise OSError if
     # the file already exists before it's opened.
     OS_OPEN_FLAGS = os.O_WRONLY | os.O_CREAT | os.O_EXCL | getattr(os, "O_BINARY", 0)
+    _default_open_flags = (
+        os.O_WRONLY | os.O_CREAT | os.O_EXCL | getattr(os, "O_BINARY", 0)
+    )
```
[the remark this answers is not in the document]
bcail: OK, if you let me know how you want it, I'll change it.

django/core/files/storage/filesystem.py
```diff
@@ -24,19 +29,30 @@ class FileSystemStorage(Storage, StorageSettingsMixin):
     # The combination of O_CREAT and O_EXCL makes os.open() raise OSError if
     # the file already exists before it's opened.
     OS_OPEN_FLAGS = os.O_WRONLY | os.O_CREAT | os.O_EXCL | getattr(os, "O_BINARY", 0)
```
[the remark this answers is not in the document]
bcail: I added instructions on what to remove.

docs/releases/5.1.txt
```diff
@@ -454,6 +455,10 @@ Miscellaneous
 * The ``check`` keyword argument of ``CheckConstraint`` is deprecated in favor
   of ``condition``.
 
+* The ``OS_OPEN_FLAGS`` property of ``FileSystemStorage`` has been deprecated.
+  To allow overwriting files in storage, set the new ``allow_overwrite`` option
```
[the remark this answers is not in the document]
bcail: I updated the links, except that I got a warning making a link to `OS_OPEN_FLAGS` - I think because it's not documented.

docs/releases/5.1.txt
```diff
@@ -204,7 +204,8 @@ Error Reporting
 File Storage
 ~~~~~~~~~~~~
 
-* ...
+* The ``allow_overwrite`` parameter has been added to ``FileSystemStorage``,
```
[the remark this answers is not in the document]
bcail: I updated these links.

docs/ref/files/storage.txt
```diff
@@ -55,6 +55,13 @@ The ``FileSystemStorage`` class
         The file system permissions that the directory will receive when it is
         saved. Defaults to :setting:`FILE_UPLOAD_DIRECTORY_PERMISSIONS`.
 
+    .. attribute:: allow_overwrite
+
+        Flag to control allowing saving a new file over an existing one.
+        Defaults to ``False``.
+
+        .. versionadded:: 5.1
```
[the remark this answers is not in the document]
bcail: I moved it over.

django/core/files/storage/filesystem.py
bcail: I updated the tests.

docs/releases/5.1.txt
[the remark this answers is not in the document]
sarahboyce: Ah yes sorry, leave `OS_OPEN_FLAGS` 👍

jschneier:
Kind of derailed the original ticket so would rename the title

django/core/files/storage/filesystem.py
sarahboyce: I think instead of overwriting `get_available_name` here, we can overwrite `exists` to something like:
```python
    def exists(self, name):
        exists = os.path.lexists(self.path(name))
        if self._allow_overwrite:
            return False
        return exists
```
It requires some test updates (these need a think) but something like:
```diff
@@ -99,14 +99,14 @@ class FileStorageTests(SimpleTestCase):
         f = self.storage.open("storage_test", "w")
         f.write("storage contents")
         f.close()
-        self.assertTrue(self.storage.exists("storage_test"))
+        self.assertTrue(os.path.exists(os.path.join(self.temp_dir, "storage_test")))

         f = self.storage.open("storage_test", "r")
         self.assertEqual(f.read(), "storage contents")
         f.close()

         self.storage.delete("storage_test")
-        self.assertFalse(self.storage.exists("storage_test"))
+        self.assertFalse(os.path.exists(os.path.join(self.temp_dir, "storage_test")))

     def _test_file_time_getter(self, getter):
         # Check for correct behavior under both USE_TZ=True and USE_TZ=False.
@@ -275,10 +275,10 @@ class FileStorageTests(SimpleTestCase):
         """
         Saving a pathname should create intermediate directories as necessary.
         """
-        self.assertFalse(self.storage.exists("path/to"))
+        self.assertFalse(os.path.exists(os.path.join(self.temp_dir, "path/to")))
         self.storage.save("path/to/test.file", ContentFile("file saved with path"))

-        self.assertTrue(self.storage.exists("path/to"))
+        self.assertTrue(os.path.exists(os.path.join(self.temp_dir, "path/to")))
         with self.storage.open("path/to/test.file") as f:
             self.assertEqual(f.read(), b"file saved with path")

@@ -694,13 +694,11 @@ class OverwritingStorageTests(FileStorageTests):
         stored_name_1 = self.storage.save(name, f_1)
         try:
             self.assertEqual(stored_name_1, name)
-            self.assertTrue(self.storage.exists(name))
             self.assertTrue(os.path.exists(os.path.join(self.temp_dir, name)))
             with self.storage.open(name) as fp:
                 self.assertEqual(fp.read(), content_1)
             stored_name_2 = self.storage.save(name, f_2)
             self.assertEqual(stored_name_2, name)
-            self.assertTrue(self.storage.exists(name))
             self.assertTrue(os.path.exists(os.path.join(self.temp_dir, name)))
             with self.storage.open(name) as fp:
                 self.assertEqual(fp.read(), content_2)
@@ -722,13 +720,11 @@ class OverwritingStorageTests(FileStorageTests):
         stored_name_1 = self.storage.save(name, f_1)
         try:
             self.assertEqual(stored_name_1, name)
-            self.assertTrue(self.storage.exists(name))
             self.assertTrue(os.path.exists(os.path.join(self.temp_dir, name)))
             with self.storage.open(name) as fp:
                 self.assertEqual(fp.read(), content_1)
             stored_name_2 = self.storage.save(name, f_2)
             self.assertEqual(stored_name_2, name)
-            self.assertTrue(self.storage.exists(name))
             self.assertTrue(os.path.exists(os.path.join(self.temp_dir, name)))
             with self.storage.open(name) as fp:
```

django/core/files/storage/filesystem.py
```diff
@@ -98,12 +130,22 @@ def _save(self, name, content):
...
-                    # The current umask value is masked out by os.open!
-                    fd = os.open(full_path, self.OS_OPEN_FLAGS, 0o666)
+                    if (
+                        self._default_open_flags == self.OS_OPEN_FLAGS
+                        and self._allow_overwrite
+                    ):
+                        open_flags = self._default_open_flags & ~os.O_EXCL
+                    else:
+                        open_flags = self.OS_OPEN_FLAGS
+                    fd = os.open(full_path, open_flags, 0o666)
```
sarahboyce, line 148: This is missing a comment as to how this will simplify after the OS_OPEN_FLAGS is removed.
If you search for `# RemovedInDjango60Warning: when the deprecation ends, replace with:` in the codebase I think you'll get what I mean 👍

django/core/files/storage/filesystem.py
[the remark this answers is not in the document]
sarahboyce: I'd rather have it defined in the places it's being used than be a private attribute

docs/releases/5.1.txt
```diff
@@ -456,6 +459,12 @@ Miscellaneous
 * The ``check`` keyword argument of ``CheckConstraint`` is deprecated in favor
   of ``condition``.
 
+* The ``OS_OPEN_FLAGS`` property of
```
sarahboyce, line 462:
```suggestion
* The undocumented ``OS_OPEN_FLAGS`` property of
```

docs/ref/files/storage.txt
```diff
@@ -55,6 +55,13 @@ The ``FileSystemStorage`` class
         The file system permissions that the directory will receive when it is
         saved. Defaults to :setting:`FILE_UPLOAD_DIRECTORY_PERMISSIONS`.
 
+    .. attribute:: allow_overwrite
+
+        Flag to control allowing saving a new file over an existing one.
+        Defaults to ``False``.
+
+    .. versionadded:: 5.1
```
sarahboyce, line 63:
```suggestion
        .. versionadded:: 5.1
        
        Flag to control allowing saving a new file over an existing one.
        Defaults to ``False``.
```

django/core/files/storage/filesystem.py
bcail: I added comments about the deprecation end.

django/core/files/storage/filesystem.py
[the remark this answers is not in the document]
bcail: I removed the private attribute.

sarahboyce:
Thank you for the updates @bcail, have you seen [this suggestion](https://github.com/django/django/pull/18020#discussion_r1590011076)?
As it would change how `exists` works (a method we should overwrite/define) rather than `get_available_name` (that in theory we shouldn't overwrite/define), I think this might be a better way to go 🤔

bcail:
@sarahboyce I wonder if there would be any downstream code that would want to call `exists()`, to see if a file is in storage or not, and would get confused by us changing `exists()` to always return `False` for overwriting storage?

But I'll do it whichever way you think is best.

sarahboyce:
Thank you @bcail 🥳 this looks great, I pushed up small edits and will seek a second opinion before merging 👍

bcail:
Thanks, @sarahboyce.

sarahboyce (approved):

dopry:
@sarahboyce the changes to .exists has impacted a large number of developers and production systems that rely on the django ecosystem. .exists should only return whether a file exists. the allow_overwrite flag should not impact the response of the exists method. It is both a breaking change and leads to unexpected behavior for users who want to know if a file exists in the storage.  Please be mindful when reviewing this kind of low level code... No change at this low of a level is a small change.

sarahboyce:
@dopry you may have noticed that we fixed this in Django 6 months ago after the feedback from the django-storages issue (also commented on the issue).
That the issue is still active is due to a pending fix and release to django-storages

Please be kind. I don't see how this comment is supposed to be helpful

dopry:
No need to be defensive, I'm just trying to point out some things you could have kept in mind that might have prevented this error from entering the code base to begin with. Maybe I was too terse in my haste. I wouldn't take the time if I didn't care. 

In the course of your review, as a maintainer you recommended the exists change while the developer working on the code was questioning whether it would have unintended impact. From you explanation you made that decision based on your reading of the documentation. This is a fine in a pre-release scenario as no one is depending on the code yet. Once the code is in the wild and developers are using it, they're using it as it is... Many of us will look at the implementation first and not the documentation. After release it is better to update the documentation to match the code rather than the other way around.  There are a few principles that I tend to keep in mind when I face these kinds of decisions as a maintainer. 

1. The open-closed principal, basically we shouldn't change what is in the wild, code is open to extension (wrappers, proxies, overrides) but closed to modification.
2. The stable dependency principle - low level things should change more infrequently than high level things... Meaning one should always second guess themselves anytime they're making a change to a low level library. 

Obviously these aren't hard and fast rules and we have tools to facilitate stability when there is good cause to violate them like deprecation cycles and major releases/breaking change notifications.  As an OSS maintainer, it is important to understand that large audiences depend on you and that you are working in public view. It is important to be mindful of your decisions, to  apply best practices and community guidelines to your decision making, and to look for opportunities to grow as a maintainer.

bcail:
@dopry I don't think "trying to point out some things you could have kept in mind" is helpful in these comments. I would suggest that a better way to show your care is to work on open tickets, and look for tasks you can complete that will support the fellows' work.

nessita:
@dopry Your last comment comes across as dismissive and patronizing. You are implying that the Django Fellows did not consider certain important aspects, without providing any evidence to support such a claim.

Additionally, it seems you are trying to "educate" the Django Fellows about what it means to be a good maintainer (one could even say "mansplaining" :roll_eyes:). While it's important to share knowledge, this approach feels condescending and dismissive, especially considering the expertise and effort that the Django Fellows contribute to this project. This kind of behavior and messaging are neither helpful nor welcome here.

It was already pointed out that this issue was addressed months ago, and the fix is pending in `django-storages`. So, I find it hard to see how your comment is constructive at this point. Please remember that the Django community values respect and collaboration. If this pattern of comments continues, we will have no choice but to report the behavior according to the [Django Code of Conduct](https://www.djangoproject.com/conduct/).

PR #18025, opened 2024-03-27
Fixed Ticket #35330 (bug): The update of related objects fails in the admin when the related model is camel case.

devin13cox:
ticket-35330

# Branch description

This fix introduces `model_name` to use as the check for `data-model-ref` instead of `model` so that in the event that a model utilizes CamelCasing, we are able to correctly update related fields within the admin.


devin13cox:
I realize that I forgot to tag in the Co-Authored commit message for @nessita . I can re-introduce the PR if that is the easiest resolution.

nessita:
> I realize that I forgot to tag in the Co-Authored commit message for @nessita . I can re-introduce the PR if that is the easiest resolution.

No worries, I can add it later before merging.

One note though: when creating the PRs, please make sure to remove all square brackets from the placeholders, so links to tickets are properly converted. See, the following (raw):

```
ticket-[35330] vs ticket-35330
```

Produces: ticket-[35330] vs ticket-35330

nessita (approved):
Thank you @devin13cox, this looks great!

I added a unit test, will merge soon.

PR #18026, opened 2024-03-27
Fixed #35331 -- Updated dropdown lists with new entries added via the "+" sign from M2M field.

devin13cox:
# Trac ticket number
ticket-35331

# Branch description
When adding a new related entry on a filter horizontal vie the "+" sign from an M2M field, it used to not updated related fields. Now we call `UpdateRelatedSelectOptions` for all related fields except the `from` box. This also introduces a fix to ensure that the cache always accounts for items inside the selectBox so that they may be moved without requiring a refresh.

### DIFF
--- a/django/contrib/admin/static/admin/js/admin/RelatedObjectLookups.js
+++ b/django/contrib/admin/static/admin/js/admin/RelatedObjectLookups.js
@@ -87,7 +87,7 @@
         }
     }
 
-    function updateRelatedSelectsOptions(currentSelect, win, objId, newRepr, newId) {
+    function updateRelatedSelectsOptions(currentSelect, win, objId, newRepr, newId, skipIds = []) {
         // After create/edit a model from the options next to the current
         // select (+ or :pencil:) update ForeignKey PK of the rest of selects
         // in the page.
@@ -100,7 +100,7 @@
         const selectsRelated = document.querySelectorAll(`[data-model-ref="${modelName}"] [data-context="available-source"]`);
 
         selectsRelated.forEach(function(select) {
-            if (currentSelect === select) {
+            if (currentSelect === select || skipIds && skipIds.includes(select.id)) {
                 return;
             }
 
@@ -109,6 +109,11 @@
             if (!option) {
                 option = new Option(newRepr, newId);
                 select.options.add(option);
+                // Update SelectBox cache for related fields.
+                if (window.SelectBox !== undefined && !SelectBox.cache[currentSelect.id]) {
+                    SelectBox.add_to_cache(select.id, option);
+                    SelectBox.redisplay(select.id);
+                }
                 return;
             }
 
@@ -136,9 +141,14 @@
             $(elem).trigger('change');
         } else {
             const toId = name + "_to";
+            const toElem = document.getElementById(toId);
             const o = new Option(newRepr, newId);
             SelectBox.add_to_cache(toId, o);
             SelectBox.redisplay(toId);
+            if (toElem && toElem.nodeName.toUpperCase() === 'SELECT') {
+                const skipIds = [name + "_from"];
+                updateRelatedSelectsOptions(toElem, win, null, newRepr, newId, skipIds);
+            }
         }
         const index = relatedWindows.indexOf(win);
         if (index > -1) {
--- a/tests/admin_views/test_related_object_lookups.py
+++ b/tests/admin_views/test_related_object_lookups.py
@@ -3,6 +3,8 @@ from django.contrib.auth.models import User
 from django.test import override_settings
 from django.urls import reverse
 
+from .models import CamelCaseModel
+
 
 @override_settings(ROOT_URLCONF="admin_views.urls")
 class SeleniumTests(AdminSeleniumTestCase):
@@ -100,6 +102,8 @@ class SeleniumTests(AdminSeleniumTestCase):
         self.wait_until(lambda d: len(d.window_handles) == 1, 1)
         self.selenium.switch_to.window(self.selenium.window_handles[0])
 
+        id_value = CamelCaseModel.objects.get(interesting_name=interesting_name).id
+
         # Check that both the "Available" m2m box and the "Fk" dropdown now
         # include the newly added CamelCaseModel instance.
         fk_dropdown = self.selenium.find_element(By.ID, "id_fk")
@@ -107,7 +111,7 @@ class SeleniumTests(AdminSeleniumTestCase):
             fk_dropdown.get_attribute("innerHTML"),
             f"""
             <option value="" selected="">---------</option>
-            <option value="1" selected>{interesting_name}</option>
+            <option value="{id_value}" selected>{interesting_name}</option>
             """,
         )
         # Check the newly added instance is not also added in the "to" box.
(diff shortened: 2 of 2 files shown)


nessita:
Hey @devin13cox, thank you for your continued and steady contributions!

I haven't tried this yet but I think this is not the proper solution, in my [original debugging I mentioned](https://github.com/django/django/pull/17897#issuecomment-2000288000):

> Basically the id used in the `document.getElementById` is wrong for the M2M widget (but *correct* for the single FK field, so we may need to fix the call site to pass the proper name)

So the proposed adding `_from` will break the current use cases that *are* working, so this fix would need to change the call sites of this code and (perhaps) pass the element identifier. Does this make sense?

devin13cox:
> Hey @devin13cox, thank you for your continued and steady contributions!
> 
> I haven't tried this yet but I think this is not the proper solution, in my [original debugging I mentioned](https://github.com/django/django/pull/17897#issuecomment-2000288000):
> 
> > Basically the id used in the `document.getElementById` is wrong for the M2M widget (but _correct_ for the single FK field, so we may need to fix the call site to pass the proper name)
> 
> So the proposed adding `_from` will break the current use cases that _are_ working, so this fix would need to change the call sites of this code and (perhaps) pass the element identifier. Does this make sense?

Yeah I realized that after the initial push, hence the re-check after initial failure if applicable added in the second commit. Might be cleaner to pass the proper name from wherever it is initialized. I'll see if I can hunt that spot down.

nessita:
> > Hey @devin13cox, thank you for your continued and steady contributions!
> > I haven't tried this yet but I think this is not the proper solution, in my [original debugging I mentioned](https://github.com/django/django/pull/17897#issuecomment-2000288000):
> > > Basically the id used in the `document.getElementById` is wrong for the M2M widget (but _correct_ for the single FK field, so we may need to fix the call site to pass the proper name)
> > 
> > 
> > So the proposed adding `_from` will break the current use cases that _are_ working, so this fix would need to change the call sites of this code and (perhaps) pass the element identifier. Does this make sense?
> 
> Yeah I realized that after the initial push, hence the re-check after initial failure if applicable added in the second commit. 

Oh, I see now, thanks for mentioning it.

> Might be cleaner to pass the proper name from wherever it is initialized. I'll see if I can hunt that spot down.

Yeah, I would much rather we do this, if it's possible. Let me know! I'll set the ticket as "patch needs improvement", when this is ready for re-review, please update the ticket flags accordingly. Thank you!

django/contrib/admin/static/admin/js/admin/RelatedObjectLookups.js
```diff
@@ -118,7 +118,9 @@
     }
 
     function dismissAddRelatedObjectPopup(win, newId, newRepr) {
-        const name = removePopupIndex(win.name);
+        const urlParams = new URLSearchParams(win.location.search);
+        const tackOn = urlParams.has('_name_alteration') ? urlParams.get('_name_alteration') : '';
```
sarahboyce, line 122: This can be simplified to:
```suggestion
        const tackOn = urlParams.get('_name_alteration') || '';
```

tests/admin_views/test_related_object_lookups.py
```diff
@@ -117,3 +117,46 @@ def test_related_object_update_with_camel_casing(self):
...
+
+        # Check that the state is in the Available Source box
+        self.assertHTMLEqual(
+            _get_HTML_inside_element_by_id("id_m2m_from"),
+            f"""
+            <option selected value="1">{name}</option>
+            """,
+        )
+
+        # Now check that the state shows in the fk box
```
sarahboyce, line 155: I think these comments are not needed 👍 
```suggestion
        self.selenium.find_element(By.ID, "add_id_m2m").click()
        self.wait_for_and_switch_to_popup()

        self.selenium.find_element(By.ID, "id_interesting_name").send_keys(name)
        self.selenium.find_element(By.NAME, "_save").click()

        self.wait_until(lambda d: len(d.window_handles) == 1, 1)
        self.selenium.switch_to.window(self.selenium.window_handles[0])

        self.assertHTMLEqual(
            _get_HTML_inside_element_by_id("id_m2m_from"),
            f"""
            <option selected value="1">{name}</option>
            """,
        )
```

sarahboyce (commented):
There's a bug with the current solution

image

When I add via the M2M widget, the new 'State' object appears in both places (both available and chosen source). If I then select it to move it to chosen source, they both vanish from the widget. It does however appear in the dropdown. 
Can you take a look?

There is also a selenium test failure which needs investigating

sarahboyce:
Sorry @devin13cox I didn't realise the bug was being handled in a different PR and already had a ticket (sorry about that)
Only minor formating changes required here 👍

devin13cox:
Not really sure what is causing the Selenium test to fail. Locally this is not the case, (including an up-to-date merge) and the test as-is is what I would expect the actual functionality to be. The value getting set to two almost sounds like an extra item is being created--which would be an issue--but I am having a hard time reproducing it. 

Any chance either of you @nessita @sarahboyce have an idea of why it's failing here but not locally?

devin13cox:
Is it possible to re-run the checks? I'm fairly certain this change does not impact cache tests.

sarahboyce:
Hi @devin13cox I have merged in https://github.com/django/django/pull/17897 and rebased this PR on main
Unfortunately we have a bug here, when I add a new entry using the M2M "+" action, the new instance is added to the "Available" box rather than the "Chosen" box. If I then select it and click the arrow to move it to chosen, it disappears from "Available" but does not appear in "Chosen".
We should add asserts to the test that the item exists in the chosen box and not the available box and push up a fix 👍

devin13cox:
I found a new bug in master while testing. When adding a foreign key, it correctly shows up in the "from" box, but when clicked to move over to the "chosen" box, it disappears. This is due to a related change, but not this PR.

devin13cox:
@sarahboyce Sorry for the delay in getting back to this! I ended up changing the approach a bit to a new direction that makes more sense to me (it's also simpler). Basically we ensure the we call `updateRelatedSelectsOptions` while also explicitly denying an update to the `from` box (this all is happening inside of an `else` statement where we are handling the `to` box anyways). I made the parameter accept an array to make it more generic in the event of future use cases--although technically we could get away with a singular string value for this PR. 

Additionally, the issue involving disappearing items when moving between the SelectBox was actually also an issue in main; it is relevant to this ticket--but also enlarges the scope--so not sure if you'd prefer a separate PR here or maybe we can just expand the scope of the ticket's title.

The solution there was to ensure that we are updating the cache during instances where we would have skipped a cache update (this is particularly important when adding a FK when a M2M select box exists). 

Let me know what you think, thanks!

sarahboyce (commented):
Thank you for the updates @devin13cox - been testing and looks really good!

django/contrib/admin/static/admin/js/admin/RelatedObjectLookups.js
```diff
@@ -109,6 +110,11 @@
             if (!option) {
                 option = new Option(newRepr, newId);
                 select.options.add(option);
+                // This is needed to properly update SelectBox cache from related fields.
+                if ((window.SelectBox !== undefined) && (!SelectBox.cache[currentSelect.id] || !SelectBox.cache_contains(currentSelect.id, option.value))) {
```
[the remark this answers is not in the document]
devin13cox: I just tested and you are correct in that we do not need `!SelectBox.cache_contains(currentSelect.id, option.value)`.

sarahboyce (approved):
Thank you @devin13cox - this is great ⭐ 
I pushed minor tweaks

PR #18027, opened 2024-03-28
Fixed #35336 -- Addressed crash when adding a GeneratedField with % literals.

charettes:
ticket-35336

--- 

A longer term solution is likely to have a better separation of parametrized DDL altogether to handle checks, constraints, defaults, and generated fields but such a change would require a significant refactor that isn't suitable for a backport.

Thanks Adrian Garcia for the report.

### DIFF
--- a/django/db/backends/base/schema.py
+++ b/django/db/backends/base/schema.py
@@ -317,9 +317,9 @@ class BaseDatabaseSchemaEditor:
             if default_value is not None:
                 column_default = "DEFAULT " + self._column_default_sql(field)
                 if self.connection.features.requires_literal_defaults:
-                    # Some databases can't take defaults as a parameter (Oracle).
-                    # If this is the case, the individual schema backend should
-                    # implement prepare_default().
+                    # Some databases can't take defaults as a parameter
+                    # (Oracle, SQLite). If this is the case, the individual
+                    # schema backend should implement prepare_default().
                     yield column_default % self.prepare_default(default_value)
                 else:
                     yield column_default
@@ -333,7 +333,9 @@ class BaseDatabaseSchemaEditor:
         ):
             null = True
         if field.generated:
-            yield self._column_generated_sql(field)
+            generated_sql, generated_params = self._column_generated_sql(field)
+            params.extend(generated_params)
+            yield generated_sql
         elif not null:
             yield "NOT NULL"
         elif not self.connection.features.implied_column_null:
@@ -420,7 +422,7 @@ class BaseDatabaseSchemaEditor:
         compiler = query.get_compiler(connection=self.connection)
         default_sql, params = compiler.compile(db_default)
         if self.connection.features.requires_literal_defaults:
-            # Some databases doesn't support parameterized defaults (Oracle,
+            # Some databases don't support parameterized defaults (Oracle,
             # SQLite). If this is the case, the individual schema backend
             # should implement prepare_default().
             default_sql %= tuple(self.prepare_default(p) for p in params)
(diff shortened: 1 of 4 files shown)


tests/schema/tests.py
```diff
@@ -886,6 +895,39 @@ class Meta:
...
+        field.contribute_to_class(GeneratedFieldContainsModel, "contains_foo")
+
+        with connection.schema_editor() as editor:
+            editor.add_field(GeneratedFieldContainsModel, field)
+
+        obj = GeneratedFieldContainsModel.objects.create()
+        obj.refresh_from_db()
+        self.assertEqual(obj.text, "foo")
+        self.assertEqual(obj.generated, "foo%")
+        self.assertIs(obj.contains_foo, True)
```
charettes, line 929: This would likely benefit to be moved somewhere else and have extended coverage like it was added in #16840.

django/db/backends/base/schema.py
```diff
@@ -431,9 +433,7 @@ def _column_generated_sql(self, field):
         """Return the SQL to use in a GENERATED ALWAYS clause."""
         expression_sql, params = field.generated_sql(self.connection)
         persistency_sql = "STORED" if field.db_persist else "VIRTUAL"
-        if params:
-            expression_sql = expression_sql % tuple(self.quote_value(p) for p in params)
-        return f"GENERATED ALWAYS AS ({expression_sql}) {persistency_sql}"
+        return f"GENERATED ALWAYS AS ({expression_sql}) {persistency_sql}", params
```
charettes, line 436: The crux of the issue here is that
1. This part was never `%` escaped (`%` -> `%%`) which mean that things were working properly during model creation unless the model included a database default which resulted in `table_sql` returning `params` and then interpolation kicked (see the `params or None` in `create_model`). As for `add_field` it systematically crashed because a list was always passed to `execute(params)` which triggered interpolation which is what is what was reported on ticket-35336.
2. Postgres simply doesn't support parametrized generated expressions, it will crash with a very clear message about it.

charettes:
This one is going to be a fun one to solve without being too invasive. I'll revisit tomorrow.

nessita:
> This one is going to be a fun one to solve without being too invasive. I'll revisit tomorrow.

Thank you Simon for working on this. I'm currently planning on delaying the next release for a few days, even a week if necessary, for two reasons:

1. to have @sarahboyce  pairing with me to do the release, and since her first day would be next Tue, it feels a bit too much to start off with a release :exploding_head:
2. to wait for the release blocker fix

Let me know what you think!

charettes:
Sounds good, I'll try to find another solution that isn't as invasive for the backport with further improvement separated to target `main`.

tests/schema/tests.py
charettes: This should do for now.

charettes:
buildbot, test on oracle.

charettes:
buildbot, test on oracle.

django/db/backends/oracle/schema.py
```diff
@@ -200,7 +200,7 @@ def _generate_temp_name(self, for_name):
     def prepare_default(self, value):
         # Replace % with %% as %-formatting is applied in
         # FormatStylePlaceholderCursor._fix_for_params().
```
charettes, line 202: Assuming Oracle tests pass I'll push another comment that removes this value. Otherwise I'll have to dust out my local Oracle setup 😅

nessita:
buildbot test on oracle.

charettes:
It has already been tested on Oracle @nessita I just happened to delete the comment. You should see the successful status the checks.

nessita:
> It has already been tested on Oracle @nessita I just happened to delete the comment. You should see the successful status the checks.

I see it now, thank you!

django/db/backends/oracle/schema.py
nessita: @charettes so following our conversation about the Oracle tests passing, the comment above still needs removing? Thanks!

charettes: It does yes, I spotted it mid-Oracle test run so I wanted to avoid pushing a change that would require me to start over. I'll push a commit to this effect now that we know tests are passing.

sarahboyce:
Thank you Simon!
@AdrianGarcia-StayWithReside does it work for you?

AdrianGarcia-StayWithReside:
@sarahboyce I just gave it a try, both the example in my ticket and our actual production code worked without issue!

nessita (approved):
@charettes Thank you for putting together this PR!

Do you think we should `Refs` the related(ish) tickets in the commit message? I'm thinking about the three you mentioned (ticket-30408, ticket-34553, ticket-32369).

charettes:
@nessita I think it's fine to don't ref them since while they are related to the same fundamental problem they are for different features.

PR #18028, opened 2024-03-28
Fixed #35306 -- Improved documentation for the various date format settings

lufafajoshua:
# Trac ticket number

ticket-35306

# Branch description
Updated documentation on localization formats usage and precedence.

# Checklist

### DIFF
--- a/docs/ref/settings.txt
+++ b/docs/ref/settings.txt
@@ -1842,9 +1842,7 @@ standard :term:`language ID format <language code>`. For example, U.S. English
 is ``"en-us"``. See also the `list of language identifiers`_ and
 :doc:`/topics/i18n/index`.
 
-:setting:`USE_I18N` must be active for this setting to have any effect.
-
-It serves two purposes:
+It serves three purposes:
 
 * If the locale middleware isn't in use, it decides which translation is served
   to all users.
@@ -1852,6 +1850,11 @@ It serves two purposes:
   user's preferred language can't be determined or is not supported by the
   website. It also provides the fallback translation when a translation for a
   given literal doesn't exist for the user's preferred language.
+* If localization is explicitly disabled via the :tfilter:`unlocalize` filter
+  or the :ttag:`{% localize off %}<localize>` tag, it provides fallback
+  localization formats which will be applied instead. See
+  :ref:`controlling localization in templates <topic-l10n-templates>` for
+  details.
 
 See :ref:`how-django-discovers-language-preference` for more details.
 
--- a/docs/topics/i18n/formatting.txt
+++ b/docs/topics/i18n/formatting.txt
@@ -89,6 +89,9 @@ To activate or deactivate localization for a template block, use:
         {{ value }}
     {% endlocalize %}
 
+When localization is disabled, the :ref:`localization settings <settings-l10n>`
+formats are applied.
+
 See :tfilter:`localize` and :tfilter:`unlocalize` for template filters that will
 do the same job on a per-variable basis.
 
@@ -133,8 +136,9 @@ To force localization of a single value, use :tfilter:`localize`. To
 control localization over a large section of a template, use the
 :ttag:`localize` template tag.
 
-Returns a string representation for unlocalized numbers  (``int``, ``float``,
-or ``Decimal``).
+Returns a string representation for numbers  (``int``, ``float``, or
+``Decimal``) with the :ref:`localization settings <settings-l10n>` formats
+applied.
 
 .. _custom-format-files:
 


nessita:
@lufafajoshua thank you for your work.

When this PR is ready for review, please update the checklist from the PR description, and don't forget to set the proper ticket flags for this to be listed in the branches needing review.

docs/topics/i18n/formatting.txt
```diff
@@ -60,6 +60,9 @@ to be machine-readable, you will always want unlocalized values. You
 may also want to use localization in selected templates, rather than
 using localization everywhere.
 
+The format used is the one from the settings and under normal circumstances
+the locale dictated format has a higher precedence and will be applied instead.
```
claudep, line 64: What format? Which settings? Sorry, but I don't think this adds much clarity to these docs.

Maybe something like (once again I'm not a skilled English writer):
`When localization is disabled through one of the tag or filter described below, the localization formats used are taken from the general formatting settings (list of settings here?).`

claudep (changes requested):

docs/topics/i18n/formatting.txt
lufafajoshua: Hey @claudep , thanks for your review and response on this, Just for clarity, is it okay to pass a reference to the [globalization settings](https://docs.djangoproject.com/en/5.0/ref/settings/#globalization-i18n-l10n)  as the general formatting settings instead of listing them.

claudep: Good idea for this link! Now that I see that list, maybe it would be nice to distinguish in that list the settings that may overriding by locale formats, and those that don't, and then suggest (maybe in a different PR) separating that list in two sub-lists. Then we would have a perfect section to link to in this PR.

docs/ref/settings.txt
```diff
@@ -1838,21 +1838,23 @@ A list of IP addresses, as strings, that:
...
-A string representing the language code for this installation. This should be in
-standard :term:`language ID format <language code>`. For example, U.S. English
-is ``"en-us"``. See also the `list of language identifiers`_ and
-:doc:`/topics/i18n/index`.
-
-:setting:`USE_I18N` must be active for this setting to have any effect.
+A string representing the language code for this installation.
+This should be in standard :term:`language ID format <language code>`.
+For example, U.S. English is ``"en-us"``. See also the
+`list of language identifiers`_ and :doc:`/topics/i18n/index`.
```
sarahboyce, line 1844: This is the same, can you revert these changes?

docs/ref/settings.txt
```diff
@@ -1838,21 +1838,23 @@ A list of IP addresses, as strings, that:
...
 * If the locale middleware isn't in use, it decides which translation is served
   to all users.
 * If the locale middleware is active, it provides a fallback language in case the
   user's preferred language can't be determined or is not supported by the
   website. It also provides the fallback translation when a translation for a
   given literal doesn't exist for the user's preferred language.
+* If localization is explicitly disabled via the :tfilter:`localize` and
+  :tfilter:`unlocalize` template filters or tags, it provides fallback
+  localization formats which will be applied instead. see
+  :ref:`controlling localization in templates <topic-l10n-templates>`.
```
sarahboyce, line 1857: I think as we're referencing explicitly disabling localization, we should reference 
```
:tfilter:`unlocalize`
```
and 
```
:ttag:`{% localize off %} <localize>`
```
I think it might be wrong to reference
```
:tfilter:`localize`
```
when talking about disabling localization

docs/ref/settings.txt
```diff
@@ -1838,21 +1838,22 @@ A list of IP addresses, as strings, that:
...
 
 * If the locale middleware isn't in use, it decides which translation is served
   to all users.
 * If the locale middleware is active, it provides a fallback language in case the
   user's preferred language can't be determined or is not supported by the
   website. It also provides the fallback translation when a translation for a
   given literal doesn't exist for the user's preferred language.
+* If localization is explicitly disabled via the :tfilter:`localize` and
+  :tfilter:`unlocalize` template filters or tags. see
+  :ref:`controlling localization in templates <topic-l10n-templates>`.
```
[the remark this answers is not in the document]
sarahboyce: Yes, "see" needs to be capitalized and it would be good to end with "for [insert something]".  
You can just say "for more details" but you might want to be more specific like "for further explanation on [insert something]."
Rough example:
```
See :ref:`controlling localization in templates <topic-l10n-templates>` for more details.
```

docs/topics/i18n/formatting.txt
sarahboyce: It sounds like we still want to do this, it can be a separate commit or another PR 👍

docs/ref/settings.txt
```diff
@@ -3666,6 +3668,9 @@ Forms
 * :setting:`FORM_RENDERER`
 * :setting:`FORMS_URLFIELD_ASSUME_HTTPS`
 
+
+.. _globalization-settings:
```
sarahboyce, line 3672: This is a new unused reference, I believe the start of implementing [this comment](https://github.com/django/django/pull/18028#discussion_r1567427192).

sarahboyce (commented):
Thank you for this @lufafajoshua ⭐ I have a few comments
Can you also update the PR title to better represent what this PR is trying to acheive?

sarahboyce (approved):
Thank you @lufafajoshua 👍 this looks good to me!
I have squashed commits and applied minor updates, want to check other people are happy before merging but I think we're very close

claudep (approved):
LGTM, thanks!

PR #18033, opened 2024-03-29
Fixed #35342 -- Increased Integer clean reliability.

thiper01:
# Trac ticket number

ticket-35342

# Branch description
During the full_clean( ) routine, when an IntegerField is being cleaned and the code calls the field's to_python( ) method, if the value of the field is a float, it will just be converted to an integer using int(value). As it is characteristic of python's native int( ) function, it converts any float to int regardless of decimal places and approximations necessary, allowing for examples like this to happen:
```
exModel = ModelWithIntegerField(value=1.2)
>>   exModel.value     #1.2
exModel.clean_fields() #no validation error raised
>>   exModel.value     #1
```

I don't know if this behaviour is because of ticket [#24229](https://code.djangoproject.com/ticket/24229), but I believe there are better ways than just allowing any number through IntegerField verification as long as it's a number, regardless of how it affects the output.
My suggestions to improve this are:
1# Check first if it's a float type then verify if the numbers after the decimal point are 0 if value%1 == 0, if they are, no problem converting, if not raise a TypeError as it would distort the original value without warning.

# Checklist

### DIFF
--- a/django/db/models/fields/__init__.py
+++ b/django/db/models/fields/__init__.py
@@ -2141,6 +2141,9 @@ class IntegerField(Field):
         if value is None:
             return value
         try:
+            if isinstance(value, float):
+                if value%1 != 0:
+                    raise TypeError
             return int(value)
         except (TypeError, ValueError):
             raise exceptions.ValidationError(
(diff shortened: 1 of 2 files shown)


shangxiao:
Closing per ticket

PR #18039, opened 2024-04-02
Fixed #25947 -- Query's str() method worked normally when 'default' database is empty

jayvynl:
`sql_with_params`, `build_lookup` and `is_nullable` in `Query` would use the correct connection and generate more reasonable SQL.

# Trac ticket number

ticket-25947

# Branch description
This fix makes output of `print(queryset.query)` more reasonable when the queryset use a different database from `default`.

# Checklist

### DIFF
--- a/django/db/models/sql/query.py
+++ b/django/db/models/sql/query.py
@@ -289,8 +289,9 @@ class Query(BaseExpression):
 
     explain_info = None
 
-    def __init__(self, model, alias_cols=True):
+    def __init__(self, model, alias_cols=True, queryset=None):
         self.model = model
+        self.queryset = queryset
         self.alias_refcount = {}
         # alias_map is the most important data structure regarding joins.
         # It's used for recording which joins exist in the query and what
@@ -316,6 +317,20 @@ class Query(BaseExpression):
         self.extra = {}  # Maps col_alias -> (col_sql, params).
 
         self._filtered_relations = {}
+
+    # Remove queryset from pickle result.
+    # Refer https://code.djangoproject.com/ticket/27159
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        state["queryset"] = None
+        return state
+
+    @property
+    def db(self):
+        # Must not use ``if self.queryset``, that will trigger evaluation of queryset.
+        if self.queryset is not None:
+            return self.queryset.db
+        return DEFAULT_DB_ALIAS
 
     @property
     def output_field(self):
@@ -346,7 +361,7 @@ class Query(BaseExpression):
         Return the query as an SQL string and the parameters that will be
         substituted into the query.
         """
-        return self.get_compiler(DEFAULT_DB_ALIAS).as_sql()
+        return self.get_compiler(self.db).as_sql()
 
     def __deepcopy__(self, memo):
         """Limit the amount of work when a Query is deepcopied."""
@@ -1399,13 +1414,11 @@ class Query(BaseExpression):
             return lhs.get_lookup("isnull")(lhs, True)
 
         # For Oracle '' is equivalent to null. The check must be done at this
-        # stage because join promotion can't be done in the compiler. Using
-        # DEFAULT_DB_ALIAS isn't nice but it's the best that can be done here.
-        # A similar thing is done in is_nullable(), too.
+        # stage because join promotion can't be done in the compiler.
         if (
             lookup_name == "exact"
             and lookup.rhs == ""
-            and connections[DEFAULT_DB_ALIAS].features.interprets_empty_strings_as_nulls
+            and connections[self.db].features.interprets_empty_strings_as_nulls
         ):
             return lhs.get_lookup("isnull")(lhs, True)
 
@@ -2624,14 +2637,9 @@ class Query(BaseExpression):
         nullable for those backends. In such situations field.null can be
         False even if we should treat the field as nullable.
         """
-        # We need to use DEFAULT_DB_ALIAS here, as QuerySet does not have
-        # (nor should it have) knowledge of which connection is going to be
-        # used. The proper fix would be to defer all decisions where
-        # is_nullable() is needed to the compiler stage, but that is not easy
-        # to do currently.
         return field.null or (
             field.empty_strings_allowed
-            and connections[DEFAULT_DB_ALIAS].features.interprets_empty_strings_as_nulls
+            and connections[self.db].features.interprets_empty_strings_as_nulls
         )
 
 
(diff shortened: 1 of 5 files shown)


tests/queries/test_query.py
```diff
@@ -160,6 +160,11 @@ def test_filter_non_conditional(self):
         with self.assertRaisesMessage(TypeError, msg):
             query.build_where(Func(output_field=CharField()))
 
+    def test_use_correct_db(self):
```
sarahboyce, line 163: This tests the implementation without covering the original issue as raised in the ticket, I think having something like:

```diff
diff --git a/tests/multiple_database/tests.py b/tests/multiple_database/tests.py
index 9587030a46..acda154f48 100644
--- a/tests/multiple_database/tests.py
+++ b/tests/multiple_database/tests.py
@@ -2,12 +2,12 @@ import datetime
 import pickle
 from io import StringIO
 from operator import attrgetter
-from unittest.mock import Mock
+from unittest.mock import Mock, patch

 from django.contrib.auth.models import User
 from django.contrib.contenttypes.models import ContentType
 from django.core import management
-from django.db import DEFAULT_DB_ALIAS, router, transaction
+from django.db import DEFAULT_DB_ALIAS, router, transaction, connections
 from django.db.models import signals
 from django.db.utils import ConnectionRouter
 from django.test import SimpleTestCase, TestCase, override_settings
@@ -1235,6 +1235,21 @@ class QueryTestCase(TestCase):
         val = Book.objects.raw("SELECT id FROM multiple_database_book").using("other")
         self.assertQuerySetEqual(val, [dive.pk], attrgetter("pk"))

+    def test_query_string(self):
+        with (
+            patch.object(connections["default"].ops, "compiler")
+            as default_db_compiler
+        ):
+            queryset = Person.objects.using("other").all().order_by("id")
+            self.assertEqual(
+                str(queryset.query),
+                'SELECT "multiple_database_person"."id", '
+                '"multiple_database_person"."name" '
+                'FROM "multiple_database_person" '
+                'ORDER BY "multiple_database_person"."id" ASC'
+            )
+            default_db_compiler.assert_not_called()
+
     def test_select_related(self):
         """
```
might be better

django/db/models/sql/query.py
```diff
@@ -289,8 +289,9 @@ class Query(BaseExpression):
 
     explain_info = None
 
-    def __init__(self, model, alias_cols=True):
+    def __init__(self, model, alias_cols=True, queryset=None):
```
sarahboyce, line 292: I think I would prefer not to pass in the queryset and then we can be less worried about evaluating it, maybe we can take a look at the db alias?
Something like:
```diff
--- a/django/db/models/sql/query.py
+++ b/django/db/models/sql/query.py
@@ -288,10 +288,10 @@ class Query(BaseExpression):
     deferred_loading = (frozenset(), True)

     explain_info = None
+    using = None

-    def __init__(self, model, alias_cols=True, queryset=None):
+    def __init__(self, model, alias_cols=True):
         self.model = model
-        self.queryset = queryset
         self.alias_refcount = {}
         # alias_map is the most important data structure regarding joins.
         # It's used for recording which joins exist in the query and what
@@ -318,19 +318,9 @@ class Query(BaseExpression):

         self._filtered_relations = {}

-    # Remove queryset from pickle result.
-    # Refer https://code.djangoproject.com/ticket/27159
-    def __getstate__(self):
-        state = self.__dict__.copy()
-        state["queryset"] = None
-        return state
-
     @property
     def db(self):
-        # Must not use ``if self.queryset``, that will trigger evaluation of queryset.
-        if self.queryset is not None:
-            return self.queryset.db
-        return DEFAULT_DB_ALIAS
+        return self.using or DEFAULT_DB_ALIAS
```
and then we set `query.using` 🤔 maybe we call it `_db` or `db` (naming is hard)

sarahboyce (commented):
Thank you for being patient @jayvynl 👍 
I have some thoughts (and I'm really sorry, I started this a while ago without submitting the review 😭)

tests/queries/test_query.py
jayvynl: You're right, I have add your test in the new commit.

django/db/models/sql/query.py
jayvynl: Thank you for your suggestion, in the last commit, I have pass QuerySet.db instead of QuerySet itself to Query.

django/db/models/sql/query.py
```diff
@@ -2624,14 +2631,9 @@ def is_nullable(self, field):
         nullable for those backends. In such situations field.null can be
         False even if we should treat the field as nullable.
         """
-        # We need to use DEFAULT_DB_ALIAS here, as QuerySet does not have
-        # (nor should it have) knowledge of which connection is going to be
```
charettes, line 2628: I think [this comment nails it](https://code.djangoproject.com/ticket/25947#comment:1), _teaching_ `sql.Query` about db alias seems wrong as this abstraction was designed to be backend agnostic.

Moving forward with this change gives the impression that `self.db` will always be the right choice over `DEFAULT_DB_ALIAS` but in most cases the proper database to use won't be known until execution time as it will be determine dynamically from `django.db.router`. All this code does is _fix_ the crash when `using()` explicitly used.

The impact of the changes proposed here go well beyond fixing what was brought up in ticket-25947 which is that `Query.__str__()` would crash if there was no `DATABASES["default"]` alias defined. If we really want to prevent that from happening (lots of things are going to break if you don't set `DATABASES["default"]` btw) I suggest we define a property that does

```python
@cached_property
def _default_db_alias(self):
    if DEFAULT_DB_ALIAS in connections:
        return DEFAULT_DB_ALIAS
    # Pick the first defined alias if "default" is mising
    return next(iter(connections), DEFAULT_DB_ALIAS)
```

Trying to solve this particular use case of supporting a lack of `DATABASES["default"]` is a slippery slope though as tens of features expect it to be defined. Addressing this issue sends a signal that this is a blessed configuration and that we commit to solving other `DEFAULT_DB_ALIAS` usages.

---

By looking at this ticket and [ticket-25705](https://code.djangoproject.com/ticket/25705#comment:12) recently I'm starting to believe that we should combine these two tickets into a single one that keep things as it for `Query.__str__()` and document `Query.sql_with_params(using: str = DEFAULT_DB_ALIAS)` instead.

That would allow users that have exotic `DATABASES` configurations to use this method instead to target a particular aliases, avoid venturing down the path of safe and correct parameters escaping for all backends (and force third party ones to do so as well), enforce the safe usages of `QuerySet.raw`.

jayvynl: How about also set Query.db in QuerySet.__init__ ? This will fix the crash when default db is empty .

jayvynl: OK, thank you for all the explanation.

PR #18041, opened 2024-04-02
Early adopt UTF-8 Mode on Windows.

smithdc1:
Python 3.15 will enable UTF-8 mode by default

https://peps.python.org/pep-0686/#enable-utf-8-mode-by-default

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,6 +19,9 @@ permissions:
 
 jobs:
   windows:
+    env:
+      # Python 3.15 will make UTF-8 Mode the default.
+      PYTHONUTF8: 1
     runs-on: windows-latest
     strategy:
       matrix:


nessita:
Hey @smithdc1, I find this PR valuable and I think we should merge it. Is there something else that needs to be checked before marking it ready for review?

smithdc1:
Hi @nessita this is ready for review. 👍

(Do you want me to create / find a ticket to reference?)

adamchainz:
Is there much risk that we will miss bugs that only occur in non-UTF-8 mode?

nessita:
> Is there much risk that we will miss bugs that only occur in non-UTF-8 mode?

Do you have any more details about your thinking?

To understand more about this, I spent some time reading thru:
* https://discuss.python.org/t/pep-597-enable-utf-8-mode-by-default-on-windows/3122
* https://discuss.python.org/t/pep-686-make-utf-8-mode-default/14435
* https://discuss.python.org/t/pep-686-make-utf-8-mode-default-round-2/14737
* https://docs.python.org/3/using/windows.html#utf-8-mode

My understanding so far is that using `UTF-8` as the default encoding for writing and reading is the right thing to do for Windows users to have the best experience, and if Python did not switch the default sooner was only to be super cautious.
The biggest concern as I understood it was reading and interacting with config files or similar that would have been written to the FS without an explicit encoding, specifically when they include non-ascii characters. I'm not sure this would be of relevance for Django? The main counterargument is that writing a file without passing an encoding explicitly should be treated as a bug on its own, and that that scenario has other problems on their own.

smithdc1:
Thank you for sharing the links. 

I think this [comment](https://discuss.python.org/t/pep-686-make-utf-8-mode-default-round-2/14737/15) from Steve Dower makes me believe this is the right thing to do. 

> One of my conditions (as much as I can apply it) is that the UTF-8 mode be evangelised widely before the default is changed.

Re Adam's comment about missing bugs in non-UTF-8 mode maybe we could run the tests once a week or so in non-UTF-8 mode?

nessita:
> Thank you for sharing the links.
> 
> I think this [comment](https://discuss.python.org/t/pep-686-make-utf-8-mode-default-round-2/14737/15) from Steve Dower makes me believe this is the right thing to do.
> 
> > One of my conditions (as much as I can apply it) is that the UTF-8 mode be evangelised widely before the default is changed.
> 
> Re Adam's comment about missing bugs in non-UTF-8 mode maybe we could run the tests once a week or so in non-UTF-8 mode?

I'm undecided on the value of this, but of course I would not object to add an entry for this for our scheduled tests. Would you be up to give that a try?

@adamchainz any further thoughts?

Thank you for the follow up! :star2:

PR #18045, opened 2024-04-03
 Fixed 35339 - Fixed PostgreSQL aggregate order by and filter params order 

camuthig:
# Trac ticket number

ticket-35339

# Branch description
This change fixes a bug in the PostgreSQL aggregate SQL generation when both the filter and order by expressions require parameters.

Additional changes are made to the `Aggregate` class to more consistently return the `source_expressions` for use in generating the SQL.

# Checklist

### DIFF
--- a/django/contrib/postgres/aggregates/mixins.py
+++ b/django/contrib/postgres/aggregates/mixins.py
@@ -17,13 +17,10 @@ class OrderableAggMixin:
         return super().resolve_expression(*args, **kwargs)
 
     def get_source_expressions(self):
-        if self.order_by is not None:
-            return super().get_source_expressions() + [self.order_by]
-        return super().get_source_expressions()
+        return super().get_source_expressions() + [self.order_by]
 
     def set_source_expressions(self, exprs):
-        if isinstance(exprs[-1], OrderByList):
-            *exprs, self.order_by = exprs
+        *exprs, self.order_by = exprs
         return super().set_source_expressions(exprs)
 
     def as_sql(self, compiler, connection):
--- a/django/contrib/postgres/aggregates/mixins.py
+++ b/django/contrib/postgres/aggregates/mixins.py
@@ -1,3 +1,4 @@
+from django.core.exceptions import FullResultSet
 from django.db.models.expressions import OrderByList
 
 
@@ -24,9 +25,23 @@ class OrderableAggMixin:
         return super().set_source_expressions(exprs)
 
     def as_sql(self, compiler, connection):
-        if self.order_by is not None:
-            order_by_sql, order_by_params = compiler.compile(self.order_by)
-        else:
-            order_by_sql, order_by_params = "", ()
-        sql, sql_params = super().as_sql(compiler, connection, ordering=order_by_sql)
-        return sql, (*sql_params, *order_by_params)
+        *source_exprs, filtering_expr, ordering_expr = self.get_source_expressions()
+
+        order_by_sql = ""
+        order_by_params = []
+        if ordering_expr is not None:
+            order_by_sql, order_by_params = compiler.compile(ordering_expr)
+
+        filter_params = []
+        if filtering_expr is not None:
+            try:
+                _, filter_params = compiler.compile(filtering_expr)
+            except FullResultSet:
+                pass
+
+        source_params = []
+        for source_expr in source_exprs:
+            source_params += compiler.compile(source_expr)[1]
+
+        sql, _ = super().as_sql(compiler, connection, ordering=order_by_sql)
+        return sql, (*source_params, *order_by_params, *filter_params)
--- a/django/db/models/aggregates.py
+++ b/django/db/models/aggregates.py
@@ -50,12 +50,10 @@ class Aggregate(Func):
 
     def get_source_expressions(self):
         source_expressions = super().get_source_expressions()
-        if self.filter:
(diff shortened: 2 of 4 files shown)


tests/aggregation/tests.py
```diff
@@ -1291,7 +1291,7 @@ class MyMax(Max):
 
             def as_sql(self, compiler, connection):
                 copy = self.copy()
-                copy.set_source_expressions(copy.get_source_expressions()[0:1])
+                copy.set_source_expressions(copy.get_source_expressions()[0:1] + [None])
```
charettes, line 1294: That's good enough given how convoluted this test is.

django/contrib/postgres/aggregates/mixins.py
```diff
@@ -17,19 +18,29 @@ def resolve_expression(self, *args, **kwargs):
...
+        *source_exprs, filtering_expr, ordering_expr = self.get_source_expressions()
+
+        order_by_sql, order_by_params = "", []
+        if ordering_expr is not None:
+            order_by_sql, order_by_params = compiler.compile(ordering_expr)
+
+        filter_params = []
+        if filtering_expr is not None:
+            try:
+                _, filter_params = compiler.compile(filtering_expr)
```
charettes, line 37: I wish there was a way to avoid the double compilation but without a significant refactor I don't see how. This should be good enough for now.

django/contrib/postgres/aggregates/mixins.py
```diff
@@ -17,19 +18,29 @@ def resolve_expression(self, *args, **kwargs):
...
     def get_source_expressions(self):
-        if self.order_by is not None:
-            return super().get_source_expressions() + [self.order_by]
-        return super().get_source_expressions()
+        return super().get_source_expressions() + [self.order_by]
 
     def set_source_expressions(self, exprs):
-        if isinstance(exprs[-1], OrderByList):
-            *exprs, self.order_by = exprs
+        *exprs, self.order_by = exprs
```
charettes, line 24: Excellent!

charettes (commented):
This is looking promising @camuthig nice work!

I suspect that the failing admin facets tests are due to the fact they wrongly pass empty `Q` objects to `Aggregate.filter` and previously the code was doing `if self.filter` in `get_source_expressions` and preventing it from being used `get_refs`.

I suggest that you either do

```diff
diff --git a/django/db/models/aggregates.py b/django/db/models/aggregates.py
index 0cbffacd1b..5cf59b22cd 100644
--- a/django/db/models/aggregates.py
+++ b/django/db/models/aggregates.py
@@ -40,7 +40,7 @@ def __init__(
         if default is not None and self.empty_result_set_value is not None:
             raise TypeError(f"{self.__class__.__name__} does not allow default.")
         self.distinct = distinct
-        self.filter = filter
+        self.filter = filter or None
         self.default = default
         super().__init__(*expressions, **extra)
```

or

```
index 0cbffacd1b..49190d8335 100644
@@ -63,8 +63,10 @@ def resolve_expression(
     ):
         # Aggregates are not allowed in UPDATE queries, so ignore for_save
         c = super().resolve_expression(query, allow_joins, reuse, summarize)
-        c.filter = c.filter and c.filter.resolve_expression(
-            query, allow_joins, reuse, summarize
+        c.filter = (
+            c.filter.resolve_expression(query, allow_joins, reuse, summarize)
+            if c.filter
+            else None
         )
         if summarize:
             # Summarized aggregates cannot refer to summarized aggregates.
```

The second is likely less invasive.

django/db/models/aggregates.py
```diff
@@ -105,7 +105,7 @@ def resolve_expression(
     @property
     def default_alias(self):
         expressions = self.get_source_expressions()
-        if len(expressions) == 1 and hasattr(expressions[0], "name"):
+        if len(list(filter(None, expressions))) == 1 and hasattr(expressions[0], "name"):
```
camuthig, line 108: One issue I faced was that I couldn't really make the results of `get_source_expressions` non-variadic. It causes issues when what we need to get is the source expressions, excluding filtering and ordering. This might be easier if we merged the `OrderableAggMixin` directly into the `Aggregate` class and could know there are always exactly two extra, nullable expressions on the end, but for now, it requires taking a bit of an educated guess with things like `filter(None, expressions)` to get around the variadic nature.

I explored using an `ExpressionList` for the source expressions instead, but ran into two major issues. The first being that importing it caused a cyclic import loop. The second, bigger issue being that so much of `Func` depends on `self.source_expressions` being variadic that collapsing it into a non-iterable `ExpressionList` created a lot of problems.

camuthig (commented):
Thanks for the input, @charettes . I was going to ping you today to help unblock me on those failing tests and some other questions/thoughts I had. It looks like your understanding of the failing tests was correct. At least in my local environment, the admin facets seem to be working again.

charettes:
@camuthig this is looking good to me.

If you could address the linting violations, mark the MR as non-draft, and change the commit / MR to `Fixed #35339 -- ` (we can tackle the refactor in a separate ticket / MR if you're interested to keep working on it) it should be ready for review of the fellows.

nessita (approved):
Thank you @camuthig for the PR! And thank you Simon for the review.

I have updated tiny changes and reworded the commit messages a bit. Will land when CI is green!

PR #18046, opened 2024-04-04
Refs #30224 -- Added test for MySQL datetime casting.

sarahboyce:
This is just to make existing work available for others to validate or use as a starting point 👍


sarahboyce:
Closing after ticket discussion resulted in a won't fix 👍

PR #18055, opened 2024-04-06
Fixed #35356 -- Deferred self-referential foreign key fields adequately.

charettes:
# Trac ticket number

ticket-35356

# Branch description

While ticket-34612 surfaced issues with reverse one-to-one fields deferrals it missed that switching to storing remote fields would break self-referential relationships.

This change switches to storing related objects in the select mask instead of remote fields to prevent collisions when dealing with self-referential relationships that might have a different directional mask.

While fixing ticket-21204 introduced a crash under some self-referential deferral conditions it was simply not working before as it aggregated the sets of deferred fields by model.

Thanks @OdinsPlasmaRifle for the report.


django/db/models/sql/compiler.py
```diff
@@ -1253,21 +1253,21 @@ def get_related_klass_infos(klass_info, related_klass_infos):
 
         if restricted:
             related_fields = [
-                (o.field, o.related_model)
+                (o, o.field, o.related_model)
```
charettes, line 1256: Capture the related object to use it in the loop below.

django/db/models/sql/compiler.py
```diff
@@ -1253,21 +1253,21 @@ def get_related_klass_infos(klass_info, related_klass_infos):
...
             ]
-            for related_field, model in related_fields:
-                related_select_mask = select_mask.get(related_field) or {}
+            for related_object, related_field, model in related_fields:
                 if not select_related_descend(
                     related_field,
                     restricted,
                     requested,
-                    related_select_mask,
+                    select_mask,
```
charettes, line 1265: `related_field` must be checked againt `select_mask` and not `select_mask.get(related_field)` just like any field. See how it's done in the `for f in opts.fields` loop above.

django/db/models/sql/compiler.py
```diff
@@ -1253,21 +1253,21 @@ def get_related_klass_infos(klass_info, related_klass_infos):
...
                     reverse=True,
                 ):
                     continue
 
+                related_select_mask = select_mask.get(related_object) or {}
```
charettes, line 1270: Now that we know that we can drill deeper from `select_related_descend` capture the related select mask to pass it down recursively.

django/db/models/sql/query.py
```diff
@@ -815,19 +815,17 @@ def _get_defer_select_mask(self, opts, mask, select_mask=None):
             if filtered_relation := self._filtered_relations.get(field_name):
                 relation = opts.get_field(filtered_relation.relation_name)
                 field_select_mask = select_mask.setdefault((field_name, relation), {})
-                field = relation.field
             else:
-                reverse_rel = opts.get_field(field_name)
+                relation = opts.get_field(field_name)
```
charettes, line 819: Store the related object directly instead of the remote field to be able to differentiate between the two when dealing with self-referential relationship.

django/db/models/sql/query.py
```diff
@@ -840,13 +838,7 @@ def _get_only_select_mask(self, opts, mask, select_mask=None):
...
         for field_name, field_mask in mask.items():
             field = opts.get_field(field_name)
-            # Retrieve the actual field associated with reverse relationships
-            # as that's what is expected in the select mask.
-            if field in opts.related_objects:
-                field_key = field.field
-            else:
-                field_key = field
-            field_select_mask = select_mask.setdefault(field_key, {})
+            field_select_mask = select_mask.setdefault(field, {})
```
charettes, line 841: Same thing for the `only` case instead of `defer`

tests/defer_regress/tests.py
```diff
@@ -309,6 +309,27 @@ def test_only_reverse_many_to_many_ignored(self):
...
 
+    def test_self_referential_one_to_one(self):
+        first = Item.objects.create(name="first", value=1)
+        second = Item.objects.create(name="second", value=2, source=first)
+        with self.assertNumQueries(1):
+            deferred_first, deferred_second = (
+                Item.objects.select_related("source", "destination")
+                .only("name", "source__name", "destination__value")
+                .order_by("pk")
+            )
```
charettes, line 320: To give you an idea of how this was broken in >4.2 @OdinsPlasmaRifle

This is the query that was performed on Django 4.1. Notice how the select mask is a union of all the references to the same model

```sql
SELECT 
  "defer_regress_item"."id", 
  "defer_regress_item"."name", 
  "defer_regress_item"."value", 
  "defer_regress_item"."source_id", 
  T2."id", 
  T2."name", 
  T2."value", 
  T2."source_id", 
  T3."id", 
  T3."name", 
  T3."value", 
  T3."source_id" 
FROM 
  "defer_regress_item" 
  LEFT OUTER JOIN "defer_regress_item" T2 ON (
    "defer_regress_item"."source_id" = T2."id"
  ) 
  LEFT OUTER JOIN "defer_regress_item" T3 ON (
    "defer_regress_item"."id" = T3."source_id"
  ) 
ORDER BY 
  "defer_regress_item"."id" ASC
```

This now doing the proper thing

```sql
SELECT 
  "defer_regress_item"."id", 
  "defer_regress_item"."name", 
  "defer_regress_item"."source_id", 
  T2."id", 
  T2."name", 
  T3."id", 
  T3."value" 
FROM 
  "defer_regress_item" 
  LEFT OUTER JOIN "defer_regress_item" T2 ON (
    "defer_regress_item"."source_id" = T2."id"
  ) 
  LEFT OUTER JOIN "defer_regress_item" T3 ON (
    "defer_regress_item"."id" = T3."source_id"
  ) 
ORDER BY 
  "defer_regress_item"."id" ASC
```

django/db/models/query_utils.py
```diff
@@ -340,38 +340,37 @@ def _unregister_instance_lookup(self, lookup, lookup_name=None):
...
-    if restricted:
-        if reverse and field.related_query_name() not in requested:
-            return False
-        if not reverse and field.name not in requested:
-            return False
-    if not restricted and field.null:
+    # When `select_related()` is used without a *requested mask all
+    # relationships are descended unless they are nullable.
+    if not restricted:
+        return not field.null
```
charettes, line 366: [Per the docs](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#django.db.models.query.QuerySet.select_related)

> There may be some situations where you wish to call select_related() with a lot of related objects, or where you don’t know all of the relations. In these cases it is possible to call select_related() with no arguments. This will follow all **non-null foreign keys** it can find - **nullable foreign keys must be specified**.

django/db/models/query_utils.py
```diff
@@ -340,38 +340,37 @@ def _unregister_instance_lookup(self, lookup, lookup_name=None):
...
+     * `select_mask`
     """
+    # Only relationships can be descended.
     if not field.remote_field:
         return False
-    if field.remote_field.parent_link and not reverse:
+    # Forward MTI parent links should not be explicitly descended as they are
+    # always JOIN'ed against unless excluded by `select_mask`.
+    if getattr(field.remote_field, "parent_link", False):
         return False
```
charettes, line 362: In other words, for the following models

```python
class Parent(models.Model):
   pass

class Child(Parent):
    pass
```

There's no need to do anything special for `Child.parent_ptr` (the implicit `parent_link` primary key) as it will be always be implicitly followed when doing `Child.objects.all()`.

django/db/models/query_utils.py
```diff
@@ -340,38 +340,37 @@ def _unregister_instance_lookup(self, lookup, lookup_name=None):
...
-            return False
-    if not restricted and field.null:
+    # When `select_related()` is used without a *requested mask all
+    # relationships are descended unless they are nullable.
+    if not restricted:
+        return not field.null
+    # When `select_related(*requested)` is used only fields that are part
+    # of `requested` should be descended.
+    if field.name not in requested:
         return False
```
charettes, line 370: Self explanatory, `Book.publisher` should not be followed on `Book.objects.select_related("author")`.

django/db/models/sql/query.py
```diff
@@ -791,46 +791,44 @@ def _get_defer_select_mask(self, opts, mask, select_mask=None):
...
-                # alteration time will be implemented eventually.
-                if not hasattr(reverse_rel, "field"):
-                    continue
-                field = reverse_rel.field
-                field_select_mask = select_mask.setdefault(field, {})
-            related_model = field.model._meta.concrete_model
-            self._get_defer_select_mask(
-                related_model._meta, field_mask, field_select_mask
-            )
+                opts.get_field(field_name)
```
charettes, line 831: All merged in the above loop by doing `opts.concrete_fields + opts.related_objects`.

felixxm (approved):
@charettes Thanks :+1: Looks great :rocket: 

I rebased, squashed the last two commits, and fixed tiny typos.

charettes:
Thanks @felixxm, appreciated 🙇 🎉

nessita (commented):
Thank you @charettes for this work and all the explanatory comments, those really helped me follow the code.

While trying to go deeper in some parts of these changes, I crafted a test to see how select-relating on many "chained" relation would behave (heavily based on your test and would definitely need a better name and polish if we were to include it):

```python
    def test_self_referential_one_to_one_2(self):
        first = Item.objects.create(name="first", value=1)
        second = Item.objects.create(name="second", value=2, source=first)
        third = Item.objects.create(name="third", value=2, source=second)
            result = (
                Item.objects.select_related("source", "destination", "destination__source")
                .only("name", "source__name", "destination__value", "destination__source__value")
                .order_by("pk")
            )
            self.assertEqual(len(result), 3)

        deferred_first, deferred_second, deferred_third = result

            self.assertEqual(deferred_first.destination.source.value, first.value)
            self.assertEqual(deferred_second.destination.source.value, second.value)
```

The test above fails with:

```
FAIL: test_self_referential_one_to_one_2 (defer_regress.tests.DeferRegressionTest.test_self_referential_one_to_one_2)
  File "/home/nessita/fellowship/django/tests/defer_regress/tests.py", line 347, in test_self_referential_one_to_one_2
  File "/home/nessita/fellowship/django/django/test/testcases.py", line 128, in __exit__
    self.test_case.assertEqual(
AssertionError: 2 != 0 : 2 queries executed, 0 expected
Captured queries were:
1. SELECT "defer_regress_item"."id", "defer_regress_item"."value" FROM "defer_regress_item" WHERE "defer_regress_item"."id" = 1 LIMIT 21
2. SELECT "defer_regress_item"."id", "defer_regress_item"."value" FROM "defer_regress_item" WHERE "defer_regress_item"."id" = 2 LIMIT 21
```

Could you confirm if, given how "nontraditional" would be to select-related on `destination__source` (which is the Item instance itself), this makes sense or not?

charettes:
@nessita I've started looking into the failure you've provided and I'm not sure what to think of it. It can be reduced to

```diff
diff --git a/tests/defer_regress/tests.py b/tests/defer_regress/tests.py
index 1209325f21..eb76c843fe 100644
--- a/tests/defer_regress/tests.py
+++ b/tests/defer_regress/tests.py
@@ -330,6 +330,22 @@ def test_self_referential_one_to_one(self):
         with self.assertNumQueries(1):
             self.assertEqual(deferred_first.destination.name, second.name)

+    def test_self_referential_one_to_one_2(self):
+        first = Item.objects.create(name="first", value=1)
+        second = Item.objects.create(name="second", value=2, source=first)
+        with self.assertNumQueries(1):
+            result = (
+                Item.objects.select_related("destination__source")
+                .only("destination__source__value")
+                .order_by("pk")
+            )
+            self.assertEqual(len(result), 2)
+
+        deferred_first, deferred_second = result
+
+        with self.assertNumQueries(0):
+            self.assertEqual(deferred_first.destination.source.value, first.value)
+

 class DeferDeletionSignalsTests(TestCase):
     senders = [Item, Proxy]
```

The interesting part is that the proper SQL is generated

```sql
SELECT "defer_regress_item"."id",
       T2."id",
       T2."source_id",
       T3."id",
       T3."value"
FROM "defer_regress_item"
LEFT OUTER JOIN "defer_regress_item" T2 ON ("defer_regress_item"."id" = T2."source_id")
LEFT OUTER JOIN "defer_regress_item" T3 ON (T2."source_id" = T3."id")
ORDER BY "defer_regress_item"."id" ASC
```

So I suspected a problem in `klass_info` generation or how it's interpreted by the related populator in `ModelIterable.__iter__`. From a quick glance I suspect that it's a problem in the latter in [this particular bit of logic that special cases _known related objects_](https://github.com/django/django/blob/f0d50a937910571fc4adb7e6fb91045cdc215aa8/django/db/models/query.py#L106-L122).

What I suspect happens is that because `deferred_first.destination.source` is basically `deferred_first` and no fields are selected for it in the first case then it's simply not loaded in the second case. You can assert this is the case by adding the following assertion

```python
self.assertIs(deferred_first.destination.source, deferred_first)
```

The thing is `deferred_first.destination.source` should be a distinct Python objects than `deferred_first` if we want the two objects to be able to have a different set of fields deferred but the _know related objects_ caching breaks that. Due to how we also cache one-to-one relationships between objects I'm not even sure if it's possible or a good idea to support this use case.

TL;DR we could open a Trac ticket to track and discuss whether we want to error out or deal with this issue differently but it seems like a very exotic edge case that is distinct from the problem being solved here.

As for the comments you added during the review I agree with all of them. LMK if you'd like me to apply them myself or you'd prefer to take care of that. Thank you!

nessita:
Thank you @charettes for the response, it makes sense.

> we could open a Trac ticket to track and discuss whether we want to error out or deal with this issue differently but it seems like a very exotic edge case that is distinct from the problem being solved here.

While I would like that if the objects are the same because of the _identity_ property, then the resulting SQL would also be as smart and just fetch the data from the "outter" `SELECT` (instead of navigating `JOIN`s against the same row), I agree with your assessment and I don't see the need right now to create a new ticket or do any fix.

> LMK if you'd like me to apply them myself or you'd prefer to take care of that.

I can take care of that tomorrow. Thank you!

nessita (approved):
LGTM, landing when CI is green.

PR #18058, opened 2024-04-08
Fixed #35350 -- Fixed save() with pk set on models with GeneratedFields.

sarahboyce:
Thanks Matt Hegarty for the report.

Regression in f333e35.

# Trac ticket number

ticket-35350

# Checklist

### DIFF
--- a/django/db/models/base.py
+++ b/django/db/models/base.py
@@ -1063,12 +1063,16 @@ class Model(AltersData, metaclass=ModelBase):
         for a single table.
         """
         meta = cls._meta
-        non_pks = [f for f in meta.local_concrete_fields if not f.primary_key]
+        non_pks_non_generated = [
+            f
+            for f in meta.local_concrete_fields
+            if not f.primary_key and not f.generated
+        ]
 
         if update_fields:
-            non_pks = [
+            non_pks_non_generated = [
                 f
-                for f in non_pks
+                for f in non_pks_non_generated
                 if f.name in update_fields or f.attname in update_fields
             ]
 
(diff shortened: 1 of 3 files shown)


django/db/models/base.py
```diff
@@ -1063,12 +1063,16 @@ def _save_table(
         for a single table.
         """
         meta = cls._meta
-        non_pks = [f for f in meta.local_concrete_fields if not f.primary_key]
+        non_pks_non_generated = [
+            f
+            for f in meta.local_concrete_fields
+            if not f.primary_key and not f.generated
+        ]
```
charettes, line 1070: This doesn't have to done here as this is meant as a non-invasive backport but since `non_pks_non_generated` is only used for containment checks and doesn't need to preserve ordering it might be a good candidate to use a `set` instead of a `list`.

charettes (approved):

sarahboyce:
Note that I have just finished testing this against the test project that @matthewhegarty (thank you for proving that ⭐) and this works for me there too. 👍

docs/releases/5.0.5.txt
```diff
@@ -9,4 +9,6 @@ Django 5.0.5 fixes several bugs in 5.0.4.
 Bugfixes
 ========
 
-* ...
+* Fixed a bug in Django 5.0 that caused a crash of ``Model.save()`` for
+  models with a ``GeneratedField`` when creating a model instance while
+  providing a primary key (:ticket:`35350`).
```
nessita, line 14: Just an optional suggestion!

```suggestion
* Fixed a bug in Django 5.0 that caused a crash of ``Model.save()`` when
  creating an instance of a model with a ``GeneratedField`` and providing a
  primary key (:ticket:`35350`).
```

nessita (commented):

nessita (approved):
Thank you!

django/db/models/base.py
adamchainz: It could also be cached up on `Options`, it doesn’t need regenerating on every `save()` call.

PR #18062, opened 2024-04-10
 Fixed -- Inconsistency in tutorial 7.

milinsoft:
# Trac ticket number
N/A

# Branch description
Currently description text is inconsistent with the code example.  List vs Tuple

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -128,6 +128,7 @@ answer newbie questions, and generally made Django that much better:
     av0000@mail.ru
     Axel Haustant <noirbizarre@gmail.com>
     Aymeric Augustin <aymeric.augustin@m4x.org>
+    Aleksander Milinkevich <milinsoft@gmail.com>
     Bahadır Kandemir <bahadir@pardus.org.tr>
     Baishampayan Ghose
     Baptiste Mispelon <bmispelon@gmail.com>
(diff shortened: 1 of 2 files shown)


AgaevDanik (approved):

nessita (commented):
Hello! Thank you for taking the time to propose this PR.

I see where the confusion comes from, but instead of using tuples, I would recommend that we adjust the description, for example:

> which is a sequence of field names to display (list, tuple, etc.), as columns, on the ...

What do you think?

milinsoft:
> Hello! Thank you for taking the time to propose this PR.
> 
> I see where the confusion comes from, but instead of using tuples, I would recommend that we adjust the description, for example:
> 
> > which is a sequence of field names to display (list, tuple, etc.), as columns, on the ...
> 
> What do you think?

Hi,
I think both options are great, I also considered that. If we go this way, I would adjust the description to `iterable` as it is more universal. 
Pertaining to the code part, as long as we don't need dynamically modify the content, a `tuple` would fit better. 

With that being said, I propose update the description adding `iterable` (`tuple`, `list`, etc...) and show tuple as example.
@nessita  What are your thoughts?

milinsoft:
@nessita 
Thank you for the explanation! 
I've updated the PR.

nessita:
> @nessita Thank you for the explanation! I've updated the PR.

Thank you @milinsoft! Would you go the extra mile and follow the PR template when it asks for the commit message to be written in past tense? You can see examples in the current git log history:

```
$ git log | grep tutorial
    Fixed #35347 -- Clarified choice_set attribute in tutorial 2.
    Restored django.db.models.F import in final code snippet added at the beginning of tutorial 4.
    ...
    Removed distracting note from tutorial 4.
    ...
    Corrected method/function wording in tutorial 3.
    Added backticks around name argument in tutorial 3.
    ...
```

milinsoft:
@nessita sure! Please take a look!

PR #18066, opened 2024-04-12
Doc'd that PostgreSQL specifies CASCADE when dropping columns.

shangxiao:
Submitting an admonition for a gotcha that a colleague spent some time on – they had some views go missing and were wondering what caused it.

I understand that CASCADE may be based on the fact that PG will abort any ALTER TABLE … DROP COLUMN statement if there are any objects dependent on that column 👍  Though probably not a bad idea to just clarify that here.  (We could also adjust the wording to say this?)

### DIFF
--- a/docs/ref/migration-operations.txt
+++ b/docs/ref/migration-operations.txt
@@ -195,6 +195,16 @@ if the field is nullable or if it has a default value that can be used to
 populate the recreated column. If the field is not nullable and does not have a
 default value, the operation is irreversible.
 
+.. admonition:: PostgreSQL
+
+    ``RemoveField`` will also delete any additional database objects that are
+    related to the removed field (like views, for example). This is because the
+    resulting ``DROP COLUMN`` statement will include the ``CASCADE`` clause to
+    ensure `dependent objects outside the table are also dropped`_.
+
+.. _dependent objects outside the table are also dropped: https://www.postgresql.org/docs/current/sql-altertable.html#SQL-ALTERTABLE-PARMS-CASCADE
+
+
 ``AlterField``
 --------------
 


docs/ref/migration-operations.txt
```diff
@@ -195,6 +195,15 @@ if the field is nullable or if it has a default value that can be used to
 populate the recreated column. If the field is not nullable and does not have a
 default value, the operation is irreversible.
 
+.. admonition:: PostgreSQL
+
+    ``RemoveField`` specifies the ``CASCADE`` suffix to ensure `dependent
+    objects outside the table are also dropped`_. This will affect any
+    additional database objects created like views, etc.
```
nessita, line 202: Hey @shangxiao, thank you for this branch!

Question/comment: when I first read this admonition, I thought the message meant that one could decide whether `CASCADE` is set or not, but after reading docs and seeing code, this is not optional, is a fact. Do you think we could rephrase this to be more explicit about it? What about something like:

```suggestion
    ``RemoveField`` will also delete any additional database objects that are
    related to the field being removed (like views for example). This is
    because the resulting SQL for PostgreSQL will use the ``CASCADE`` modifier
    to ensure `dependent objects outside the table are also dropped`_.
```

shangxiao: > when I first read this admonition, I thought the message meant that one could decide whether CASCADE is set or not

Ah yep when I reread it today I myself saw that I wasn't clear this is something specified in the SQL statement and not a feature users can toggle 🙂

You're rephrasing is good, though I _believe_ "clause" is actually the correct term to use 🤔 (I had "suffix" & you've got "modifier")

shangxiao: We should also clarify "DROP" somewhere like "because the resulting `DROP COLUMN` SQL statement" or something to that effect.

nessita: Pushed a version I'm happy with, if you agree, we can land! :star2:

shangxiao: Thanks so much, looks good!

nessita (approved):

PR #18068, opened 2024-04-12
Fixed #35359 -- Made autodetector add fields referenced by GeneratedField.expression before it.

DevilsAutumn:
# Trac ticket number

[Ticket #35359](https://code.djangoproject.com/ticket/35359)

# Branch description
While adding a `GeneratedField` field along with a new field, the `AddField` operations were being generated in the wrong order due to which it was breaking the migrations. The `AddField` operation for fields referenced by `GeneratedField.expression` should be added before the `GeneratedField` itself.

My solution modifies the `_generate_added_field` method in the autodetector to check if a field is a `GeneratedField`, then extract the referenced fields from `.expression` and recursively call `_generate_added_field` to add `AddField` operation for the fields referenced by `.expression` before `GeneratedField`.

# Checklist

### DIFF
--- a/django/db/migrations/autodetector.py
+++ b/django/db/migrations/autodetector.py
@@ -1126,6 +1126,8 @@ class MigrationAutodetector:
                     self.to_state,
                 )
             )
+        if field.generated:
+            dependencies.extend(self._get_dependencies_for_generated_field(field))
         # You can't just add NOT NULL fields with no default or fields
         # which don't allow empty strings as default.
         time_fields = (models.DateField, models.DateTimeField, models.TimeField)
@@ -1545,6 +1547,27 @@ class MigrationAutodetector:
                     OperationDependency.Type.CREATE,
                 )
             )
+        return dependencies
+
+    def _get_dependencies_for_generated_field(self, field):
+        dependencies = []
+        referenced_base_fields = models.Q(field.expression).referenced_base_fields
+        newly_added_fields = sorted(self.new_field_keys - self.old_field_keys)
+        for app_label, model_name, added_field_name in newly_added_fields:
+            added_field = self.to_state.models[app_label, model_name].get_field(
+                added_field_name
+            )
+            if (
+                added_field.remote_field and added_field.remote_field.model
+            ) or added_field.name in referenced_base_fields:
+                dependencies.append(
+                    OperationDependency(
+                        app_label,
+                        model_name,
+                        added_field.name,
+                        OperationDependency.Type.CREATE,
+                    )
+                )
         return dependencies
 
     def _get_dependencies_for_model(self, app_label, model_name):
--- a/tests/migrations/test_autodetector.py
+++ b/tests/migrations/test_autodetector.py
@@ -13,6 +13,7 @@ from django.db.migrations.graph import MigrationGraph
 from django.db.migrations.loader import MigrationLoader
 from django.db.migrations.questioner import MigrationQuestioner
 from django.db.migrations.state import ModelState, ProjectState
+from django.db.models.functions import Concat, Lower
 from django.test import SimpleTestCase, TestCase, override_settings
 from django.test.utils import isolate_lru_cache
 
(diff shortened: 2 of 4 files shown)


django/db/migrations/autodetector.py
```diff
@@ -1153,6 +1157,25 @@ def _generate_added_field(self, app_label, model_name, field_name):
...
+            f_references = {
+                expr.name for expr in expressions if isinstance(expr, models.F)
+            }
+            q_references = {
+                child[0].split(models.constants.LOOKUP_SEP, 1)[0]
+                for expr in expressions
+                if isinstance(expr, models.Q)
+                for child in expr.children
+                if isinstance(child, tuple)
+            }
```
sarahboyce, line 1172: Nice, this looks to use the [approach](https://github.com/django/django/blob/7029b17315d74ea4a9e253f43485b8d73cd91067/django/db/models/query_utils.py#L179) from https://github.com/django/django/pull/16054/ 👍 

Maybe...
- @shangxiao you can pull out `referenced_base_fields` and it's tests as a separate commit
- I can merge that to main first
- You can both rebase your PRs and use it
- When I backport this, I would backport the commit for `referenced_base_fields` as well

sarahboyce (commented):
Thank you @DevilsAutumn this looks good ⭐ 

Can you add a release note to 5.0.5? This will be included in the upcoming release 😁

tests/migrations/test_autodetector.py
```diff
@@ -1369,6 +1370,37 @@ def test_add_date_fields_with_auto_now_add_asking_for_default(
...
+                ),
+            ],
+        )
+        changes = self.get_changes([initial_state], [updated_state])
+        self.assertNumberMigrations(changes, "testapp", 1)
+        self.assertOperationTypes(changes, "testapp", 0, ["AddField", "AddField"])
+        self.assertOperationFieldAttributes(
+            changes, "testapp", 0, 1, expression=Concat(Lower("name"), Lower("surname"))
+        )
+
```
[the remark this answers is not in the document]
DevilsAutumn: I was trying to think of a case that might fail during foreign keys additions along with a generated field , do you have one in mind? 🤔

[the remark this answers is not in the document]
charettes: @DevilsAutumn 

1. Add a new model (e.g. `Foo`)
2. Add a foreign key to this model (`foo = models.ForeignKey("Foo", models.CASCADE)`)
3. Add a generated field that uses that attribute / column name of the foreign key (`foo_id_str = GeneratedField(Cast("foo_id", models.TextField()), ...)`.

It might already be working but my concerns here are that

1. Foreign keys are usually added last
2. Do we have logic in place so that `foo_id` is correctly mapped to the `foo` field addition by `attname`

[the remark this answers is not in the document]
DevilsAutumn: @charettes Thanks 👍 
Adding a fk along with a generatedfield was failing, I have fixed it and added a test case for the same.

I investigated for the `foo_id` mapping to `foo` as well and it seems to work as expected, Do you think its worth adding a test for this case as well?

django/db/migrations/autodetector.py
shangxiao: ok I can take a look 👍

sarahboyce: On reflection, I think this is not the best plan. Let's keep this as it is, and David when you rebase with these changes later on you can refact this to use `referenced_base_fields`

tests/migrations/test_operations.py
```diff
@@ -1379,6 +1379,54 @@ def test_add_field(self):
         self.assertEqual(definition[1], [])
         self.assertEqual(sorted(definition[2]), ["field", "model_name", "name"])
 
+    @skipUnlessDBFeature("supports_stored_generated_columns")
+    def _test_add_generate_field(self):
```
charettes, line 1383: Have we confirmed that this test even run since it has a `_` prefix? I thought `unittest` only picked up method with a `test` prefix.

tests/migrations/test_autodetector.py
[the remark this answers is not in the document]
charettes: The tests added looks great @DevilsAutumn thanks for that!

django/db/migrations/autodetector.py
shangxiao: Now in PR #18093 👍

sarahboyce: I'm very good at questioning myself. Wasn't sure it is necessary to backport two commits and a (slightly) larger change set 🤔 do you agree with the original approach?

nessita: The original approach makes sense to me because the code that is being added in this PR is the exact same logic as the one factored out by david, and while there are surrounding tests, there are not specific tests for it. I don't see an issue in backporting the two commits as long as the one from David `Refs` this release blocker.

charettes (approved):

nessita (approved):
Looks great! I pushed small edits, will merge and backport when CI is green.

PR #18070, opened 2024-04-12
Fix #8307 -- Saving a model with an ImageField with width_field or height_field no longer results in an extra read operation

john-parton:
# Trac ticket number

ticket-8307

# Branch description

This is an alternate approach from https://github.com/django/django/pull/17775 to solving 8307.

# Checklist

### DIFF
--- a/django/db/models/fields/files.py
+++ b/django/db/models/fields/files.py
@@ -88,10 +88,15 @@ class FieldFile(File, AltersData):
     # to further manipulate the underlying file, as well as update the
     # associated model instance.
 
+    def _set_instance_attribute(self, name, content):
+        # Existence of this method is primarily motivated by custom logic
+        # in ImageFieldFile._set_instance_attribute
+        setattr(self.instance, self.field.attname, name)
+
     def save(self, name, content, save=True):
         name = self.field.generate_filename(self.instance, name)
         self.name = self.storage.save(name, content, max_length=self.field.max_length)
-        setattr(self.instance, self.field.attname, self.name)
+        self._set_instance_attribute(self.name, content)
         self._committed = True
 
         # Save the object because it has changed, unless save is False
@@ -380,6 +385,12 @@ class ImageFileDescriptor(FileDescriptor):
 
 
 class ImageFieldFile(ImageFile, FieldFile):
+    def _set_instance_attribute(self, name, content):
+        setattr(self.instance, self.field.attname, content)
+        # Update the name in case generate_filename() or storage.save() changed
+        # it, but bypass the descriptor to avoid re-reading the file.
+        self.instance.__dict__[self.field.attname] = self.name
+
     def delete(self, save=True):
         # Clear the image dimensions cache
         if hasattr(self, "_dimensions_cache"):
--- a/docs/ref/models/fields.txt
+++ b/docs/ref/models/fields.txt
@@ -1361,6 +1361,13 @@ can change the maximum length using the :attr:`~CharField.max_length` argument.
 The default form widget for this field is a
 :class:`~django.forms.ClearableFileInput`.
 
+.. versionchanged:: 5.1
+
+    In older versions, auto-populating ``height_field`` and ``width_field`` 
+    would always result in re-reading the file from the storage backend. If 
+    your storage backend resizes images, the actual image's width and height 
+    will not match the auto-populated fields.
+
 ``IntegerField``
 ----------------
 
--- a/tests/model_fields/models.py
+++ b/tests/model_fields/models.py
@@ -13,6 +13,8 @@ from django.db.models.functions import Lower
 from django.utils.functional import SimpleLazyObject
 from django.utils.translation import gettext_lazy as _
 
+from .storage import NoReadFileSystemStorage
+
 try:
     from PIL import Image
 except ImportError:
@@ -372,6 +374,23 @@ if Image:
             height_field="headshot_height",
             width_field="headshot_width",
         )
+
+    class PersonNoReadImage(models.Model):
+        """
+        Model that defines an ImageField with a storage backend that does not
+        support reading.
+
+        This supports a test to avoid a double-read performance problem.
+        """
+
+        mugshot = models.ImageField(
+            upload_to="tests",
+            storage=NoReadFileSystemStorage(),
+            width_field="mugshot_width",
+            height_field="mugshot_height",
+        )
+        mugshot_width = models.IntegerField()
+        mugshot_height = models.IntegerField()
 
 
 class CustomJSONDecoder(json.JSONDecoder):
--- a/tests/model_fields/storage.py
+++ b/tests/model_fields/storage.py
@@ -1,0 +1,6 @@
+from django.core.files.storage.filesystem import FileSystemStorage
+
+
+class NoReadFileSystemStorage(FileSystemStorage):
+    def open(self, *args, **kwargs):
+        raise AssertionError("This storage class does not support reading.")
--- a/tests/model_fields/test_imagefield.py
+++ b/tests/model_fields/test_imagefield.py
@@ -18,6 +18,7 @@ if Image:
     from .models import (
         Person,
         PersonDimensionsFirst,
+        PersonNoReadImage,
         PersonTwoImages,
         PersonWithHeight,
         PersonWithHeightAndWidth,
@@ -30,7 +31,7 @@ else:
         pass
 
     PersonWithHeight = PersonWithHeightAndWidth = PersonDimensionsFirst = Person
-    PersonTwoImages = Person
+    PersonTwoImages = PersonNoReadImage = Person
 
 
 class ImageFieldTestMixin(SerializeMixin):
@@ -469,3 +470,28 @@ class TwoImageFieldTests(ImageFieldTestMixin, TestCase):
         # Dimensions were recalculated, and hence file should have opened.
         self.assertIs(p.mugshot.was_opened, True)
         self.assertIs(p.headshot.was_opened, True)
+
+
+@skipIf(Image is None, "Pillow is required to test ImageField")
+class NoReadTests(ImageFieldTestMixin, TestCase):
+    def test_width_height_correct_name_mangling_correct(self):
+        instance1 = PersonNoReadImage()
+
+        instance1.mugshot.save("mug", self.file1)
+
+        self.assertEqual(instance1.mugshot_width, 4)
+        self.assertEqual(instance1.mugshot_height, 8)
+
+        instance1.save()
+
+        self.assertEqual(instance1.mugshot_width, 4)
+        self.assertEqual(instance1.mugshot_height, 8)
+
+        instance2 = PersonNoReadImage()
+        instance2.mugshot.save("mug", self.file1)
+        instance2.save()
+
+        self.assertNotEqual(instance1.mugshot.name, instance2.mugshot.name)
+
+        self.assertEqual(instance1.mugshot_width, instance2.mugshot_width)
+        self.assertEqual(instance1.mugshot_height, instance2.mugshot_height)
--- a/docs/releases/5.1.txt
+++ b/docs/releases/5.1.txt
@@ -419,6 +419,11 @@ Miscellaneous
 * The undocumented ``django.urls.converters.get_converter()`` function is
   removed.
 
+* :class:`~django.db.models.ImageField` no longer re-reads the file from the
+  storage backend after ``width_field`` and ``height_field`` are set. If
+  your storage backend resizes images, the actual image's width and height will
+  not match the auto-populated fields.
+
 * The minimum supported version of SQLite is increased from 3.27.0 to 3.31.0.
 
 .. _deprecated-features-5.1:


john-parton:
Updated docs here.

The reason I moved the logic to the descriptor, is because in order to get the behavior correct, the logic needs both the value _before_ and _after_ the value is set. A previous pull request I had open instead modified the ImageField.update_image_dimensions method, requiring that the initial file be passed in as an additional kwarg. It also required an additional container class to tell the descriptor when to pass the original file.

docs/ref/models/fields.txt
```diff
@@ -1361,6 +1361,14 @@ can change the maximum length using the :attr:`~CharField.max_length` argument.
 The default form widget for this field is a
 :class:`~django.forms.ClearableFileInput`.
```
sarahboyce, line 1363: We also need a release note in 5.1, something like:
```
* :class:`~django.db.models.ImageField` no longer re-reads the file from the
  storage backend after ``width_field`` and ``height_field`` are set. If
  your storage backend resizes images, the actual image's width and height will
  not match the auto-populated fields.
```
I think within Backwards incompatible changes in 5.1 and in Miscellaneous maybe

tests/model_fields/test_imagefield.py
```diff
@@ -469,3 +470,32 @@ def test_dimensions(self):
         # Dimensions were recalculated, and hence file should have opened.
         self.assertIs(p.mugshot.was_opened, True)
         self.assertIs(p.headshot.was_opened, True)
+
+
+@skipIf(Image is None, "Pillow is required to test ImageField")
+class NoReadTests(ImageFieldTestMixin, TestCase):
+
+    NoReadImageModel = NoReadImageModel
```
sarahboyce, line 478: If pillow is not installed we get an error: `NameError: name 'NoReadImageModel' is not defined`
At the top of the file there is handling for this, see...
```
# Pillow not available, create dummy classes (tests will be skipped anyway)
```

Can you do the same?

django/db/models/fields/files.py
```diff
@@ -91,7 +91,16 @@ def open(self, mode="rb"):
...
-        setattr(self.instance, self.field.attname, self.name)
+
+        # The presumption is that IO for content has already been dealt with
+        setattr(self.instance, self.field.attname, content)
+
+        # Update the name if generate_filename or the storage backend changed
+        # the name, side-stepping the descriptor to avoid re-running logic.
+        # Part of the fix for #8307
+        if self.name != getattr(self.instance, self.field.attname).name:
+            self.instance.__dict__[self.field.attname] = self.name
```
sarahboyce, line 102:
```suggestion
        # Update the name in case generate_filename or the storage backend 
        # changed the name. Avoids re-running the descriptor logic.
        self.instance.__dict__[self.field.attname] = self.name
```
🤔 maybe, the test still passes for me and I don't see the harm of always setting it similar to before

django/db/models/fields/files.py
```diff
@@ -374,12 +383,20 @@ def __set__(self, instance, value):
...
         if previous_file is not None:
-            self.field.update_dimension_fields(instance, force=True)
+
+            # If the original value passed in was a string, it's usually necessary
+            # to convert it to a file object, here done using the descriptor's
+            # __get__ method.
+            if isinstance(value, str):
+                value = getattr(instance, self.field.attname)
+
+            self.field._set_dimensions(instance, value)
```
sarahboyce, line 395: I can revert this and the tests still pass

django/db/models/fields/files.py
```diff
@@ -374,12 +383,20 @@ def __set__(self, instance, value):
...
 
 
 class ImageFieldFile(ImageFile, FieldFile):
+
```
sarahboyce, line 399:
```suggestion
class ImageFieldFile(ImageFile, FieldFile):
```
Avoid adding formatting changes (we try to keep the PRs as small as possible)

sarahboyce (commented):
Hi @john-parton thank you for this ⭐ I have initial comments

- Close the other PR if this is the preferred approach?
- I wanted to start reviewing this PR as I know you were waiting for some feedback, can you give an overview on how these two relate https://code.djangoproject.com/ticket/35384?

There are some small stylistic things that we need to update [about our Python style](https://docs.djangoproject.com/en/5.0/internals/contributing/writing-code/coding-style/#python-style). Including ending comments with full stops etc.

john-parton:
> Hi @john-parton thank you for this ⭐ I have initial comments
> 
>     * Close the other PR if this is the preferred approach?

Done

>     * I wanted to start reviewing this PR as I know you were waiting for some feedback, can you give an overview on how these two relate https://code.djangoproject.com/ticket/35384?

https://code.djangoproject.com/ticket/35384 is unrelated technically. It's just another issue I found while digging through the code. This PR doesn't close that ticket.

> There are some small stylistic things that we need to update [about our Python style](https://docs.djangoproject.com/en/5.0/internals/contributing/writing-code/coding-style/#python-style). Including ending comments with full stops etc.

Understood

django/db/models/fields/files.py
john-parton: I'm not exactly sure what you want me to do with this. This is a very tricky code path. What would a test look like?

django/db/models/fields/files.py
john-parton: I think I have to change it, otherwise CI fails due to black?

docs/ref/models/fields.txt
```diff
@@ -1361,6 +1361,14 @@ can change the maximum length using the :attr:`~CharField.max_length` argument.
...
 
+.. versionchanged:: 5.1
+
+    Prior to version 5.1, auto-populating height_field and width_field would always
+    result in re-reading the file from the storage backend. If your storage backend
+    resizes images, the actual image's width and height will not match the 
+    auto-populated fields.
+
```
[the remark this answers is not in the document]
john-parton: I committed the suggested change.

docs/ref/models/fields.txt
john-parton: I believe I have resolved this.

tests/model_fields/test_imagefield.py
john-parton: I believe I have resolved this.

django/db/models/fields/files.py
john-parton: I went ahead and committed the suggested change.

django/db/models/fields/files.py
john-parton: I think I got this. File is formatted correctly now.

django/db/models/fields/files.py
sarahboyce: I'm trying to establish the smallest changeset to achieve the desired result, which we will likely squash into a single commit 🤔 
As you have a test checking the extra read isn't happening, do we need these changes?

john-parton: Yes, the extra read is covered by `NoReadTests`. If you port the `NoReadTests` back to main/trunk, you'll see that they fail.

john-parton: If you go look at the signature of `_set_dimensions` you'll see that the required type of `value` is a file-like object or `None`. That's what is has to be coerced here.

sarahboyce: Ah I'm not sure if I was clear, I think we can revert this specific bit to what it was before 
```python
self.field.update_dimension_fields(instance, force=True)
```
I see you have updated `update_dimension_fields` to use `_set_dimensions` and when I do this change, the test still passes

django/db/models/fields/files.py
```diff
@@ -91,7 +91,14 @@ def open(self, mode="rb"):
...
         name = self.field.generate_filename(self.instance, name)
         self.name = self.storage.save(name, content, max_length=self.field.max_length)
-        setattr(self.instance, self.field.attname, self.name)
+
+        # The presumption is that IO for content has already been dealt with.
+        setattr(self.instance, self.field.attname, content)
+
+        # Update the name in case generate_filename or the storage backend
+        # changed the name. Avoids re-running the descriptor logic.
+        self.instance.__dict__[self.field.attname] = self.name
```
sarahboyce, line 100: This appears to be the core of the fix and all other changes in the file can be reverted and the test still passes. 🤔 
Either we need more tests to show the refactoring is needed or we should revert the other changes

john-parton: Lol. Maybe.  I only arrived at that after rewriting about a five times though.

sarahboyce: Yes! John you've put loads of work into this and helped improve the understanding of an area not been touched in years ⭐️
Changes that end up being a few lines are never easy and take hours and hours to come to. Thank you for the work you've invested in this

john-parton:
@sarahboyce I went ahead and removed all of the changes other than the "core" part of the fix you identified. It seems so obvious now, but it definitely isn't. I adjusted the comment slightly to help guide someone reading that part of the code.

I also removed all of the changes to other comments, so it's just those specific lines.

In referring to the documentation change I made:

 `Name of a model field which will be auto-populated with the height of the image each time the image attribute is set.` 

This is not because the behavior changed. This has been the behavior forever, but I wanted it to be clear that the height and width fields do not get updated when the model is saved if the image field isn't changed. That might seem obvious, but if you have a cron job that scans all your images and resizes them, then they won't get the updated sizes.

tests/model_fields/models.py
```diff
@@ -373,6 +375,22 @@ class PersonTwoImages(models.Model):
...
+        This supports a test to avoid a double-read performance problem."""
+
+        testimage = models.ImageField(
+            upload_to="tests",
+            storage=NoReadFileSystemStorage(),
+            width_field="width",
+            height_field="height",
+        )
+        width = models.IntegerField()
+        height = models.IntegerField()
```
sarahboyce, line 392: Optional - we could also make a Person-y thing if we want
```suggestion
    class PersonNoReadImage(models.Model):
        """
        Model that defines an ImageField with a storage backend that does not
        support reading.

        This supports a test to avoid a double-read performance problem."""

        mugshot = models.ImageField(
            upload_to="tests",
            storage=NoReadFileSystemStorage(),
            width_field="width",
            height_field="height",
        )
        mugshot_width = models.IntegerField()
        mugshot_height = models.IntegerField()
```

tests/model_fields/storage.py
```diff
@@ -0,0 +1,17 @@
...
+class NoReadStorageMixin:
+    def open(self, *args, **kwargs):
+        raise NoReadException("This storage class does not support reading.")
+
+
+class NoReadFileSystemStorage(NoReadStorageMixin, FileSystemStorage):
+    """A storage backend which does not support reading.
+
+    Use this to test the behavior of Django when a storage backend is not
+    supposed to do any additional reads."""
```
sarahboyce, line 17:
```suggestion
class NoReadFileSystemStorage(FileSystemStorage):
    def open(self, *args, **kwargs):
        raise NoReadException("This storage class does not support reading.")
```
I think the exception is self documenting and we don't reuse the mixin.

tests/model_fields/test_imagefield.py
```diff
@@ -32,6 +33,9 @@ class Person:
     PersonWithHeight = PersonWithHeightAndWidth = PersonDimensionsFirst = Person
     PersonTwoImages = Person
 
+    class NoReadImageModel:
+        pass
+
```
sarahboyce, line 38:
```suggestion
    PersonTwoImages = NoReadImageModel = Person
```
I think we can do something like this

docs/ref/models/fields.txt
```diff
@@ -1361,6 +1361,14 @@ can change the maximum length using the :attr:`~CharField.max_length` argument.
...
 :class:`~django.forms.ClearableFileInput`.
 
+.. versionchanged:: 5.1
+
+    In older versions, auto-populating ``height_field`` and ``width_field`` 
+    would always result in re-reading the file from the storage backend. If 
+    your storage backend resizes images, the actual image's width and height 
+    will not match the auto-populated fields.
+
+
```
sarahboyce, line 1371:
```suggestion

```
(only 1 blank line needed)

sarahboyce (commented):
> @sarahboyce I went ahead and removed all of the changes other than the "core" part of the fix you identified. It seems so obvious now, but it definitely isn't. I adjusted the comment slightly to help guide someone reading that part of the code.
> I also removed all of the changes to other comments, so it's just those specific lines.

Thank you so much this looks great 🥳 
I have a few minor comments for us to add the final polish but it's looking really good

> This is not because the behavior changed. This has been the behavior forever, but I wanted it to be clear that the height and width fields do not get updated when the model is saved if the image field isn't changed. That might seem obvious, but if you have a cron job that scans all your images and resizes them, then they won't get the updated sizes.

Ok thank you for highlighting. I think, let's pull those couple of lines into a separate PR. The reason being is sometimes document clarifications get back ported (to 5.0 for example) and if this is unrelated to your change, let's pull it out and deal with it separately.

After that, can you squash all commits into the one commit that has a message something like:
```
Fixed #8307 -- Prevented a read operation when saving ImageFields with width_field and height_field.
```

I will also get a second opinion on the release notes, docs and the decision not to deprecate.

john-parton:
I moved the doc change which wasn't directly related to the bugfix here: https://github.com/django/django/pull/18112

I accepted all of your suggestions, and then rewrote the tests to use the model you suggested. Fixed linter errors, rebased and force-pushed with the commit message you suggested.

sarahboyce:
Thank you John 

> I will also get a second opinion on the release notes, docs and the decision not to deprecate.

I have [started a discussion here to get feedback on whether we need to deprecate](https://forum.djangoproject.com/t/to-deprecate-or-not-to-deprecate-proposed-behaviour-change-when-saving-an-imagefield/).
For context, I am still quite new to the Fellow role so apologise I'm a little cautious/unsure about how best to proceed.

The PR looks great! If we do need to add a deprecation path, it shouldn't be too much work 👍

tests/model_fields/storage.py
```diff
@@ -0,0 +1,10 @@
+from django.core.files.storage.filesystem import FileSystemStorage
+
+
+class NoReadException(Exception):
+    pass
+
+
+class NoReadFileSystemStorage(FileSystemStorage):
+    def open(self, *args, **kwargs):
+        raise NoReadException("This storage class does not support reading.")
```
adamchainz, line 10: It’s for tests, we can use `AssertionError`, no need for the extra noise creating a custom exception class.

```suggestion
class NoReadFileSystemStorage(FileSystemStorage):
    def open(self, *args, **kwargs):
        raise AssertionError("This storage class does not support reading.")
```

tests/model_fields/test_imagefield.py
```diff
@@ -469,3 +470,30 @@ def test_dimensions(self):
         # Dimensions were recalculated, and hence file should have opened.
         self.assertIs(p.mugshot.was_opened, True)
         self.assertIs(p.headshot.was_opened, True)
+
+
+@skipIf(Image is None, "Pillow is required to test ImageField")
+class NoReadTests(ImageFieldTestMixin, TestCase):
+    PersonNoReadImage = PersonNoReadImage
+
```
adamchainz, line 478: Afaict, there’s no need to copy this pattern from the above tests, because there are no inherited tests to run. Remove it and the `self.` for the class in the tests.

```suggestion
```

django/db/models/fields/files.py
```diff
@@ -91,7 +91,14 @@ def open(self, mode="rb"):
...
         self.name = self.storage.save(name, content, max_length=self.field.max_length)
-        setattr(self.instance, self.field.attname, self.name)
+
+        # Set the file-like content using the descriptors __set__ method.
+        setattr(self.instance, self.field.attname, content)
+
+        # Update the name in case generate_filename or the storage backend
+        # changed the name. Avoids re-running the descriptor logic.
+        self.instance.__dict__[self.field.attname] = self.name
+
```
adamchainz, line 101: It’s great to comment this, but I think the current content slightly overexplains. I propose this:

```suggestion
        setattr(self.instance, self.field.attname, content)
        # Update the name in case generate_filename() or storage.save() changed
        # it, but bypass the descriptor to avoid re-reading the file.
        self.instance.__dict__[self.field.attname] = self.name
```

adamchainz (commented):
Great work on this old ticket!! 

I have some small suggestions.

john-parton:
- Merged in suggested changes
- tweaked tests (as requested) so they continue to pass

Looks like based on our discussion in [the forum](https://forum.djangoproject.com/t/to-deprecate-or-not-to-deprecate-proposed-behaviour-change-when-saving-an-imagefield/30640/12) that most likely we'll want an example of how to restore the old behavior. Again, I know I'm a broken record: I can't imagine why you would WANT to do that, but at least we'll have our bases covered.

john-parton:
Force pushed because CI failed (error in `main` or a bad merge? Not sure.)

sarahboyce:
> Force pushed because CI failed (error in `main` or a bad merge? Not sure.)

No problem, agreed unrelated to your changes 👍

john-parton:
Mostly a note for myself: I mentioned adding docs somewhere to guide users on how to "force" django to re-read the image after it's been persisted to update the width and height fields. Still need to do that.

adamchainz:
Please squash commits and retitle `Fixed #8307 -- Prevented re-read when saving an ImageField.` to follow the [commit guidelines](https://docs.djangoproject.com/en/dev/internals/contributing/committing-code/#committing-guidelines).

john-parton:
Rebased, squashed, changed commit message.

In the future, should I go ahead and squash every time I patch? I usually leave the extra commits on in between reviews so that my individual changes can be seen.

adamchainz:
I always squash every time I change, I think it’s what’s general done on GitHub. Otherwise the merger has to do it for you. The diff can be seen on GitHub by clicking the “Compare” link next to the message “John Parton force-pushed the bugfix/8307-anti-pattern-removal branch from 6dea56e to 9c4a122”.

john-parton:
That's good info. Thanks. I'll make sure to be more religious with my squashing.

sarahboyce:
@john-parton just to let you know the 5.1 feature freeze is planned for [May 22nd](https://code.djangoproject.com/wiki/Version5.1Roadmap)

> I mentioned adding docs somewhere to guide users on how to "force" django to re-read the image after it's been persisted to update the width and height fields. Still need to do that.

I discussed this with @nessita, another option is to have this as a comment in the ticket. It doesn't have to be in the docs/release notes but having this publicly available (so in Trac) would let us point other users to it if they have an issue.
If it is actually quite hard to revert this behaviour, we can reconsider the setting that gets immediately deprecated. This gives users time to refactor accordingly.

If we miss 5.1, this will go in 5.2. There is 0 rush but wanted to make sure you were made aware 👍

john-parton:
Perhaps the most straightforward way to force the old behavior at an instance level is to do this:

```python
instance = Model()
instance.image = ...
instance.save()  # Image is saved to the storage backend, width/height are set using the image before it's committed
instance.image = instance.image  # To a python newbie, this is just nonsense, but this runs the descriptor logic
instance.save(update_fields=["width", "height"])
```

Another way

```python
instance = Model()
instance.image = ...
instance.save()
Model.image.update_dimension_fields(instance, force=True)  # This is what the descriptor calls internally
instance.save(update_fields=["width", "height"])
```

If you want to force the behavior automatically at a model level, you can patch its `save()` method with either of those techniques.

```python
class Model(models.Model):
    ....
    def save(
        self, force_insert=False, force_update=False, using=None, update_fields=None
    ):
        super().save(
            force_insert=force_insert,
            force_update=force_update,
            using=using,
            update_fields=update_fields,
        )

        if update_fields is None or "image" in update_fields:
            self.image = self.image
            super().save(using=using, update_fields=["width", "height"])
```

A `post_save` signal would also work, perhaps it could be made generically enough to apply to all model saves.

django/db/models/fields/files.py
```diff
@@ -91,7 +91,10 @@ def open(self, mode="rb"):
     def save(self, name, content, save=True):
         name = self.field.generate_filename(self.instance, name)
         self.name = self.storage.save(name, content, max_length=self.field.max_length)
-        setattr(self.instance, self.field.attname, self.name)
+        setattr(self.instance, self.field.attname, content)
+        # Update the name in case generate_filename() or storage.save() changed
+        # it, but bypass the descriptor to avoid re-reading the file.
+        self.instance.__dict__[self.field.attname] = self.name
```
sarahboyce, line 97: Right now we have updated `save` in `FieldFile` due to an issue in `ImageField` due to the complexity of `ImageFileDescriptor` and `ImageField`. Then we documented the potential impact to `ImageField` but actually all FieldFields might be affected. So I think we should isolate the solution to the problem being solved.

Easy option is to overwrite the `save`  of `ImageFieldFile`.
Later I would love us to simplify `update_dimension_fields` which I think is a bit of a mess

john-parton: Presumably the concern is that someone out there has subclassed FileField and relies on the descriptor getting called in this specific case?

john-parton: Let's supposed you have a Model with a FileField, not an image field

```python3
class Model(models.Model):
    file = models.FileField(...)
```

And you do the following:

```python
instance = Model()
instance.file.save("myfile", ContentFile(b"foo"))
instance.file.read()
```

BEFORE my change this would result in the file being read off the storage backend. AFTER this would result in directly reading the value from `ContentFile`.

So you're right that there's a change in behavior for non-image fields, but I'm wondering if that's the more "correct" or "desirable" behavior in the general case.

john-parton: I just deleted a comment because I'm not sure it's right, and need to double check some things.

I'm wondering what happens if you call `read()` on a filefield after saving it: does it hit the storage backend again or not?

john-parton: OK, so with this patch, reading a file immediately after saving it will involve hitting the storage backend, even if the file _could_ be cached in memory

Example code:


```python
class Model(models.Model):
    file = models.FileField(...)

instance = Model()
instance.file = ContentFile(b"foo", name="bar.txt")
instance.file.read() == b"foo"  # No read
instance.save()
# instance.__dict__["file"] is now an instance of str() and the underlying ContentFile is dicsarded
instance.file.read() == b"foo"  # Will always result in a read to the storage backend
```

john-parton: If we guard the code like this


```python3
  setattr(self.instance, self.field.attname, content)
  # Update the name in case generate_filename() or storage.save() changed
  # it, but bypass the descriptor to avoid re-reading the file.
  # Only update in the case that it's strictly necessary
  if getattr(self.instance, self.field.attname).name != self.name:
      self.instance.__dict__[self.field.attname] = self.name
 ```
 
 Now, the flow control is more complicated, and there's two different scenarios
 
 ```python
instance = Model()
instance.file = ContentFile(b"foo", name="bar.txt")
instance.file.read() == b"foo"  # No hit to the storage backend
instance.save()
# If the storage backend is "cool" with "bar.txt" as a filename then
# instance.__dict__["file"] is not an instance of str(), but rather a file-like object
instance.file.read() == b"foo"  # Won't necessarily hit the storage backend
```

However, if the storage backend changes the filename for any reason (either the file already exists, it normalized whitespace, literally any reason), then it will result in the final read() hitting the storage backend.

john-parton: I've got something like this

```python3
    def _set_instance_attribute(self, name, content):
        # Existence of this method is primarily motivated by custom logic
        # in ImageFieldFile._set_instance_attribute
        setattr(self.instance, self.field.attname, name)

    def save(self, name, content, save=True):
        name = self.field.generate_filename(self.instance, name)
        self.name = self.storage.save(name, content, max_length=self.field.max_length)
        self._set_instance_attribute(self.name, content)
        self._committed = True

        # Save the object because it has changed, unless save is False
        if save:
            self.instance.save()
  ```
  
  And then in `ImageFieldFile`
  
  
  ```python3
      def _set_instance_attribute(self, name, content):
        setattr(self.instance, self.field.attname, content)
        # Update the name in case generate_filename() or storage.save() changed
        # it, but bypass the descriptor to avoid re-reading the file.
        self.instance.__dict__[self.field.attname] = self.name
   ```
   
   That way I don't have to just wholesale copy/paste the save method when overwriting it

sarahboyce: Nice idea 💡

john-parton: @sarahboyce Updated and force-pushed

sarahboyce (approved):
Thank you for this!
I have made very tiny edits and will merge in this PR: https://github.com/django/django/pull/18184

sarahboyce:
Moved to https://github.com/django/django/pull/18184

PR #18072, opened 2024-04-13
Refs #31405 -- Moved redirect_to_login to utils.py.

Hisham-Pak:
# Trac ticket number

ticket-31405

# Branch description
This pull request moves `redirect_to_login` to a new file `utils.py` from `views.py` in order to avoid circular import instead of importing locally in a function.
https://github.com/django/django/pull/17792#discussion_r1549554865

# Checklist

### DIFF
--- a/django/contrib/auth/views.py
+++ b/django/contrib/auth/views.py
@@ -1,5 +1,3 @@
-from urllib.parse import urlparse, urlunparse
-
 from django.conf import settings
 
 # Avoid shadowing the login() and logout() views below.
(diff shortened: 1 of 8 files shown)


django/contrib/auth/views.py
```diff
@@ -15,9 +13,12 @@
     SetPasswordForm,
 )
 from django.contrib.auth.tokens import default_token_generator
+
+# For backwards compatibility with Django < 5.1
+from django.contrib.auth.utils import redirect_to_login  # NOQA: F401
```
sarahboyce, line 18: I'm wondering if we have to do this, if so we should have a deprecation message.
Looking at this PR which is a little similar (https://github.com/django/django/pull/17050) looks like we don't necessarily need to. 🤔

sarahboyce (commented):
This also needs a release note in 5.1, see [this PR](https://github.com/django/django/pull/17050) as an example of a similar change 👍

sarahboyce:
@Hisham-Pak I've had a discussion with some other contributors on Discord and, as this refactor requires a deprecation without a meaningful gain to a user, we've decided it doesn't make sense.

I appreciate you did this after my suggestion and thank you for doing so.
It takes time to build a picture of what makes the most sense, so please be patient with me 🌟

PR #18073, opened 2024-04-13
Added check_key in RedisCache incr

cyx2000:
# Trac ticket number
N/A

# Branch description
In redis, incr can don't need check key, so add check_key to this command. The defauult value is true

# Checklist
- [y] This PR targets the `main` branch. 
- [y] The commit message is written in past tense, mentions the ticket number, and ends with a period.
- [y] I have added or updated relevant **tests**.
- [y] I have added or updated relevant **docs**, including release notes if applicable.
- [y] For UI changes, I have attached **screenshots** in both light and dark modes.

### DIFF
--- a/django/core/cache/backends/redis.py
+++ b/django/core/cache/backends/redis.py
@@ -129,9 +129,9 @@ class RedisCacheClient:
         client = self.get_client(key)
         return bool(client.exists(key))
 
-    def incr(self, key, delta):
+    def incr(self, key, delta, check_key: bool):
         client = self.get_client(key, write=True)
-        if not client.exists(key):
+        if check_key and not client.exists(key):
             raise ValueError("Key '%s' not found." % key)
         return client.incr(key, delta)
 
@@ -209,9 +209,9 @@ class RedisCache(BaseCache):
         key = self.make_and_validate_key(key, version=version)
         return self._cache.has_key(key)
 
-    def incr(self, key, delta=1, version=None):
+    def incr(self, key, delta=1, version=None, check_key=True):
         key = self.make_and_validate_key(key, version=version)
-        return self._cache.incr(key, delta)
+        return self._cache.incr(key, delta, check_key)
 
     def set_many(self, data, timeout=DEFAULT_TIMEOUT, version=None):
         if not data:
--- a/tests/cache/tests.py
+++ b/tests/cache/tests.py
@@ -1803,14 +1803,6 @@ class RedisCacheTests(BaseCacheTests, TestCase):
     def incr_decr_type_error(self):
         return self.lib.ResponseError
 
-    def test_incr_write_connection(self):
-        cache.set("number", 42)
-        with mock.patch(
-            "django.core.cache.backends.redis.RedisCacheClient.get_client"
-        ) as mocked_get_client:
-            cache.incr("number")
-            self.assertEqual(mocked_get_client.call_args.kwargs, {"write": True})
-
     def test_incr_check_key(self):
         with self.assertRaises(ValueError):
             cache.incr("check_key")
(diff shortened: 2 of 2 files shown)


charettes (commented):
I think that the fundamental problem here is that `cache.incr` / `decr` [are documented to raise an exception if they key is missing](https://docs.djangoproject.com/en/5.0/topics/cache/#django.core.cache.cache.decr) which I believe is a bit strange in retrospective given how often this method is used in the wild.

It means that practically every call to it must be paired with a either a systematic `add`

```python
if not cache.add(key, delta):
    cache.incr(key, delta)
```

Or be performed as a fallback

```python
try:
    cache.incr(key, delta)
except ValueError:
    if not cache.add(key, delta):
        cache.incr(key, delta)
```

Add to that the underlying `exists` call performed by the Redis backend and it means that it order to use `add` effectively there is no way around doing

```
SET key delta NX
--- Do the following in most cases
EXISTS key
INCRBY key delta
```

or

```
EXISTS key
INCRBY key delta
--- Do the following in fallback cases
SET key delta NX
EXISTS key
INCRBY key delta
```

That's up to 3-5 operations which are non-pipelined (meaning they'll each need to roundtrip over the network serially) and non-atomic. Things is this could all be done in a single `INCRBY key delta` assuming `incr` and `decr` supported a way to specify a default value (which I assume `0` would be the case most of the time).

I focused on Redis here but from eye balling all the other existing backend it seems it wouldn't be hard to adapt them to augment the interface to accept a kwarg (alike to the `check_key` proposed here) so it simply assigns `delta` instead of erroring out when a key is missing. This seems more valuable and likely to be accepted that hacking a solution solely for the Redis backend.

I suggest that you raise this point on the forum @cyx2000 [as documented for new features proposal of this type](https://docs.djangoproject.com/en/5.0/internals/contributing/bugs-and-features/#requesting-features) to gather feedback.

I'll close this MR in the meantime as it's not in a receivable form due to the lack of consensus on the proper solution.

cyx2000:
> I think that the fundamental problem here is that `cache.incr` / `decr` [are documented to raise an exception if they key is missing](https://docs.djangoproject.com/en/5.0/topics/cache/#django.core.cache.cache.decr) which I believe is a bit strange in retrospective given how often this method is used in the wild to perform moving window throttling.
> 
> It means that practically every call to it must be paired with a either a systematic `add`
> 
> ```python
> if not cache.add(key, delta):
>     cache.incr(key, delta)
> ```
> 
> Or be performed as a fallback
> 
> ```python
> try:
>     cache.incr(key, delta)
> except ValueError:
>     if not cache.add(key, delta):
>         cache.incr(key, delta)
> ```
> 
> Add to that the underlying `exists` call performed by the Redis backend and it means that it order to use `incr` properly there is no way around doing
> 
> ```
> SET key delta NX
> --- Do the following in most cases
> EXISTS key
> INCRBY key delta
> ```
> 
> or
> 
> ```
> EXISTS key
> INCRBY key delta
> --- Do the following in fallback cases
> SET key delta NX
> EXISTS key
> INCRBY key delta
> ```
> 
> That's up to 3-5 operations which are non-pipelined (meaning they'll each need to roundtrip over the network serially) and non-atomic. Things is this could all be done in a single `INCRBY key delta` assuming `incr` and `decr` supported a way to specify a fallback to pre-existing value (which I assume `0` would be the case most of the time anyway).
> 
> I focused on Redis here but from eye balling all the other existing backend it seems it wouldn't be hard to adapt them to augment the interface to accept a kwarg (alike to the `check_key` proposed here) so it simply assigns `delta` instead of erroring out when a key is missing. This seems more valuable and likely to be accepted that hacking a solution solely for the Redis backend.
> 
> I suggest that you raise this point on the forum @cyx2000 [as documented for new features proposal of this type](https://docs.djangoproject.com/en/5.0/internals/contributing/bugs-and-features/#requesting-features) to gather feedback.
> 
> I'll close this MR in the meantime as it's not in a receivable form due to the lack of consensus on the proper solution.

Thanks for your reply and review.

PR #18077, opened 2024-04-15
Fixed #35375 -- Fixed tabular inline admin original line position to use `inset-inline-start`. (solution 2)

eyalch:
# Trac ticket number

ticket-35375

# Branch description

As mentioned in the issue, this changes the way the original line of tabular inlines is positioned so that it's correctly positioned on the right in RTL languages _without_ any special/additional styles specific for RTL.

Please see #18076 as this solution is an alternative.

# Checklist

### DIFF
--- a/django/contrib/admin/static/admin/css/forms.css
+++ b/django/contrib/admin/static/admin/css/forms.css
@@ -455,7 +455,7 @@ body.popup .submit-row {
 
 .inline-group .tabular td.original p {
     position: absolute;
-    left: 0;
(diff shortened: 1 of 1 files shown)


sarahboyce:
Just been informed that there is a wider plan to use logical properties to improve our RTL support: https://github.com/django/django/pull/17560
In light of that information, I will close this and we will adopt option 1 while the refactor and discussion on logical properties is ongoing 👍 
Thank you for this

PR #18078, opened 2024-04-15
Fixed TypeError comparing datetime.datetime to datetime.date.

jsantos42:
# Trac ticket number
N/A

# Branch description
On part 2 of the tutorial, it is suggested to add a `was_published_recently` custom method. However, this method throws a type error when called, since it's comparing two different types. By converting the `now()` instance to a Date, it fixes the problem.

# Checklist

### DIFF
--- a/docs/intro/tutorial02.txt
+++ b/docs/intro/tutorial02.txt
@@ -463,7 +463,7 @@ Let's also add a custom method to this model:
     class Question(models.Model):
         # ...
         def was_published_recently(self):
-            return self.pub_date >= timezone.now() - datetime.timedelta(days=1)
+            return self.pub_date >= timezone.now().date() - datetime.timedelta(days=1)
 
 Note the addition of ``import datetime`` and ``from django.utils import
 timezone``, to reference Python's standard :mod:`datetime` module and Django's


docs/intro/tutorial02.txt
```diff
@@ -463,7 +463,7 @@ Let's also add a custom method to this model:
     class Question(models.Model):
         # ...
         def was_published_recently(self):
-            return self.pub_date >= timezone.now() - datetime.timedelta(days=1)
+            return self.pub_date >= timezone.now().date() - datetime.timedelta(days=1)
```
adamzap, line 466: It seems to me that your change introduces a `TypeError` and that the tutorial is fine as-is:

1. `pub_date` is a `DateTimeField` which is represented as a `datetime.datetime` in Python
2. `timezone.now()` returns a `datetime.datetime`
3. A `datetime.timedelta` can be subtracted from a `datetime.datetime` instance

This is what I'm getting:

```python
>>> import datetime
>>> from django.utils import timezone
>>>
>>> datetime.datetime.now() >= timezone.now() - datetime.timedelta(days=1)
True
>>> datetime.datetime.now() >= timezone.now().date() - datetime.timedelta(days=1)
  File "<console>", line 1, in <module>
TypeError: can't compare datetime.datetime to datetime.date
```

Are you seeing something different?

jsantos42: That's funny, I also have an error on the first one...
Screenshot 2024-04-15 at 17 37 45

Since we're at it, here's the output I have on the console. I get true with the `.date()`, but the error otherwise.
Screenshot 2024-04-15 at 17 41 26

adamzap: I have `USE_TZ` set to `False` for simplicity. I think that's why you got the first error.

I'll run through the tutorial soon and compare.

adamzap: I can't reproduce your error from the tutorial instructions. What does `print(foo.pub_date)` give you?

jsantos42: I don't have `USE_TZ` defined.

Screenshot 2024-04-15 at 18 25 50

jsantos42: I followed the tutorial by the book, IIRC.

Here's my models.py file:
```
import datetime

from django.db import models
from django.utils import timezone


class Question(models.Model):
    question_text = models.CharField(max_length=200)
    pub_date = models.DateField("date published")

    def __str__(self):
        return self.question_text

    def was_published_recently(self):
        return self.pub_date >= timezone.now() - datetime.timedelta(days=1)


class Choice(models.Model):
    question = models.ForeignKey(Question, on_delete=models.CASCADE)
    choice_text = models.CharField(max_length=200)
    votes = models.IntegerField(default=0)

    def __str__(self):
        return self.choice_text
```

jsantos42: I have Django version 5.0.4, working on a Mac M1. But then again, I'm in a virtual env, so I wonder if this would really make a difference.

adamzap: Thank you for posting your code! You defined the `pub_date` field as a `DateField` instead of a `DateTimeField`. That's the problem.

https://docs.djangoproject.com/en/5.0/intro/tutorial02/#creating-models

adamzap: Also, `USE_TZ` is a Django setting. No need to think about that right now. The confusion was my fault!

nessita: @jsantos42 `USE_TZ` is a Django setting, not an environment variable.

nessita:
@jsantos42 While we appreciate your effort and time preparing this PR, I'll close it since `pub_date` is indeed a datetime and the example is correct. It seems that you may be having trouble following the tutorial, you could try to get some extra help in any of our [amazing community channels](https://www.djangoproject.com/community/).

Thank you and hopefully you can contribute again in the near future! :star2:

docs/intro/tutorial02.txt
jsantos42: Thank you guys! My bad, sorry for wasting your time, I was just trying to help 🙈
Cheers!

adamzap: No worries! Glad to help.

PR #18080, opened 2024-04-16
Added a high contrast mode to screenshot cases.

sarahboyce:
# Branch description

To extend our accessibility tests, added screenshots in forced-colors mode.

# Checklist

### DIFF
--- a/django/test/selenium.py
+++ b/django/test/selenium.py
@@ -191,6 +191,26 @@ class SeleniumTestCase(LiveServerTestCase, metaclass=SeleniumTestCaseBase):
             finally:
                 self.selenium.execute_script("localStorage.removeItem('theme');")
 
+    def set_emulated_media(self, features, media=""):
+        if self.browser != "chrome":
+            self.skipTest("Emulated media controls are only supported on Chrome.")
+        # Chrome Dev Tools Protocol Emulation.setEmulatedMedia
+        # https://chromedevtools.github.io/devtools-protocol/1-3/Emulation/#method-setEmulatedMedia
+        self.selenium.execute_cdp_cmd(
+            "Emulation.setEmulatedMedia", {"media": media, "features": features}
+        )
+
+    @contextmanager
+    def high_contrast(self):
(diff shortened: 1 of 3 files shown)


sarahboyce:
Screenshots can be seen in the Artifacts here: https://github.com/django/django/actions/runs/8701792464/
They are the ones with `forced_colors` in the file name

Here are a couple of examples:
test_collapsible_fieldset_forced_colors-collapsed

test_first_field_focus_forced_colors-focus-multi-widget

This aims to aid the review process.
We may need to skip this on browsers that do not support forced-colors mode.

knyghty:
Super cool 👍🏻

sarahboyce:
> Firstly, congratulations on your new role! 🥳 

Thank you very much 😊 

- Renamed to `high_contrast`
- This will still generate screenshots on other browsers but they will just be the same as the desktop (we only run screenshots for chrome in the CI)

MHLut:
FYI, I have added some screenshots in https://github.com/django/django/pull/17910/ per @sarahboyce's suggestion.

sabderemane (approved):
Look good to me, nice addition ❤️

django/test/selenium.py
```diff
@@ -191,6 +191,26 @@ def dark(self):
             finally:
                 self.selenium.execute_script("localStorage.removeItem('theme');")
 
+    def set_emulated_media(self, features, media=""):
+        if self.browser != "chrome":
+            return
```
ngnpope, line 196: I guess we could skip on unsupported browsers to avoid generating misleading screenshots? 🤔

```suggestion
            self.skipTest("Emulated media controls are only supported on Chrome.")
```

ngnpope (approved):
Looks good. I have a little suggestion to skip the test on non-Chrome browsers which will work without affecting other screenshot cases as we're duplicating the test method.

Also need to update the PR title; perhaps:

```
Added a high contrast mode to screenshot cases.
```

django/test/selenium.py
sarahboyce: That's much nicer thank you

PR #18081, opened 2024-04-17
Allow support for postgres CREATE INDEX CONCURRENTLY IF NOT EXISTS

marcelofern:
# Trac ticket number
N/A

# Branch description
Prior to this commit, the only way of creating concurrent indexes with the postgres database backend was _without_ the `IF NOT EXISTS` conditional.

In certain cases, it is useful to have an idempotent migration. Examples are when a migration must be run during out-of-business-hours and a migration file is subsequently created to propagate the same changes to other environments (including developer ones).

This commit adds support for CREATE INDEX CONCURRENTLY IF NOT EXISTS, allowing the caller to decide whether or not to use the conditional.

The same support wasn't added to CREATE INDEX (without the CONCURRENTLY key), because further changes into the AddIndex class would take place, leaking Postgres requirements into a class that is used for multiple database backends.
# Checklist


shangxiao:
Thanks but all changes require a ticket

shangxiao:
(Also please first raise a thread on the forum before creating a ticket)

marcelofern:
@shangxiao I have a ticket now. https://code.djangoproject.com/ticket/35383#ticket
Could you please re-open the PR.

PR #18082, opened 2024-04-17
Added JS wait in test_inline_add_another_widgets.

MHLut:
# Trac ticket number
N/A

# Branch description
This Selenium test occasionally fails. We (@sarahboyce and I) suspect it could be due to Selenium being faster than JavaScript while testing.

Sadly, it is hard to test if this fix works due to the random nature of this bug occurring.

# Checklist


MHLut:
@sarahboyce As promised. Can you add the Selenium tag to this PR?

sarahboyce (approved):
Thank you for picking this up! 
Re-ran multiple times and didn't experience a failure (about the best I can do here).

PR #18092, opened 2024-04-20
Fixed #20744 -- Removed generic kwargs from form Field class docs.

adamzap:
# Trac ticket number

ticket-20744

# Branch description

I changed `.. class:: Field(**kwargs)` to  `.. class:: Field` in the form `Field` docs because the ticket says having `**kwargs` there implies that any keyword arguments can be passed to the class, which is not true.

image

This seems to be how similar classes like `Form` are [documented](https://github.com/django/django/blob/53719d6b5b745dd99b1ab9315afb242f706ebbf1/docs/ref/forms/api.txt#L26), so I thought this simple fix would be sufficient.

# Checklist

### DIFF
--- a/docs/ref/forms/fields.txt
+++ b/docs/ref/forms/fields.txt
@@ -7,7 +7,7 @@ Form fields
 
 .. currentmodule:: django.forms
 
-.. class:: Field(**kwargs)
+.. class:: Field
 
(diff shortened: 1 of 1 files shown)


nessita (approved):
Thank you @adamzap for this work! I agree this is an acceptable solution and that is consistent with how the `Form` class is documented.

PR #18101, opened 2024-04-25
Fixed #35405 -- Used @cached_property in FieldCacheMixin.

adamchainz:
# Trac ticket number

ticket-35405

# Branch description

Optimization as described on ticket.

# Checklist

### DIFF
--- a/django/db/models/fields/mixins.py
+++ b/django/db/models/fields/mixins.py
@@ -1,31 +1,52 @@
+import warnings
+
 from django.core import checks
+from django.utils.deprecation import RemovedInDjango60Warning
+from django.utils.functional import cached_property
 
 NOT_PROVIDED = object()
 
 
 class FieldCacheMixin:
-    """Provide an API for working with the model's fields value cache."""
+    """
+    An API for working with the model's fields value cache.
 
+    Subclasses must set self.cache_name to a unique entry for the cache -
+    typically the field’s name.
+    """
+
+    # RemovedInDjango60Warning.
     def get_cache_name(self):
         raise NotImplementedError
 
-    def get_cached_value(self, instance, default=NOT_PROVIDED):
+    @cached_property
+    def cache_name(self):
+        # RemovedInDjango60Warning: when the deprecation ends, replace with:
+        # raise NotImplementedError
         cache_name = self.get_cache_name()
+        warnings.warn(
+            f"Override {self.__class__.__qualname__}.cache_name instead of "
+            "get_cache_name().",
+            RemovedInDjango60Warning,
+        )
+        return cache_name
+
+    def get_cached_value(self, instance, default=NOT_PROVIDED):
         try:
-            return instance._state.fields_cache[cache_name]
+            return instance._state.fields_cache[self.cache_name]
         except KeyError:
             if default is NOT_PROVIDED:
                 raise
             return default
 
     def is_cached(self, instance):
-        return self.get_cache_name() in instance._state.fields_cache
+        return self.cache_name in instance._state.fields_cache
 
     def set_cached_value(self, instance, value):
-        instance._state.fields_cache[self.get_cache_name()] = value
+        instance._state.fields_cache[self.cache_name] = value
 
     def delete_cached_value(self, instance):
-        del instance._state.fields_cache[self.get_cache_name()]
+        del instance._state.fields_cache[self.cache_name]
 
 
 class CheckFieldDefaultMixin:
--- a/tests/model_fields/test_mixins.py
+++ b/tests/model_fields/test_mixins.py
@@ -1,0 +1,75 @@
+from django.db.models.fields.mixins import FieldCacheMixin
+from django.test import SimpleTestCase
+from django.utils.deprecation import RemovedInDjango60Warning
+from django.utils.functional import cached_property
+
+from .models import Foo
+
+
+# RemovedInDjango60Warning.
+class ExampleOld(FieldCacheMixin):
+    def get_cache_name(self):
+        return "example"
+
+
+class Example(FieldCacheMixin):
+    @cached_property
+    def cache_name(self):
+        return "example"
+
+
+class FieldCacheMixinTests(SimpleTestCase):
+    def setUp(self):
+        self.instance = Foo()
+        self.field = Example()
+
+    # RemovedInDjango60Warning: when the deprecation ends, replace with:
+    # def test_cache_name_not_implemented(self):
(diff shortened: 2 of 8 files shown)


django/db/models/fields/mixins.py
```diff
@@ -1,31 +1,37 @@
...
+    Provide an API for working with the model's fields value cache.
 
-    def get_cache_name(self):
+    Subclasses must set self.cache_name to a unique entry for the cache -
+    typically the field’s name.
+    """
+
+    @cached_property
+    def cache_name(self):
         raise NotImplementedError
```
charettes, line 17: Should we have a deprecation shim here that has `cache_name` call `get_cache_name` if not overriden?


```python
def get_cache_name(self):
    raise NotImplementedError

@cached_property
def cache_name(self):
    cache_name = self.get_cache_name()
    warnings.warn("cache_name must be overriden instead of get_cache_name", ...)
    return cache_name
```

adamchainz: Done!

django/db/models/fields/mixins.py
```diff
@@ -1,31 +1,52 @@
...
+from django.utils.deprecation import RemovedInDjango60Warning
+from django.utils.functional import cached_property
 
 NOT_PROVIDED = object()
 
 
 class FieldCacheMixin:
-    """Provide an API for working with the model's fields value cache."""
+    """
+    Provide an API for working with the model's fields value cache.
```
nessita, line 12: Since we are already modifying this docstring, I would advice we follow PEP and define a class as a noun and not as a verb:
```suggestion
    An API for working with the model's fields value cache.
```

adamchainz:
I have added some tests specific to the mixin, including for the deprecation path.

django/db/models/fields/mixins.py
adamchainz: Thanks, done.

tests/model_fields/test_mixins.py
```diff
@@ -0,0 +1,69 @@
+from django.db.models.fields.mixins import FieldCacheMixin
+from django.test import SimpleTestCase
+from django.utils.deprecation import RemovedInDjango60Warning
+from django.utils.functional import cached_property
+
+from .models import Foo
+
+
+class ExampleOld(FieldCacheMixin):
```
nessita, line 9: These are great tests, thank you for adding them. But, I think we should add some `RemovedInDjango60Warning` comments through this file to ensure we do the proper cleanup also at the test level when we bootstrap Django 6.0. Would you agree?

nessita (commented):
Branch looks great! I think is ready to checkin pending the adding of deprecation comments to the new test file so we can later cleanup accordingly. Thank you @adamchainz!

nessita (approved):
Thank you Adam, Simon and Sarah!

asottile-sentry:
fwiw this patch appears to break calls to `.get_cache_name()` since the mixin's `NotImplementedError` version is not overridden any more:

```pytb
../sentry/src/sentry/db/models/base.py:124: in is_field_cached
    name = self._get_relational_field(field_name).get_cache_name()
.venv/lib/python3.12/site-packages/django/db/models/fields/mixins.py:20: in get_cache_name
    raise NotImplementedError
E   NotImplementedError
```

there isn't a clear forward/backward compatible way to write this either (as `.cache_name` isn't available in 5.0) and there wasn't a deprecation period

nessita:
> fwiw this patch appears to break calls to `.get_cache_name()` since the mixin's `NotImplementedError` version is not overridden any more:
> 
> ```
> ../sentry/src/sentry/db/models/base.py:124: in is_field_cached
>     name = self._get_relational_field(field_name).get_cache_name()
> .venv/lib/python3.12/site-packages/django/db/models/fields/mixins.py:20: in get_cache_name
>     raise NotImplementedError
> E   NotImplementedError
> ```
> 
> there isn't a clear forward/backward compatible way to write this either (as `.cache_name` isn't available in 5.0) and there wasn't a deprecation period

Hey @asottile-sentry, could you please create a ticket in the [Django Trac System](https://code.djangoproject.com/) with details on how to reproduce? We did land this with a deprecation path so I would like to understand more why is not working for you/your project.

Thank you!

asottile-sentry:
that's an annoying barrier to entry, can you file one for me?  here's a pretty minimal bit of code (minus getting the django app working)

```python
from django.db import models

class A(models.Model): ...

class B(models.Model):
    a = models.ForeignKey(A, on_delete=models.CASCADE)

B._meta.get_field('a').get_cache_name()
```

```
  File "/Users/asottile/workspace/sentry/src/sentry/models/t.py", line 8, in <module>
    B._meta.get_field('a').get_cache_name()
  File "/Users/asottile/workspace/sentry/.venv/lib/python3.12/site-packages/django/db/models/fields/mixins.py", line 20, in get_cache_name
    raise NotImplementedError
NotImplementedError
```

this code succeeds in django 5.0.8 and fails in django 5.1 -- the graceful deprecation is done incorrectly as the original method was removed

adamchainz:
> that's an annoying barrier to entry, can you file one for me?

It’s not a very high barrier. Log in with GitHub at https://code.djangoproject.com/ , click “New ticket”, fill in fields as appropriate.

asottile-sentry:
> > that's an annoying barrier to entry, can you file one for me?
> 
> It’s not a very high barrier. Log in with GitHub at https://code.djangoproject.com/ , click “New ticket”, fill in fields as appropriate.

it would be much nicer if they were just on github to begin with -- why force me to sign up for yet another thing (which I can't do on my work account anyway)

nessita:
> > > that's an annoying barrier to entry, can you file one for me?
> > 
> > 
> > It’s not a very high barrier. Log in with GitHub at https://code.djangoproject.com/ , click “New ticket”, fill in fields as appropriate.
> 
> it would be much nicer if they were just on github to begin with -- why force me to sign up for yet another thing (which I can't do on my work account anyway)

You don't need a new account, you can login with your existing GitHub account.

asottile-sentry:
> > > > that's an annoying barrier to entry, can you file one for me?
> > > 
> > > 
> > > It’s not a very high barrier. Log in with GitHub at https://code.djangoproject.com/ , click “New ticket”, fill in fields as appropriate.
> > 
> > 
> > it would be much nicer if they were just on github to begin with -- why force me to sign up for yet another thing (which I can't do on my work account anyway)
> 
> You don't need a new account, you can login with your existing GitHub account.

that requires me to _authorize an app_ to my work account

PR #18102, opened 2024-04-25
Changed [yN] prompt to [y/N] in squashmigrations

dylancdavis:
Added a slash between the yes/no options printed from `squashmigrations` command to make it consistent with other commands that print similar prompts. Updated the corresponding documentation of `squashmigration` to match.

### DIFF
--- a/django/core/management/commands/squashmigrations.py
+++ b/django/core/management/commands/squashmigrations.py
@@ -123,7 +123,7 @@ class Command(BaseCommand):
             if self.interactive:
                 answer = None
                 while not answer or answer not in "yn":
-                    answer = input("Do you wish to proceed? [yN] ")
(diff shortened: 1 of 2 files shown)


nessita (approved):
Thank you @dylancdavis, this makes sense, and as you say is consistent with other prompts such as:

```
db/migrations/questioner.py:219:        msg = "Was %s.%s renamed to %s.%s (a %s)? [y/N]"
db/migrations/questioner.py:234:        msg = "Was the model %s.%s renamed to %s? [y/N]"
db/migrations/questioner.py:245:            + "Should these migration branches be merged? [y/N]",
contrib/auth/management/commands/createsuperuser.py:188:                            "Bypass password validation and create user anyway? [y/N]: "
```

PR #18104, opened 2024-04-26
Fixed #35407 -- Cached model's Options.swapped.

adamchainz:
# Trac ticket number

ticket-35407

# Branch description

Optimization per ticket.

# Checklist

### DIFF
--- a/django/db/models/options.py
+++ b/django/db/models/options.py
@@ -5,6 +5,7 @@ from collections import defaultdict
 from django.apps import apps
 from django.conf import settings
 from django.core.exceptions import FieldDoesNotExist, ImproperlyConfigured
+from django.core.signals import setting_changed
 from django.db import connections
 from django.db.models import AutoField, Manager, OrderWrt, UniqueConstraint
 from django.db.models.query_utils import PathInfo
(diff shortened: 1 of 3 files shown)


nessita:
Thank you @adamchainz! Could you please add a test to ensure the new approach works properly? Thank you!

django/db/models/options.py
```diff
@@ -230,6 +231,9 @@ def contribute_to_class(self, cls, name):
                 self.db_table, connection.ops.max_name_length()
             )
 
+        if self.swappable:
+            setting_changed.connect(self.setting_changed)
```
charettes, line 235: Cool that was my main concern when I saw the ticket.

Since `weak` defaults to `True` we'll make sure not to keep a strong reference to all the model classes created during migrations (hundreds of them in tests).

charettes (approved):

adamchainz:
> Thank you @adamchainz! Could you please add a test to ensure the new approach works properly? Thank you!

Done. The existing tests covered most or all cases already, implicitly, but I added explicit tests in the `model_meta` module.

django/db/models/options.py
adamchainz: Yup. I considered being a little more efficient by having a single, batch signal receiver that maps settings to “live” swappable models. But I don’t think that’s worth it as most projects use, at most, one swappable model (`auth.User`).

sarahboyce (approved):
Thank you!

PR #18110, opened 2024-04-28
Fixed #35378 -- Added maximum header length for long addresses.

lufafajoshua:
# Trac ticket number

ticket-35378 
# Branch description
Incorrect folding of long address headers with special characters when using 7bit Content-Transfer-Encoding in EmailMessage.

# Checklist

### DIFF
--- a/tests/mail/tests.py
+++ b/tests/mail/tests.py
@@ -902,7 +902,7 @@ class MailTests(HeadersCheckMixin, SimpleTestCase):
             "Subject",
             "Long address with special characters",
             "from@example.com",
-            ['"Người nhận a very very long, name" <to@example.com>']
+            ['"Người nhận a very very long, name" <to@example.com>'],
         )
         s = msg.message().as_bytes()
         self.assertIn(b"Content-Transfer-Encoding: 7bit", s)
--- a/tests/mail/tests.py
+++ b/tests/mail/tests.py
@@ -894,6 +894,21 @@ class MailTests(HeadersCheckMixin, SimpleTestCase):
         )
         self.assertNotIn(b">From the future", email.message().as_bytes())
 
+    def test_long_address_folding(self):
+        # Ticket #35378
+        # should encode long addresses with special characters using
+        # 7bit Content-Transfer-Encoding
+        msg = EmailMessage(
+            "Subject",
+            "Long address with special characters",
+            "from@example.com",
+            ['"Người nhận a very very long, name" <to@example.com>']
+        )
+        s = msg.message().as_bytes()
+        self.assertIn(b"Content-Transfer-Encoding: 7bit", s)
+        s = msg.message().as_string()
+        self.assertIn("Content-Transfer-Encoding: 7bit", s)
+
     def test_dont_base64_encode(self):
         # Ticket #3472
         # Shouldn't use Base64 encoding at all
--- a/django/core/mail/message.py
+++ b/django/core/mail/message.py
@@ -131,7 +131,7 @@ class MIMEMixin:
         lines that begin with 'From '. See bug #13433 for details.
         """
         fp = StringIO()
-        g = generator.Generator(fp, mangle_from_=False)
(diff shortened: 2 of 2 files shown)


tests/mail/tests.py
```diff
@@ -894,6 +894,21 @@ def test_dont_mangle_from_in_body(self):
         )
         self.assertNotIn(b">From the future", email.message().as_bytes())
 
+    def test_long_address_folding(self):
+        # Ticket #35378
+        # should encode long addresses with special characters using
+        # 7bit Content-Transfer-Encoding
```
DevilsAutumn, line 900: We no longer add comments in the tests until its really important. So maybe these can be removed.

tests/mail/tests.py
```diff
@@ -894,6 +894,21 @@ def test_dont_mangle_from_in_body(self):
...
+        msg = EmailMessage(
+            "Subject",
+            "Long address with special characters",
+            "from@example.com",
+            ['"Người nhận a very very long, name" <to@example.com>'],
+        )
+        s = msg.message().as_bytes()
+        self.assertIn(b"Content-Transfer-Encoding: 7bit", s)
+        s = msg.message().as_string()
+        self.assertIn("Content-Transfer-Encoding: 7bit", s)
```
DevilsAutumn, line 910: I reverted your changes from `message.py` and was expecting the this test to fail but surprisingly it passed.
The main issues mentioned in the ticket was the incorrect folding of long address headers when using 7bit Content-Transfer-Encoding , So if you'll check for the `Content-Transfer-Encoding: 7bit` in message string it will always be there. 

After you set `maxheaderlen=0`, it will treat all the long addresses as one string and there wont be any new line (\n) in between the address. So, in my opinion it will be better if you check the encoded long address for a new line symbol in somewhere between the address. If its not there then it means that its no longer folding the long addresses and treating it as one single string. Does that make sense to you?
```suggestion
        s = msg.message().as_bytes()
        self.assertIn(b"=?utf-8?b?TmfGsOG7nWkgbmjhuq1uIGEgdmVyeSB2ZXJ5IGxvbmcsIG5hbWU=?= <to@example.com>\n", s)
        s = msg.message().as_string()
        self.assertIn("=?utf-8?b?TmfGsOG7nWkgbmjhuq1uIGEgdmVyeSB2ZXJ5IGxvbmcsIG5hbWU=?= <to@example.com>\n", s)
```

DevilsAutumn (commented):
@lufafajoshua Thanks for this patch 👍, left small comments.

tests/mail/tests.py
sarahboyce: Yes, a test update is required 👍
Something like this might work
```python
    def test_long_address_folding(self):
        msg = EmailMessage(
            "Subject",
            "Long address with special characters",
            "from@example.com",
            ['"Người nhận a very very long, name" <to@example.com>'],
        ).message()
        msg.policy = policy.default.clone(cte_type="7bit")
        self.assertIn(
            "=?utf-8?b?TmfGsOG7nWkgbmjhuq1uIGEgdmVyeSB2ZXJ5IGxvbmcs?= name <to@example.com>",
            msg.as_bytes().decode("ascii")
        )
```
This would use the [example from the discussion](https://github.com/anymail/django-anymail/issues/369#issuecomment-2053731851).

This also fails and I am not sure adding `maxheaderlen=0` is the right solution. It goes against previous design decisions (see ticket-31784) and [RFC 2822](https://datatracker.ietf.org/doc/html/rfc2822#section-2.1.1) stating that 
> Each line of characters MUST be no more than 998 characters, and SHOULD be no more than 78 characters, excluding the CRLF.

Allowing longer line length was discussed and discouraged in [this comment](https://github.com/anymail/django-anymail/issues/369#issuecomment-2052683478).
> Allowing longer lines would solve the problem (no folding needed), but has similar compatibility concerns to using 8bit (or worse).

In short, I think more investigation is needed 🤔

PR #18112, opened 2024-04-29
Adjust docs slightly so that behavior of width/height fields is documented more precisely.

john-parton:
…ented more precisely.

# Trac ticket number
N/A

# Branch description
This is a trivial change which adjust the docs regarding the `width` and `height` fields for `ImageField`. 

If you do the following:

1. Save a model with an image field and width/height fields
2. Modify the underlying file outside of django.
3. Load the instance from the DB and re-save it.

Then the width/height fields will not be updated. This might seem obvious to a more experienced dev, but the current documentation isn't exactly clear on the behavior.

See discussion: https://github.com/django/django/pull/18070

# Checklist

### DIFF
--- a/docs/ref/models/fields.txt
+++ b/docs/ref/models/fields.txt
@@ -1342,13 +1342,13 @@ following optional arguments:
 
 .. attribute:: ImageField.height_field
 
-    Name of a model field which will be auto-populated with the height of the
-    image each time the model instance is saved.
+    Name of a model field which is auto-populated with the height of the image
+    each time an image object is set.
 
 .. attribute:: ImageField.width_field
 
-    Name of a model field which will be auto-populated with the width of the
-    image each time the model instance is saved.
+    Name of a model field which is auto-populated with the width of the image
+    each time an image object is set.
 
 Requires the `Pillow`_ library.
 


docs/ref/models/fields.txt
```diff
@@ -1343,12 +1343,12 @@ following optional arguments:
 .. attribute:: ImageField.height_field
 
     Name of a model field which will be auto-populated with the height of the
-    image each time the model instance is saved.
+    image each time the image attribute is set.
```
sarahboyce, line 1346: Wondering if "image attribute" is the best way to say this. What do you think about...
```suggestion
    image each time the image is set.
```
or 
```suggestion
    image each time the :class:`ImageField` instance is set.
```

john-parton: I'd like to avoid directly referencing the term `ImageField`, because it's _annoyingly_ complicated exactly what's going on.

I chose to use the word "attribute" because the code specifically calls ImageFieldFile `attr_class`.

If you omit the word attribute, you can still "set the image" and not trigger the descriptor or the width/height logic.

Here's an example of "setting the image" where you're not "setting the image attribute":

```python
MyModel.objects.filter(id=3).update(photo="myphoto.jpg")
```

So I'm `-1` on the second change and `-0` on the first one.

sarahboyce: I should have clarified why I am not keen on "image attribute" here. 
In the context of an attribute definition (`height_field`) referring to an "image attribute" makes it reasonable for readers to believe there is an undocumented attribute called `image`.

`attr_class` is referenced in the docs in a section "Writing a ``FileField`` subclass", but in a different page, so I think reader knowledge of `attr_class` is unlikely in order to make this connection. I understand why you chose to name it as such.

`FieldFile` is documented but not `ImageFieldFile` which might have been nice to reference (as there is a section refering to `FieldFile` vs `FileField`). 

I feel like that `.update` bypasses logic is fairly well know but that's a valid point. 
Let me try again, how about...
```
    Name of a model field which is auto-populated with the height of the image 
    each time an image object is set.
```
If you are 0+ I can push up the tweaks and merge 👍

john-parton: I think that's "good enough" / `+0`. Go ahead and push/merge

sarahboyce (approved):
Thank you! Pushed small edits 👍

PR #18056, opened 2024-04-07
Fixed #373 -- Added CompositePrimaryKey.

csirmazbendeguz:
# Trac ticket number

ticket-373

# Branch description

This branch adds the `CompositePrimaryKey` field. If present, Django will create a composite primary key.

Please refer to the [docs](https://github.com/django/django/pull/18056/files#diff-cca8870fcaec19104d999f61553ba925c72e2eb19b4933068c4849f2ce58a6f6) for a more in-depth explanation.

[Proposal](https://forum.djangoproject.com/t/gsoc-2024-proposal-django-orm-support-for-composite-primary-keys/29146)
[Previous PR](https://github.com/django/django/pull/18031)
[Composite FK](https://github.com/django/django/pull/18205)
[Admin](https://github.com/csirmazbendeguz/django/pull/10)
[Composite Generic FK](https://github.com/csirmazbendeguz/django/pull/9)
~~[Model._is_pk_set()](https://github.com/django/django/pull/18450)~~ ✅ 
~~[Tuple Lookups](https://github.com/django/django/pull/18404)~~ ✅ 
~~[Serial Fields](https://github.com/django/django/pull/18123)~~ ❌ 

```python
class Tenant(models.Model):
    pass


class User(models.Model):
    pk = models.CompositePrimaryKey("tenant_id", "id", primary_key=True)
    tenant = models.ForeignKey(Tenant, on_delete=models.CASCADE)
    id = models.IntegerField()


class Comment(models.Model):
    pk = models.CompositePrimaryKey("tenant_id", "id")
    tenant = models.ForeignKey(Tenant, on_delete=models.CASCADE)
    id = models.IntegerField()
    user_id = models.IntegerField()
    user = models.ForeignObject(
        User,
        on_delete=models.CASCADE,
        from_fields=("tenant_id", "user_id"),
        to_fields=("tenant_id", "id"),
        related_name="+",
    )
```

# Checklist
- [X] This PR targets the `main` branch. 
- [X] The commit message is written in past tense, mentions the ticket number, and ends with a period.
- [X] I have checked the "Has patch" **ticket flag** in the Trac system.
- [X] I have added or updated relevant **tests**.
- [x] I have added or updated relevant **docs**, including release notes if applicable.
- [x] For UI changes, I have attached **screenshots** in both light and dark modes.

PR #18123, opened 2024-05-03
Fixed #27452 -- Added serial fields to contrib.postgres.

csirmazbendeguz:
# Trac ticket number
ticket-27452

# Branch description
Added `SmallSerialField`, `SerialField`, `BigSerialField` to `contrib.postgres`.

[Previous PR](https://github.com/django/django/pull/12989)
[Composite PK](https://github.com/django/django/pull/18056)

!!! A more basic implementation can be found in [this PR](https://github.com/django/django/pull/18280). !!!

# Checklist

### DIFF
--- a/django/contrib/postgres/fields/serial.py
+++ b/django/contrib/postgres/fields/serial.py
@@ -1,9 +1,17 @@
 from django.db import models
-from django.db.models import NOT_PROVIDED
-from django.db.models.expressions import DatabaseDefault
+from django.db.models.expressions import DatabaseDefault, Expression
 from django.utils.translation import gettext_lazy as _
 
 __all__ = ("BigSerialField", "SmallSerialField", "SerialField")
+
+
+class NextSerialSequence(Expression):
+    allowed_default = True
+
+    def as_sql(self, compiler, connection):
+        table = self.field.model._meta.db_table
+        column = self.field.column
+        return "nextval(pg_get_serial_sequence(%s, %s))", [table, column]
 
 
 class SerialFieldMixin:
@@ -11,6 +19,7 @@ class SerialFieldMixin:
 
     def __init__(self, *args, **kwargs):
         default = DatabaseDefault()
+        db_default = NextSerialSequence(self)
 
         if not kwargs.setdefault("blank", True):
             raise ValueError(f"{self.__class__.__name__} must be blank.")
@@ -18,7 +27,7 @@ class SerialFieldMixin:
             raise ValueError(f"{self.__class__.__name__} must not be null.")
         if kwargs.setdefault("default", default) is not default:
             raise ValueError(f"{self.__class__.__name__} cannot have a default.")
-        if kwargs.get("db_default", NOT_PROVIDED) is not NOT_PROVIDED:
+        if kwargs.setdefault("db_default", db_default) is not db_default:
             raise ValueError(
                 f"{self.__class__.__name__} cannot have a database default."
             )
@@ -28,6 +37,7 @@ class SerialFieldMixin:
     def deconstruct(self):
         name, path, args, kwargs = super().deconstruct()
         del kwargs["default"]
+        del kwargs["db_default"]
         return name, path, args, kwargs
 
     def get_prep_value(self, value):
--- a/django/contrib/postgres/fields/serial.py
+++ b/django/contrib/postgres/fields/serial.py
@@ -1,0 +1,67 @@
+from django.db import models
+from django.db.models import NOT_PROVIDED
+from django.db.models.expressions import DatabaseDefault
+from django.utils.translation import gettext_lazy as _
+
+__all__ = ("BigSerialField", "SmallSerialField", "SerialField")
+
+
+class SerialFieldMixin:
+    db_returning = True
+
+    def __init__(self, *args, **kwargs):
+        default = DatabaseDefault()
+
+        if not kwargs.setdefault("blank", True):
+            raise ValueError(f"{self.__class__.__name__} must be blank.")
+        if kwargs.setdefault("null", False):
+            raise ValueError(f"{self.__class__.__name__} must not be null.")
+        if kwargs.setdefault("default", default) is not default:
+            raise ValueError(f"{self.__class__.__name__} cannot have a default.")
+        if kwargs.get("db_default", NOT_PROVIDED) is not NOT_PROVIDED:
+            raise ValueError(
+                f"{self.__class__.__name__} cannot have a database default."
+            )
+
+        super().__init__(*args, **kwargs)
+
+    def deconstruct(self):
+        name, path, args, kwargs = super().deconstruct()
+        del kwargs["default"]
+        return name, path, args, kwargs
+
+    def get_prep_value(self, value):
+        value = super().get_prep_value(value)
+        if value is None:
+            value = self.get_default()
+        return value
+
+
+class BigSerialField(SerialFieldMixin, models.BigIntegerField):
+    description = _("Big serial")
+
+    def get_internal_type(self):
+        return "BigSerialField"
+
+    def db_type(self, connection):
+        return "bigserial"
+
+
+class SmallSerialField(SerialFieldMixin, models.SmallIntegerField):
+    description = _("Small serial")
+
+    def get_internal_type(self):
+        return "SmallSerialField"
+
+    def db_type(self, connection):
+        return "smallserial"
+
+
+class SerialField(SerialFieldMixin, models.IntegerField):
+    description = _("Serial")
+
+    def get_internal_type(self):
+        return "SerialField"
+
+    def db_type(self, connection):
+        return "serial"
--- a/django/contrib/postgres/fields/serial.py
+++ b/django/contrib/postgres/fields/serial.py
@@ -46,6 +46,9 @@ class BigSerialField(SerialFieldMixin, models.BigIntegerField):
     def db_type(self, connection):
         return "bigserial"
 
+    def rel_db_type(self, connection):
+        return "bigint"
+
 
 class SmallSerialField(SerialFieldMixin, models.SmallIntegerField):
     description = _("Small serial")
@@ -56,6 +59,9 @@ class SmallSerialField(SerialFieldMixin, models.SmallIntegerField):
     def db_type(self, connection):
         return "smallserial"
 
+    def rel_db_type(self, connection):
+        return "smallint"
+
 
 class SerialField(SerialFieldMixin, models.IntegerField):
     description = _("Serial")
@@ -65,3 +71,6 @@ class SerialField(SerialFieldMixin, models.IntegerField):
 
     def db_type(self, connection):
         return "serial"
+
+    def rel_db_type(self, connection):
+        return "integer"
--- a/django/core/management/commands/inspectdb.py
+++ b/django/core/management/commands/inspectdb.py
@@ -1,5 +1,6 @@
 import keyword
 import re
+from collections import namedtuple
 
 from django.core.management.base import BaseCommand, CommandError
 from django.db import DEFAULT_DB_ALIAS, connections
@@ -84,6 +85,19 @@ class Command(BaseCommand):
                 types.add("v")
             table_info = connection.introspection.get_table_list(cursor)
             table_info = {info.name: info for info in table_info if info.type in types}
+            table_inspects = []
+            TableInspect = namedtuple(
+                "TableInspect",
+                (
+                    "table_name",
+                    "relations",
+                    "constraints",
+                    "primary_key_columns",
+                    "primary_key_column",
+                    "unique_columns",
+                    "table_description",
+                ),
+            )
 
             for table_name in options["table"] or sorted(name for name in table_info):
                 if table_name_filter is not None and callable(table_name_filter):
@@ -118,11 +132,54 @@ class Command(BaseCommand):
                     table_description = connection.introspection.get_table_description(
                         cursor, table_name
                     )
+                    table_inspects.append(
+                        TableInspect(
+                            table_name=table_name,
+                            relations=relations,
+                            constraints=constraints,
+                            primary_key_columns=primary_key_columns,
+                            primary_key_column=primary_key_column,
+                            unique_columns=unique_columns,
+                            table_description=table_description,
+                        )
+                    )
                 except Exception as e:
                     yield "# Unable to inspect table '%s'" % table_name
                     yield "# The error was: %s" % e
                     continue
 
+            for table_inspect in table_inspects:
+                table_name, table_description = (
+                    table_inspect.table_name,
+                    table_inspect.table_description,
+                )
+
+                def is_postgres_row(row):
+                    field_type, _, _ = self.get_field_type(connection, table_name, row)
+                    return field_type.startswith("postgres.fields.")
+
+                if any(is_postgres_row(row) for row in table_description):
+                    yield "from django.contrib import postgres"
+                    break
+
+            for table_inspect in table_inspects:
+                (
+                    table_name,
+                    relations,
+                    constraints,
+                    primary_key_columns,
+                    primary_key_column,
+                    unique_columns,
+                    table_description,
+                ) = (
+                    table_inspect.table_name,
+                    table_inspect.relations,
+                    table_inspect.constraints,
+                    table_inspect.primary_key_columns,
+                    table_inspect.primary_key_column,
+                    table_inspect.unique_columns,
+                    table_inspect.table_description,
+                )
                 model_name = self.normalize_table_name(table_name)
                 yield ""
                 yield ""
@@ -362,6 +419,16 @@ class Command(BaseCommand):
             else:
                 field_params["max_digits"] = row.precision
                 field_params["decimal_places"] = row.scale
+
+        serial_fields = {
+            "postgres.fields.SerialField": "AutoField",
+            "postgres.fields.SmallSerialField": "SmallAutoField",
+            "postgres.fields.BigSerialField": "BigAutoField",
+        }
+
+        if connection.vendor == "postgresql" and field_type in serial_fields:
+            auto_field = serial_fields[field_type]
+            field_notes.append(f"You may want to consider using {auto_field} instead.")
 
         return field_type, field_params, field_notes
 
--- a/django/db/backends/postgresql/introspection.py
+++ b/django/db/backends/postgresql/introspection.py
@@ -40,17 +40,22 @@ class DatabaseIntrospection(BaseDatabaseIntrospection):
 
     def get_field_type(self, data_type, description):
         field_type = super().get_field_type(data_type, description)
-        if description.is_autofield or (
-            # Required for pre-Django 4.1 serial columns.
-            description.default
-            and "nextval" in description.default
-        ):
+
+        if description.is_autofield:
             if field_type == "IntegerField":
                 return "AutoField"
             elif field_type == "BigIntegerField":
                 return "BigAutoField"
             elif field_type == "SmallIntegerField":
                 return "SmallAutoField"
+        elif description.default and "nextval" in description.default:
+            if field_type == "IntegerField":
+                return "postgres.fields.SerialField"
+            elif field_type == "BigIntegerField":
+                return "postgres.fields.BigSerialField"
+            elif field_type == "SmallIntegerField":
+                return "postgres.fields.SmallSerialField"
+
         return field_type
 
     def get_table_list(self, cursor):
--- a/django/db/backends/postgresql/operations.py
+++ b/django/db/backends/postgresql/operations.py
@@ -2,6 +2,7 @@ import json
 from functools import lru_cache, partial
 
 from django.conf import settings
+from django.contrib.postgres.fields.serial import SerialFieldMixin
 from django.db.backends.base.operations import BaseDatabaseOperations
 from django.db.backends.postgresql.psycopg_any import (
     Inet,
@@ -43,6 +44,15 @@ class DatabaseOperations(BaseDatabaseOperations):
         "AutoField": "integer",
         "BigAutoField": "bigint",
         "SmallAutoField": "smallint",
+        "SerialField": "integer",
+        "BigSerialField": "bigint",
+        "SmallSerialField": "smallint",
+    }
+    integer_field_ranges = {
+        **BaseDatabaseOperations.integer_field_ranges,
+        "SmallSerialField": (-32768, 32767),
+        "SerialField": (-2147483648, 2147483647),
+        "BigSerialField": (-9223372036854775808, 9223372036854775807),
     }
 
     if is_psycopg3:
@@ -55,6 +65,9 @@ class DatabaseOperations(BaseDatabaseOperations):
             "PositiveSmallIntegerField": numeric.Int2,
             "PositiveIntegerField": numeric.Int4,
             "PositiveBigIntegerField": numeric.Int8,
+            "SmallSerialField": numeric.Int2,
+            "SerialField": numeric.Int4,
+            "BigSerialField": numeric.Int8,
         }
 
     def unification_cast_sql(self, output_field):
@@ -250,7 +263,7 @@ class DatabaseOperations(BaseDatabaseOperations):
             # underlying sequence name from the table name and column name.
 
             for f in model._meta.local_fields:
-                if isinstance(f, models.AutoField):
+                if isinstance(f, (models.AutoField, SerialFieldMixin)):
                     output.append(
                         "%s setval(pg_get_serial_sequence('%s','%s'), "
                         "coalesce(max(%s), 1), max(%s) %s null) %s %s;"
@@ -265,9 +278,7 @@ class DatabaseOperations(BaseDatabaseOperations):
                             style.SQL_TABLE(qn(model._meta.db_table)),
                         )
                     )
-                    # Only one AutoField is allowed per model, so don't bother
-                    # continuing.
-                    break
+
         return output
 
     def prep_for_iexact_query(self, x):
--- a/django/db/backends/postgresql/schema.py
+++ b/django/db/backends/postgresql/schema.py
@@ -367,3 +367,17 @@ class DatabaseSchemaEditor(BaseDatabaseSchemaEditor):
             )
             row = cursor.fetchone()
             return row[0] if row else None
+
+    def skip_default(self, field):
+        return field.get_internal_type() in {
+            "SerialField",
+            "BigSerialField",
+            "SmallSerialField",
+        }
+
+    def skip_default_on_alter(self, field):
+        return field.get_internal_type() in {
+            "SerialField",
+            "BigSerialField",
+            "SmallSerialField",
+        }
--- a/django/db/backends/postgresql/schema.py
+++ b/django/db/backends/postgresql/schema.py
@@ -13,6 +13,16 @@ class DatabaseSchemaEditor(BaseDatabaseSchemaEditor):
     )
     sql_alter_sequence_type = "ALTER SEQUENCE IF EXISTS %(sequence)s AS %(type)s"
     sql_delete_sequence = "DROP SEQUENCE IF EXISTS %(sequence)s CASCADE"
+    sql_create_sequence = (
+        "CREATE SEQUENCE IF NOT EXISTS %(sequence)s "
+        "AS %(data_type)s "
+        "OWNED BY %(table)s.%(column)s"
+    )
+    sql_reset_sequence = (
+        "SELECT setval(pg_get_serial_sequence(%s, %s), "
+        "coalesce(max(%s), 1), max(%s) IS NOT NULL) FROM %s;"
+    )
+    sql_rename_sequence = "ALTER SEQUENCE %s RENAME TO %s;"
 
     sql_create_index = (
         "CREATE INDEX %(name)s ON %(table)s%(using)s "
@@ -38,6 +48,11 @@ class DatabaseSchemaEditor(BaseDatabaseSchemaEditor):
         "ALTER TABLE %(table)s DROP CONSTRAINT %(name)s"
     )
     sql_delete_procedure = "DROP FUNCTION %(procedure)s(%(param_types)s)"
+    serial_field_types = {
+        "SerialField": "integer",
+        "BigSerialField": "bigint",
+        "SmallSerialField": "smallint",
+    }
 
     def execute(self, sql, params=()):
         # Merge the query client-side, as PostgreSQL won't do it server-side.
@@ -167,15 +182,51 @@ class DatabaseSchemaEditor(BaseDatabaseSchemaEditor):
         old_internal_type = old_field.get_internal_type()
         # Make ALTER TYPE with IDENTITY make sense.
         table = strip_quotes(model._meta.db_table)
+
         auto_field_types = {
-            "AutoField",
-            "BigAutoField",
-            "SmallAutoField",
+            "AutoField": "integer",
+            "BigAutoField": "bigint",
+            "SmallAutoField": "smallint",
         }
         old_is_auto = old_internal_type in auto_field_types
         new_is_auto = new_internal_type in auto_field_types
-        if new_is_auto and not old_is_auto:
-            column = strip_quotes(new_field.column)
+
+        old_is_serial = old_internal_type in self.serial_field_types
+        new_is_serial = new_internal_type in self.serial_field_types
+        column = strip_quotes(new_field.column)
+
+        if old_is_auto and new_is_serial:
+            data_type = self.serial_field_types[new_internal_type]
+            self.execute(self._drop_identity_sql(table, column))
+            self.execute(self._create_sequence_sql(table, column, data_type))
+            fragment, other_actions = super()._alter_column_type_sql(
+                model, old_field, new_field, data_type, old_collation, new_collation
+            )
+
+            return (
+                fragment,
+                other_actions
+                + [
+                    (self._reset_sequence_sql(table, column), []),
+                ],
+            )
+        elif old_is_serial and new_is_auto:
+            sequence_name = self._get_sequence_name(table, column)
+            self.execute(self._delete_sequence_sql(sequence_name))
+            data_type = auto_field_types[new_internal_type]
+            fragment, other_actions = super()._alter_column_type_sql(
+                model, old_field, new_field, data_type, old_collation, new_collation
+            )
+
+            return (
+                fragment,
+                other_actions
+                + [
+                    (self._add_identity_sql(table, column), []),
+                    (self._reset_sequence_sql(table, column), []),
+                ],
+            )
+        elif new_is_auto and not old_is_auto:
             return (
                 (
                     self.sql_alter_column_type
@@ -187,27 +238,13 @@ class DatabaseSchemaEditor(BaseDatabaseSchemaEditor):
                     [],
                 ),
                 [
-                    (
-                        self.sql_add_identity
-                        % {
-                            "table": self.quote_name(table),
-                            "column": self.quote_name(column),
-                        },
-                        [],
-                    ),
+                    (self._add_identity_sql(table, column), []),
                 ],
             )
         elif old_is_auto and not new_is_auto:
             # Drop IDENTITY if exists (pre-Django 4.1 serial columns don't have
             # it).
-            self.execute(
-                self.sql_drop_indentity
-                % {
-                    "table": self.quote_name(table),
-                    "column": self.quote_name(strip_quotes(new_field.column)),
-                }
-            )
-            column = strip_quotes(new_field.column)
+            self.execute(self._drop_identity_sql(table, column))
             fragment, _ = super()._alter_column_type_sql(
                 model, old_field, new_field, new_type, old_collation, new_collation
             )
@@ -215,41 +252,56 @@ class DatabaseSchemaEditor(BaseDatabaseSchemaEditor):
             # have it).
             other_actions = []
             if sequence_name := self._get_sequence_name(table, column):
-                other_actions = [
-                    (
-                        self.sql_delete_sequence
-                        % {
-                            "sequence": self.quote_name(sequence_name),
-                        },
-                        [],
-                    )
-                ]
+                other_actions = [(self._delete_sequence_sql(sequence_name), [])]
             return fragment, other_actions
         elif new_is_auto and old_is_auto and old_internal_type != new_internal_type:
             fragment, _ = super()._alter_column_type_sql(
                 model, old_field, new_field, new_type, old_collation, new_collation
             )
-            column = strip_quotes(new_field.column)
-            db_types = {
-                "AutoField": "integer",
-                "BigAutoField": "bigint",
-                "SmallAutoField": "smallint",
-            }
             # Alter the sequence type if exists (Django 4.1+ identity columns
             # don't have it).
             other_actions = []
             if sequence_name := self._get_sequence_name(table, column):
+                data_type = auto_field_types[new_internal_type]
                 other_actions = [
-                    (
-                        self.sql_alter_sequence_type
-                        % {
-                            "sequence": self.quote_name(sequence_name),
-                            "type": db_types[new_internal_type],
-                        },
-                        [],
-                    ),
+                    (self._alter_sequence_type_sql(sequence_name, data_type), []),
                 ]
             return fragment, other_actions
+        elif new_is_serial and not old_is_serial:
+            data_type = self.serial_field_types[new_internal_type]
+            self.execute(self._create_sequence_sql(table, column, data_type))
+            self.execute(self._reset_sequence_sql(table, column))
+
+            return super()._alter_column_type_sql(
+                model, old_field, new_field, data_type, old_collation, new_collation
+            )
+        elif old_is_serial and not new_is_serial:
+            sequence_name = self._get_sequence_name(table, column)
+            fragment, other_actions = super()._alter_column_type_sql(
+                model, old_field, new_field, new_type, old_collation, new_collation
+            )
+
+            return (
+                fragment,
+                other_actions
+                + [
+                    (self._delete_sequence_sql(sequence_name), []),
+                ],
+            )
+        elif old_is_serial and new_is_serial and old_internal_type != new_internal_type:
+            data_type = self.serial_field_types[new_internal_type]
+            sequence_name = self._get_sequence_name(table, column)
+            fragment, other_actions = super()._alter_column_type_sql(
+                model, old_field, new_field, data_type, old_collation, new_collation
+            )
+
+            return (
+                fragment,
+                other_actions
+                + [
+                    (self._alter_sequence_type_sql(sequence_name, data_type), []),
+                ],
+            )
         else:
             return super()._alter_column_type_sql(
                 model, old_field, new_field, new_type, old_collation, new_collation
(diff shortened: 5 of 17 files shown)


django/contrib/postgres/fields/serial.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault("default", DatabaseDefault())
+        super().__init__(
+            *args,
+            **{
+                **kwargs,
+                "blank": True,
+            },
+        )
```
LilyFoote, line 20: Assuming we want to force `blank=True` we can write this as:

```suggestion
        kwargs["blank"] = True
        super().__init__(*args, **kwargs)
```

django/contrib/postgres/fields/serial.py
```diff
@@ -0,0 +1,87 @@
...
+        if self.default != DatabaseDefault():
+            return [
+                checks.Error(
+                    "SerialFields do not accept default values.",
+                    obj=self,
+                    id="fields.E911",
+                ),
+            ]
+        else:
+            return []
```
LilyFoote, line 57: We can check for the `NOT_PROVIDED` singleton here instead of using `DatabaseDefault` for something it wasn't meant for:

```suggestion
        if self.default is NOT_PROVIDED:
            return []
        return [
            checks.Error(
                "SerialFields do not accept default values.",
                obj=self,
                id="fields.E911",
            ),
        ]
```

django/contrib/postgres/fields/serial.py
```diff
@@ -0,0 +1,87 @@
...
+from django.utils.translation import gettext_lazy as _
+
+__all__ = ("BigSerialField", "SmallSerialField", "SerialField")
+
+
+class SerialFieldMixin:
+    db_returning = True
+
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault("default", DatabaseDefault())
```
LilyFoote, line 13: The default value of `default` is `NOT_PROVIDED`, which is more appropriate, so we don't need this line:

```suggestion
```

django/contrib/postgres/fields/serial.py
```diff
@@ -0,0 +1,87 @@
...
+            **{
+                **kwargs,
+                "blank": True,
+            },
+        )
+
+    def deconstruct(self):
+        name, path, args, kwargs = super().deconstruct()
+        kwargs.pop("blank")
+        kwargs.pop("default")
```
LilyFoote, line 25:
```suggestion
```

django/db/backends/postgresql/operations.py
```diff
@@ -265,9 +277,7 @@ def sequence_reset_sql(self, style, model_list):
                             style.SQL_TABLE(qn(model._meta.db_table)),
                         )
                     )
-                    # Only one AutoField is allowed per model, so don't bother
-                    # continuing.
-                    break
```
LilyFoote, line 270: Is there a check for two `AutoField`s? If not perhaps we should add one, though perhaps in a separate ticket/PR.

django/db/backends/postgresql/schema.py
```diff
@@ -174,8 +185,41 @@ def _alter_column_type_sql(
...
+                    self.sql_set_default_sequence
+                    % {
+                        "table": self.quote_name(table),
+                        "column": self.quote_name(column),
+                        "sequence": self.quote_value(sequence),
+                    },
+                    [],
+                ),
+                [],
+            )
```
LilyFoote, line 212: I would pull some of this out into a new variable:

```suggestion
            set_default_sequence = self.sql_set_default_sequence % {
                "table": self.quote_name(table),
                "column": self.quote_name(column),
                "sequence": self.quote_value(sequence),
            }
            return ((set_default_sequence, []), [])
```

And similarly below.

django/db/backends/postgresql/schema.py
```diff
@@ -174,8 +185,41 @@ def _alter_column_type_sql(
         }
         old_is_auto = old_internal_type in auto_field_types
         new_is_auto = new_internal_type in auto_field_types
-        if new_is_auto and not old_is_auto:
```
LilyFoote, line 177: Do we need to check for `serial` fields before `auto` fields? If not I'd prefer to add the new code on the end of the `if/elif` chain - I think it'd make checking the diff a lot more straightforward.

tests/schema/tests.py
```diff
@@ -5719,3 +5738,350 @@ def test_ci_cs_db_collation(self):
...
+
+        self.serial_to_integer_test(BigSerialField, IntegerField)
+
+    @unittest.skipUnless(connection.vendor == "postgresql", "PostgreSQL specific")
+    @isolate_apps("schema")
+    @tag("serial")
+    def test_big_serial_to_small_integer(self):
+        from django.contrib.postgres.fields import BigSerialField
+
+        self.serial_to_integer_test(BigSerialField, SmallIntegerField)
```
LilyFoote, line 5812: This feels like a good use case for `subTest` to me.

tests/schema/tests.py
```diff
@@ -5719,3 +5738,350 @@ def test_ci_cs_db_collation(self):
...
+
+    def serial_to_integer_test(self, old_model, new_model):
```
LilyFoote, line 5814: I think these are fields, not models:

```suggestion
    def serial_to_integer_test(self, old_field, new_field):
```

LilyFoote (commented):
This looks like a great start!

In addition to the specific comments I added, we'll need some migrations tests to cover adding/removing serial fields and converting between serial and auto fields.

django/contrib/postgres/fields/serial.py
csirmazbendeguz: Yes, you're correct, this is how it's done in the `AutoFieldMixin` as well, I was just using an old branch that had this. I'll fix!

django/db/backends/postgresql/operations.py
csirmazbendeguz: Currently, `AutoField`s are always primary keys, so it's impossible to have multiple auto fields. I re-opened the ticket to revisit this topic recently [#8576](https://code.djangoproject.com/ticket/8576).

django/contrib/postgres/fields/serial.py
csirmazbendeguz: Yes, I agree, this is better.

django/contrib/postgres/fields/serial.py
csirmazbendeguz: Maybe we could plug the `DatabaseDefault` in `get_prep_value`?

tests/schema/tests.py
csirmazbendeguz: Yeah, good point.

tests/schema/tests.py
csirmazbendeguz: Yes, the issue is we would either need to generate the test models with a dynamic name, or clean up the test models after the subtests - IMO it's more trouble than worth it.

django/db/backends/postgresql/schema.py
csirmazbendeguz: We need to check the `auto -> serial` and `serial -> auto` cases, otherwise the `auto -> non-auto` and `non-auto -> auto` branches will execute.

django/db/backends/postgresql/schema.py
csirmazbendeguz: Yes, this part looks messy... maybe I could put this in a function, e.g. `def _set_default_sequence_sql`, seems that's what others did in the other schema editor classes.

csirmazbendeguz:
> This looks like a great start!
> 
> In addition to the specific comments I added, we'll need some migrations tests to cover adding/removing serial fields and converting between serial and auto fields.

@LilyFoote , thanks for the review! Could you point me to some tests I could copy to get started? I'm not sure how to test the migrations other than with the `tests/schema` tests.

LilyFoote:
For database defaults I added tests in [tests/migrations/test_autodetector.py](https://github.com/django/django/blob/main/tests/migrations/test_autodetector.py) and [tests/migrations/test_operations.py](https://github.com/django/django/blob/main/tests/migrations/test_operations.py).

csirmazbendeguz:
I have managed to test these operations (add, alter, remove serial field) in the `schema.tests`, I'm not sure if more tests are really needed.

django/db/backends/postgresql/introspection.py
```diff
@@ -40,17 +40,20 @@ class DatabaseIntrospection(BaseDatabaseIntrospection):
...
                 return "BigAutoField"
             elif field_type == "SmallIntegerField":
                 return "SmallAutoField"
+        elif description.default and "nextval" in description.default:
+            if field_type == "IntegerField":
+                return "SerialField"
+            elif field_type == "BigIntegerField":
+                return "BigSerialField"
+            elif field_type == "SmallIntegerField":
+                return "SmallSerialField"
```
ngnpope, line 56: ⚠️  There potential for breakage of existing projects here...

This check on default value was kept for backward compatibility as, prior to Django 4.1, the implementation of `AutoField` on PostgreSQL didn't use identity columns but serial columns with sequences. (See [here](https://docs.djangoproject.com/en/stable/releases/4.1/#models).)

This change will mean that all of those `AutoField`s that were created before Django 4.1, will now be treated as `SerialField`s and break a lot of stuff. We will need some way to distinguish between these cases. See [my comment](https://github.com/django/django/pull/15542#issuecomment-1403697290) which highlights that it wasn't necessary to migrate to identity columns, but that it was achievable with a manual migration. Many will have upgraded Django and not done this.

django/db/backends/postgresql/operations.py
```diff
@@ -237,20 +251,18 @@ def tablespace_sql(self, tablespace, inline=False):
...
             # Use `coalesce` to set the sequence for each model to the max pk
             # value if there are records, or 1 if there are none. Set the
             # `is_called` property (the third argument to `setval`) to true if
             # there are records (as the max pk value is already in use),
             # otherwise set it to false. Use pg_get_serial_sequence to get the
             # underlying sequence name from the table name and column name.
-
             for f in model._meta.local_fields:
-                if isinstance(f, models.AutoField):
+                if isinstance(f, (AutoFieldMixin, SerialFieldMixin)):
```
ngnpope, line 265: Small aside: This is an interesting one. There's a wider issue around replacing `isinstance(..., AutoField)` with something else - most likely some kind of `is_generated` or ` is_auto` or something - using a flag, rather than `isinstance()`. I'd have to look up the details as it was a number of years ago. It's probably not going to go in the direction of using `isinstance(..., AutoFieldMixin)` though... 🤔

docs/ref/checks.txt
```diff
@@ -252,6 +252,8 @@ Model fields
   in Django 5.1. *This check appeared in Django 4.2 and 5.0*.
 * **fields.E907**: ``django.contrib.postgres.fields.CITextField`` is removed
   except for support for historical migrations.
+* **fields.E910**: ``SerialField``\s do not accept null values.
+* **fields.E911**: ``SerialField``\s do not accept default values.
```
ngnpope, line 256: It looks as though `[EW]9??` have all been used for the deprecation of fields, so these could probably do with a different number? It looks as though `[EW]22?` has been used for `GeneratedField`, so maybe use `E300` and `E301` for these? _(We could use that for fields from `django.contrib.postgres`?)_

Better yet, these seem fairly generic, so can we reuse an existing code? `E910` is almost equivalent to `E110`, but perhaps we can just go for:

```
* **fields.E013**: ``<field>`` does not accept null values.
* **fields.E014**: ``<field>`` does not accept default values.
```

ngnpope (changes requested):
We will need to address the backward compatibility with pre-Django 4.1 `AutoField` on PostgreSQL to avoid breaking many old projects that didn't manually migrate to identity columns.

docs/ref/checks.txt
csirmazbendeguz: Thanks @ngnpope , I was not aware. I agree with your generic solution.

django/db/backends/postgresql/operations.py
csirmazbendeguz: Yeah, having a flag for this would make this code more extensible. Does it have to be addressed in this PR though?
By the way, as far as I understand, this function is broken for `BigAutoField` and `SmallAutoField`, that's why I replaced it with the mixin.

django/db/backends/postgresql/introspection.py
csirmazbendeguz: I see, thanks for this @ngnpope . You're right, this is backwards incompatible.

I suppose the correct solution would be to force users to migrate first.

As a workaround, I could attach a comment to `SerialField`s storing the current Django version and check for this comment's presence on introspection. This is kind of a hack, as it would mean users can't attach a comment to `SerialField`s, but it would work as a temporary solution (unless someone has this very specific comment on their `AutoField` already, but I highly doubt that).

csirmazbendeguz:
@ngnpope , thank you for reviewing my changes ! 🙏 I have pushed a fix for the issues you mentioned.

django/db/backends/postgresql/operations.py
ngnpope: > Does it have to be addressed in this PR though?

No, was just pondering whether we wanted to do this isinstance of mixin things. I guess it's internal, so it's fine.

django/db/backends/postgresql/introspection.py
ngnpope: > I suppose the correct solution would be to force users to migrate first.

I'm not sure that's viable. It would be a hard break for something that worked fine and there may be other ramifications if other non-Django code is doing things with those tables...

> ...but it would work as a temporary solution...

It's a case of break one feature to enable another. And if a DBA manually removed or changed a comment then the new feature would break. And at what point does this "temporary" solution get replaced with a permanent one?

I haven't had much time to dig into potential answers to this I'm afraid.

csirmazbendeguz: Ok, in that case, we can leave it the way it is. The feature will work fine without this `get_field_type` function. This function is only used by the `inspectdb` command, which is why I was thinking it was okay to break backwards-compatibility in the first place.

ngnpope: 🙈 Ah, I was having an off day yesterday and missed that this was introspection-only.

Perhaps it is fine to make this change then as, while backward-incompatible, we don't make guarantees that introspection will give a perfect result.

In `django.core.management.commands.inspectdb.Command.get_field_types()` we have some `field_notes` that get output as comments in the generated code. It'd be nice to add a comment for `SerialField`, etc. that it may be intended that it is an `AutoField` or something if we were to go ahead with this change. I'm not sure how that'd work exactly as we probably don't want to reference `django.contrib.postgres` fields directly there though.

I would say that this caution still applies in general, however, e.g. we likely want to test your migration changes with a model that starts out as a pre-Django 4.1 `AutoField` for PostgreSQL to see how it behaves.

csirmazbendeguz: I just realized `inspectdb` cannot output Postgres-specific fields. It will generate `models.SerialField` which is incorrect. Let me see if that's possible to fix...

csirmazbendeguz: @ngnpope , I pushed the changes addressing this comment.

csirmazbendeguz:
@ngnpope , I think all concerns has been addressed, let me know what you think.

LilyFoote (approved):
I'm loving the progress you've made here. I have a few minor suggestions, but overall looks good to me!

django/contrib/postgres/fields/serial.py
```diff
@@ -0,0 +1,86 @@
...
+                    "SerialFields do not accept null values.",
+                    obj=self,
+                    id="fields.E910",
+                ),
+            ]
+        else:
+            return []
+
+    def _check_default(self):
+        if self.default is NOT_PROVIDED:
```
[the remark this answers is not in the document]
csirmazbendeguz: Yes, that's a great catch. We wouldn't want people to set any kind of default value.

django/db/backends/postgresql/schema.py
```diff
@@ -167,14 +189,57 @@ def _alter_column_type_sql(
...
+                [
+                    (self._drop_identity_sql(table, column), []),
+                    (self._create_sequence_sql(table, column, data_type), []),
+                    (self._reset_sequence_sql(table, column), []),
+                    (self._set_default_sequence_sql(table, column), []),
+                ],
+            )
+        elif old_is_serial and new_is_auto:
+            column = strip_quotes(new_field.column)
+            sequence_name = self._get_sequence_name(table, column)
```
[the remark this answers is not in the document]
csirmazbendeguz: I think it's better to keep this one nested, because it has to execute a database query.

csirmazbendeguz:
Thanks a lot @LilyFoote ! I addressed your suggestions.

django/contrib/postgres/fields/serial.py
```diff
@@ -0,0 +1,98 @@
...
+            *self._check_not_null(),
+            *self._check_default(),
+            *self._check_database_default(),
+        ]
+
+    def _check_not_null(self):
+        if self.null:
+            return [
+                checks.Error(
+                    "SerialFields do not accept null values.",
```
sarahboyce, line 34:
```suggestion
                f"{self.__class__.__name__} must not have ``null=True``.",
```
Saying "must not have `null=True`" aligns more to `fields.E007`.

Also this is currently documented as 
```
``<field>`` does not accept null values.
```
But always says `SerialFields`, maybe we should do something similar to `fields.E010` and update the error message depending on the field. If so, update each of these, otherwise I would document the check to be:
```
SerialFields does not accept null values.
```

django/contrib/postgres/fields/serial.py
```diff
@@ -0,0 +1,98 @@
...
+                id="fields.E014",
+            ),
+        ]
+
+    def _check_database_default(self):
+        if self.db_default is NOT_PROVIDED:
+            return []
+        return [
+            checks.Error(
+                "SerialFields do not accept default database values.",
```
sarahboyce, line 58:
```suggestion
                "SerialFields do not accept default database values (db_default).",
```
This aligns better to `fields.E011`

docs/ref/contrib/postgres/fields.txt
```diff
@@ -920,3 +920,28 @@ define :class:`~django.contrib.postgres.constraints.ExclusionConstraint`. See
 docs/current/rangetypes.html#RANGETYPES-INCLUSIVITY>`_.
 
 .. _different bounds: https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-IO
+
+``Serial`` fields
+=================
+
+.. versionadded:: 5.1
```
sarahboyce, line 927:
```suggestion
.. versionadded:: 5.2
```

docs/releases/5.1.txt
```diff
@@ -123,6 +123,11 @@ Minor features
 * :class:`~django.contrib.postgres.indexes.BTreeIndex` now supports the
   ``deduplicate_items`` parameter.
 
+* :class:`~django.contrib.postgres.fields.SerialField`,
+  :class:`~django.contrib.postgres.fields.BigSerialField` and
+  :class:`~django.contrib.postgres.fields.SmallSerialField` have been added
+  corresponding to their respective PostgreSQL database types.
```
sarahboyce, line 129: Please rebase and move this to 5.2 👍

docs/ref/contrib/postgres/fields.txt
```diff
@@ -920,3 +920,28 @@ define :class:`~django.contrib.postgres.constraints.ExclusionConstraint`. See
...
+
+.. class:: SerialField(**options)
+
+    A field for storing an automatically incremented sequence of integers.
+
+    It's similar to :class:`AutoField <django.db.models.AutoField>`, but it
+    doesn't have to be a primary key.
+
+    It's a 32-bit integer that is is guaranteed to fit numbers from ``1`` to
+    ``2147483647``.
```
sarahboyce, line 937:
```suggestion
    A 32-bit integer that is is guaranteed to fit numbers from ``1`` to
    ``2147483647``.
```
I don't know why I don't like "it's" here

docs/ref/contrib/postgres/fields.txt
```diff
@@ -920,3 +920,28 @@ define :class:`~django.contrib.postgres.constraints.ExclusionConstraint`. See
...
+=================
+
+.. versionadded:: 5.1
+
+.. class:: SerialField(**options)
+
+    A field for storing an automatically incremented sequence of integers.
+
+    It's similar to :class:`AutoField <django.db.models.AutoField>`, but it
+    doesn't have to be a primary key.
```
sarahboyce, line 934:
```suggestion
    An automatically incrementing :class:`IntegerField`, much like :class:`AutoField`
    except that it does not have to store a primary key.
```
Maybe

sarahboyce (commented):
Thank you @csirmazbendeguz
I haven't reviewed this in depth but will hope to review again soon, as a first look through, this looks nice

django/contrib/postgres/fields/serial.py
csirmazbendeguz: @sarahboyce , I was following `fields.E110` which has:
```
``BooleanField``\s do not accept null values
```
It was @ngnpope 's idea to make this error message general, and I agree with it, because this way we wouldn't need to define a new error code every time a new field doesn't accept a `null` value.

docs/ref/contrib/postgres/fields.txt
csirmazbendeguz: done, thanks 👍

docs/ref/contrib/postgres/fields.txt
csirmazbendeguz: `it does not have to ***be*** a primary key.` sounds more natural to me, but I'm not a native English speaker.

docs/ref/contrib/postgres/fields.txt
csirmazbendeguz: Maybe it's not necessary to include this sentence here at all. Similar to `AutoField`, since we say it's an `IntegerField`, it can be concluded it's 32-bit.

docs/releases/5.1.txt
csirmazbendeguz: thanks, done 👍

django/contrib/postgres/fields/serial.py
csirmazbendeguz: yes, I suppose it's clearer this way, thanks!

csirmazbendeguz:
Thanks a lot @sarahboyce , I addressed your comments, let me know what you think.

docs/ref/contrib/postgres/fields.txt
sarahboyce: Yes I like it 👍

docs/ref/contrib/postgres/fields.txt
sarahboyce: Agreed 👍

django/contrib/postgres/fields/serial.py
```diff
@@ -0,0 +1,99 @@
...
+        if self.null:
+            return [
+                checks.Error(
+                    f"{self.__class__.__name__} does not accept null values.",
+                    obj=self,
+                    id="fields.E013",
+                ),
+            ]
+        else:
+            return []
```
sarahboyce, line 40:
```suggestion
        if self.null is False:
            return []
        return [
            checks.Error(
                f"{self.__class__.__name__} does not accept null values.",
                obj=self,
                id="fields.E013",
            ),
        ]
```
(nit) let's make this the same structure as the other checks

django/contrib/postgres/fields/serial.py
```diff
@@ -0,0 +1,99 @@
...
+    db_returning = True
+
+    def __init__(self, *args, **kwargs):
+        kwargs["blank"] = True
+        super().__init__(*args, **kwargs)
+
+    def deconstruct(self):
+        name, path, args, kwargs = super().deconstruct()
+        kwargs.pop("blank")
+        return name, path, args, kwargs
```
sarahboyce, line 20: I can remove the `__init__` and `deconstruct` methods without any tests failing, can you add a test on why these are needed?

django/core/management/commands/inspectdb.py
```diff
@@ -123,6 +123,13 @@ def handle_inspection(self, options):
                     yield "# The error was: %s" % e
                     continue
 
+                def is_postgres_row(row):
+                    field_type, _, _ = self.get_field_type(connection, table_name, row)
+                    return field_type.startswith("postgres.fields.")
+
+                if any(is_postgres_row(row) for row in table_description):
+                    yield "from django.contrib import postgres"
```
sarahboyce, line 131: This is untested

django/db/backends/postgresql/operations.py
```diff
@@ -265,9 +278,7 @@ def sequence_reset_sql(self, style, model_list):
                             style.SQL_TABLE(qn(model._meta.db_table)),
                         )
                     )
-                    # Only one AutoField is allowed per model, so don't bother
-                    # continuing.
-                    break
```
sarahboyce, line 270: If I revert this, no tests fail

sarahboyce (commented):
Thank you for the updates @csirmazbendeguz
I think there are a few tests/asserts we need to add 👍

django/contrib/postgres/fields/serial.py
csirmazbendeguz: Okay, adjusted

django/core/management/commands/inspectdb.py
csirmazbendeguz: To be fair the other import is untested as well. But actually, this might be incorrect, let me check and fix.

django/db/backends/postgresql/operations.py
csirmazbendeguz: It should fail, did you run `test_sequence_reset_sql_statements` ?

csirmazbendeguz:
Thanks @sarahboyce , fair enough, I made some adjustments.

django/db/backends/postgresql/operations.py
InvalidInterrupt: I'm finally catching up on this PR; @ngnpope are you referring to the `uses_sequence` flag from this old PR for the same issue?
https://github.com/django/django/pull/12989/files#diff-ef0fe18fcc05a65a19d03870445490ac8a49a12bcfe3c1f8b6879d7b9cdc1d82R177

docs/ref/contrib/postgres/fields.txt
```diff
@@ -920,3 +920,24 @@ define :class:`~django.contrib.postgres.constraints.ExclusionConstraint`. See
...
+``Serial`` fields
+=================
+
+.. versionadded:: 5.2
+
+.. class:: SerialField(**options)
+
+    An automatically incrementing :class:`~django.db.models.IntegerField`,
+    much like :class:`~django.db.models.AutoField` except that it does not have
+    to be a primary key.
```
InvalidInterrupt, line 933:
```suggestion
    to be a primary key, and is not limited to a single instance per model.
```
This is also a significant difference.

InvalidInterrupt (commented):
I haven't really read many of the tests yet; not much has caught my eye that isn't already being discussed.

django/contrib/postgres/fields/serial.py
```diff
@@ -0,0 +1,98 @@
...
+    def _check_default(self):
+        if self.default is NOT_PROVIDED:
+            return []
+        return [
+            checks.Error(
+                f"{self.__class__.__name__} does not accept default values.",
+                obj=self,
+                id="fields.E014",
+            ),
+        ]
```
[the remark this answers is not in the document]
csirmazbendeguz: Thanks @InvalidInterrupt !

I don't think this is an issue, it's possible to create an object without using the sequence:

```
class Dummy(Model):
    serial = SerialField()

Dummy.objects.create(serial=0).serial  # 0
Dummy.objects.create(serial=0).serial  # 0
Dummy.objects.create(serial=0).serial  # 0
Dummy.objects.create().serial  # 1
Dummy.objects.create().serial  # 2
Dummy.objects.create().serial  # 3
Dummy.objects.create(serial=1).serial  # 1
Dummy.objects.create().serial  # 4
```

Is this what you're suggesting?

```
class Dummy(Model):
    serial = SerialField(default=0)

Dummy.objects.create().serial  # 0
Dummy.objects.create().serial  # 0
Dummy.objects.create().serial  # 0
Dummy.objects.create(serial=DatabaseDefault()).serial  # 1
Dummy.objects.create(serial=DatabaseDefault()).serial  # 2
Dummy.objects.create(serial=DatabaseDefault()).serial  # 3
Dummy.objects.create(serial=1).serial  # 1
Dummy.objects.create(serial=DatabaseDefault()).serial  # 4
```

I think the first one is better because it's consistent with how `AutoField` works.

docs/ref/contrib/postgres/fields.txt
csirmazbendeguz: Thanks!

django/contrib/postgres/fields/serial.py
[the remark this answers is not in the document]
sarahboyce: I've been testing on a little project, if we add a `SerialField` to an existing model I cannot make it nullable or assign a default and so I have to give a temporary default in order to create a migration and add the field. 

However, **whatever I give here is ignored** and the field is populated everywhere with as a serial starting from 1
I feel like it shouldn't prompt us in that case (forcing my input and then ignoring it to me is a bug).
Also considering this, a default option might be nice 🤔 (cc @LilyFoote)

django/core/management/commands/inspectdb.py
```diff
@@ -75,6 +76,7 @@ def handle_inspection(self, options):
                 "field names."
             )
             yield "from %s import models" % self.db_module
+
```
sarahboyce, line 79: @csirmazbendeguz these are minor clean up changes that will happen on the PR either by you or by a merger
- we want to limit the changed lines to keep the diff really tight to what's important, occasionally you have added a new line or removed a line, I will probably revert some of this (you can too if you like). This is considered formatting changes
- you have added `@tag("serial")` for your tests, this is not how we organise tests generally and needs to be removed
- the commits need to be squashed. This will likely be 1 commit (or if we can think of a nice way of splitting them we will)

sarahboyce: I added this for informational purposes 👍 

I forgot to say that I have tested this with a local project and tried out a number of things to try and see if I can find any issues (had a look in the admin,  reversed migrations, used them in `GeneratedFields`) and [this comment](https://github.com/django/django/pull/18123#discussion_r1616847144) is the only thing I'm concerned about. It looks really good ⭐

django/contrib/postgres/fields/serial.py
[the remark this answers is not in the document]
csirmazbendeguz: I see 🤔 Thanks for catching that @sarahboyce ! I tested it too but I didn't notice the the input was ignored. 😱 I'll investigate.

Yes, `default` makes sense in the context of migrations.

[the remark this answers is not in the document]
LilyFoote: Good catch! I think the least invasive change would be to adapt the migration generation code to know that `SerialField` can be created without an explicit default.

For a user-chosen default my question would be "why a `SerialField` and not an `IntegerField`?".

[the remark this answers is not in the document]
sarahboyce: Yes, I think that makes sense. 
Thinking some more, I think default would be misleading and I think what is being discussed is an optional `start` parameter. If that's wanted, it could be added in later 🤔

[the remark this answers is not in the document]
csirmazbendeguz: I added the `has_db_default` function to solve this the way @LilyFoote suggested.

csirmazbendeguz:
I realized instead of adding new system checks, I could just raise a `ValueError` in `__init__`. This is how `GeneratedField` does it. WDYT?

django/contrib/postgres/fields/serial.py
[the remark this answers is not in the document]
InvalidInterrupt: @csirmazbendeguz Yes, your second example is what I propose to allow. I don't think it's a question of choosing one or the other, your first example can continue to work as well.

> For a user-chosen default my question would be "why a `SerialField` and not an `IntegerField`?".

With `IntegerField`, the Django ORM/Migrations code won't manage the sequence for you, and you still want to use that sequence sometimes. `IntegerField` is only really a replacement if you're never going to use the sequence, not just if you don't want it to be the default.

> Thinking some more, I think default would be misleading and I think what is being discussed is an optional `start` parameter. If that's wanted, it could be added in later

A `start` parameter was not what I was proposing, but rather the ability to set a default other than "use the sequence by default". A `start` parameter might be a good idea for future development though.


I agree that this is a strange case (I don't think I would ever use it myself), but why should we impose more restrictions than the database, rather than simply setting reasonable defaults that the user could override?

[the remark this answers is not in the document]
csirmazbendeguz: Thanks for explaining @InvalidInterrupt !

I think the only place this could be useful is in migrations. So when adding a serial field, it could be pre-filled with a default number.

I would find this API confusing otherwise, users may confuse it with the `start` parameter for example.
I don't think it's too restrictive since users can still opt out of using the sequence by providing an explicit value.

That said, it's always easier to remove a restriction than to add one.

In my opinion, the use-case of migrations is an edge-case and it can be solved by first creating an `IntegerField` then changing it to a `SerialField`, so it wouldn't justify enabling this.

If someone can provide a good use-case this restriction could be removed in the future, but for now I have to disagree.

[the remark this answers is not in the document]
InvalidInterrupt: My description of the features potential use may have been confusing; but using `default` in this manner would be just like (every?) other field type, so I'm sceptical it would be confused for something like `start` outside of the context of this thread. In regards to API design, as it stands `SerialField` will be the exception. Eliminating that exception would probably also fix the migrations without adding a kludge to the migration code for this.

I've just finished mocking up the removal of this restriction to confirm that it works as expected and doesn't break anything else. I'm going to finish cleaning up the code and then push it to my fork for reference as a PoC.

That being said, I don't think this is a big issue; and it shouldn't hold up this PR. If I'm not persuading anyone that we should remove this guardrail and trust the Django users, then we should just move along for now.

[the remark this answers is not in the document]
InvalidInterrupt: Also, even `AutoField` allows you to set a `default`, even though that means that you'll only be able to save a single instance of the model without specifying an explicit value for the PK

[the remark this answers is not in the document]
csirmazbendeguz: I checked and `AutoField` allows a `default` too, it will cause integrity errors of course but it allows it. So yes you are raising a valid point, this would be an exception, thank you for pointing that out. It may be worth enabling for consistency's sake. @LilyFoote , @sarahboyce what do you think?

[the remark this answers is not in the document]
csirmazbendeguz: Actually, there's a problem with migrating a `SerialField` with a `default` value. It's impossible to do this in Postgres, since `serial` already implies the field has a default sequence.

django/db/backends/postgresql/schema.py
```diff
@@ -13,6 +13,21 @@ class DatabaseSchemaEditor(BaseDatabaseSchemaEditor):
...
+    sql_create_sequence = (
+        "CREATE SEQUENCE IF NOT EXISTS %(sequence)s "
+        "AS %(data_type)s "
+        "OWNED BY %(table)s.%(column)s"
+    )
+    sql_set_default_sequence = (
+        "ALTER TABLE %s ALTER COLUMN %s "
+        "SET DEFAULT nextval(pg_get_serial_sequence(%s, %s))"
+    )
+    sql_drop_default = "ALTER TABLE %(table)s ALTER COLUMN %(column)s DROP DEFAULT"
```
sarahboyce, line 25: I wonder if we need this, this isn't all that different to `BaseDatabaseSchemaEditor.sql_alter_column_no_default`

django/db/migrations/autodetector.py
```diff
@@ -1134,7 +1134,7 @@ def _generate_added_field(self, app_label, model_name, field_name):
         preserve_default = (
             field.null
             or field.has_default()
-            or field.db_default is not models.NOT_PROVIDED
+            or field.has_db_default()
             or field.many_to_many
             or (field.blank and field.empty_strings_allowed)
             or (isinstance(field, time_fields) and field.auto_now)
```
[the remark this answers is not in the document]
csirmazbendeguz: I still haven't opened a PR. The reason is I don't think it would help. If `has_db_default` is true, some code would try to use the `db_default`,  which would raise errors.

I'm looking for alternatives.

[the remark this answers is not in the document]
sarahboyce: No problem 👍

django/db/backends/postgresql/schema.py
csirmazbendeguz: I suppose I could combine that with `sql_alter_column`

csirmazbendeguz:
So I decided to stop fighting against what's already in place and try to conform to the existing migration scripts. I managed to achieve the following:
1. no `default` is allowed
2. adding a serial field to an existing table doesn't trigger a prompt
3. adding a serial field to an existing table will populate it with a sequence
4. changing a nullable integer to a serial field will populate the null values with a sequence starting from 1. it will also reset the sequence to the highest value after.
**QUESTIONS**: what would be the most sensible default to you?
1. on changing a nullable integer to a serial field, should we reset the sequence to the highest number?
2. is populating the null values with a sequence starting from 1 ok? or should we try populate it starting with max(field)?

django/db/backends/postgresql/schema.py
csirmazbendeguz: You were right. I ended up removing the default value management code, Django already has a way to do it so there's no need to do it here.

django/db/migrations/autodetector.py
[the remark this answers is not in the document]
csirmazbendeguz: Instead of hacking around `default` / `db_default`, I just set them to a real expression when creating the object. So there's no need to have a `had_db_default` function.

tests/inspectdb/tests.py
```diff
@@ -246,6 +248,28 @@ def test_number_field_types(self):
...
         )
 
+        if connection.vendor == "postgresql":
+            assertFieldType(
+                "serial_field",
+                "postgres.fields.SerialField()  "
+                "# You may want to consider using AutoField instead.",
+            )
+
+    @skipUnless(connection.vendor == "postgresql", "PostgreSQL specific SQL")
```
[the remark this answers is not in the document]
csirmazbendeguz: Good idea, thanks!

tests/schema/tests.py
```diff
@@ -5719,3 +5738,649 @@ def test_ci_cs_db_collation(self):
             if connection.vendor == "postgresql":
                 with connection.cursor() as cursor:
                     cursor.execute("DROP COLLATION IF EXISTS case_insensitive")
+
+    @unittest.skipUnless(connection.vendor == "postgresql", "PostgreSQL specific")
+    @isolate_apps("schema")
+    def test_serial_to_integer(self):
+        from django.contrib.postgres.fields import SerialField
+
+        self.serial_to_integer_test(SerialField, IntegerField)
```
charettes, line 5747: I kind of wish we tested all these in something like `postgres_tests/test_schema` but given this is a pre-existing pattern here import testing against `django.contrib.postgres` this doesn't have to be done in this PR.

django/contrib/postgres/fields/serial.py
```diff
@@ -0,0 +1,77 @@
...
+__all__ = ("BigSerialField", "SmallSerialField", "SerialField")
+
+
+class SerialFieldSequence(Expression):
+    allowed_default = True
+
+    def as_sql(self, compiler, connection):
+        table = self.field.model._meta.db_table
+        column = self.field.column
+        return f"nextval(pg_get_serial_sequence('{table}', '{column}'))", []
```
charettes, line 14: This can be an SQL injection vector and will crash if any table or column name includes a quote.

```suggestion
        table = self.field.model._meta.db_table
        column = self.field.column
        return "nextval(pg_get_serial_sequence(%s, %s))", [table, column]
```

I would personally use `Func` instead to avoid this error prone boilerplate

```python
class NextSerialSequence(Func):
    template = "nextval(pg_get_serial_sequence(%(expressions)s))"
```

And override `SerialFieldMixin._db_default_expression` to return

```python
@cached_property
def _db_default_expression(self):
    return NextSerialSequence(
        Value(self.model._meta.db_table),
        Value(self.column),
        output_field=self,
    )
```

Which should also avoid the `db_default` injection in `__init__` and removal in `deconstruct`.

django/contrib/postgres/fields/serial.py
```diff
@@ -0,0 +1,77 @@
...
+        name, path, args, kwargs = super().deconstruct()
+        del kwargs["default"]
+        del kwargs["db_default"]
+        return name, path, args, kwargs
+
+    def get_prep_value(self, value):
+        value = super().get_prep_value(value)
+        if value is None:
+            value = self.get_default()
+        return value
```
charettes, line 47: I'm surprised this is required given we set `default = DatabaseDefault()`?

csirmazbendeguz: It's for supporting `None` arguments:

`SerialModel.objects.create(serial=None)`

Since it's possible to provide `None` to auto fields, it should also be possible to provide it for serial fields.

If that's not necessary, we can remove it.

django/contrib/postgres/fields/serial.py
csirmazbendeguz: Thanks @charettes , fair enough. We should really set `db_default` to something though, overriding `_db_default_expression` is not enough because this variable is checked during migrations.

django/db/backends/postgresql/schema.py
```diff
@@ -13,6 +13,16 @@ class DatabaseSchemaEditor(BaseDatabaseSchemaEditor):
...
+    sql_create_sequence = (
+        "CREATE SEQUENCE IF NOT EXISTS %(sequence)s "
+        "AS %(data_type)s "
+        "OWNED BY %(table)s.%(column)s"
+    )
+    sql_reset_sequence = (
+        "SELECT setval(pg_get_serial_sequence(%s, %s), "
+        "coalesce(max(%s), 1), max(%s) IS NOT NULL) FROM %s;"
+    )
+    sql_rename_sequence = "ALTER SEQUENCE %s RENAME TO %s;"
```
charettes, line 25: Thinking more about it isn't there something fundamentally wrong here about not using the native `serial` types?

Why are we going the extra length of manually emulating through schema definition and explicitly using `nextval(pg_get_serial_sequence(...))` on `INSERT` what simply using `serial` and friends does?

I could be missing something here but it seems that most of the complexity of this patch stems from trying to emulate what Postgres does very well already?

csirmazbendeguz: @charettes , it's because Postgres doesn't support it very well. `serial` is not a true type, and it's not possible to alter a field's type to `serial`. we have to emulate it because that's the only way it can work.

charettes: Could you elaborate on how Postgres doesn't support `serial` very well? It has been used to power `AutoField` for over a decade before we switched to `IDENTITY` so I'm surprised to learn that it poses problems.

From my understanding the main reason for the manual sequence management and `db_default` assignment is to support altering back and forth between integer, auto, and serial fields. Is that really going to be a common enough use case that it warrants all the complexity it requires?

I think it's a question worth asking because there are no mentions of supporting this use case in the accepted ticket-27452 and I personally don't think that it will be common enough operation to warrant that much extra complexity to the schema editor.

Don't get me wrong, it is impressive that you got it working and I'm sure you learned a ton doing so. I'm just doubtful that the code complexity tradeoff is worth it (1.4k additional line) considering that this could be simply implemented using mixins including `db_returning = True` and using `get_db_type = "serial"` at first.

csirmazbendeguz: You can use `serial` when creating a table or when adding a field, but you can't use it for anything else, e.g. you can't alter a column's type to serial and you can't cast to serial. But do let me know if I'm wrong, it didn't work when I tested it.

I agree it adds some complexity, but the majority of that 1.4k lines is just tests, the schema editor received 160 additions and 44 removals only. I assumed the functionality to transition between integer fields <-> serial fields <-> auto fields would be a requirement because of the similarities between these fields. If all we wanted to support was creating serial fields, I'm not sure this would need any changes to core, it could be a package.

csirmazbendeguz: > From my understanding the main reason for the manual sequence management and db_default assignment is to support altering back and forth between integer, auto, and serial fields.

Yes this is correct.

csirmazbendeguz:
We had a discussion about this yesterday with @charettes about this:
1. We would like a 3rd opinion whether altering between serial <-> integer <-> auto fields is needed.
2. @charettes pointed out a potential race condition. When altering an integer field to a serial field, for example, there may be new records inserted before we reset the sequence. To fix this we would need to lock the column.
3. I haven't tested what happens if one of the operations fail. For example, what if the sequence is created but the program crashes? Are the operations rolled back in this case (is the sequence deleted)? This would require further testing.

django/db/backends/postgresql/schema.py
LilyFoote: I think a migration story is important here - I'm sure there will be people who would have chosen `SerialField` if it was available and now want to migrate over.

@csirmazbendeguz can you share the raw SQL that didn't work when you tested? Maybe that will inspire a better solution.

csirmazbendeguz: @LilyFoote , yes, I mean this one for example:

```sql
ALTER TABLE "schema_foo" ALTER COLUMN "field" TYPE bigserial USING "field"::bigserial
-- type "serial" does not exist
```

csirmazbendeguz:
Thanks @LilyFoote , @sarahboyce for the meeting.
Notes:
1. Let's open a separate PR adding the basic serial fields as @charettes suggests, we can merge that, and we can argue later if support for other functionality is necessary (can be added in a follow-up PR).
2. Is resetting the sequence after migration really necessary? Removing support for this extra functionality would remove the possibility of a race condition.

csirmazbendeguz:
Please refer this [PR](https://github.com/django/django/pull/18280) for a more basic implementation.

csirmazbendeguz:
[forum discussion](https://forum.djangoproject.com/t/serialfields-yay-or-nay/32541)

PR #18132, opened 2024-05-05
`prefetch_related_objects` fails to cache UUID FKs when the string representation of a UUID is used

selcuk:
# Trac ticket number

ticket-35434

# Branch description
Added unit test to reproduce the issue in the title.

# Checklist

### DIFF
--- a/tests/prefetch_related/test_uuid.py
+++ b/tests/prefetch_related/test_uuid.py
@@ -57,9 +57,7 @@ class UUIDPrefetchRelated(TestCase):
         pet = Pet.objects.create(name="Fifi")
         toy_ball = Toy(pet_id=pet.id, name="Ball")
         toy_bone = Toy(pet_id=str(pet.id), name="Bone")
-        prefetch_related_objects(
-            [toy_ball, toy_bone], "pet"
-        )
+        prefetch_related_objects([toy_ball, toy_bone], "pet")
         self.assertEqual(pet, toy_ball.pet)
         self.assertEqual(pet, toy_bone.pet)
 
--- a/tests/prefetch_related/test_uuid.py
+++ b/tests/prefetch_related/test_uuid.py
@@ -1,6 +1,7 @@
+from django.db.models import prefetch_related_objects
 from django.test import TestCase
 
-from .models import Flea, House, Person, Pet, Room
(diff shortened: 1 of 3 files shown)


tests/prefetch_related/test_uuid.py
```diff
@@ -52,6 +53,14 @@ def test_prefetch_related_from_uuid_model_to_uuid_model_with_values_flat(self):
             [pet.id],
         )
 
+    def test_prefetch_related_objects_str_as_uuid_fk(self):
+        pet = Pet.objects.create(name="Fifi")
+        toy_ball = Toy(pet_id=pet.id, name="Ball")
+        toy_bone = Toy(pet_id=str(pet.id), name="Bone")
+        prefetch_related_objects([toy_ball, toy_bone], "pet")
+        self.assertEqual(pet, toy_ball.pet)
+        self.assertEqual(pet, toy_bone.pet)
```
jacobtylerwalls, line 62:
```suggestion
        with self.assertNumQueries(0):
            self.assertEqual(pet, toy_ball.pet)
            self.assertEqual(pet, toy_bone.pet)
```

jacobtylerwalls (commented):
Thanks for the patch.

Had a look at whether the model with ManyToManyField could be used to test this and didn't find a way, so I think the new test model is fine.

One suggested change to clarify that the underlying functionality still works (and not merely doesn't crash).

Also verified this fixes the issue I encountered in my project (mentioned on ticket) 🚀 .

tests/prefetch_related/test_uuid.py
selcuk: That makes perfect sense. Thanks, applied the suggestion.

jacobtylerwalls:
I can mark this "ready for checkin" on Trac if you can address two things from the [patch review checklist](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/submitting-patches/#patch-review-checklist): 

- commit message & pr title: use ticket number and period, like: Fixed #XXXXX -- Avoided bad thing.
- release note in docs/releases/5.0.5.txt. Although it's not a regression in a recent version (I had this problem on 4.2), `ObjectDoesNotExist` inappropriately escaping qualifies as a "crashing bug" under my understanding of the backport policy. Someone may correct me on this point.

PR #18138, opened 2024-05-07
Update timezone documentation to reflect actual behavior (#35430).

koupro0204:
# Trac ticket number

ticket-35430

# Branch description
This branch addresses an inconsistency in the Django documentation regarding how time zones are handled in templates and forms. The existing documentation suggests that time zone-aware datetime objects are translated to the end user’s time zone. However, it has been observed that Django actually uses the TIME_ZONE settings. The proposed changes correct this documentation to accurately reflect the actual behavior: when time zone support is enabled, Django stores datetime information in UTC in the database, uses time-zone-aware datetime objects internally, and translates them to the :ref:default current time zone <default-current-time-zone>` in templates and forms. This update ensures that the documentation matches the implemented behavior, reducing potential confusion for developers.

# Checklist


docs/topics/i18n/timezones.txt
```diff
@@ -9,7 +9,7 @@ Overview
 
 When support for time zones is enabled, Django stores datetime information in
 UTC in the database, uses time-zone-aware datetime objects internally, and
-translates them to the end user's time zone in templates and forms.
+translates them to the :ref:`default time zone <default-current-time-zone>` in templates and forms.
```
sarahboyce, line 12:
```suggestion
converts them to the :ref:`current time zone <default-current-time-zone>` in
templates and forms.
```
Reading further below, I think "converts them to the current time zone" is what is most consistent.
We also need to wrap the docs at 80 characters wide (see [our docs guidelines](https://docs.djangoproject.com/en/5.0/internals/contributing/writing-documentation/#guidelines-for-restructuredtext-files)). 

Finally we will need a single commit (so make sure to squash/rebase any edits here) and use [our commit message format](https://docs.djangoproject.com/en/5.0/internals/contributing/committing-code/#committing-guidelines).

This is an example commit message
```
Fixed #35430 -- Corrected docs on timezone conversion in templates and forms.
```

@koupro0204 would you like to give that a go? If not, I can push up small updates later 👍

koupro0204: As stated, the default time zone is determined by the :setting:`TIME_ZONE` setting. Therefore, it should be referred to as "default" rather than "current". From my testing, it was found that it is determined by the setting's `TIME_ZONE` rather than the end user's actual time zone.

sarahboyce: https://docs.djangoproject.com/en/5.0/topics/i18n/timezones/#default-current-time-zone
It depends on whether the timezone is activated or not
> You should set the current time zone to the end user’s actual time zone with [activate()](https://docs.djangoproject.com/en/5.0/ref/utils/#django.utils.timezone.activate). Otherwise, the default time zone is used.

So unfortunately it's a little complicated for a small sentence. Maybe...

```suggestion
converts them to the end user's time zone in forms. By default, templates will
convert to the :ref:`default time zone <default-current-time-zone>`, but this
can be updated to the end user's time zone through the use of
:ref:`filters and tags <time-zones-in-templates>`.
```

koupro0204:
I'm new to this, and I'm not sure if the commits are correct. Could you please confirm if everything is okay?

sarahboyce:
> I'm new to this, and I'm not sure if the commits are correct. Could you please confirm if everything is okay?

Don't worry, it's all quite complicated. Ideally we'd have 1 commit, right now we have 4 (including a merge commit). If you have a read about git rebase/squash/reset etc I think that would help 
But I will push edits to the git history (don't worry about this too much)

koupro0204:
> Don't worry, it's all quite complicated. Ideally we'd have 1 commit, right now we have 4 (including a merge commit). If you have a read about git rebase/squash/reset etc I think that would help But I will push edits to the git history (don't worry about this too much)

thank you.
I will be able to study more and contribute more from now on!

sarahboyce (approved):

PR #18142, opened 2024-05-08
Fixed #35438 --Failed to insert an model instance into db that utilizes a UUID as its primary key and incorporates a GeneratedField.

zshongyi:
# Trac ticket number
[ticket-35438](https://code.djangoproject.com/ticket/35438#ticket)

# Branch description

I have defined a model that employs a UUIDField as its primary key, accompanied by a JSONField and several GeneratedFields, which are intended for indexing the JSONFields. However, when attempting to load data via loaddata from a JSON fixture, the process fails during the execution of the execute_sql method within django/db/models/SQL/compiler.py. The problematic code snippet is as follows:
```
def execute_sql(self, returning_fields=None):
    ....
            if not self.returning_fields:
                return []
   ....
            else:
                cols = [opts.pk.get_col(opts.db_table)]
                rows = [
                    (
                        self.connection.ops.last_insert_id(
                            cursor,
                            opts.db_table,
                            opts.pk.column,
                        ),
                    )
                ]
        converters = self.get_converters(cols)
        if converters:
            rows = list(self.apply_converters(rows, converters))
        return rows
```

Here, returning_fields comprises GeneratedFields, thereby preventing the method from directly returning an empty list at the outset of its logical execution. Moreover, since the model's primary key is not an AUTO_INCREMENT column (it's a UUID), last_insert_id() erroneously returns 0. This subsequently triggers an error within self.apply_converters(rows, converters) as it attempts to treat 0 as a UUID.

To address this issue, the else block logic requires adjustment to accommodate the presence of GeneratedField. Alternatively, within django/db/models/base.py's _save_table method, one could exclude GeneratedField from returning_fields. A modified approach might resemble:

# Checklist

### DIFF
--- a/django/db/models/base.py
+++ b/django/db/models/base.py
@@ -1138,7 +1138,7 @@ class Model(AltersData, metaclass=ModelBase):
                 for f in meta.local_concrete_fields
                 if not f.generated and (pk_set or f is not meta.auto_field)
             ]
-            returning_fields = meta.db_returning_fields
+            returning_fields = [f for f in meta.db_returning_fields if not f.generated]
             results = self._do_insert(
                 cls._base_manager, using, fields, returning_fields, raw
             )


sarahboyce:
Hi @zshongyi, this will need a test to be accepted 👍

jacobtylerwalls:
Thanks for the report. This issue was fixed in #19024.

PR #18146, opened 2024-05-08
Fixed #34034 -- Allow setting HTML attributes on Select widget options.

marianaasilva:
# Trac ticket number

Related to ticket-34034

# Branch description
Picks up on @khink's previous work to add the `option_attrs` parameter for widgets inheriting from ChoiceWidget.
I've updated the `ChoiceWidget.__deepcopy__` method and the test for it `ChoiceWidgetTest.test_deepcopy` and added a test for RadioSelect `RadioSelectTest.test_constructor_option_attrs` to ensure that `option_attrs` takes precedence.

# Checklist


nessita:
Hello @marianaasilva! Thank you for this work :star2: 

This PR needs some code style fixing (see CI errors). Once you solved those, please unset the relevant ticket flags following [these docs](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/submitting-patches/#patch-style) so this PR gets added back to the "branches needing review" section of the [Django Developer Dahsboard](https://dashboard.djangoproject.com/) .

docs/ref/forms/widgets.txt
```diff
@@ -195,6 +195,18 @@ Django will then include the extra attributes in the rendered output:
...
+For example, if you wish to set custom CSS classes for a RadioSelect's options, you could
+do as follows:
+
+.. code-block:: pycon
+
+    >>> from django.forms.widgets import RadioSelect
+    >>> widget = RadioSelect(choices=(("J", "John"),), option_attrs={"class": "special"})
+    >>> widget.render(name="beatle", value=["J"])
+    '<div><div>\n    <label><input type="radio" name="beatle" value="J" class="special" checked>\n John</label>\n\n</div>\n</div>'
+
```
sarahboyce, line 209: I am debating about whether this section makes most sense here or in [widgets inheriting from Select](https://docs.djangoproject.com/en/5.0/ref/forms/widgets/#widgets-inheriting-from-the-select-widget). If it remains here, it might be nice to link it

This also needs a `versionchanged:: 5.2` note to state that `option_attrs` is new

django/contrib/admin/widgets.py
```diff
@@ -500,6 +500,7 @@ def __init__(self, field, admin_site, attrs=None, choices=(), using=None):
         self.db = using
         self.choices = choices
         self.attrs = {} if attrs is None else attrs.copy()
+        self.option_attrs = {}
```
sarahboyce, line 503: Should we add a test in `admin_widgets/test_autocomplete_widget.py` where `option_attrs` are set?

sarahboyce (commented):
Thank you for this @marianaasilva and for being patient with us ❤️ 

I think we need to document `option_attrs` in our [`Select` widget docs](https://docs.djangoproject.com/en/5.0/ref/forms/widgets/#django.forms.Select) with a `.. versionadded:: 5.2` note and add a release note to Django 5.2

django/contrib/admin/widgets.py
marianaasilva: I'm not sure about this one 🤔 If I understood correctly, this change was added so it wouldn't break other tests, otherwise, it doesn't have any additional effect. Should we remove it from here and try to handle it differently?

docs/ref/forms/widgets.txt
```diff
@@ -744,6 +758,20 @@ that specifies the template used to render each choice. For example, for the
...
+        An optional dictionary containing option-specific HTML attributes to
+        be set on the rendered widget.
+
+        .. code-block:: pycon
+
+            >>> from django.forms.widgets import RadioSelect
+            >>> widget = RadioSelect(choices=(("J", "John"),), option_attrs={"class": "special"})
+            >>> widget.render(name="beatle", value=["J"])
+            '<div><div>\n    <label><input type="radio" name="beatle" value="J" class="special" checked>\n John</label>\n\n</div>\n</div>'
```
sarahboyce, line 774:
```suggestion
        be set on the options of the rendered widget.

        .. code-block:: pycon

            >>> from django.forms.widgets import Select
            >>> widget = Select(choices=(("J", "John"),), option_attrs={"class": "special"})
            >>> widget.render(name="beatle", value=["J"])
            '<select name="beatle">\n  <option value="J" class="special" selected>John</option>\n\n</select>'
```
As this example is on the Select widget, let's use the Select widget :+1:

sarahboyce (commented):
Thank you for the updates @marianaasilva :star: 
Added a few more thoughts :+1:

django/forms/widgets.py
```diff
@@ -709,9 +711,14 @@ def create_option(
...
     ):
         index = str(index) if subindex is None else "%s_%s" % (index, subindex)
-        option_attrs = (
-            self.build_attrs(self.attrs, attrs) if self.option_inherits_attrs else {}
-        )
+
+        if self.option_attrs:
+            option_attrs = self.build_attrs(self.option_attrs)
+        elif self.option_inherits_attrs:
+            option_attrs = self.build_attrs(self.attrs, attrs)
```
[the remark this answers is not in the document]
smithdc1: > I can't find why we have this and why it's False for Select but True for RadioSelect

The case of `required=True` is a good example here. With Select you want the HTML `required` attribute on the `<select>` element and not each `<option>`. For RadioSelect the `required` attribute needs to be on at least one "option" / `<input>`. 

Building on Sarah's observation above -- in addition to overriding a given `attrs` dictionary this patch also results in `required` and `id` attributes no longer being shown. 

``` python
class MyForm(forms.Form):
    select = forms.ChoiceField(
        choices=((None, "---"), ("1", "1"), ("2", "2")),
        widget=forms.RadioSelect(option_attrs={"option-attr": "test"}),
    )

print(MyForm())
```

before
``` HTML
<div>
<fieldset>
   <legend>Select:</legend>
   <div id="id_select">
      <div>
         <label for="id_select_0"><input type="radio" name="select" value="" required id="id_select_0" checked>---</label>
      </div>
      <div>
         <label for="id_select_1"><input type="radio" name="select" value="1" required id="id_select_1">1</label>
      </div>
      <div>
         <label for="id_select_2"><input type="radio" name="select" value="2" required id="id_select_2">2</label>
      </div>
   </div>
</fieldset>
```

after
``` html
<div>
<fieldset>
   <legend>Select:</legend>
   <div id="id_select">
      <div>
         <label><input type="radio" name="select" value="" option-attr="test" checked>---</label>
      </div>
      <div>
         <label><input type="radio" name="select" value="1" option-attr="test">1</label>
      </div>
      <div>
         <label><input type="radio" name="select" value="2" option-attr="test">2</label>
      </div>
   </div>
</fieldset>
```

Wasif-Shahzad:
I have this assigned this issue to myself on trac and was understanding things since a day. Currently, I have added a test for the issue in `RadioChoice` pointed out by @smithdc1 but I am not sure about how to work towards resolving it. Could there be any direction/guide? @sarahboyce 
Just to revise the bug
```py
        option_attrs = (
            self.build_attrs(self.attrs, attrs) if self.option_inherits_attrs else {}
        )
        # if self.option_attrs:
        #     option_attrs = self.build_attrs(self.option_attrs)
        # elif self.option_inherits_attrs:
        #     option_attrs = self.build_attrs(self.attrs, attrs)
        # else:
        #     option_attrs = {}
```
The commented code was added instead of the version above as per the last discussion but this removes the HTML attributes `required` and `id` from the `<input>` tag when making a `RadioSelect` widget on a `ChoiceField`.

smithdc1:
Closing in favour of follow up PR #21500

PR #18147, opened 2024-05-09
Updated grammar for "more-about-django-form-classes" documentation.

laaksomavrick:
# Trac ticket number
N/A

# Branch description

Minor grammatical nitpicks while reading the documentation re: forms.

# Checklist

### DIFF
--- a/docs/topics/forms/index.txt
+++ b/docs/topics/forms/index.txt
@@ -379,7 +379,7 @@ detail is rarely important.
 
 .. admonition:: Models and Forms
 
-    In fact if your form is going to be used to directly add or edit a Django
+    If your form is going to be used to directly add or edit a Django
     model, a :doc:`ModelForm </topics/forms/modelforms>` can save you a great
     deal of time, effort, and code, because it will build a form, along with the
     appropriate fields and their attributes, from a ``Model`` class.


AUTHORS
```diff
@@ -693,6 +693,7 @@ answer newbie questions, and generally made Django that much better:
     Mattia Procopio <promat85@gmail.com>
     Mattias Loverot <mattias@stubin.se>
     mattycakes@gmail.com
+    Mavrick Laakso <laakso.mavrick@gmail.com>
```
shangxiao, line 696: Unfortunately significant changes to the codebase are required to request adding oneself to the authors list, sorry 🤷‍♂️

docs/topics/forms/index.txt
```diff
@@ -379,9 +379,9 @@ detail is rarely important.
 
 .. admonition:: Models and Forms
 
-    In fact if your form is going to be used to directly add or edit a Django
+    In fact, if your form is going to be used to directly add or edit a Django
     model, a :doc:`ModelForm </topics/forms/modelforms>` can save you a great
-    deal of time, effort, and code, because it will build a form, along with the
+    deal of time, effort, and code. It will build a form, along with the
```
shangxiao, line 384: The only change I would make here is to remove the "In fact" altogether (and leave in the "because").

Rationale: Tech docs should be direct and to-the-point.

"If you from is going to be used directly…" reads more directly.  The "because" is actually beneficial here imho because it ties the statement with the reasoning.

shangxiao (changes requested):

shangxiao:
Hi 👋 thanks for the suggestion 🏆, though I'd do it a bit differently 🙂

AUTHORS
laaksomavrick: Understood. I will remove myself.

docs/topics/forms/index.txt
laaksomavrick: Addressed - thank you.

shangxiao (approved):
Thanks 👍 Will need to wait to see what a merger thinks.

sarahboyce:
Thank you both for this discussion. 
This is slightly clearer, but considering this would update the translation strings, I'm not sure it is worth the work it would give to others 🤔 
@laaksomavrick if you're looking to make a documentation update, picking up https://code.djangoproject.com/ticket/35441 might be a good start 👍

PR #18156, opened 2024-05-10
Fixed #35434 -- Advised validating unfetched instances before testing result cache.

jacobtylerwalls:
# Trac ticket number

ticket-35434

# Branch description
Clarify in the discussion of queryset caching that validating the primary key field on unfetched instances may be necessary when testing against the result cache.

# Checklist

### DIFF
--- a/docs/topics/db/queries.txt
+++ b/docs/topics/db/queries.txt
@@ -896,6 +896,40 @@ reuse it:
     >>> print([p.headline for p in queryset])  # Evaluate the query set.
     >>> print([p.pub_date for p in queryset])  # Reuse the cache from the evaluation.
 
+.. admonition:: Validate unfetched instances before comparing to the cache
+
+    Because the result cache consists of Python objects, you need may need to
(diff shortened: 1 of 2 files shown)


docs/topics/db/queries.txt
```diff
@@ -896,6 +896,40 @@ reuse it:
...
+
+            def __str__(self):
+                return f"{self.headline} ({self.day})"
+
+        >>> from datetime import date
+        >>> entry1 = DailyEntry.objects.create(
+                day=date(2020, 1, 1),
+                headline="Happy New Year!",
+            )
+        >>> entry2 = DailyEntry(day="2020-01-01", headline="Happy New Year!")
```
nessita, line 921: I'm not comfortable adding this to the docs because it could be seen as if this is a valid date field assignment when it's not (and it's my understanding that this is exactly Simon's point in the ticket). You could filter using the string but you should not do attribute assignment using any incorrect type.

PR #18168, opened 2024-05-15
Update SelectBox.js DOM text reinterpreted as HTML

Shivam7-1:
By using innerText, it will avoid the risk of HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML.

# Checklist


Shivam7-1:
Hi @sarahboyce Could You Please Review This PR 
Thanks

sarahboyce:
This is assigning the content to an empty string rather than some users input. So I don't see how this is a vulnerability.
If you believe you have found a security vulnerability, you should send an email to security@djangoproject.com (see [details](https://docs.djangoproject.com/en/5.0/internals/security/)) and keep it private.

PR #18172, opened 2024-05-16
Fixed #35443 -- Changed ordinal to return negative integers as is.

S-Tornqvist:
# Trac ticket number

ticket-35443

# Branch description
"-1st", "-2nd" and so on are incorrect words. Function `ordinal` in module `humanize` has been changed to return negative number "as is", so that for example `-1` is converted to `"-1"`.

The changes corresponds to alternative 1 of the suggested solutions in ticket #35443, which was accepted by Sarah Boyce.

# Checklist

### DIFF
--- a/django/contrib/humanize/templatetags/humanize.py
+++ b/django/contrib/humanize/templatetags/humanize.py
@@ -24,12 +24,14 @@ register = template.Library()
 def ordinal(value):
     """
     Convert an integer to its ordinal as a string. 1 is '1st', 2 is '2nd',
-    3 is '3rd', etc. Works for any integer.
+    3 is '3rd', etc. Works for any non-negative integer.
     """
     try:
         value = int(value)
     except (TypeError, ValueError):
         return value
+    if value < 0:
+        return str(value)
     if value % 100 in (11, 12, 13):
         # Translators: Ordinal format for 11 (11th), 12 (12th), and 13 (13th).
         value = pgettext("ordinal 11, 12, 13", "{}th").format(value)
--- a/docs/ref/contrib/humanize.txt
+++ b/docs/ref/contrib/humanize.txt
@@ -143,3 +143,4 @@ Examples:
 * ``3`` becomes ``3rd``.
 
 You can pass in either an integer or a string representation of an integer.
+Negative integers are returned unchanged.
--- a/tests/humanize_tests/tests.py
+++ b/tests/humanize_tests/tests.py
@@ -55,6 +55,9 @@ class HumanizeTests(SimpleTestCase):
             "102",
             "103",
             "111",
+            "-0",
+            "-1",
+            "-105",
             "something else",
             None,
         )
@@ -70,6 +73,9 @@ class HumanizeTests(SimpleTestCase):
             "102nd",
             "103rd",
             "111th",
+            "0th",
+            "-1",
+            "-105",
             "something else",
             None,
         )


Siburg:
You beat me to it. I like the test with `-0`. As per the first comment on the ticket, you may want to tweak `docs/ref/contrib/humanize.txt` as well.

S-Tornqvist:
> You beat me to it. I like the test with `-0`. As per the first comment on the ticket, you may want to tweak `docs/ref/contrib/humanize.txt` as well.

Thanks :)

I committed a fixup ([8130b21](https://github.com/django/django/pull/18172/commits/8130b218f83db6861fe9315534b439a7a2447988)) for the docs. Unsure what the policy for this is, but I assume everything is squashed when accepted? Didn't want to alter the history with a force push.

Robinrswanson (changes requested):
Thank you for your contribution! I’ve reviewed the code, tests, and documentation, and they all appear to follow the guidelines. However, for the pull request itself, the commits need to be squashed into a single commit. Once that’s done, it can proceed with the review.

S-Tornqvist:
> Thank you for your contribution! I’ve reviewed the code, tests, and documentation, and they all appear to follow the guidelines. However, for the pull request itself, the commits need to be squashed into a single commit. Once that’s done, it can proceed with the review.

Great! The commits are now squashed. Thanks for the help.

django/contrib/humanize/templatetags/humanize.py
```diff
@@ -24,12 +24,14 @@
...
     """
     Convert an integer to its ordinal as a string. 1 is '1st', 2 is '2nd',
-    3 is '3rd', etc. Works for any integer.
+    3 is '3rd', etc. Works for any non-negative integer.
     """
     try:
         value = int(value)
     except (TypeError, ValueError):
         return value
+    if value < 0:
```
sarahboyce, line 33:
```suggestion
    if value <= 0:
```
I don't think we should support `0th` as it doesn't really make sense

tests/humanize_tests/tests.py
```diff
@@ -70,6 +73,9 @@ def test_ordinal(self):
             "102nd",
             "103rd",
             "111th",
+            "0th",
```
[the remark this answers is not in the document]
Siburg: I don't agree with that. I do think there are use cases for 0th. For example: the 0th law of thermodynamics. It may not be common, but it is valid usage.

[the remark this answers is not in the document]
S-Tornqvist: I agree with @Siburg: "0th" is uncommon but correct.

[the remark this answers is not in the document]
sarahboyce: Alright, as we already have 0 -> 0th let's keep it in case someone's using it.

docs/ref/contrib/humanize.txt
```diff
@@ -143,3 +143,5 @@ Examples:
 * ``3`` becomes ``3rd``.
 
 You can pass in either an integer or a string representation of an integer.
+Only non-negative integers are converted. Anything else, including negative
+integers and non-integers, is returned unchanged.
```
sarahboyce, line 147:
```suggestion
```
I think instead of this, let's take inspiration from the [`apnumber` docs](https://docs.djangoproject.com/en/5.0/ref/contrib/humanize/#apnumber)

```
Converts a non-negative integer to its ordinal as a string. Otherwise, returns the number.

Examples:

* ``0`` becomes ``0th``.
* ``1`` becomes ``1st``.
* ``2`` becomes ``2nd``.
* ``3`` becomes ``3rd``.
* ``1.5`` becomes ``1.5``.
* ``-4`` becomes ``-4``.
```

Siburg: That's actually trickier than you might think.
`1.5` becomes `"1st"`
`"1.5"` indeed stays as `"1.5"`

sarahboyce: Sounds like we should also make an update and test cases here 👍

Siburg: I may be overthinking this, but my current suggestion for documentation text start is:
"You can pass in a non-negative number, or a string representation for a non-negative integer. Anything else is returned unchanged".

I don't know how many examples and test cases we want to add. For example, it will work for `Decimal` and `Fraction` instances as well. Then again, we probably don't want to get silly.

sarahboyce: > You can pass in a non-negative number, or a string representation for a non-negative integer. Anything else is returned unchanged

This doesn't hold true for 1.5 -> "1st"

S-Tornqvist: Not sure I follow the logic for ordinals and floating points, but that sounds like another issue. Marking this conversation as resolved.

sarahboyce: You're adding docs here says that non-integers will be returned unchanged, this is not true. Either this should be removed or you need go handle floats.

S-Tornqvist: True, I'll get on removing it

S-Tornqvist: Replaced with:
```diff
You can pass in either an integer or a string representation of an integer.
+ Negative integers are returned unchanged.
```

sarahboyce (approved):
Thank you @S-Tornqvist ⭐

PR #18178, opened 2024-05-18
Fixed #35449 -- Fixed validation issue with SplitArrayField when remove_trailing_nulls=True

saJaeHyukc:
# Trac ticket number

[ticket-35449](https://code.djangoproject.com/ticket/35449)

# Branch description
The proposed changes ensure that trailing null values are correctly removed before validation occurs, and that errors are appropriately handled.

# Checklist

### DIFF
--- a/docs/releases/5.0.7.txt
+++ b/docs/releases/5.0.7.txt
@@ -9,4 +9,5 @@ Django 5.0.7 fixes several bugs in 5.0.6.
 Bugfixes
 ========
 
-* ...
+* Fixed a bug in ``SplitArrayField`` where validation did not work properly 
+  when ``remove_trailing_nulls=True`` (:ticket:`35449`).
--- a/tests/postgres_tests/test_array.py
+++ b/tests/postgres_tests/test_array.py
@@ -1339,6 +1339,22 @@ class TestSplitFormField(PostgreSQLSimpleTestCase):
             ],
         )
 
+    def test_invalid_char_length_with_remove_trailing_nulls(self):
+        field = SplitArrayField(
+            forms.CharField(max_length=2),
+            size=3,
+            remove_trailing_nulls=True,
+        )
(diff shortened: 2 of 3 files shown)


tests/postgres_tests/test_array.py
```diff
@@ -1339,6 +1339,22 @@ def test_invalid_char_length(self):
...
+        )
+        with self.assertRaises(exceptions.ValidationError) as cm:
+            field.clean(["abc", "", ""])
+        self.assertEqual(
+            cm.exception.messages,
+            [
+                "Item 1 in the array did not validate: Ensure this value has at most 2 "
+                "characters (it has 3).",
+            ],
+        )
```
sarahboyce, line 1356:
```suggestion
            field.clean(["abc", "", "", ""])
        self.assertEqual(
            cm.exception.messages,
            [
                "Item 1 in the array did not validate: Ensure this value has at most 2 "
                "characters (it has 3).",
                "Item 2 in the array did not validate: This field is required.",
                "Item 3 in the array did not validate: This field is required.",
            ],
        )
```
Sorry, I think the suggested test case wasn't exactly right.
I think because `required=True` we should have errors for missing Item 2 and 3 but because `remove_trailing_nulls=True` we can ignore that the list is too long 🤔

docs/releases/5.0.7.txt
```diff
@@ -9,4 +9,5 @@ Django 5.0.7 fixes several bugs in 5.0.6.
 Bugfixes
 ========
 
-* ...
+* Fixed a bug in ``SplitArrayField`` where validation did not work properly 
+  when ``remove_trailing_nulls=True`` (:ticket:`35449`).
```
sarahboyce, line 13: Only release blocker bugs get back ported, so in this case, a release note is not needed 👍

sarahboyce (commented):
Thank you for this @saJaeHyukc!
I have a couple of comments 👍

PR #18181, opened 2024-05-20
Refs #35189 -- Moved CSS details' cursor declaration to base.css.

nessita:
# Trac ticket number

ticket-35189

# Branch description
Split CSS change where the `cursor: pointer` CSS declaration is moved to the toplevel base.css file in preparation for landing PR #17910.

# Checklist

### DIFF
--- a/django/contrib/admin/static/admin/css/changelists.css
+++ b/django/contrib/admin/static/admin/css/changelists.css
@@ -159,7 +159,6 @@
     font-weight: 400;
     padding: 0 15px;
     margin-bottom: 10px;
-    cursor: pointer;
 }
 
 #changelist-filter details summary > * {
--- a/django/contrib/admin/static/admin/css/base.css
+++ b/django/contrib/admin/static/admin/css/base.css
@@ -219,6 +219,10 @@ fieldset {
     border-top: 1px solid var(--hairline-color);
 }
 
+details summary {
+    cursor: pointer;
+}
+
 blockquote {
     font-size: 0.6875rem;
     color: #777;


ngnpope (approved):
🪄

django/contrib/admin/static/admin/css/changelists.css
```diff
@@ -159,7 +159,6 @@
     font-weight: 400;
     padding: 0 15px;
     margin-bottom: 10px;
-    cursor: pointer;
```
MHLut, line 162: Does this break the cursor for `#changelist-filter h3`?

I wasn't sure about this yet, hence I hadn't removed it in the other PR.

ngnpope: Oh, missed that 🤦🏻

Although, should the `<h3>` have a pointer? I see that there are `<a>` tags within the `<h3>` tags anyway...

nessita: I don't think so, could you share a bit more what's your thinking?

The H3 under `#changelist-filter` looks like this with this change:
image

nessita: Good point Nick on why those are headers, initially I thought it was so screen readers announce that following the hierarchy:

```
H2 Filters -> H3 Show Counts
          |-> H3 By Author
...
```

But the details/summary in that sidebar are not using H3 :thinking:

nessita: As far as my testing goes, this does not break the existing `details summary` in the changelist filter, so I'll land and continue the work in PR #17910.

PR #18187, opened 2024-05-22
Fixed #35467 -- Replaced `urlparse` with `urlsplit`.

RealOrangeOne:
# Trac ticket number

[ticket-35467](https://code.djangoproject.com/ticket/35467)

# Branch description

Replaces `urllib.parse.urlparse` with `urllib.parse.urlsplit`. From testing, it's ~6x faster:

```
In [1]: import urllib.parse

In [2]: %timeit urllib.parse.urlparse("https://example.com")
1.52 µs ± 37.9 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [3]: %timeit urllib.parse.urlsplit("https://example.com")
258 ns ± 1.14 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

See also https://www.youtube.com/watch?v=ABJvdsIANds.

I've reviewed each changed call, see https://github.com/django/django/pull/18187#issuecomment-2125349295 for details.

# Checklist

### DIFF
--- a/django/test/client.py
+++ b/django/test/client.py
@@ -8,7 +8,7 @@ from functools import partial
 from http import HTTPStatus
 from importlib import import_module
 from io import BytesIO, IOBase
-from urllib.parse import unquote_to_bytes, urljoin, urlparse, urlsplit
+from urllib.parse import unquote_to_bytes, urljoin, urlsplit
 
 from asgiref.sync import sync_to_async
 
@@ -458,11 +458,7 @@ class RequestFactory:
         return json.dumps(data, cls=self.json_encoder) if should_encode else data
 
     def _get_path(self, parsed):
-        path = parsed.path
-        # If there are parameters, add them
-        if parsed.params:
-            path += ";" + parsed.params
-        path = unquote_to_bytes(path)
+        path = unquote_to_bytes(parsed.path)
         # Replace the behavior where non-ASCII values in the WSGI environ are
         # arbitrarily decoded with ISO-8859-1.
         # Refs comment in `get_bytes_from_wsgi()`.
@@ -647,7 +643,7 @@ class RequestFactory:
         **extra,
     ):
         """Construct an arbitrary HTTP request."""
-        parsed = urlparse(str(path))  # path can be lazy
+        parsed = urlsplit(str(path))  # path can be lazy
         data = force_bytes(data, settings.DEFAULT_CHARSET)
         r = {
             "PATH_INFO": self._get_path(parsed),
@@ -671,8 +667,7 @@ class RequestFactory:
         # If QUERY_STRING is absent or empty, we want to extract it from the URL.
         if not r.get("QUERY_STRING"):
             # WSGI requires latin-1 encoded strings. See get_path_info().
-            query_string = parsed[4].encode().decode("iso-8859-1")
-            r["QUERY_STRING"] = query_string
+            r["QUERY_STRING"] = parsed.query.encode().decode("iso-8859-1")
         return self.request(**r)
 
 
@@ -748,7 +743,7 @@ class AsyncRequestFactory(RequestFactory):
         **extra,
     ):
         """Construct an arbitrary HTTP request."""
-        parsed = urlparse(str(path))  # path can be lazy.
+        parsed = urlsplit(str(path))  # path can be lazy.
         data = force_bytes(data, settings.DEFAULT_CHARSET)
         s = {
             "method": method,
@@ -772,7 +767,7 @@ class AsyncRequestFactory(RequestFactory):
         else:
             # If QUERY_STRING is absent or empty, we want to extract it from
             # the URL.
-            s["query_string"] = parsed[4]
+            s["query_string"] = parsed.query
         if headers:
             extra.update(HttpHeaders.to_asgi_names(headers))
         s["headers"] += [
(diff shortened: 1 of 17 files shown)


charettes:
> There are a few places where this will likely change functionality, such that "params" (not query params) will be included in the "path" component of a parsed URL, which may impact matching based on path. URL routing doesn't appear to be impacted by this change, and most uses of urlparse don't touch the .path component. I'm not sure how much of an issue this is, and whether we want to keep using urlparse in locations where the .path value is used?

IMO there should not be any change of functionality. In areas that expect to deal with `.path` in any manner `urlparse` should continue to be used otherwise `urlsplit` can be used for performance reasons when it's not the case.

> After this change, urlparse is no longer used in the codebase.

I think the ticket was accepted on the basis that there might be areas where we unnecessarily use `urlparse` and could switch to `urlsplit` and not to necessarily get rid of all instances of `urlparse`.

RealOrangeOne:
> In areas that expect to deal with `.path` in any manner `urlparse` should continue to be used

Makes sense to me! I'll get this updated shortly.

> not to necessarily get rid of all instances of `urlparse`

That wasn't my intention, it was more a statement that I'd checked and updated each call site.

RealOrangeOne:
Most use cases of `urlparse` didn't touch the path, so they can be converted to `urlsplit` just fine. Most of those which do use `.path` simply parse the URL, mutate the querystring, then stick it back together, which is also fine (so long as `urlunsplit` is used).

There was 1 edge cases in the test client, where previously we were appending parameters back into `PATH_INFO`, which now gets simpler thanks to `urlsplit`. If I implement it incorrectly (ie don't add the parameters back with `urlparse`), unit tests catch it, so I'm confident this behaviour is maintained.

The only case I was less happy was in static file handling:

https://github.com/django/django/blob/b7c7209c67f742eda8184c46f139e0e1cb16a1f4/django/contrib/staticfiles/handlers.py#L39

We check the path starts with a parsed version of `STATIC_URL`. If `STATIC_URL` contained parameters, this could change behaviour. I've therefore left it as-is with `urlparse`. There's little performance gain here anyway, as it's only called once during app startup. This feels pretty niche, and unlikely to cause issues, but there's also no real benefit to changing anyway. I have however changed the code to use the named indexes, so it's clearer what's going on.

adamchainz (approved):
Thank you for picking this up @RealOrangeOne , and thoroughly auditing all the call sites. LGTM, although I didn’t check each site.

django/test/client.py
```diff
@@ -458,11 +458,7 @@ def _encode_json(self, data, content_type):
         return json.dumps(data, cls=self.json_encoder) if should_encode else data
 
     def _get_path(self, parsed):
-        path = parsed.path
-        # If there are parameters, add them
-        if parsed.params:
-            path += ";" + parsed.params
-        path = unquote_to_bytes(path)
+        path = unquote_to_bytes(parsed.path)
```
nessita, line 461: I'm a bit concerned about this change because the *original* logic is untested. In `main`, if you fully remove the two lines that re-adds params:

```python
        # If there are parameters, add them
        if parsed.params:
            path += ";" + parsed.params
```

No test fail. Would you be willing to have a separated, initial commit that adds a test for this logic?

docs/ref/urlresolvers.txt
```diff
@@ -215,7 +215,7 @@ view would raise a ``Http404`` error before redirecting to it::
         # modify the request and response as required, e.g. change locale
         # and set corresponding locale cookie
 
-        view, args, kwargs = resolve(urlparse(next)[2])
+        view, args, kwargs = resolve(urlsplit(next)[2])
```
nessita, line 218: Since we are changing this bit, could we also use `.path`  for clarity?
```suggestion
        view, args, kwargs = resolve(urlsplit(next).path)
```

nessita (commented):
I think this looks great! Thank you for your contribution :star2: 

Added some minor comments, let me know what you think!

django/test/client.py
RealOrangeOne: From my testing, I was able to make this code fail unit tests when it was configured incorrectly, although our cases might have been different. Given this is the test client, and thus quite important to be well-tested, I'll have a check through and see if there's some more opportunity to harden said tests.

nessita: Thanks @RealOrangeOne! Given your comment I retried my test, using a fully clean `main` checkout and I think you are right. With this change:
```diff
diff --git a/django/test/client.py b/django/test/client.py
index aa42c1f60a..49d616b1d0 100644
--- a/django/test/client.py
+++ b/django/test/client.py
@@ -460,8 +460,8 @@ class RequestFactory:
     def _get_path(self, parsed):
         path = parsed.path
         # If there are parameters, add them
-        if parsed.params:
-            path += ";" + parsed.params
+        # if parsed.params:
+        #     path += ";" + parsed.params
         path = unquote_to_bytes(path)
         # Replace the behavior where non-ASCII values in the WSGI environ are
         # arbitrarily decoded with ISO-8859-1.
```
There is at least one unit test that fails:
```
FAIL: test_url_parameters (test_client.tests.ClientTest.test_url_parameters)
Make sure that URL ;-parameters are not stripped.
...
AssertionError: '/unknown_view/' != '/unknown_view/;some-parameter'
- /unknown_view/
+ /unknown_view/;some-parameter
```
So I believe we are covered in this regard. Thanks for checking nevertheless!

PR #18190, opened 2024-05-22
Fixed #35472 -- Used temporary directory in test_imagefield.NoReadTests.

sarahboyce:
# Trac ticket number

ticket-35472

# Branch description

This should probably be backported to 5.1 👍


nessita (approved):
Looks and works great!

PR #18195, opened 2024-05-24
Fixed #35477 -- Fixed field required validation in forms inheriting from SetPassword.

fsbraun:
# Trac ticket number

ticket-35477

# Branch description

When adding additional validation to the `SetPassword` form in `django.contrib.auth` and they fail, the from's clean method assumed they were empty since it did only check for truthiness of the passwords.

This PR changes the check for empty passwords to them being empty strings.

@sarahboyce Thanks for providing those helpful tests!!

# Checklist

### DIFF
--- a/django/contrib/auth/forms.py
+++ b/django/contrib/auth/forms.py
@@ -154,14 +154,14 @@ class SetPasswordMixin:
         if not usable_password:
             return self.cleaned_data
 
-        if not password1:
+        if not password1 and password1_field_name not in self.errors:
             error = ValidationError(
                 self.fields[password1_field_name].error_messages["required"],
                 code="required",
             )
             self.add_error(password1_field_name, error)
 
-        if not password2:
+        if not password2 and password2_field_name not in self.errors:
             error = ValidationError(
                 self.fields[password2_field_name].error_messages["required"],
                 code="required",
--- a/tests/auth_tests/test_forms.py
+++ b/tests/auth_tests/test_forms.py
@@ -60,6 +60,21 @@ class TestDataMixin:
         )
 
 
+class ExtraValidationFormMixin:
+    def __init__(self, *args, failing_fields=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.failing_fields = failing_fields or {}
+
+    def failing_helper(self, field_name):
+        if field_name in self.failing_fields:
+            errors = [
+                ValidationError(error, code="invalid")
+                for error in self.failing_fields[field_name]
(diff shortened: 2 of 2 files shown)


sarahboyce:
Thank you for the PR @fsbraun ⭐ 
@uri-rodberg does this work for you?

nessita (commented):
Hey @fsbraun, thank you for your speedy work on this!

Could you please ensure that there is a test for every form inheriting from `SetPasswordMixin`? (`BaseUserCreationForm`, `SetPasswordForm` and `AdminPasswordChangeForm`)

uri-rodberg:
> Thank you for the PR @fsbraun ⭐ @uri-rodberg does this work for you?

How do I install this version with pip?

fsbraun:
Try `pip install git+https://github.com/fsbraun/django@ticket_35477`

uri-rodberg:
> Thank you for the PR @fsbraun ⭐ @uri-rodberg does this work for you?

This version works. Thank you.

django/contrib/auth/forms.py
```diff
@@ -154,14 +154,14 @@ def validate_passwords(
...
-        if not password1:
+        if password1 == "":
             error = ValidationError(
                 self.fields[password1_field_name].error_messages["required"],
                 code="required",
             )
             self.add_error(password1_field_name, error)
 
-        if not password2:
+        if password2 == "":
```
nessita, line 164: If this line is undone, no test fail! :scream:
Could you please ensure that there are assertions for this logic change?

nessita (changes requested):
Hey @fsbraun, thanks for adding tests! Could you please rename and move each test to its specific test class? Otherwise we get tests for other forms inside ` SetPasswordFormTest` which is about the form and not the mixin.

The relevant test cases are:

* `SetPasswordFormTest` (for `SetPasswordForm`)
*  `BaseUserCreationFormTest` (for `BaseUserCreationForm`)
*  `AdminPasswordChangeFormTest` (for `AdminPasswordChangeForm`)

I also added a comment about an untested change. Thank you! :star2:

django/contrib/auth/forms.py
fsbraun: @nessita That would be another three test cases. OK?

nessita: Yes! The more the merrier?

fsbraun: Actually, it also works by extending the single test per form. 👍

PR #18201, opened 2024-05-26
Improved email address normalisation

huncut2016:
# Branch description
### Problem Description

When working with email addresses, especially those hosted on Gmail, there are specific nuances that can cause logically equivalent email addresses to appear different. This can lead to issues such as duplicated accounts or unnecessary complexity in user management systems. The primary challenges include:

1. **Dots in Local Part**: Gmail ignores dots in the local part of the email address. For example, `john.doe@gmail.com` and `johndoe@gmail.com` are considered the same by Gmail, but they might be treated as different addresses by a system that doesn't recognize this rule.

2. **Plus Addressing**: Gmail allows users to add tags to their email addresses using a plus sign. For example, `john.doe+newsletter@gmail.com` is delivered to `john.doe@gmail.com`. However, this can create apparent duplicates if the tags are not stripped before processing.

3. **Case Sensitivity**: While email addresses are generally case-insensitive, the domain part should always be lowercased to ensure uniformity. Systems might incorrectly treat `John.Doe@gmail.com` and `john.doe@gmail.com` as different addresses due to case differences.

### Solution Description

The provided code aims to normalize email addresses, particularly addressing the peculiarities of Gmail addresses. This normalization process ensures that logically equivalent email addresses are treated as identical within a system. 

### Explanation

1. **Default Value Handling**: Ensures `email` is a non-empty string.
2. **Splitting Email**: Attempts to split the email into `email_name` (local part) and `domain_part` at the '@' character. If the split fails, the original email is returned as a lowercased string.
3. **Lowercasing the Domain**: The domain part is lowercased to ensure uniformity.
4. **Gmail-specific Normalization**: If the domain is either `gmail.com` or `googlemail.com`, the local part undergoes further normalization by removing dots and any substring after a plus sign, followed by lowercasing the local part.
5. **Reconstructing the Normalized Email**: The normalized email address is reconstructed by combining the processed local part and the lowercased domain part.

This function ensures that Gmail addresses are normalized according to Gmail-specific rules, while other domains are lowercased to maintain consistency.

# Checklist

### DIFF
--- a/django/contrib/auth/base_user.py
+++ b/django/contrib/auth/base_user.py
@@ -22,7 +22,11 @@ class BaseUserManager(models.Manager):
     @classmethod
     def normalize_email(cls, email):
         """
-        Normalize the email address by lowercasing the domain part of it.
+        Normalize the email address by:
+        - Lowercasing the domain.
+        - Lowercasing Gmail address entirely.
+        - Ignoring dots in the local part for Gmail addresses.
+        - Ignoring anything after a plus sign in the local part for Gmail addresses.
         """
         email = email or ""
         try:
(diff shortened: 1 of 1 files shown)


sarahboyce:
Changes such as this required an accepted ticket

PR #18203, opened 2024-05-28
Fixed #35402 -- Fixed crash in django_test_skips for unimported submodules.

jacobtylerwalls:
# Trac ticket number
ticket-35402

# Branch description
The logic in `django_test_skips()` was relying on undocumented behavior of `import_string()` to mostly succeed when given a module. (It is doc'd to accept classes/attributes.)

Failures could happen if given a module that hasn't yet been imported.

Alternative to adjusting import_string() to reliably accept modules, as #18103 does, which isn't its documented behavior.

# Checklist

### DIFF
--- a/django/db/backends/base/creation.py
+++ b/django/db/backends/base/creation.py
@@ -337,12 +337,23 @@ class BaseDatabaseCreation:
         from unittest import expectedFailure, skip
 
         for test_name in self.connection.features.django_test_expected_failures:
+            # test_method_name might actually be a class.
+            # And either way, test_case_name might actually be a module.
             test_case_name, _, test_method_name = test_name.rpartition(".")
             test_app = test_name.split(".")[0]
             # Importing a test app that isn't installed raises RuntimeError.
             if test_app in settings.INSTALLED_APPS:
-                test_case = import_string(test_case_name)
-                test_method = getattr(test_case, test_method_name)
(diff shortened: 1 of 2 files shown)


django/db/backends/base/creation.py
```diff
@@ -337,21 +337,37 @@ def mark_expected_failures_and_skips(self):
         from unittest import expectedFailure, skip
 
         for test_name in self.connection.features.django_test_expected_failures:
+            # test_method_name might actually be a class.
+            # And either way, test_case_name might actually be a module.
```
jacobtylerwalls, line 341: I wanted to refactor to avoid this going forward, but short of changing `import_string()` to start throwing when given a module, I don't think we have enough information to the tell the two cases apart.

timgraham:
I think it's worth trying this solution.

tests/backends/base/test_creation.py
```diff
@@ -296,6 +297,9 @@ def test_mark_expected_failures_and_skips(self):
                 "backends.base.test_creation.skip_test_function",
             },
         }
+        popped_module = sys.modules.pop("backends.base")
+        self.addCleanup(sys.modules.__setitem__, "backends.base", popped_module)
```
timgraham, line 301: I guess this adds test coverage for the new code? (but does the original path remained covered?) An explanatory comment is warranted.

PR #18211, opened 2024-05-30
Fixed #35380 -- Automated the screenshot generation process for docs.

nessita:
This is a draft! And this is far from a final version. Known things that needs improvement:

* the `news` app should not be under `tests`, it should not be run with every selenium test run
* the update of the screenshots should be a manual action and not something triggered without user control
* investigation is needed for "showing the mouse pointer"
* investigation is needed for "dropdown shown in the wrong screen place"
* window size tweaks may be necessary on a case by case basis (i.e. per test)

# Trac ticket number

ticket-35380

# Branch description
DRAFT: Initial attempt to programmatically generate screenshots for docs.

# Checklist
- [] I have added or updated relevant docs, including release notes if applicable.
- [] I have attached screenshots in both light and dark modes for any UI changes.


tests/news/tests.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def change_header_display(self, display):
+        self.selenium.execute_script(
+            f'document.getElementById("header").style.display = "{display}";'
+        )
+        self.selenium.execute_script(
+            f'document.getElementsByTagName("nav")[0].style.display = "{display}";'
+        )
+
+    @screenshot_cases(SIZES)
```
sarahboyce, line 62: As `take_screenshot` has been overwritten to always take a screenshot, I think we should use the `ChangeWindowSize` context manager directly. 
I suspect that because we change the window size happens first and then the top bar is hidden after, the screenshots are smaller than expected. I might be wrong, but if we hide first and then resize the screen, we might get closer to 800x800px
Also in some cases (such as the flat pages one) we may want to increase the screenshot size slightly, so this might give us more flexibility

EDIT: I see you've already mentioned this in your PR summary 👍

tests/news/tests.py
```diff
@@ -0,0 +1,155 @@
...
+    def test_user_actions(self):
+        from selenium.webdriver import ActionChains
+        from selenium.webdriver.common.by import By
+
+        users = [
+            ("Adrian", "Holovaty", False),
+            ("Admin", "User", True),
+            ("Jacob", "Kaplan-Moss", True),
+            ("Simon", "Willison", False),
+        ]
```
sarahboyce, line 72: (side topic to be ignored in the context of this PR) I know these are here because they are the creators of Django, I wonder if everyone knows this, or they just look like a list of male English-y names 🤔

django/contrib/admin/static/admin/css/base.css
```diff
@@ -223,6 +223,10 @@ details summary {
     cursor: pointer;
 }
 
+details summary:hover {
+    color: var(--link-hover-color);
+}
```
sarahboyce, line 228: I guess this needs to be pulled out? Not sure what this is for 🙂

tests/news/tests.py
```diff
@@ -0,0 +1,155 @@
...
+            )
+        self.selenium.get(self.live_server_url + reverse("admin:auth_user_changelist"))
+        self.hide_nav_sidebar()
+        self.change_header_display("none")
+
+        filter_link = self.selenium.find_element(By.CSS_SELECTOR, "details summary")
+        # Simulate a hover.
+        ActionChains(self.selenium).move_to_element(filter_link).perform()
+        self.take_screenshot("list_filter")
+
```
sarahboyce, line 92: I think we want to select a person before we click the actions dropdown 👍

sarahboyce (commented):
Thank you so much for starting this ⭐ 
This already looks fantastic 🚀 

I appreciate this is a WIP 👍 
Comments are (hopefully) helpful thoughts

On the mouse display, from what I've been reading, this doesn't seem to be supported by selenium as selenium is using "[a representation of a pointer device]( https://www.selenium.dev/documentation/webdriver/actions_api/mouse/)" rather than your actual mouse. Looks like [hacking in some javascript](https://stackoverflow.com/questions/67453285/the-way-to-show-mouse-cursor-using-selenium-in-python) _might_ work but we should check if the cursor is all that important

nessita:
From Sarah, regarding the misplaced dropdown in the screenshot:

> I think we need to trigger a redraw. I think because an element is hidden via javascript, we need the dropdown to be redrawn but it isn't as it's not a child of what got hidden (something like this)
Apparently resizing the window might cause a "redraw" and then the dropdown might appear in the right position.
So I think, hide stuff, resize, click dropdown, cross fingers, screenshot 📸 
https://stackoverflow.com/questions/8840580/force-dom-redraw-refresh-on-chrome-mac

sarahboyce:
Elements of the questions raised in https://github.com/django/django/pull/18290 might be relevent here

nessita:
Closing following forum and ticket conversations.

PR #18215, opened 2024-05-30
Fixed #15727 -- Added Content Security Policy (CSP) support.

robhudson:
#### Trac ticket number

ticket-15727

#### Branch description
This adds CSP support for both the enforced header and report-only header.

There are a number of `TODO` notes within this PR that I'm looking for feedback on. 

#### Checklist
- [x] I have checked the "Has patch" ticket flag in the Trac system (was already set from prior efforts.


django/middleware/security.py
```diff
@@ -63,4 +120,22 @@ def process_response(self, request, response):
                 "Cross-Origin-Opener-Policy",
                 self.cross_origin_opener_policy,
             )
+
+        # TODO: Consider if we want to exclude CSP from a default set of views in
+        # DEBUG mode, similar to django-csp.
+        # TODO: Consider if we want to ship with the `@csp_exempt` decorator, to exclude
+        # views from CSP headers.
+        # TODO: Consider if we want to allow per-view configuration of CSP headers via
+        # decorators, similar to django-csp.
```
robhudson, line 129: The django-csp project ships with 3 extra decorators, apart from `csp_exempt`:
* `csp` to set a CSP on a view, ignoring what is configured in the Django settings
* `csp_replace` to replace the value of a specified directive
* `csp_update` to add a value to a specified directive
 
I'm not sure if all 3 are fit for Django but I think `csp` would be useful, especially for 3rd party apps to be able to set CSP directives for their own views (e.g. something like wagtail admin, or even Django's own admin).

This does open up security concerns, however, if a 3rd party app has a weaker CSP and potentially allows a XSS attack vector.

docs/spelling_wordlist
```diff
@@ -238,6 +238,7 @@ iterable
 iterables
 iteratively
 ize
+javascript
```
adamchainz, line 241: Use correct capitalization instead: JavaScript

django/middleware/security.py
```diff
@@ -1,8 +1,58 @@
...
+# define their own CSP on their own views. There are security implications to
+# consider here, so this would need to be carefully considered.
+
+# TODO: Consider adding support for CSP hash sources, similar to nonce sources.
+
+# TODO: Consider adding a security check to warn if the CSP settings are empty
+# and no CSP headers will be sent. (docs/refs/checks.txt)
+
+
+def build_csp(config, nonce=None):
```
adamchainz, line 29: investigate caching at least parts of this, like:

https://github.com/adamchainz/django-permissions-policy/blob/cde2f2a9242819882621974d13b28099d23eb6eb/src/django_permissions_policy/__init__.py#L150-L151

Python is not the fastest for string manipulation, and since the policy won't change request-to-request, it's ripe for caching.

I guess the nonce part can't be cached - that can be glued on per-request.

docs/ref/middleware.txt
```diff
@@ -464,6 +466,69 @@ in the :setting:`SECURE_REDIRECT_EXEMPT` setting.
 
 .. _nginx: https://nginx.org/
 
+.. _content-security-policy:
+
+Content Security Policy
+~~~~~~~~~~~~~~~~~~~~~~~
```
adamchainz, line 472: I think all CSP information should be a separate howto page, like the clickjacking and csrf topics.

django/middleware/security.py
robhudson: With decorators now added to this PR, the method arguments can vary based on view decorator and per request, making it more challenging to cache this string. I can explore it more later once things stabilize.

django/middleware/security.py
robhudson: I decided to go with these sets of decorators, leaving out `csp_replace` and `csp_update` since they seem redundant and also a bit confusing.

To exempt a view from either or both CSP headers:
```python
@csp_exempt(enforced: bool = False, report_only: bool = False)
```

To override the configured Django CSP settings for custom CSP per view, where `policy` is the same as the Django settings:
```python
@csp_enforced(policy: dict[str, Any])
@csp_report_only(policy: dict[str, Any])
```

This highlights 2 different ways we can go with the 2 different CSP dispositions: enforced vs report-only...
1. Specify as method keyword arguments
2. Specify with separate methods

I'm indifferent on which is best but do feel that they should be consistent. Open for discussion on which option is preferred.

robhudson:
I refactored the code today after review from Adam. 

- The middleware is now its own middleware, not part of `SecurityMiddleware`
- Decorators were added. See comment above looking for feedback.
- I removed the docs for now but will continue to work on them.

What's left:

- Documentation
- A context processor test
- Possibly a security check to warn if the CSP settings are empty that no CSP headers will be sent
- Possibly verify that all specified directives match the spec
- - Pro: Could help prevent typos or invalid directives appearing in the header
- - Con: Django would need to keep up with the CSP spec as new directives get added
- Possibly review what views may need to be exempted from CSP
- Possibly review what views want to set their own CSP, but this could come in later commits

django/middleware/security.py
robhudson: I ended up circling around to 2 decorators:

1: `csp_exempt`


```python
def csp_exempt(enforced=True, report_only=True)
```

This can be used with or without args. E.g.
```python
@csp_exempt
def view(request):
   return HttpResponse("Exempt both CSP headers")

@csp_exempt(report_only=False)
def view(request):
   return HttpResponse("Enforced header is exempt")
```

2. `csp_override`

```python
def csp_override(config, enforced=True, report_only=True)
```

Usage example:
```python
@csp_override({"DIRECTIVES": {"img-src": [csp.SELF, "data:"]}})
def view(request):
    return HttpResponse("Overrides the policy for both enforced and report-only")

@csp_override({"DIRECTIVES": {"img-src": [csp.SELF, "data:"]}}, enforced=False)
def view(request):
    return HttpResponse("Overrides the policy only for report-only")
```

I think this approach maintains consistency in how they're used -- both decorators accept boolean arguments to configure which CSP header (enforced or report-only) they apply to. By consolidating the previously proposed override decorators into a single decorator, it simplifies usage and avoids the need to remember or stack multiple decorators.

robhudson:
I'm going to set this as ready for review. I think it's feature complete for a first that includes the middleware to set the CSP headers, a couple decorators to fine tune the CSP policy for views, and a context processor for accessing the nonce easily.

Inline there are a few `TODO` comments for further discussion. Feel free to add Github comments on those.

docs/ref/csp.txt
```diff
@@ -0,0 +1,279 @@
+.. _content-security-policy:
+
+=============================
+Content Security Policy (CSP)
+=============================
+
```
sarahboyce, line 6:
```suggestion

.. versionadded:: 5.2

```

docs/ref/settings.txt
```diff
@@ -2353,6 +2353,30 @@ Unless set to ``None``, the
 :ref:`cross-origin-opener-policy` header on all responses that do not already
 have it to the value provided.
 
+.. setting:: SECURE_CSP
+
+``SECURE_CSP``
+--------------
+
```
sarahboyce, line 2360:
```suggestion

.. versionadded:: 5.2

```

docs/ref/settings.txt
```diff
@@ -2353,6 +2353,30 @@ Unless set to ``None``, the
...
+If configured with Content Security Policy directives, the
+:class:`~django.middleware.csp.ContentSecurityPolicyMiddleware` sets the
+:ref:`Content-Security-Policy <content-security-policy>` header on all responses
+that do not already have it.
+
+.. setting:: SECURE_CSP_REPORT_ONLY
+
+``SECURE_CSP_REPORT_ONLY``
+--------------------------
+
```
sarahboyce, line 2372:
```suggestion

.. versionadded:: 5.2

```

sarahboyce (commented):
Hi @robhudson - I think you've not had a review as the PR was not in the review queue (see "Patches that need review" within [Reviewing patches](https://code.djangoproject.com/wiki/Reports#Reviewingpatches))
That's because this ticket currently has "Needs documentation" marked as "yes"

Can you add a release note to Django 5.2, add a couple `.. versionadded:: 5.2` notes, and then mark "Needs documentation" as "no" in the ticket so it appears in the review queue?

docs/ref/csp.txt
```diff
@@ -0,0 +1,279 @@
...
+1. Mitigating XSS attacks by preventing the execution of inline scripts and restricting which external
+   scripts can be loaded.
+2. Controlling which external resources (like images, fonts, or stylesheets) can be loaded by your site.
+3. Preventing unwanted framing of your site, protecting against clickjacking attacks.
+4. Reporting policy violations to a specified URL, allowing you to monitor potential attacks or
+   misconfigured directives.
+
+CSP provides a robust and flexible framework for securing your web application, giving you fine-grained
+control over the resources your site can use and interact with. By leveraging Django's CSP middleware,
+you can easily implement and manage these policies in your project.
```
shangxiao, line 29: My 2¢ which is likely to start a flamewar 😅: One of the main criticisms I often hear about the Django docs is that too much introductory/explanatory material is put into the reference section. Experienced devs just often want to get to the guts without having to scroll/skim past explanations.

My suggestion is this sort of thing ^ could potentially be moved to the `/topics/security` section with a bit dedicated to CSP, and this part here could just be a super super brief "CSP is xxx … for more info see <link to topic>"

django/views/decorators/csp.py
```diff
@@ -0,0 +1,67 @@
...
+        if report_only:
+            response._csp_config_ro = config
+        return response
+
+    def decorator(view_func):
+        @wraps(view_func)
+        async def _wrapped_async_view(request, *args, **kwargs):
+            response = await view_func(request, *args, **kwargs)
+            response = _set_config(response, config, enforced, report_only)
+            return response
```
sarahboyce, line 19: Please add test coverage for the async implementation

django/views/decorators/csp.py
```diff
@@ -0,0 +1,67 @@
...
+        if report_only:
+            response._csp_exempt_ro = report_only
+        return response
+
+    def decorator(view_func):
+        @wraps(view_func)
+        async def _wrapped_async_view(request, *args, **kwargs):
+            response = await view_func(request, *args, **kwargs)
+            response = _set_exempt(response, enforced, report_only)
+            return response
```
sarahboyce, line 48: Please add test coverage for the async implementation

sarahboyce (commented):
Thank you @robhudson - I added initial comments

docs/ref/csp.txt
```diff
@@ -0,0 +1,279 @@
...
+  impact due to the additional processing required for each request.
+
+* Third-party Scripts: If your application relies heavily on third-party scripts or services,
+  implementing a strict CSP can be challenging and may require careful configuration.
+
+* Report-Only Mode Limitations: While useful for testing, the report-only mode
+  (``SECURE_CSP_REPORT_ONLY``) doesn't actually prevent any violations, it only reports them.
+
+* Nonce Limitations: Nonces are single-use tokens. If you're caching pages that include nonces,
+  you'll need to implement additional logic to ensure the nonces are refreshed appropriately.
```
RealOrangeOne, line 268: Nit: My reading is that this isn't quite true. The main issue is that the nonce must match in both the header and cached content, rather than the token being only valid for a single request. If a HTML page serves a nonce, and uses it in a referenced JS file, it should be fine regardless of caching.

docs/ref/csp.txt
```diff
@@ -0,0 +1,279 @@
...
+
+2. In your templates, use the ``request.csp_nonce``:
+
+   .. code-block:: html
+
+      <script nonce="{{ request.csp_nonce }}">
+          // Your inline JavaScript
+      </script>
+
+The CSP middleware will automatically generate and include a nonce in the CSP header for each request.
```
RealOrangeOne, line 113: Suggestion: My reading of the middleware is that the nonce is only added if the nonce value is read at least once during the request. It might be worth calling that out to aid the "I've set up a nonce but it's not in the header" debugging.

docs/ref/csp.txt
```diff
@@ -0,0 +1,279 @@
...
+            enforced=False,
+        )
+        def report_only_view(request):
+            return HttpResponse("Only the report-only CSP header will be overridden")
+
+These decorators provide fine-grained control over the Content Security Policy
+on a per-view basis, allowing you to customize the policy as needed for specific
+parts of your application.
+
+Limitations and Considerations
```
RealOrangeOne, line 250: Question: `django-csp` has a warning about the potential impact of disabling CSP for a specific path prefix (or view, since that's not implemented right now). Should that warning be duplicated here? 

> Excluding any path on your site will eliminate the benefits of CSP everywhere on your site. The typical browser security model for JavaScript considers all paths alike. A Cross-Site Scripting flaw on, e.g., `excluded-page/` can therefore be leveraged to access everything on the same origin.

My understanding is that any page view which is exempt has the same impact as excluding a prefix?

tests/middleware/test_csp.py
```diff
@@ -0,0 +1,285 @@
...
+        )
+
+    def test_config_with_empty_directive(self):
+        policy = {"DIRECTIVES": {"default-src": []}}
+        self.assertPolicyEqual(self.build_policy(policy), "")
+
+    def test_nonce_sentinel(self):
+        assert csp.Nonce() == csp.Nonce()
+        assert csp.NONCE == csp.Nonce()
+        assert repr(csp.Nonce()) == "django.middleware.csp.NONCE"
```
RealOrangeOne, line 141: Nit: `assertEqual`

docs/ref/settings.txt
```diff
@@ -2353,6 +2353,30 @@ Unless set to ``None``, the
...
 
+.. setting:: SECURE_CSP
+
+``SECURE_CSP``
+--------------
+
+Default: ``{}``
+
+If configured with Content Security Policy directives, the
+:class:`~django.middleware.csp.ContentSecurityPolicyMiddleware` sets the
```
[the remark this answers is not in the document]
robhudson: Is that what's throwing the docs errors? I couldn't figure that out and was sad not all the checks passed.

docs/ref/csp.txt
robhudson: Yes, the wording could be updated here. I think the _intention_ behind the nonce is single-use so I think it's worth mentioning potential considerations for caching, but to the browser as long as the nonce in the header matches the nonce in the HTML, it won't cause a CSP violation.

docs/ref/csp.txt
robhudson: Very true. I'll add a note about that. Thanks for pointing it out. 

I'll also add that if this is something we don't want to do in Django, it could be changed. I think it's helpful that if the nonce is never referenced it doesn't get added to the header, but it can also be a source of confusion. It could also be a source of confusion if the nonce is in header and not the content. But perhaps less so. I won't change anything now but just offering it up as open to discussion.

docs/ref/settings.txt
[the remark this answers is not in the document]
sarahboyce: Yes, this should be it 👍

docs/ref/csp.txt
RealOrangeOne: I believe the same principle is true for the CSRF token - it's possible there's some wording from those docs you can use as a basis.

docs/ref/csp.txt
RealOrangeOne: I also believe the implementation is that if the nonce is read, but not included in settings, it'll be added to the `script` tag, but not the header. In those cases, I assume the nonce is just ignored and no load restriction is put in. That could also be a potential source of confusion.

robhudson: Added the note below.

docs/ref/csp.txt
robhudson: I updated this to instead bring up the caching consideration and point out both header and content nonces need to match.

django/middleware/csp.py
```diff
@@ -0,0 +1,124 @@
...
+
+# TODO: Do we want to exclude CSP from a default set of views in DEBUG mode?
+
+# TODO: Should we add a security check to warn if the CSP settings are empty
+# that no CSP headers will be sent. (docs/refs/checks.txt)
+
+# TODO: Should we verify that all specified directives match the spec?
+# - Pros: helpful to developers in case of typos
+# - Cons: requires staying on top of the CSP spec.
+#         outdated Django versions may not be able to set newer directives.
```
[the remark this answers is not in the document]
robhudson: Thank you. Removed this TODO item. ✅

django/middleware/csp.py
```diff
@@ -0,0 +1,124 @@
+import base64
+import os
+from functools import partial
+
+from django.conf import settings
+from django.utils.deprecation import MiddlewareMixin
+from django.utils.functional import SimpleLazyObject
+
+# TODO: Do we want to exclude CSP from a default set of views in DEBUG mode?
```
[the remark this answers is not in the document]
robhudson: The django-csp project disables CSP headers when DEBUG is True for 404 and 500 responses, as Django's debug views may require features like inline styles that CSP could block (I haven't actually checked on this). Since DEBUG mode isn't for production, this seems reasonable. For other response codes, applying CSP headers in DEBUG mode helps with local testing and development.

docs/ref/csp.txt
robhudson: I added a bullet point about this and also reworked the text a bit. I also ordered the concerns from what I considered highest risk to lowest.

docs/ref/csp.txt
robhudson: I moved the intro to the `topics/security` section.

benjaoming: okay I'll close this one then: https://github.com/django/django/pull/18212

docs/ref/csp.txt
```diff
@@ -0,0 +1,296 @@
...
+      SECURE_CSP_REPORT_ONLY = {
+          "DIRECTIVES": {
+              "default-src": [csp.SELF],
+              "script-src": [csp.SELF],
+              "style-src": [csp.SELF],
+              # Add more directives as needed
+          }
+      }
+
+3. To use CSP nonces for inline scripts or styles, include ``csp.NONCE`` in your
```
benjaoming, line 56: Can we have a reference to a resource on what a "CSP nonce" even is? The documentation starts out with a nice and friendly welcoming to CSP, but this one stands out as something that's very un-defined.

Perhaps link to the subsequent "Nonce Usage" chapter and reference/introduce it there?

docs/ref/csp.txt
```diff
@@ -0,0 +1,296 @@
...
+      MIDDLEWARE = [
+          # ...
+          "django.middleware.csp.ContentSecurityPolicyMiddleware",
+          # ...
+      ]
+
+2. Configure your CSP policies in your ``settings.py`` using either
+   ``SECURE_CSP`` or ``SECURE_CSP_REPORT_ONLY``::
+
+      from django.middleware import csp
```
benjaoming, line 35: Is it safe/common to import from a module containing a middleware in this fashion or perhaps better to have a separate module like `django.middleware.csp.directives`?

I'm speculating that even if it works now, it could impede limitations on the future development of `django.middleware.csp` that it's supposed to be imported directly in the settings module.

docs/ref/csp.txt
```diff
@@ -0,0 +1,296 @@
...
+These decorators provide fine-grained control over the Content Security Policy
+on a per-view basis, allowing you to customize the policy as needed for specific
+parts of your application.
+
+Limitations and Considerations
+==============================
+
+While Content Security Policy (CSP) is a powerful security feature, it's
+important to be aware of its limitations and considerations, especially in the
+context of Django's implementation:
```
benjaoming, line 257: Great chapter! A bit long, but I get the intention and the ordering is good to have! Maybe a note about alternatively implementing CSP in the HTTP server layer?

Do you think this is too much _explanation_? I would think it's a tad heavy, so could consider moving it to the Security Topic. Other topics could be allowed to expand in the same fashion IMO.

docs/ref/settings.txt
```diff
@@ -2353,6 +2353,34 @@ Unless set to ``None``, the
...
+--------------------------
+
+.. versionadded:: 5.2
+
+Default: ``{}``
+
+If configured with Content Security Policy directives, the
+:class:`~django.middleware.csp.ContentSecurityPolicyMiddleware` sets the
+:ref:`Content-Security-Policy-Report-Only <content-security-policy-report-only>`
+header on all responses that do not already have it.
```
benjaoming, line 2382: These two settings currently have identical descriptions, is there a way to merge them and explain the difference in the description?

(if not, keeping them separate is fine)

How do we deal with conflicts between `SECURE_CSP_REPORT_ONLY` and `SECURE_CSP`? Which has the higher priority?

benjaoming (commented):
Exciting! Did a pass at documentation.

docs/ref/csp.txt
```diff
@@ -0,0 +1,296 @@
...
+    accessed at some point between when the middleware processes the request and
+    the response. This ensures the nonce is generated and included only when
+    necessary.
+
+    If the Django CSP settings do not explicitly include the ``csp.NONCE``
+    directive and ``request.csp_nonce`` is accessed in a template, the nonce
+    will appear in the content but not in the CSP header. This discrepancy may
+    lead to confusion or unexpected behavior.
+
+    To avoid such issues:
```
hvdklauw, line 110: Maybe when debug is True there should be some way to check this (check for the csp.NONCE in the settings if request.csp_nonce is accessed, that might still mismatch between where csp_nonce is used and where it is defined in the header, but it prevents the obvious mistake of not including it in any header.

Silently failing will just lead to lot of questions on discord about why doesn't this work.

tests/middleware/views.py
```diff
@@ -1,6 +1,11 @@
+import sys
+
 from django.http import HttpResponse
+from django.middleware import csp
```
carltongibson, line 4: Are we sure about this usage, or perhaps where this code lives? It feels a bit strange to be asking folks to import a middleware in order to use one of its constants in defining a policy. 🤔

docs/ref/csp.txt
```diff
@@ -0,0 +1,296 @@
...
+1. Add the CSP middleware to your ``MIDDLEWARE`` setting::
+
+      MIDDLEWARE = [
+          # ...
+          "django.middleware.csp.ContentSecurityPolicyMiddleware",
+          # ...
+      ]
+
+2. Configure your CSP policies in your ``settings.py`` using either
+   ``SECURE_CSP`` or ``SECURE_CSP_REPORT_ONLY``::
```
carltongibson, line 33: What happens if I set both? Should we be saying `and` here, one for each header, both of which can be used at the same time?

docs/ref/csp.txt
carltongibson: Yes, this point would seem better as part of the section below. 🤔

carltongibson (commented):
This looks good overall — nice stuff @robhudson 🎁 

I'm a little concerned about importing the middleware to use the constants... — are we sure about that? 

I didn't think the relationship between the two settings was 100% clear either. 

Not sure if this sneaks in for 5.2 or not, but it's going to be a super addition when it does — long time brewing 💃

Flimm:
How are third-party Django apps meant to modify the CSP settings of the Django project they are contained in? I'm interested in understanding the intended best practise, here.

Are they meant to modify their installation instructions in their documentation? Something like this:

> 1. Install the Django app `foobar` from PyPI: `pip install django-foobar`
> 2. Add `foobar` to your `INSTALLED_APPS` variable in your `settings.py` file:
>
>        INSTALLED_APPS = [
>            # ....
>            'foobar',
>        ]
>
> 3. Modify your `SECURE_CSP` or `SECURE_CSP_REPORT_ONLY` variables in your `settings.py` if you are using CSP to allow Foobar's dependencies to load:
>
>        from django.middleware import csp
>        SECURE_CSP = {
>            "DIRECTIVES": {
>                "script-src": [csp.SELF, 'https://example.og/foobar'],
>            },
>        }

Or are third-party Django apps meant to modify `SECURE_CSP` settings automatically in their code? They could use code  like this:

```

if not hasattr(settings, 'SECURE_CSP') or settings.SECURE_CSP is None:
    settings.SECURE_CSP = {}

settings.SECURE_CSP.setdefault('DIRECTIVES', {}).setdefault('script-src', [])
settings.SECURE_CSP['DIRECTIVES']['script-src'].append('https://example.org/foobar')
```

Also, are third-party Django apps meant to include the nonce in their templates, like this?

```
<script {% if CSP_NONCE %}nonce="{{ CSP_NONCE }}" {% endif %}>
    console.log("foobar");
</script>
```

I would love it if third-party Django apps made supporting CSP as easy as possible, and so I would be tempted to encourage those developers to modify `SECURE_CSP` automatically if needed, but I'm not sure if that is the intention of the developers of this configuration. I'm not sure if there are any drawbacks to this approach. I'm not excited about asking deevlopers to include more documentation in their installation instructions. If the settings need to change, then the installation instructions need to change too, and upgrading becomes a more manual process.

docs/ref/csp.txt
```diff
@@ -0,0 +1,296 @@
...
+        def my_view(request):
+            return HttpResponse("This view will only have an enforced CSP header")
+
+
+        @csp_exempt(enforced=False)
+        def another_view(request):
+            return HttpResponse("This view will only have a report-only CSP header")
+
+
+.. function:: csp_override(config, enforced=True, report_only=True)(view)
```
SebCorbin, line 203: For complex policies, should we provide an utility function to merge two policy dictionaries? This would be most useful for this decorator

collinanderson: Right, in a lot of cases I only want to add an additional `script-src` just for this view, without re-declaring the whole policy. It would be nice to be able to say something like `add=True`.

I suppose to avoid scope creep, that feature could be always added later and probably doesn't need to hold back getting this merged.

tests/middleware/views.py
collinanderson: I personally agree having to do two different import lines `middleware.csp` and `views.decorators.csp` would be super annoying.

But at the same time, I'd probably just use 'self', 'unsafe-inline', 'none', without importing the constants, but maybe that's somehow doing it wrong?

docs/ref/csp.txt
robhudson: I struggled with where to put these constants, while also making them relatively short to type. I do like that `csp.NONCE` both is short and also signifies the constant is related to `csp`. I originally wanted something like `from django.constants import csp` but couldn't find anything similar and didn't want to start a precedent on my own, so tried to follow convention as much as possible.

tests/middleware/views.py
robhudson: I mentioned in another comment that I was originally looking for a general place where Django has constants, something like `from django.constants import csp` but didn't see anything like that. Happy to move those wherever it makes the most sense if there's a suggestion?

docs/ref/csp.txt
robhudson: Setting both is good! Using the report-only CSP as the test bed for new policies is a good idea. The wording should be updated to suggest that - it wasn't intended to be one or the other, more that you configure either one as needed.

docs/ref/csp.txt
benjaoming: I see what you mean and how that could be a neat way to prefix things :bulb: 

Very tricky!

And no one would like to have to do deprecation and move around stuff like this in the future, so probably good to decide from all the options.

#### Alternative 1

Now I'm getting a bit too deep for my own knowledge, in the past, I often saw `import os.path` and then `os.path.exits(...)` - not sure if that's still considered good practice of if that would also call `csp.__init__.py` and thus nothing can really live in `middleware.csp` without being loaded in the settings module.

```python
from django.middleware import csp.constants

... = csp.constants.NONCE
```

#### Alternative 2

```python
from django.middleware.csp import constants as csp_constants

... = csp_constants.NONCE
```

#### Alternative 3

Or we can also have the very very verbose

```python
from django.middleware.csp.constants import CSP_NONCE

... = CSP_NONCE
```

More?

docs/ref/csp.txt
robhudson: Very true. On one of my projects we are testing new directives in the report-only policy and I imagine it would be a common case when doing so that you want to use the enforced policy as the "base" and just make a tweak here and there.

We have this little helper method we use to do that...
```python
def _override_csp(csp, append: dict[str, list[str]] = None, replace: dict[str, list[str]] = None):
    csp = deepcopy(csp)

    if append is not None:
        for key, value in append.items():
            if key in csp["DIRECTIVES"]:
                value = csp["DIRECTIVES"][key] + value
            csp["DIRECTIVES"][key] = value

    if replace is not None:
        for key, value in replace.items():
            csp["DIRECTIVES"][key] = value

    return csp
```

This isn't "complete", only what we've needed to do so far. A `remove` might also be helpful there to remove an item from one of the lists.

Since this is something that could be used for both the decorator and settings, perhaps it lives more as a utility method that could be used for both.

I also somewhat agree with the scope creep comment. A settings dict manipulation helper seems easy enough to throw together by a developer if needed. And could be something that is added later if enough need is gauged.

docs/ref/settings.txt
robhudson: There is no conflict between the two. A site can have both headers configured and it is recommended to do so when testing policy changes:
* `Content-Security-Policy` - the header for the *enforced* policy
* `Content-Security-Policy-Report-Only` - the header for the *report-only* policy

Since they are distinct settings it feels like they should have distinct sections here? There is a very slight difference in the `:ref:` pointing to the different headers.

tests/middleware/views.py
robhudson: Just wanted to point out that the constants can help typos as well as mistakenly missing the quoting of the headers like this:

```python
SECURE_CSP = {
    "DIRECTIVES": {
        "default-src": ["self"]
        # vs the correct ["'self'"]
    }
}
```

This is incorrect and would lead to a header of `default-src self` as opposed to `default-src 'self'`.

The constants make it a little easier to avoid that.

docs/ref/csp.txt
collinanderson: I also kinda like the ease of how `models.ForeignKey('MyModel', models.CASCADE)` works. Like if it was possible to allow importing the constants from the same spot as everything else. And then you'd have `@csp.override()`, etc.

tests/middleware/views.py
collinanderson: Ohh right excellent point. So weird that CSP domain names aren't quoted but 'self' is quoted.

robhudson:
I appreciate all the attention given today by various reviewers. I know it takes time and focus out of your day so I wanted to acknowledge that and say thank you. 🙏 

In an effort to give back I'd be happy to _quid pro quo_ and offer some attention or focus on anything you're working on. Feel free to DM me on the Django discord. 🤝 🌱

collinanderson:
> I appreciate all the attention given today by various reviewers. I know it takes time and focus out of your day so I wanted to acknowledge that and say thank you. 🙏
> 
> In an effort to give back I'd be happy to _quid pro quo_ and offer some attention or focus on anything you're working on. Feel free to DM me on the Django discord. 🤝 🌱

Thank _you_ for all your work on this!

docs/ref/csp.txt
robhudson: Since these constants are related to configuration in the settings, perhaps

```python
from django.conf import csp

... = csp.SELF
```

I just tested this in my local branch and it works fine. There's a few constants defined in `django/conf/__init__.py` so this doesn't feel like a _new_ precedent, per se, but adding a `django/conf/csp.py` file would be. But that feels like a more "right" place for these to live that I hadn't considered before.

robhudson: As of the last commit, the CSP constants are now defined in `django/conf/csp.py`.

docs/ref/csp.txt
robhudson: Which part is heavy? Just this section or the overall csp.txt file?

I built it with Sphinx and opened the HTML and it feels okay to me. I assume readers scan the headers mostly and dig in where needed. But open to suggestions on better structure or reducing text.

docs/ref/csp.txt
robhudson: In the latest commit I restructure the top level sections to flow a bit better, and moved this example into the section on nonces, as well as added some explanation on what nonces are and why one would use them.

tests/middleware/views.py
robhudson: > Are we sure about this usage, or perhaps where this code lives? It feels a bit strange to be asking folks to import a middleware in order to use one of its constants in defining a policy. 🤔

I moved this to `django/conf/csp.py` since it is about configuration. There are a few other constant-like things in `django/conf/__init__.py` so this made sense to me.

Now we can do `from django.conf import csp` to access `csp.SELF`, `csp.NONCE`, etc.

pauloxnet (commented):
First round of review with some code change suggestions

django/middleware/csp.py
```diff
@@ -0,0 +1,98 @@
...
+from django.utils.deprecation import MiddlewareMixin
+from django.utils.functional import SimpleLazyObject
+
+
+class ContentSecurityPolicyMiddleware(MiddlewareMixin):
+    def _make_nonce(self, request):
+        # Ensure that any subsequent calls to request.csp_nonce return the same value
+        if not getattr(request, "_csp_nonce", None):
+            request._csp_nonce = base64.b64encode(os.urandom(16)).decode("ascii")
+        return request._csp_nonce
```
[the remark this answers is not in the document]
robhudson: This doesn't store the `_csp_nonce` on the `request` object which is accessed below in the `get_policy` method.

If we change `get_policy` to use `request.csp_nonce` then that triggers `SimpleLazyObject` and makes it seen as accessed which bypasses how we avoid including the nonce in the header if it is never accessed during the req/resp cycle.

docs/ref/csp.txt
```diff
@@ -0,0 +1,296 @@
...
+Decorators
+==========
+
+.. module:: django.views.decorators.csp
+
+The examples below assume you are using function-based views. If you
+are working with class-based views, you can refer to :ref:`Decorating
+class-based views<decorating-class-based-views>`.
+
+.. function:: csp_exempt(enforced=True, report_only=True)(view)
```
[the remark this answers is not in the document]
robhudson: I've renamed `@csp_exempt` to `@csp_disabled` to make it a bit clearer and added a warning to the decorator docs that weakening or excluding CSP on one path can affect the whole site.

docs/ref/csp.txt
benjaoming: I like that a lot. Great solution @robhudson, didn't know that neither but last Friday I was thinking something inconclusive while biking about this :smile: 

It raises a positive question (that we don't need to answer now IMO) about how future directives and user-defined directives are handled as shared configuration data.

Somewhere down the road, I could see reusable apps checking if specific CSP directives exist and are prohibited/allowed in configuration. Not to say that it should make the Django-based CSP layer mandatory, but simply that the configuration of any layer could be visible in configuration.

benjaoming: I think you can resolve this (a bit confused, but I can't resolve my own comment from my own review currently)

docs/ref/csp.txt
benjaoming: I meant heavy on the explanation side. That's why I suggested that it could be a better fit over in the "security topics" section. I think the other topics there would be fine to expand in similar fashion to this chapter (considerations, limitations etc).

docs/ref/settings.txt
benjaoming: > A site can have both headers configured and it is recommended to do so when testing policy changes.

So what happens here? As a reader, I get confused about what I should do.. which takes precedence over the other, if they are defining the same directives? I can't actually find an answer myself in MDN documentation.

But MDN has this really helpful tip admonition, that maybe would be nice to include somewhere. The 2 settings are mentioned side-by-side a number of times and it could give the reader a nice fundamental helping tip to know the background of why BOTH settings are equally relevant:

> Note: Before implementing any actual CSP with the Content-Security-Policy header, you are advised to first test it out using the [Content-Security-Policy-Report-Only](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only) HTTP header; see [Report-only CSPs](https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/CSP#report-only_csps) below.

A version of that text could possibly go directly in the settings reference for the 2 variables?

docs/ref/csp.txt
robhudson: I thought the ref/csp doc was where we'd include more of the details and fleshed out explanations, and the ref/security doc was meant to be a summarized version of all security features in Django? When someone searches "django csp" they will likely land here rather than the security page, or at least that's the top result for me when I search CSRF -- I get the ref/csrf doc.

I'm happy to move things around -- I need to better understand the motivation and goals behind it.

robhudson:
After the 5.2 freeze, I took a break from this PR, but I consider this feature complete. I noticed there wasn’t any communication on why it didn’t make the 5.2 cutoff, so I'm unclear on what’s required for inclusion in 6.0. I'd appreciate any feedback or clarity on the next steps.

I'm happy to update the documentation, e.g. swapping any 5.2 references to 6.0, and address any further feedback to help get this accepted. Thanks!

nessita:
> After the 5.2 freeze, I took a break from this PR, but I consider this feature complete. I noticed there wasn’t any communication on why it didn’t make the 5.2 cutoff, so I'm unclear on what’s required for inclusion in 6.0. I'd appreciate any feedback or clarity on the next steps.

The sad reality is that many things didn’t make the 5.2 cutoff simply due to a lack of maintainer resources—nothing more. The final review, cleanup, and merging are primarily handled by the Fellows, and we worked through as many PRs as possible before the feature freeze, but unfortunately, we couldn’t get to this one. The fact is, we’re a tiny team (just 1.5 paid people, one full-time, one part-time), with other priorities that come before PR reviews, such as handling security reports and ticket triage.

> I'm happy to update the documentation, e.g. swapping any 5.2 references to 6.0, and address any further feedback to help get this accepted. Thanks!

If you can make one more time investment and do that update, it will certainly help having this is a fully reviewable state for when a Fellow gets to it from the review queue.

Thank you for your patience.

robhudson:
> The sad reality is that many things didn’t make the 5.2 cutoff simply due to a lack of maintainer resources

I totally understand and sympathize. You all are doing great work!

> If you can make one more time investment and do that update, it will certainly help having this is a fully reviewable state for when a Fellow gets to it from the review queue.

Not just one more, many more, as needed. I'm happy to help.

docs/ref/csp.txt
```diff
@@ -0,0 +1,374 @@
...
+- Used to define the **report-only policy**.
+- Sets the ``Content-Security-Policy-Report-Only`` HTTP header.
+- Monitors and reports violations without blocking content, useful for testing or debugging CSP rules.
+- A ``report-uri`` directive should be set to capture these reports.
+
+Structure of the Settings
+-------------------------
+Both ``SECURE_CSP`` and ``SECURE_CSP_REPORT_ONLY`` are dictionaries that must
+include a ``"DIRECTIVES"`` key. The value of this key is another dictionary
+containing CSP directives and their corresponding policies.
```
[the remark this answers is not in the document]
robhudson: > I don't understand why we don't have a flat dictionary and instead nest everything within DIRECTIVES - do we expect other keys?

The `DIRECTIVES` key allows room for future CSP wide configuration that aren't directive specific to keep policy logic separate from middleware behavior.

As this evolves I could imagine cases where a CSP middleware configuration may come up that could be added at the root of this setting while keeping directives separated:
- `"NONCE_GENERATOR"`: perhaps later we want to allow customization of how nonces are generated, e.g., to support external cryptographic standards or hardware security modules.
- `"POLICY_OVERRIDES"`: perhaps a general way to allow project defined CSP overrides. Maybe not adding CSP to a certain URL, or adding extra CSP in the case of login pages.
- Extensibility: It may be useful to allow developers to subclass the middleware and add a setting at this level for some custom functionality they want to include.

There is the YAGNI consideration, I agree. But if a need does arise in the future it would require a backwards incompatible change.

> In the current structure, I'd rather we combined the two settings ...

I explored this a bit when I first started out. I found the combined settings approach to be a bit messier for the implementation.

Here's a few reasons why I went forward with a 2 separate settings approach:
- 1 Setting = 1 Header – This aligns with how CSP is applied, since `Content-Security-Policy` and `Content-Security-Policy-Report-Only` are separate headers.
- I found there was less worry about if the `"enforced"` or `"report-only"` keys were in the setting or not. I found it simpler when working against the settings themselves. This isn't a strong argument, but it did simplify the implementation.
- Deriving report-only CSP from the enforce CSP is more straight-forward. It is often the case that you want to test your CSP changes by basing your report-only CSP on your enforced CSP and making changes from there. Having it as a single setting makes this a bit more tedious.

e.g. I've got something like this in one of my projects...
```python
    SECURE_CSP_REPORT_ONLY = deepcopy(SECURE_CSP)
    # Change the reporting endpoint
    SECURE_CSP_REPORT_ONLY["DIRECTIVES"]["report-uri"] = ...
    # Remove unsafe-inline for styles
    SECURE_CSP_REPORT_ONLY["DIRECTIVES"]["style-src"].remove(csp.UNSAFE_INLINE)
```

django/conf/csp.py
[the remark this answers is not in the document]
robhudson: > Adding a file in conf is inconsistent with how we handle constants generally

Yes I struggled for a place to put these constants. This seemed configuration related (`"conf"`) so it seemed like an okay choice.

> I don't think we should define the other keyword constants.

One nice thing about these is it is very easy to do something like this, which won't produce a policy as you expected:

```python
SECURE_CSP = {
  "DIRECTIVES": {
    "script-src": ["self"],
  }
}
```

The constants at least take care of the double quoting needed and potential spelling mistakes. The maintainability and back porting I understand, however, the spec moves slowly and the fallback is to just use the strings if the constant isn't defined.

> I don't understand why we are using a sentinel for NONCE

It was more a way to signify that this isn't just a string like `'self'` that goes in the header. It's a placeholder to specify where the generated nonce should go.

Seeing your diff where there's code like the following was confusing to me since it doesn't differentiate the intention if the raw string should end up in the header or not. I'm also very used to seeing the sentinel so fresh eyes are good.

```python
SECURE_CSP = {
  "DIRECTIVES": {
    "script-src": ["'self'", "'nonce'"],
  }
}
```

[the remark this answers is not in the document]
sarahboyce: Perhaps we should roughly copy the messages constants approach: https://docs.djangoproject.com/en/5.1/ref/settings/#message-level
This would mean:
- a new constants file e.g. `django/middleware/constants.py` 
- move these across (an enum is also tempting for these keywords :thinking:)
- we need each one to be documented e.g. https://docs.djangoproject.com/en/5.1/ref/contrib/messages/#message-level-constants

> One nice thing about these is it is very easy to do something like this

A different option could be that we add a system check to add some validation for this setting. I think that might be a nice option anyway given that we can use the strings directly anyway.

> the spec moves slowly

It looks to me there are already multiple added to the WIP spec (https://www.w3.org/TR/2025/WD-CSP3-20250206/#grammardef-keyword-source).

I am slightly concerned that we will get into a situation similar to response codes (https://forum.djangoproject.com/t/proposal-or-discussion-add-subclasses-for-no-content-and-created-httpresponses/39020/2) where we have to argue "just use it directly".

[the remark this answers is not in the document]
sarahboyce: > It was more a way to signify that this isn't just a string like 'self' that goes in the header. It's a placeholder to specify where the generated nonce should go.

My value suggestion was poor, it could be something like `"NONCE"`.
If it's a sentinel I think it might be difficult for folks to define it directly without importing the constants :thinking:

docs/ref/csp.txt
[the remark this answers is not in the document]
sarahboyce: This is convincing thank you :+1: 
I quite like the idea of us having a system check for the setting to ensure there is a "DIRECTIVES" key and check the names of the keys within it - unless there are valid reasons to have additional keys in `DIRECTIVES` we don't handle?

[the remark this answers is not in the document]
robhudson: Based on the feedback, it sounds like 2 settings, as it is coded now, is okay.

I like the suggestion to add a check. I'll implement the following changes:
- Add a system check to verify these settings contain a "DIRECTIVES" key.

Regarding CSP directive validation, I believe we discussed earlier (pre 5.2 freeze?) that attempting to stay current with the evolving spec presents challenges. Given the existence of [CSP extensions](https://www.w3.org/TR/CSP3/#directives-elsewhere) and possible spec changes, I would suggest it's preferable not to implement validation for the directives themselves.

The current implementation allows any directive and value that matches the settings configuration to be added to the header, which provides the flexibility needed. This approach intentionally leaves the responsibility of verifying CSP header correctness to the developer, which I believe is the right balance between functionality and maintainability. There are some online tools to validate CSP headers that we could suggest in the docs, perhaps.

django/conf/csp.py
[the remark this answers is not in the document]
robhudson: Thanks for providing all these helpful pointers and links to similar discussions. This gives me some good background.

I'll focus on exploring:
- Relocating the constants to a more appropriate location in the codebase
- Adding documentation for the constants, specifically noting that any undocumented CSP keywords can be represented as strings, with attention to the double quoting
- Investigating using a string for the nonce sentinel instead of the current implementation

All good suggestions, thank you!

tests/middleware/views.py
robhudson: @sarahboyce I just wanted to share this thread since I originally had the constants in `django/middleware/csp.py` and then moved them to `django/conf/csp.py`. 

Having them live in `django/middleware/constants.py` seems okay but does Carlton's original comment above still feel strange if we have `from django.middleware.constants import ... as csp_constants` -- importing from a middleware path into a view?

sarahboyce: Perhaps `django/utils/csp.py`?

sarahboyce (commented):
Thank you for the recent updates! 
I would like us to break this up into a couple of commits/PRs (which could be merged separately) to allow an easier review. This is my current idea:

- first commit adds the settings and middleware with no nonce support
- second commit adds nonce support
- third commit adds the decorators

What do you think?

robhudson:
> I would like us to break this up into a couple of commits/PRs (which could be merged separately) to allow an easier review. This is my current idea:
> 
>     * first commit adds the settings and middleware with no nonce support
>     * second commit adds nonce support
>     * third commit adds the decorators

That sounds like a good way to break it up. 

So that would be 3 new PRs with the code all squashed and separated out in those logical groupings. The docs and middleware may need a bit of unraveling to tease apart the nonce support, but totally doable. I'll get started on that over the next week. Thanks for the reviews!

docs/ref/csp.txt
```diff
@@ -0,0 +1,459 @@
...
+
+* Policy Exclusion Risks: Weakening or excluding a CSP policy on any page can
+  compromise the security of the entire site. Due to the "same origin" policy,
+  an attacker can exploit a vulnerability on one page to access other parts of
+  the site. Therefore, excluding any path from CSP protection negates its
+  benefits across the site.
+
+* Third-party Integration Complexity: If your application relies heavily on
+  third-party scripts or services, implementing a strict CSP can be challenging
+  and may require careful configuration and ongoing maintenance.
```
sarahboyce, line 430: I'm not sure we need to document this one

docs/ref/csp.txt
sarahboyce: I agree let's move this as a subheading within the CSP section for security

docs/ref/settings.txt
sarahboyce: I think the `SECURE_CSP` setting should mention that supporting browsers will block resources based off the header values and that `SECURE_CSP_REPORT_ONLY` should mention the reporting endpoint (I think the example would help)

docs/ref/middleware.txt
```diff
@@ -691,6 +713,11 @@ Here are some hints about the ordering of various Django middleware classes:
    After any middleware that modifies the ``Vary`` header: that header is used
    to pick a value for the cache hash-key.
 
+#. :class:`~django.middleware.csp.ContentSecurityPolicyMiddleware`
+
+   Can be placed near the bottom, but ensure any middleware that accesses ``request.csp_nonce`` is
+   positioned after it, so the nonce is properly included in the response header.
```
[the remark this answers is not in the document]
robhudson: Looking at the default list of middleware for a new project, I don't see anything that would affect it being first, since it just reads the settings and writes the header. The only tricky bit is if a nonce is used during content creation will affect whether the nonce is included in the header. So if there's a middleware that updates content in some way related to the nonce (unusual), that middleware would need to go after the CSP middleware.

docs/ref/settings.txt
```diff
@@ -2363,6 +2363,34 @@ Unless set to ``None``, the
 :ref:`cross-origin-opener-policy` header on all responses that do not already
 have it to the value provided.
 
+.. setting:: SECURE_CSP
+
+``SECURE_CSP``
```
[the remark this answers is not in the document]
robhudson: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#directives may also be a decent reference also. For each one it provides docs on the possible values.

docs/ref/csp.txt
```diff
@@ -0,0 +1,459 @@
+.. _content-security-policy:
```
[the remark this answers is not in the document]
sarahboyce: Point for @nessita, in the how-to doc it would be worth having an example of something that would be blocked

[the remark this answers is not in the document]
sarahboyce: Leaning towards using a topic rather than a how-to

django/middleware/csp.py
```diff
@@ -0,0 +1,124 @@
...
+                # Support boolean directives, like `upgrade-insecure-requests`.
+                if value[0] is True:
+                    value = ""
+                elif value[0] is False:
+                    continue
+                else:
+                    value = " ".join(value)
+            else:
+                continue
+            policy.append(f"{directive} {value}".strip())
```
adamchainz, line 123: Reading this loop, I was a little thrown off the track by the `value[0] is True` / `value[0] is False` checks. Then I realized that those special values were unconditionally being put into lists, only to be removed again.

This has the weird side effect that a list like `[True, CSP.NONE]` is allowed but nonsensical.

I think it's better to handle the specialized values separately:

```suggestion
        for directive, value in directives.items():
            if value is None:
                continue
            elif directive is True:
                rendered_value = ""
            elif directive is False:
                continue
            else:
                if not isinstance(value, list | tuple):
                    value = [value]
                if CSP.NONCE in value:
                    if nonce:
                        value = [f"'nonce-{nonce}'" if v == CSP.NONCE else v for v in value]
                    else:
                        # Remove the `NONCE` sentinel value if no nonce is provided.
                        value = [v for v in value if v != CSP.NONCE]
                if not len(value):
                    continue
                rendered_value = " ".join(value)
            policy.append(f"{directive} {rendered_value}".rstrip())
```

(I also made a minor optimization using `rstrip()` instead of `strip()`.)

django/middleware/csp.py
```diff
@@ -0,0 +1,124 @@
...
+                getattr(response, "_csp_config", None) or settings.SECURE_CSP or None
+            )
+        # Only provide the nonce if it exists on the request and is not empty.
+        nonce = getattr(request, "csp_nonce", None) or None
+        return (config, nonce)
+
+    # TODO: Make this cache-able?
+    # This is challenging due to:
+    # - Having the nonce, which is unique per request
+    # - The decorators pass in their own configs
```
adamchainz, line 96: Ah, yeah, let's forget the caching for now and drop the comment.

```suggestion
```

docs/ref/csp.txt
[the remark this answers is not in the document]
adamchainz: I think it should be a how-to, like [the CSRF one](https://docs.djangoproject.com/en/5.2/howto/csrf/). It'll need to include the **how to** content on configuring the middleware and using the decorators.

docs/ref/csp.txt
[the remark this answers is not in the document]
adamchainz: > The DIRECTIVES key allows room for future CSP wide configuration that aren't directive specific to keep policy logic separate from middleware behavior.

I'm feeling some YAGNI looking at the `DIRECTIVES` sub-key. I would rather we dropped to just a dict of directives.

The concerns you listed (`NONCE_GENERATOR`, `POLICY_OVERRIDES`, and "extensibility") can already be addressed in other ways, such as overriding the middleware or using decorators.

Meanwhile, a sub-key `DIRECTIVES` prevents the setting from mapping cleanly to the header contents, while analogous security header settings, such as `SECURE_REFERRER_POLICY`, do. (Although they contain simple string values.) So, having a "just in case" substructure feels slightly less Django-ey.

If we do end up with something else to change beyond the policy directives, we could add another setting or document how to subclass the middleware and swap out the appropriate method. The latter is likely to be preferable for fairly niche features.

adamchainz (changes requested):
Thank you for your Continuous Sustained Persistence (C.S.P.) here Rob.

I have a few comments here. The only thing I'd change on the design at this point is dropping the `DIRECTIVES` key, per the below.

docs/releases/6.0.txt
```diff
@@ -37,6 +37,48 @@ compatible with Django 6.0.
 What's new in Django 6.0
 ========================
 
+Content Security Policy (CSP) Support
```
adamchainz, line 40: We don't title-case headings
```suggestion
Content Security Policy (CSP) support
```

docs/releases/6.0.txt
```diff
@@ -37,6 +37,48 @@ compatible with Django 6.0.
...
+-------------------------------------
+
+Django 6.0 introduces built-in :ref:`Content Security Policy (CSP)
+<security-csp>` support, making it easier to protect your application against
+various types of content injection attacks. CSP helps you specify which content
+sources are allowed to be loaded by the browser.
```
adamchainz, line 46:
```suggestion
The new built-in :ref:`Content Security Policy (CSP) <security-csp>` support
makes it easier to protect your application against content injection attacks.
CSP is a security standard that allows you to specify which resources browsers
are allowed to load.
```

We don't normally say "Django <VERSION>" throughout the release notes. Also, I edited for brevity clarity.

docs/releases/6.0.txt
```diff
@@ -37,6 +37,48 @@ compatible with Django 6.0.
...
+      "django.middleware.csp.ContentSecurityPolicyMiddleware",
+      # ...
+  ]
+
+You can customize your CSP policy through Django settings:
+
+* :setting:`SECURE_CSP` for the enforced policy
+* :setting:`SECURE_CSP_REPORT_ONLY` for the report-only policy
+
+Here's an an example of a basic CSP config::
```
adamchainz, line 63:
```suggestion
Here's an an example of a basic CSP configuration::
```

Use full words wherever possible.

django/middleware/csp.py
```diff
@@ -0,0 +1,124 @@
...
+            config, nonce = self.get_policy(request, response)
+            if config:
+                response.headers[HEADER] = self.build_policy(config, nonce)
+
+        no_csp_header = HEADER_REPORT_ONLY not in response
+        is_not_disabled = not getattr(response, "_csp_disabled_ro", False)
+        if no_csp_header and is_not_disabled:
+            config, nonce = self.get_policy(request, response, report_only=True)
+            if config:
+                response.headers[HEADER_REPORT_ONLY] = self.build_policy(config, nonce)
```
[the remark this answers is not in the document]
robhudson: Nice clean-up. I also removed the constant on L67 in your suggestion. Thanks!

django/middleware/csp.py
```diff
@@ -0,0 +1,124 @@
...
+
+    The `{% if %}` block will only render if the nonce has been evaluated elsewhere.
+
+    """
+
+    def __init__(self):
+        super().__init__(self._generate)
+
+    def _generate(self):
+        return base64.b64encode(os.urandom(16)).decode("ascii")
```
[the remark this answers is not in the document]
robhudson: I like this. After looking I opted to use `secrets.token_urlsafe(16)` which is a base64 string instead of hex representing the same entropy but in 22 characters instead of 32.

django/middleware/csp.py
robhudson: I incorporated these changes and also did little more clean-up here. It's looking much better and simpler to reason about as you read it. Appreciate the suggestions.

robhudson:
I've addressed all the code-specific feedback discussed so far. My next steps are:

Documentation reorganization: I'll move implementation details from the reference guide into the how-to documentation, keeping `ref/csp.txt` focused on the technical API specification.

Breaking down the PR: As suggested, I'll split this into smaller, more manageable chunks:
1. First PR: Middleware with nonce support only
2. Follow-up PR: Decorators and context processor

django/core/checks/security/csp.py
```diff
@@ -0,0 +1,30 @@
...
+
+W026 = Warning(
+    "Each enabled CSP setting (SECURE_CSP, SECURE_CSP_REPORT_ONLY) must be a "
+    "dictionary.",
+    id="security.W026",
+)
+
+
+@register(Tags.security)
+def check_csp_directives(app_configs, **kwargs):
```
adamchainz, line 13:
```suggestion
def check_csp_settings(app_configs, **kwargs):
```

I think, since we dropped `DIRECTIVES`

django/core/checks/security/csp.py
```diff
@@ -0,0 +1,30 @@
+from django.conf import settings
+
+from .. import Tags, Warning, register
+
+W026 = Warning(
+    "Each enabled CSP setting (SECURE_CSP, SECURE_CSP_REPORT_ONLY) must be a "
+    "dictionary.",
```
adamchainz, line 7: Can we put the setting name in the message? Otherwise, it leaves the user hunting for where the mistake may be. I'm thinking of complex settings configurations where production has a report-only policy separate from some base configuration.

django/core/checks/security/csp.py
```diff
@@ -0,0 +1,30 @@
...
+    errors = []
+
+    # Check both CSP settings
+    for setting_name in ("SECURE_CSP", "SECURE_CSP_REPORT_ONLY"):
+        setting_value = getattr(settings, setting_name, None)
+
+        # Only validate if the setting is explicitly set (not None or empty dict)
+        if setting_value:
+            if not isinstance(setting_value, dict):
+                errors.append(W026)
```
adamchainz, line 28: Maybe we can do the below, so we still trigger a warning for an empty list and similar falsey values?

```suggestion
        if setting_value is not None and not isinstance(setting_value, dict):
           errors.append(W026)
```

django/middleware/csp.py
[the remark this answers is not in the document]
adamchainz: Brilliant, thanks!

docs/releases/6.0.txt
adamchainz: (This comment applies to all titles in the updated docs here.)

robhudson:
A new PR has been filed at https://github.com/django/django/pull/19393

The new PR:
* splits this into a smaller reviewable chunk focusing on the middleware and adjacent components.
* squashes the commits.
* addresses the requested changes to the docs (adding a `howto`, etc).

PR #18220, opened 2024-06-03
Made translation maps consistent in NullBooleanSelect widget.

Wecros:
# Trac ticket number
N/A

# Branch description
`format_value` should be consistent with `value_from_datadict` methods. Translation map options "True" and "False" were missing from the `format_value` method. This change ensures consistency of use between the methods, as well as practical passing of "True" value to the widget when rendering.

# Checklist
  - not applicable
  - not applicable
  - not applicable


Wecros:
Hello, is anyone planning on looking at this small PR?

sarahboyce:
> `format_value` should be consistent with `value_from_datadict` methods

This [isn't documented](https://docs.djangoproject.com/en/5.0/ref/forms/widgets/#django.forms.Widget.value_from_datadict) and so I'm not sure that's quite accurate

@smithdc1 can you help explain the difference between `format_value` and `value_from_datadict` and, hence, whether this is indeed a bug?

If there is a bug here, this would require a test update and an accepted ticket (which will help in getting reviews)

smithdc1:
I'm not sure about this (it's been like this ~forever). An explanation of the driver of the change and why the current behaviour is incorrect would be helpful. That would help get a ticket accepted, which would be required for this change.

Sarah, if i understand correctly...  Broadly, `value_from_datadict()` is doing the  `request` -> python conversion  and `format_data()` is doing the conversion the opposite way.

Wecros:
> I'm not sure about this (it's been like this ~forever). An explanation of the driver of the change and why the current behaviour is incorrect would be helpful. That would help get a ticket accepted, which would be required for this change.

Our use case was like this: we were getting `status=False` from the GET query (for filtering) on new page render, and we were confused why the widget was not correctly rendering the selected choice as false, but instead was rendering as the default option.

This was kind of confusing to us, since `value_from_datadict()` correctly assigns the `"False"` value from GET query to the python `False` value, making the filter work correctly. But the widget was not rendered in UI correctly because `"False"` is not a valid value in `format_value()` function.

We fixed this by only sending lowercase "true", "false" values which are handled in same way across the `NullBooleanSelect`, and to be honest sending anything else than "true"/"false" values doesn't make much sense since everything translates to them anyway in `format_value()`.

This is not a big deal, but I would love to know a scenario where it makes sense to define "True"/"False" mapping for `value_from_datadict()` and not define it for `format_value()` since there seems to be direct reverse relation between those two functions.

Simply said, adding the "True"/"False" values to the other function or removing them completely from the other function seems like the logical way to go for me, though I understand this might be too big of a change for something so insignificant.

PR #18226, opened 2024-06-07
Added support to add floating point numbers using |add template filter.

adnanAliSDE:
# Checklist
- [x] I have added support for floating point number addition using the add filter of the template system
- [x] This will be helpful for manipulation of ratings in e-commerce applications

# Why I am doing this PR
- When I was making an ecommerce application.I needed to some floating point number in order to show half and full stars in the product cards.But when I saw my rating was 4.8 and using rating|add:0.5 gave 4 I was very confused then I went to the source code and I thought typecasting value, add to float instead of integers will be helpful.
- If you find this change appropriate then please merge it.
- And if not please let me why we should use integer instead so that I may learn from you and understand your perspective.

### DIFF
--- a/django/template/defaultfilters.py
+++ b/django/template/defaultfilters.py
@@ -728,7 +728,7 @@ def unordered_list(value, autoescape=True):
 def add(value, arg):
     """Add the arg to the value."""
     try:
-        return int(value) + int(arg)
+        return float(value) + float(arg)
     except (ValueError, TypeError):
         try:
             return value + arg


smithdc1:
Thanks for the PR. 

It is documented that this template filter first coerces values to integers. Changing this now would be a breaking change and it's been like this ~forever.

I would suggest the best approach would be to create a custom template filter for your project if you have a different need. 

You could also start a discussion on the forum to gain wider consensus on this change. This would benefit from a compelling story on the backward compatability concerns.  

https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#add

adnanAliSDE:
Thank you smith for your advice.Now I have made a custom template filter in my project that gives me the HTML for showing stars corresponding to the product rating 👍

PR #18244, opened 2024-06-08
Fixed #33579 -- Added exception to be raised when no rows are affected on Model.save().

EnriqueSoria:
# Trac ticket number

ticket-33579

# Branch description
Change `DatabaseError` to a more specialized one, that inherits from `DatabaseError`, to make detecting this specific error easily, while mantaining backwards compatibility with old code (by inheriting from old exception)

# Checklist

### DIFF
--- a/django/db/models/base.py
+++ b/django/db/models/base.py
@@ -16,10 +16,12 @@ from django.core.exceptions import (
     FieldError,
     MultipleObjectsReturned,
     ObjectDoesNotExist,
+    ObjectNotUpdated,
     ValidationError,
 )
 from django.db import (
     DJANGO_VERSION_PICKLE_KEY,
+    DatabaseError,
     connection,
     connections,
     router,
@@ -47,7 +49,6 @@ from django.db.models.signals import (
     pre_save,
 )
 from django.db.models.utils import AltersData, make_model_tuple
-from django.db.utils import NoRowsAffected
 from django.utils.deprecation import RemovedInDjango60Warning
 from django.utils.encoding import force_str
 from django.utils.hashable import make_hashable
@@ -166,6 +167,20 @@ class ModelBase(type):
                         if hasattr(x, "_meta") and not x._meta.abstract
                     )
                     or (MultipleObjectsReturned,),
+                    module,
+                    attached_to=new_class,
+                ),
+            )
+            new_class.add_to_class(
+                "NotUpdated",
+                subclass_exception(
+                    "NotUpdated",
+                    tuple(
+                        x.NotUpdated
+                        for x in parents
+                        if hasattr(x, "_meta") and not x._meta.abstract
+                    )
+                    or (ObjectNotUpdated, DatabaseError),
                     module,
                     attached_to=new_class,
                 ),
@@ -1112,9 +1127,11 @@ class Model(AltersData, metaclass=ModelBase):
                 base_qs, using, pk_val, values, update_fields, forced_update
             )
             if force_update and not updated:
-                raise NoRowsAffected("Forced update did not affect any rows.")
+                raise self.NotUpdated("Forced update did not affect any rows.")
             if update_fields and not updated:
-                raise NoRowsAffected("Save with update_fields did not affect any rows.")
+                raise self.NotUpdated(
+                    "Save with update_fields did not affect any rows."
+                )
         if not updated:
             if meta.order_with_respect_to:
                 # If this is a model with an order_with_respect_to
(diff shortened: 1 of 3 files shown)


EnriqueSoria:
I'm not sure if I should change expected error in existing test (tests.force_insert_update.tests.ForceTests.test_force_update), or maybe keep that one (to ensure retrocompatibility reasons with DatabaseError) and add another one that actually checks that new exact Exception is raised

django/db/utils.py
```diff
@@ -26,6 +26,10 @@ class DatabaseError(Error):
     pass
 
 
+class NoRowsAffected(DatabaseError):
+    pass
```
charettes, line 30: I feel like it should live next to `ObjectDoesNotExist` / `MultipleObjectsReturned` as it's virtually the same concept

Maybe something like `NoObjectAffected` (that inherits from `DataBaseError` for unfortunate historical reasons)

charettes (commented):
Any thoughts about making it so the exception also has [per-model subclasses like `Model.DoesNotExist` and `.MultipleObjectsReturned`](https://github.com/django/django/blob/2ab106a02b94b9cbad3146501b978098463ecc3c/django/db/models/base.py#L145-L172) so when the following is done


```python
child.save(update_fields=["child_field", "parent_field"])
```

Then `_save_table`  can raise `cls.NoObjectsAffected` and the caller can differentiate between a lack of affected child and parent object.

Using this per-class trick could also allow only the subclasses to inherit from `DatabaseError` to preserve the previous behaviour and thus allow `NoObjectAffected` to be defined in `django.core.exceptions` by doing

```
new_class.add_to_class(
  "NoObjectsAffected",
  subclass_exception(
      "NoObjectsAffected",
      tuple(
          x.NoObjectsAffected
          for x in parents
          if hasattr(x, "_meta") and not x._meta.abstract
      )
      or (NoObjectsAffected, DatabaseError),
      module,
      attached_to=new_class,
  ),
)
```

ngnpope:
> Maybe something like `NoObjectAffected`

I think this should still be plural and I find "affected" to be strange wording… _(I know it came from the original error message.)_

Maybe `NoObjectsUpdated` or `ObjectsNotUpdated` with `Model.ObjectsNotUpdated` or even `Model.NotUpdated` for the copy attached to the model?

charettes:
@ngnpope totally open to more appropriate name. `NoObjectsUpdated` seems like a great one.

ngnpope:
@charettes Your suggestions are good. Ignore my reference to plural form - I had `QuerySet.update()` (multiple rows) in my head and not `Model.save()` (a single row).

django/db/utils.py
EnriqueSoria: I'm having issues placing `ObjectNotUpdated` next to `ObjectDoesNotExist` because I have to import `DataBaseError` and I run into circular import issues.

Any idea how to overcome it?

charettes: @EnriqueSoria the `ObjectNotUpdated` class defined in `django.core.exception` doesn't have to subclass `DataBaseError` so there should not be any circular import errors. Only [the subclasses](https://github.com/django/django/pull/18244#pullrequestreview-2106497667) defined in `models.base` should extent both and `DataBaseError` is already available there.

EnriqueSoria: That makes a lot of sense, thanks!

charettes: This can be removed now @EnriqueSoria as it's superseded by `ObjectNotUpdated`.

django/db/models/base.py
```diff
@@ -1112,9 +1127,11 @@ def _save_table(
                 base_qs, using, pk_val, values, update_fields, forced_update
             )
             if force_update and not updated:
-                raise DatabaseError("Forced update did not affect any rows.")
+                raise self.NotUpdated("Forced update did not affect any rows.")
```
charettes, line 1130: You should raise `cls.NotUpdated` to allow differentiation between different bases.

django/db/models/base.py
```diff
@@ -1112,9 +1127,11 @@ def _save_table(
...
             if update_fields and not updated:
-                raise DatabaseError("Save with update_fields did not affect any rows.")
+                raise self.NotUpdated(
```
charettes, line 1132: You should raise `cls.NotUpdated` to allow differentiation between different bases.

In the case of multi-table inheritance (also called MTI) getting to know which base failed to update might be useful.

charettes (commented):
Thanks for the adjustments Enrique!

I left some comments for required adjustments and this PR will also need tests (look into augmenting the ones around `DatabaseError` being raised on `save` today by searching the code base for `Forced update did`) as well documentation which you can draw inspiration from [`DoesNotExist`](https://docs.djangoproject.com/en/5.1/ref/models/class/#multipleobjectsreturned).

PR #18251, opened 2024-06-08
Fixed #35505 -- Added extrabody block to admin/base.html.

stefanivic:
# Trac ticket number

ticket-35505

# Branch description
The following change introduces a new template block inside the `admin/base.html`, just before the closing `</body>` tag that can be used to add any code after the footer section, like custom `<script>` tags. 

# Checklist

### DIFF
--- a/django/contrib/admin/templates/admin/base.html
+++ b/django/contrib/admin/templates/admin/base.html
@@ -121,5 +121,6 @@
   <symbol viewBox="0 0 24 24" width="1rem" height="1rem" id="icon-sun"><path d="M0 0h24v24H0z" fill="currentColor"/><path d="M12 18a6 6 0 1 1 0-12 6 6 0 0 1 0 12zm0-2a4 4 0 1 0 0-8 4 4 0 0 0 0 8zM11 1h2v3h-2V1zm0 19h2v3h-2v-3zM3.515 4.929l1.414-1.414L7.05 5.636 5.636 7.05 3.515 4.93zM16.95 18.364l1.414-1.414 2.121 2.121-1.414 1.414-2.121-2.121zm2.121-14.85l1.414 1.415-2.121 2.121-1.414-1.414 2.121-2.121zM5.636 16.95l1.414 1.414-2.121 2.121-1.414-1.414 2.121-2.121zM23 11v2h-3v-2h3zM4 11v2H1v-2h3z"/></symbol>
 </svg>
 <!-- END SVGs -->
+{% block extrabody %}{% endblock extrabody %}
 </body>
 </html>
--- a/docs/ref/contrib/admin/index.txt
+++ b/docs/ref/contrib/admin/index.txt
@@ -2827,6 +2827,32 @@ linked to the document in ``{% block dark-mode-vars %}``.
 
 .. _prefers-color-scheme: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme
 
+.. _extrabody:
+
+``extrabody`` block
+===================
+
+.. versionadded:: 5.2
+
+You can add custom HTML, JavaScript, or other content to appear just before the
+closing ``</body>`` tag of templates that extend ``admin/base.html`` by
+extending the ``extrabody`` block. For example, if you want an alert to appear
+on page load you could add a ``admin/base.html`` template override to your
+project:
+
+.. code-block:: html+django
+
+    {% extends 'admin/base.html' %}
+
+    {% block extrabody %}
+        {{ block.super }}
+        <script>
+            document.addEventListener('DOMContentLoaded', function() {
+                window.alert('Welcome!');
+            });
+        </script>
+    {% endblock extrabody %}
+
 ``AdminSite`` objects
 =====================
 
--- a/tests/admin_views/tests.py
+++ b/tests/admin_views/tests.py
@@ -1746,6 +1746,10 @@ class AdminCustomTemplateTests(AdminViewBasicTestCase):
         response = self.client.get(reverse("admin:admin_views_section_add"))
         self.assertContains(response, "bodyclass_consistency_check ")
 
+    def test_extended_extrabody(self):
+        response = self.client.get(reverse("admin:admin_views_section_add"))
+        self.assertContains(response, "extrabody_check\n</body>")
+
     def test_change_password_template(self):
         user = User.objects.get(username="super")
         response = self.client.get(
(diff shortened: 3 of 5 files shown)


docs/releases/5.1.txt
```diff
@@ -106,13 +106,16 @@ Minor features
...
+  closing body tag.
+
 :mod:`django.contrib.auth`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 * The default iteration count for the PBKDF2 password hasher is increased from
   720,000 to 870,000.
 
-* The default ``parallelism`` of the ``ScryptPasswordHasher`` is 
+* The default ``parallelism`` of the ``ScryptPasswordHasher`` is
```
knyghty, line 118: This change should be reverted.

knyghty (changes requested):
Thanks, looking good.

A few things I can immediately think of.

knyghty:
We should also add a note into the admin documentation that this block exists and how to override it.

docs/releases/5.1.txt
stefanivic: Thanks for the feedback, reverted.

docs/releases/5.1.txt
```diff
@@ -106,13 +106,16 @@ Minor features
 * :attr:`.ModelAdmin.list_display` now supports using ``__`` lookups to list
   fields from related models.
 
+* Added :ttag: {% extra_body %} template tag to the base template, above the
+  closing body tag.
```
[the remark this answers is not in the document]
stefanivic: Thanks for the feedback, moved to 5.2.

stefanivic:
> We should also add a note into the admin documentation that this block exists and how to override it.

Do you have any tips where we should put this ? I mostly see template tags in the docs, but I can't seem to find anything about the blocks.

knyghty:
I think the most similar documentation we have is this section: https://docs.djangoproject.com/en/5.0/ref/contrib/admin/#theming-support

Perhaps we could add another section underneath it?

stefanivic:
> I think the most similar documentation we have is this section: https://docs.djangoproject.com/en/5.0/ref/contrib/admin/#theming-support
> 
> Perhaps we could add another section underneath it?

Agreed, I have added the documentation in the latest change. Let me know if I need to work a bit on the wording.

knyghty (approved):
Everything seems okay to me though I'm also not sure on the wording.

stefanivic:
> Everything seems okay to me though I'm also not sure on the wording.

Thanks for the feedback, I improved the wording a bit to match the style of Django documentation a bit better.

docs/ref/contrib/admin/index.txt
```diff
@@ -2827,6 +2827,36 @@ linked to the document in ``{% block dark-mode-vars %}``.
...
+        {{ block.super }}
+        ...
+    {% endblock %}
+
+By extending the ``admin/base.html`` template and overriding the ``extra_body``
+block, you can insert any custom HTML, JavaScript, or other content that needs
+to appear just before the closing ``</body>`` tag. The ``{{ block.super }}``
+ensures that any existing content in the ``extra_body`` block from the parent
+template is preserved.
+
```
sarahboyce, line 2858:
```suggestion
The ``{{ block.super }}`` ensures that any existing content in the ``extra_body``
block from the parent template is preserved.
```
I would then remove this bit to have less duplication (there was also an extra new line that wasn't needed).

tests/admin_views/tests.py
```diff
@@ -8928,3 +8928,19 @@ def test_non_admin_url_404_if_not_authenticated(self):
...
+        cls.superuser = User.objects.create_superuser(
+            username="super1", password="secret", email="super2@example.com"
+        )
+
+    def setUp(self):
+        self.client.force_login(self.superuser)
+
+    def test_extra_body_block(self):
+        response = self.client.get(reverse("block_view"))
+        self.assertContains(response, "Test string for the extra body block")
```
sarahboyce, line 8946: I think we can maybe use `AdminCustomTemplateTests` and add `extrabody` to `tests/admin_views/templates/admin/base_site.html` (then we don't need a new template file and view etc).
I also think it would be nice that the assert shows that this is added just before the closing body tag so adding that in here would be nice 👍

django/contrib/admin/templates/admin/base.html
```diff
@@ -121,5 +121,7 @@
   <symbol viewBox="0 0 24 24" width="1rem" height="1rem" id="icon-sun"><path d="M0 0h24v24H0z" fill="currentColor"/><path d="M12 18a6 6 0 1 1 0-12 6 6 0 0 1 0 12zm0-2a4 4 0 1 0 0-8 4 4 0 0 0 0 8zM11 1h2v3h-2V1zm0 19h2v3h-2v-3zM3.515 4.929l1.414-1.414L7.05 5.636 5.636 7.05 3.515 4.93zM16.95 18.364l1.414-1.414 2.121 2.121-1.414 1.414-2.121-2.121zm2.121-14.85l1.414 1.415-2.121 2.121-1.414-1.414 2.121-2.121zM5.636 16.95l1.414 1.414-2.121 2.121-1.414-1.414 2.121-2.121zM23 11v2h-3v-2h3zM4 11v2H1v-2h3z"/></symbol>
 </svg>
 <!-- END SVGs -->
+
+{% block extra_body %}{% endblock %}
```
sarahboyce, line 125:
```suggestion
{% block extrabody %}{% endblock %}
```
Considering we have `extrastyle` and `extrahead`, this should perhaps be called `extrabody`

sarahboyce (commented):
Thank you @stefanivic 👍 
This looks great, I have a few minor comments

django/contrib/admin/templates/admin/base.html
stefanivic: @sarahboyce i was going after `extra_js` but considering the `extrahead` block, I agree. It's a better match.

stefanivic:
@sarahboyce thank you for taking the time to write a great review. I agree with the proposed changes and have implemented them. Let me know what you think.

docs/ref/contrib/admin/index.txt
```diff
@@ -2827,6 +2827,37 @@ linked to the document in ``{% block dark-mode-vars %}``.
 
 .. _prefers-color-scheme: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme
 
+.. _extrabody:
+
+``extrabody`` template block
+=============================
+
+You can insert any custom HTML, JavaScript, or other content that needs
```
smithdc1, line 2835:
```suggestion
=============================

.. versionadded:: 5.2

You can insert any custom HTML, JavaScript, or other content that needs
```

stefanivic: Thanks for the feedback, I missed this part in the contribution docs. Added now.

sarahboyce (approved):
Thank you @stefanivic 🥳 this looks great
I pushed minor tweaks 👍

PR #18256, opened 2024-06-08
Ticket #35507 changed to search landmark in filter sidebar in order to improve accessibility.

icastellanox:
# Trac ticket number

ticket-35507

# Branch description
changed to search landmark in filter sidebar in order to improve accessibility.

# Checklist
- [ x] This PR targets the `main` branch. 

### DIFF
--- a/django/contrib/admin/templates/admin/change_list.html
+++ b/django/contrib/admin/templates/admin/change_list.html
@@ -73,7 +73,7 @@
       </div>
       {% block filters %}
         {% if cl.has_filters %}
-          <nav id="changelist-filter" aria-labelledby="changelist-filter-header">
+          <search id="changelist-filter" aria-labelledby="changelist-filter-header">
             <h2 id="changelist-filter-header">{% translate 'Filter' %}</h2>
             {% if cl.is_facets_optional or cl.has_active_filters %}<div id="changelist-filter-extra-actions">
               {% if cl.is_facets_optional %}<h3>
@@ -85,7 +85,7 @@
               </h3>{% endif %}
             </div>{% endif %}
             {% for spec in cl.filter_specs %}{% admin_list_filter cl spec %}{% endfor %}
-          </nav>
+          </search>
         {% endif %}
       {% endblock %}
     </div>
--- a/tests/admin_views/tests.py
+++ b/tests/admin_views/tests.py
@@ -873,7 +873,7 @@ class AdminViewBasicTest(AdminViewBasicTestCase):
         response = self.client.get(reverse("admin:admin_views_thing_changelist"))
         self.assertContains(
             response,
-            '<nav id="changelist-filter" aria-labelledby="changelist-filter-header">',
+            '<search id="changelist-filter" aria-labelledby="changelist-filter-header">',
             msg_prefix="Expected filter not found in changelist view",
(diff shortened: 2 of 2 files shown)


thibaudcolas (approved):
Working great, thank you @icastellanox!

django/contrib/admin/templates/admin/change_list.html
```diff
@@ -73,7 +73,7 @@
       </div>
       {% block filters %}
         {% if cl.has_filters %}
-          <nav id="changelist-filter" aria-labelledby="changelist-filter-header">
```
sarahboyce, line 76: We also have `<form id="changelist-search" method="get" role="search">` in `django/contrib/admin/templates/admin/search_form.html` should we add an `aria-label` here as described in [labelling landmarks](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Search_role#labeling_landmarks)?

sarahboyce (commented):
Roughly adding the same comment in the ticket but this basically reverts the work of ticket-34835 which was introduced in 5.1, makes me think we should backport this to 5.1 before the beta freeze and mark this as a release blocker if we are instead saying this is in fact more accurate. Otherwise, we can close this ticket and PR as wontfix.

We will also need to update the release note added in ticket-34835.
cc @django/accessibility team for advice here

django/contrib/admin/templates/admin/change_list.html
knyghty: We should yes. For some reason I thought we had this already but it looks like I was looking at the wrong thing.

MHLut:
Is there anything still blocking this backport? It looks good to me.

sarahboyce:
[This comment](https://github.com/django/django/pull/18256#discussion_r1638101793) hasn't been addressed

django/contrib/admin/templates/admin/change_list.html
MHLut: It's better to use a heading and `aria-labelledby`, because:

1. Screen reader users can navigate using headings.
2. `aria-label` doesn't get translated by translation tools like (by not limited to) Google Translate.

gtossou: Hi. I would like to give a help on this ticket. Could you give me some explanation about the change required regarding the heading and aria-labelledby please ? Thanks

knyghty: In the template Sarah mentioned there's a `<form>` element with `role="search"`, which makes this element act the same as a `<search>` element. Since these types of landmark elements need labels, we should add _something_ here. I think if we go with the approach @MHLut suggests we should add a `<h2>` with `class="sr-only"`, an `id` and some descriptive text, maybe just "Search" is enough, or maybe "Search <model.verbose_name_plural>" - I don't think we want this visible - and then we can add that ID to the form's -aria-labelledby` attribute.

gtossou: Ok if i understand well, the finally solution would replace the `nav ` element with a `form `(having a role="search). And next to it a h2, right ?

smithdc1:
Continued in https://github.com/django/django/pull/19379

PR #18258, opened 2024-06-08
Fixed #24638 -- Added support for SQL comments

viciu:
This PR reactivates work done in #4495 and #15711 -  useful addition to Django for adding traceability for queries.

# Trac ticket number

ticket-24638

# Branch description

Add queryset comment() method

- [x] Rebase against 5.x
- [x] Support SELECT statement
- [x] Support UPDATE statement
- [ ] Support INSERT statement
- [ ] Support DELETE statement
- [x] use regex instead if "*/"

# Checklist

### DIFF
--- a/django/db/models/query.py
+++ b/django/db/models/query.py
@@ -1783,6 +1783,17 @@ class QuerySet(AltersData):
         clone._db = alias
         return clone
 
+    def comment(self, message):
+        """Add a comment to the query."""
+        clone = self._clone()
+        if "/*" in message or "*/" in message:
+            raise ValueError(
+                "Cannot pass strings containing /* or */ to comment(). "
+                "Escape or strip these delimiters before calling comment()."
+            )
+        clone.query.comments = (*clone.query.comments, message)
+        return clone
+
     ###################################
     # PUBLIC INTROSPECTION ATTRIBUTES #
     ###################################
--- a/django/db/models/query.py
+++ b/django/db/models/query.py
@@ -35,12 +35,16 @@ from django.db.models.utils import (
 from django.utils import timezone
 from django.utils.deprecation import RemovedInDjango60Warning
 from django.utils.functional import cached_property, partition
+from django.utils.regex_helper import _lazy_re_compile
 
 # The maximum number of results to fetch in a get() query.
 MAX_GET_RESULTS = 21
 
 # The maximum number of items to display in a QuerySet.__repr__
 REPR_OUTPUT_SIZE = 20
+
+# Comments are forbidden inside comments.
+COMMENT_FORBIDDEN_PATTERN = _lazy_re_compile(r"/\*|\*/")
 
 
 class BaseIterable:
@@ -1786,12 +1790,13 @@ class QuerySet(AltersData):
     def comment(self, message):
         """Add a comment to the query."""
         clone = self._clone()
-        if "/*" in message or "*/" in message:
+        if COMMENT_FORBIDDEN_PATTERN.search(message):
             raise ValueError(
                 "Cannot pass strings containing /* or */ to comment(). "
                 "Escape or strip these delimiters before calling comment()."
             )
-        clone.query.comments = (*clone.query.comments, message)
+        commented_message = f"/* {message} */"
+        clone.query.comments = (*clone.query.comments, commented_message)
         return clone
 
     ###################################
--- a/django/db/models/sql/compiler.py
+++ b/django/db/models/sql/compiler.py
@@ -787,6 +787,9 @@ class SQLCompiler:
                 result = ["SELECT"]
                 params = []
 
+                if self.query.comments:
+                    result += self.comments_sql()
+
                 if self.query.distinct:
                     distinct_result, distinct_params = self.connection.ops.distinct_sql(
                         distinct_fields,
@@ -1627,6 +1630,9 @@ class SQLCompiler:
                 else:
                     yield value
 
+    def comments_sql(self):
+        return [f"/* {comment} */" for comment in self.query.comments]
+
 
 class SQLInsertCompiler(SQLCompiler):
     returning_fields = None
@@ -1976,8 +1982,11 @@ class SQLUpdateCompiler(SQLCompiler):
             else:
                 values.append("%s = NULL" % qn(name))
         table = self.query.base_table
-        result = [
-            "UPDATE %s SET" % qn(table),
+        result = ["UPDATE"]
+        if self.query.comments:
+            result += self.comments_sql()
+        result += [
+            "%s SET" % qn(table),
             ", ".join(values),
         ]
         try:
--- a/django/db/models/sql/compiler.py
+++ b/django/db/models/sql/compiler.py
@@ -787,8 +787,7 @@ class SQLCompiler:
                 result = ["SELECT"]
                 params = []
 
-                if self.query.comments:
-                    result += self.comments_sql()
+                result += self.query.comments
 
                 if self.query.distinct:
                     distinct_result, distinct_params = self.connection.ops.distinct_sql(
@@ -1630,9 +1629,6 @@ class SQLCompiler:
                 else:
                     yield value
 
-    def comments_sql(self):
-        return [f"/* {comment} */" for comment in self.query.comments]
-
 
 class SQLInsertCompiler(SQLCompiler):
     returning_fields = None
@@ -1751,7 +1747,9 @@ class SQLInsertCompiler(SQLCompiler):
         insert_statement = self.connection.ops.insert_statement(
             on_conflict=self.query.on_conflict,
         )
-        result = ["%s %s" % (insert_statement, qn(opts.db_table))]
+        result = ["%s" % insert_statement]
+        result += self.query.comments
+        result += ["%s" % qn(opts.db_table)]
         fields = self.query.fields or [opts.pk]
         result.append("(%s)" % ", ".join(qn(f.column) for f in fields))
 
@@ -1897,7 +1895,13 @@ class SQLDeleteCompiler(SQLCompiler):
         )
 
     def _as_sql(self, query):
-        delete = "DELETE FROM %s" % self.quote_name_unless_alias(query.base_table)
+        result = [
+            "DELETE",
+            *self.query.comments,
+            "FROM",
+            self.quote_name_unless_alias(query.base_table),
+        ]
+        delete = " ".join(result)
         try:
             where, params = self.compile(query.where)
         except FullResultSet:
@@ -1981,12 +1985,10 @@ class SQLUpdateCompiler(SQLCompiler):
                 update_params.append(val)
             else:
                 values.append("%s = NULL" % qn(name))
-        table = self.query.base_table
-        result = ["UPDATE"]
-        if self.query.comments:
-            result += self.comments_sql()
-        result += [
-            "%s SET" % qn(table),
+        result = [
+            "UPDATE",
+            *self.query.comments,
+            "%s SET" % qn(self.query.base_table),
             ", ".join(values),
         ]
         try:
--- a/docs/ref/models/querysets.txt
+++ b/docs/ref/models/querysets.txt
@@ -2042,7 +2042,8 @@ Adds an SQL comment to be inserted into the resultant SQL statement, after
 
     Entry.objects.comment("blog/views.py:50").all()
 
-… translates (roughly) into the following SQL::
+… translates (roughly) into the following SQL:
+
 .. code-block:: sql
 
     SELECT /* blog/views.py:50 */ blog_entry.* FROM blog_entry
@@ -2058,9 +2059,10 @@ entries, for example::
     )
 
 … translates into
+
 .. code-block:: sql
 
-    UPDATE /* blog/views.py:75 */ /* python=3.12 */ blog_entry SET "headline" = 'This is new'
+    UPDATE /* blog/views.py:50 */ /* python=3.12 */ blog_entry SET "headline" = 'This is new'
 
 Operators that return new ``QuerySet``\s
 ----------------------------------------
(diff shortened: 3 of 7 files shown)


django/db/models/sql/compiler.py
```diff
@@ -795,6 +795,9 @@ def as_sql(self, with_limits=True, with_col_aliases=False):
                     result += distinct_result
                     params += distinct_params
 
+                if self.query.comments:
```
adamchainz, line 798: Let's put the coment right after SELECT, before the DISTINCT clause which can be arbitrarily long.

django/db/models/sql/compiler.py
```diff
@@ -1627,6 +1630,9 @@ def explain_query(self):
                 else:
                     yield value
 
+    def comments_sql(self):
+        return " ".join([f"/* {comment} */" for comment in self.query.comments])
```
adamchainz, line 1634: Can we avoid the join? Callers should all be doing it already.

adamchainz (changes requested):
Great start, looking forward to your iteration tomorrow.

The PR description formatting needs a little fixing.

django/db/models/sql/compiler.py
viciu: Done, added a test as well.

django/db/models/sql/compiler.py
viciu: DONE

django/db/models/sql/compiler.py
```diff
@@ -1627,6 +1630,9 @@ def explain_query(self):
                 else:
                     yield value
 
+    def comments_sql(self):
+        return [f"/* {comment} */" for comment in self.query.comments]
```
adamchainz, line 1634: another idea: perhaps we can add the `/* ... */` wrapping to comments in `QuerySet.comment()`. Then we wouldn't even need a `comments_sql` method, the tuple can just be thrown into the SQL list.

docs/ref/models/querysets.txt
```diff
@@ -2030,6 +2030,37 @@ See the :doc:`/topics/db/sql` for more information.
...
 
+``comment()``
+~~~~~~~~~~~~~
+
+.. versionadded:: 5.2
+
+.. method:: comment(message)
+
+Adds an SQL comment to be inserted into the resultant SQL statement, after
+``SELECT``/``UPDATE``, and if applicable, ``DISTINCT``.. For example::
```
adamchainz, line 2041: Is it not possible to make it work for `INSERT`, at least when using `Model.objects.comment(...).create(...)` ?

docs/ref/models/querysets.txt
```diff
@@ -2030,6 +2030,37 @@ See the :doc:`/topics/db/sql` for more information.
...
+
+    Entry.objects.comment("blog/views.py:50").all()
+
+… translates (roughly) into the following SQL::
+
+    SELECT /* blog/views.py:50 */ blog_entry.* FROM blog_entry
+
+This is useful for adding information to the SQL statement which when logged
+can be used to correlate slow and problematic queries to the generating source
+code.
```
adamchainz, line 2052:
```suggestion
This is useful for adding logging information to the generated SQL for
correlating problematic queries with the responsible Python code.
```

docs/ref/models/querysets.txt
```diff
@@ -2030,6 +2030,37 @@ See the :doc:`/topics/db/sql` for more information.
...
+Comments can be chained and will be added to the SQL statement as separate
+entries, for example::
+
+    Entry.objects.comment("blog/views.py:50").comment("python=3.11").update(
+        headline="This is new"
+    )
+
+… translates into::
+
+    UPDATE blog_entry.* /* blog/views.py:75 */ /* python=3.11 */ SET "headline" = 'This is new'
```
adamchainz: Needs updating
```suggestion
    UPDATE /* blog/views.py:75 */ /* python=3.11 */ blog_entry SET "headline" = 'This is new'
```

docs/releases/5.2.txt
```diff
@@ -178,7 +178,9 @@ Migrations
 Models
 ~~~~~~
 
-* ...
+* The new :meth:`.QuerySet.comment` method allows for injecting comments into
+  the generated SQL statements. This can be useful for correlating troublesome
+  queries to the application code.
```
adamchainz, line 183: Try to avoid judgmental words like “troublesome”, in favour of specific terms like “slow”. It’s more precise and translates better.

```suggestion
* The new :meth:`.QuerySet.comment` method allows for adding comments into
  the generated SQL. This can be useful for correlating slow queries with
  application code.
```

django/db/models/sql/compiler.py
viciu: Done

django/db/models/query.py
```diff
@@ -1783,6 +1787,18 @@ def using(self, alias):
...
 
+    def comment(self, message):
+        """Add a comment to the query."""
+        clone = self._clone()
+        if COMMENT_FORBIDDEN_PATTERN.search(message):
+            raise ValueError(
+                "Cannot pass strings containing /* or */ to comment(). "
+                "Escape or strip these delimiters before calling comment()."
+            )
+        commented_message = f"/* {message} */".format(message=message)
```
adamchainz, line 1798: It's an f-string, the `.format` isn't doing anything
```suggestion
        commented_message = f"/* {message} */"
```

tests/queries/tests.py
```diff
@@ -4602,3 +4602,74 @@ def test_ticket_23622(self):
...
+
+    def test_stops_delimiters_in_message(self):
+        for comment in ["foo=/a/b/c*/", "/*foo=/a/b/c", "**//SELECT nothing;//**"]:
+            msg = (
+                "Cannot pass strings containing /* or */ to comment(). "
+                "Escape or strip these delimiters before calling comment()."
+            )
+
+            with self.subTest(comment), self.assertRaisesMessage(ValueError, msg):
+                NamedCategory.objects.comment(comment)
```
adamchainz, line 4661: We can hoist `msg` and split the subtests one per line, to make it easier to read and add any more in the future.

```suggestion
        msg = (
            "Cannot pass strings containing /* or */ to comment(). "
            "Escape or strip these delimiters before calling comment()."
        )
        for comment in [
            "foo=/a/b/c*/",
            "/*foo=/a/b/c",
            "**//SELECT nothing;//**",
        ]:
            with self.subTest(comment), self.assertRaisesMessage(ValueError, msg):
                NamedCategory.objects.comment(comment)
```

tests/queries/tests.py
```diff
@@ -4602,3 +4602,74 @@ def test_ticket_23622(self):
...
+
+    def test_delete(self):
+        with CaptureQueriesContext(connection) as captured_queries:
+            NamedCategory.objects.create(name="test")
+            NamedCategory.objects.comment("DELETE 1").all().delete()
+
+            NamedCategory.objects.create(name="test")
+            NamedCategory.objects.comment("DELETE 2").all().delete()
+        self.assertIn("DELETE FROM /* DELETE 1 */", captured_queries[1]["sql"])
```
adamchainz, line 4670: 1. We don't need to create any data in order to run `DELETE`
2. Only one delete query needed
3. Use `startswith` and more realistic example data

```suggestion
            NamedCategory.objects.comment("blog/views.py:50").all().delete()
        self.assertTrue(
            captured_queries[0]["sql"].startswith(
                "DELETE /* blog/views.py:50 */ FROM", 
            )
        )
```

adamchainz (changes requested):

adamchainz:
Thanks for putting sprint time in on this! I found a bit more time here to review, hope you'll be able to update soon.

tests/queries/tests.py
viciu: The problem here is that delete does not generate delete query, but select:

This

```
NamedCategory.objects.comment("DELETE 2").delete()
```

will fail:
```
FAIL: test_delete (queries.tests.SqlCommentsInQueriesTests.test_delete)
  File "/home/viciu/prog/django/tests/queries/tests.py", line 4670, in test_delete
    self.assertIn("DELETE FROM /* DELETE 1 */", [q['sql'] for q in captured_queries])
AssertionError: 'DELETE FROM /* DELETE 1 */' not found in 
['SELECT /* DELETE 2 */ "queries_dumbcategory"."id", "queries_namedcategory"."dumbcategory_ptr_id", "queries_namedcategory"."name" FROM "queries_namedcategory" INNER JOIN "queries_dumbcategory" ON ("queries_namedcategory"."dumbcategory_ptr_id" = "queries_dumbcategory"."id")']

```


therefore I was experimenting with adding some records first and then deleting them. If I understand correctly it looks like django fetches records to delete first, then updates related records with nulls, then finally deletes them one by one (including related records):

```
FAIL: test_delete (queries.tests.SqlCommentsInQueriesTests.test_delete)
  File "/home/viciu/prog/django/tests/queries/tests.py", line 4670, in test_delete
    self.assertIn("DELETE FROM /* DELETE 1 */", [q['sql'] for q in captured_queries])
AssertionError: 'DELETE FROM /* DELETE 1 */' not found in 
[
'INSERT INTO "queries_dumbcategory" ("id") VALUES (NULL)', 
'INSERT INTO "queries_namedcategory" ("dumbcategory_ptr_id", "name") VALUES (1, \'test\')', 
'SELECT /* DELETE 1 */ "queries_dumbcategory"."id", "queries_namedcategory"."dumbcategory_ptr_id", "queries_namedcategory"."name" FROM "queries_namedcategory" INNER JOIN "queries_dumbcategory" ON ("queries_namedcategory"."dumbcategory_ptr_id" = "queries_dumbcategory"."id")', 
'UPDATE "queries_tag" SET "category_id" = NULL WHERE "queries_tag"."category_id" IN (1)', 
'DELETE FROM "queries_namedcategory" WHERE "queries_namedcategory"."dumbcategory_ptr_id" IN (1)',
'DELETE FROM "queries_dumbcategory" WHERE "queries_dumbcategory"."id" IN (1)', 

'INSERT INTO "queries_dumbcategory" ("id") VALUES (NULL)', 
'INSERT INTO "queries_namedcategory" ("dumbcategory_ptr_id", "name") VALUES (2, \'test\')', 
'SELECT /* DELETE 2 */ "queries_dumbcategory"."id", "queries_namedcategory"."dumbcategory_ptr_id", "queries_namedcategory"."name" FROM "queries_namedcategory" INNER JOIN "queries_dumbcategory" ON ("queries_namedcategory"."dumbcategory_ptr_id" = "queries_dumbcategory"."id")', 
'UPDATE "queries_tag" SET "category_id" = NULL WHERE "queries_tag"."category_id" IN (2)', 
'DELETE FROM "queries_namedcategory" WHERE "queries_namedcategory"."dumbcategory_ptr_id" IN (2)', 
'DELETE FROM "queries_dumbcategory" WHERE "queries_dumbcategory"."id" IN (2)']
```

django/db/models/sql/compiler.py
```diff
@@ -1891,7 +1895,13 @@ def contains_self_reference_subquery(self):
         )
 
     def _as_sql(self, query):
-        delete = "DELETE FROM %s" % self.quote_name_unless_alias(query.base_table)
+        result = [
+            "DELETE",
+            *self.query.comments,
```
charettes, line 1900: What's the rationale for injecting comments here instead of at the end of the query?

django/db/models/sql/compiler.py
```diff
@@ -1975,9 +1985,10 @@ def as_sql(self):
                 update_params.append(val)
             else:
                 values.append("%s = NULL" % qn(name))
-        table = self.query.base_table
         result = [
-            "UPDATE %s SET" % qn(table),
+            "UPDATE",
+            *self.query.comments,
```
charettes, line 1990: Ditto.

charettes (commented):
I understand the desire to help workaround MySQL'ism but I find it strange that we default to injecting comments where it can be useful for the latter on all backends given the main purpose of these is likely to be traceability which is harder to parse when it's right in the middle SQL.

Could we have some backends level option that at least configure where they should live and default to prepending / postpending them on non-MySQL/Oracle backends?

django/db/models/sql/compiler.py
viciu: Thanks for your comment. Queries can be long and truncated - please refer to the comment https://github.com/django/django/pull/15711#issuecomment-1132633387

adamchainz: Yeah, the idea is that in a running query list, you can see the type of query first, then any tracing info. I saw this recommended by MySQL experts years ago, hence my suggestion for it in the previous PR. This is not for supporting any MySQL specific options.

I'm not so sure now that we should try to support inserting comments at any of the positions used for query hints. That would require a very complex API and might still not cover all cases. Also, database instrumentation exists to allow rewriting queries when needed.

charettes: > I saw this recommended by MySQL experts years ago.

Do you have references for that? Isn't it fair to assume that an approach that works well for MySQL and recommended by these experts might not work so well for other backends (e.g. Postgres). I've seen plenty of [solutions](https://github.com/google/sqlcommenter/blob/c447d218a1b8ff628571b22cf4e74f2a6fe2e38a/python/sqlcommenter-python/google/cloud/sqlcommenter/__init__.py#L46-L53) and APMs default to prepending the traces to queries to ensure they are never truncated.

At work we built a solution that postpends commented traces specifically for MySQL and it appears to be working well hence my curiosity here.

charettes: From an internal discussion with DBREs

1. Some tooling doesn’t support parsing in the middle of a query
2. Keeping queries starting with their respective verb makes it easier to filter
3. Putting the trace at the end of query might cause it to be truncated off on some systems

This makes me think that a proper solution would at the very least need to be configurable?

pgramaca:
Let me know if there is something i can help with.
I would love to help with this feature.

jacobtylerwalls:
Hi @viciu, we just got a new proposal in #21226. Were you planning to continue here, or should we continue with the new one?

Happy to keep iterating here if you like 👍

PR #17667, opened 2023-12-31
Fixed #35060 -- Deprecated passing positional arguments to Model.save()/asave().

salvo-polizzi:
Refs: [ticket 35060](https://code.djangoproject.com/ticket/35060)

PR #18259, opened 2024-06-08
Fixed #22712 -- Avoided name shadowing of "all" in findstatic command.

avallbona:
Fixed TICKET-22712 - Consider not using built-in functions as parameters.

# Trac ticket number

ticket-22712

# Branch description
Replaced the name that shadows the builtin funcion **all** and replaced by **fetch_all**. I'm not sure if we need the changes to be backwards compatible. 

# Checklist
- [x] I have attached screenshots in both light and dark modes for any UI changes. (N/A)

### DIFF
--- a/django/contrib/staticfiles/finders.py
+++ b/django/contrib/staticfiles/finders.py
@@ -1,5 +1,6 @@
 import functools
 import os
+import warnings
 
 from django.apps import apps
 from django.conf import settings
@@ -8,11 +9,38 @@ from django.core.checks import Error, Warning
 from django.core.exceptions import ImproperlyConfigured
 from django.core.files.storage import FileSystemStorage, Storage, default_storage
 from django.utils._os import safe_join
+from django.utils.deprecation import RemovedInDjango60Warning
 from django.utils.functional import LazyObject, empty
 from django.utils.module_loading import import_string
 
 # To keep track on which directories the finder has searched the static files.
 searched_locations = []
+
+
+# RemovedInDjango60Warning: When the deprecation ends, remove completely.
+def _check_deprecated_find_param(class_name="", find_all=False, **kwargs):
+    method_name = "find" if not class_name else f"{class_name}.find"
+    if "all" in kwargs:
+        legacy_all = kwargs.pop("all")
+        msg = (
+            "Passing the `all` argument to find() is deprecated. Use `find_all` "
+            "instead."
+        )
+        warnings.warn(msg, RemovedInDjango60Warning, stacklevel=2)
+
+        # If both `find_all` and `all` were given, raise TypeError.
+        if find_all is not False:
+            raise TypeError(
+                f"{method_name}() got multiple values for argument 'find_all'"
+            )
+
+        find_all = legacy_all
+
+    if kwargs:  # any remaining kwargs must be a TypeError
+        first = list(kwargs.keys()).pop()
+        raise TypeError(f"{method_name}() got an unexpected keyword argument '{first}'")
+
+    return find_all
 
 
 class BaseFinder:
@@ -26,12 +54,20 @@ class BaseFinder:
             "configured correctly."
         )
 
-    def find(self, path, all=False):
+    # RemovedInDjango60Warning: When the deprecation ends, remove completely.
+    def _check_deprecated_find_param(self, **kwargs):
+        return _check_deprecated_find_param(
+            class_name=self.__class__.__qualname__, **kwargs
+        )
+
+    # RemovedInDjango60Warning: When the deprecation ends, replace with:
+    # def find(self, path, find_all=False):
+    def find(self, path, find_all=False, **kwargs):
         """
         Given a relative file path, find an absolute file path.
 
-        If the ``all`` parameter is False (default) return only the first found
-        file path; if True, return a list of all found files paths.
+        If the ``find_all`` parameter is False (default) return only the first
+        found file path; if True, return a list of all found files paths.
         """
         raise NotImplementedError(
             "subclasses of BaseFinder must provide a find() method"
@@ -113,17 +149,22 @@ class FileSystemFinder(BaseFinder):
                 )
         return errors
 
-    def find(self, path, all=False):
+    # RemovedInDjango60Warning: When the deprecation ends, replace with:
+    # def find(self, path, find_all=False):
+    def find(self, path, find_all=False, **kwargs):
         """
         Look for files in the extra locations as defined in STATICFILES_DIRS.
         """
+        # RemovedInDjango60Warning.
+        if kwargs:
+            find_all = self._check_deprecated_find_param(find_all=find_all, **kwargs)
         matches = []
         for prefix, root in self.locations:
             if root not in searched_locations:
                 searched_locations.append(root)
             matched_path = self.find_location(root, path, prefix)
             if matched_path:
-                if not all:
+                if not find_all:
                     return matched_path
                 matches.append(matched_path)
         return matches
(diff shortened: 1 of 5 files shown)


django/contrib/staticfiles/finders.py
```diff
@@ -113,17 +113,22 @@ def check(self, **kwargs):
                 )
         return errors
 
-    def find(self, path, all=False):
+    def find(self, path, fetch_all=False, **kwargs):
```
EnriqueSoria, line 116: I'm not familiarized in how deprecations are handled in python/django, but I think that adding `**kwargs` is overkill. Maybe add something like:

```python
    def find(self, path, fetch_all=False, all=None):
        if all is not None:
            # TODO: Add a warning or something
            fetch_all = all
```

This sadly doesn't fix shadowing `all`, but by adding a deprecation warning it will allow us to remove it in future versions.

Maybe someone can give us some advice on how to handle this?

avallbona: Thanks for the feedback @EnriqueSoria. As per your suggestion I've added the deprecations warning for the use of the `all` param in 03b3b1faabd1970ad846f1266ffb320b1ce0508b . On the other hand, I don't see why the use of `**kwargs` is an overkill in this case. I think it fits the case, removing the use of the builtin `all` function and at the same time avoiding the errors for third party packages that may use or extend the existing finders API.

EnriqueSoria: When using a function that has `**kwargs` you don't know which parameters it does accept. That's why I dislike using it. 

But I guess in this case is worth.

EnriqueSoria (approved):

django/contrib/staticfiles/finders.py
nessita: Using `**kwargs` has the undesired side effect of allowing for arbitrary named arguments and the silently ignoring them. This is something that we should avoid so I see two options:
1. Continue using `all` for the deprecation period, is not that terrible since we have already done so for the last 10+ years. 
2. Change to use `**kwargs` but then we should check that no other named param was given and raise the corresponding `TypeError`.

I think 1 is the best option, so I would suggest to go down that path.

django/contrib/staticfiles/finders.py
```diff
@@ -26,11 +32,11 @@ def check(self, **kwargs):
             "configured correctly."
         )
 
-    def find(self, path, all=False):
+    def find(self, path, fetch_all=False, **kwargs):
```
nessita, line 35: I'm not sold on the new proposed name, I think a more accurate option would be `match_all` or `find_all`. To me, fetch implies that something will be "brought" from there to here (ie like data retrieval or similar), which is not really the case.
What do you think?

nessita (commented):
Thank you @avallbona for this work! It looks good :star2: 

I have added some comments and I also think that we should include release notes in the `5.2.txt` file, under the section `Features deprecated in 5.2`. Something like:

```
* The ``all`` argument for the ``django.contrib.staticfiles.find()`` function
  is deprecated in favor of the ``find_all`` argument.
```

django/contrib/staticfiles/finders.py
avallbona: I agree, I think `find_all` it's a better fit for this param name. Handled in 704fdb92b513ee0d6218068990bd70afa90dfdfd

avallbona:
> Thank you @avallbona for this work! It looks good 🌟
> 
> I have added some comments and I also think that we should include release notes in the `5.2.txt` file, under the section `Features deprecated in 5.2`. Something like:
> 
> ```
> * The ``all`` argument for the ``django.contrib.staticfiles.find()`` function
>   is deprecated in favor of the ``find_all`` argument.
> ```

Addressed the comments in 704fdb92b513ee0d6218068990bd70afa90dfdfd Thanks for the feedback 🙇🏿‍♂️

nessita:
Thank you @avallbona for the param rename! Looks great.

Did you see my comment about `kwargs`? Would you have some time to do some follow up with that?

django/contrib/staticfiles/finders.py
nessita: @avallbona The comment above is what I was referring in the global comment.

avallbona: Sorry I missed this comment 😬 . I'm not sure I like the option 1, we don't solve the problem right away and we pospone the solution to a future version. So, In option 1, would we: keep the the method signature as `find(path, find_all=False, all=False)` or just like `find(path, all=False)` and raise the deprecation warning?

avallbona:
> Thank you @avallbona for the param rename! Looks great.
> 
> Did you see my comment about `kwargs`? Would you have some time to do some follow up with that?

I've answered in the comment about `kwargs`. Thanks for the feedback 🙇🏿‍♂️

django/contrib/staticfiles/finders.py
avallbona: @nessita Addressed the issue in eacad9336b5225b6bd47894cd1d7d5db33367a47

django/contrib/staticfiles/finders.py
```diff
@@ -113,17 +119,25 @@ def check(self, **kwargs):
...
         """
         Look for files in the extra locations as defined in STATICFILES_DIRS.
         """
+
+        # we cover the case the method get called with the legacy param "all"
+        if all is not None:
+            warnings.warn(
+                DEPRECATION_WARNING_MSG, RemovedInDjango60Warning, stacklevel=2
+            )
+            find_all = all
```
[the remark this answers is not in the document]
avallbona: @EnriqueSoria, It may be a personal preference but, I don't like this approach. It feels a bit hacky.

nessita:
> > Thank you @avallbona for the param rename! Looks great.
> > Did you see my comment about `kwargs`? Would you have some time to do some follow up with that?
> 
> I've answered in the comment about `kwargs`. Thanks for the feedback 🙇🏿‍♂️

Thank you, overall that looks good. I think that some comments such as:

```python
# RemovedInDjango60Warning: When the deprecation ends, replace with:
# def find(self, path, find_all=False):
```

As a reference/example, we are just discussing/fixing a similar-ish issue for the `save`  and `asave` methods. You can see how that was done  in these PRs:

#17667
#18304
#18313

avallbona:
> > > Thank you @avallbona for the param rename! Looks great.
> > > Did you see my comment about `kwargs`? Would you have some time to do some follow up with that?
> > 
> > 
> > I've answered in the comment about `kwargs`. Thanks for the feedback 🙇🏿‍♂️
> 
> Thank you, overall that looks good. I think that some comments such as:
> 
> ```python
> # RemovedInDjango60Warning: When the deprecation ends, replace with:
> # def find(self, path, find_all=False):
> ```
> 
> As a reference/example, we are just discussing/fixing a similar-ish issue for the `save` and `asave` methods. You can see how that was done in these PRs:
> 
> #17667 #18304 #18313

@nessita handled in 3563132ea47cb8a0a52b2558e541bbc579535c30 Thanks for the feedback 🙇🏿‍♂️

nessita (approved):
Thank you so much @avallbona for this work :star2: 

I pushed a slight rework of the solution because, once I branched this PR to complete the review, I realized that the duplication for the deprecation check could be unified in a helper method.

I also took the opportunity to try the `kwargs` approach which I think worked well enough. Let me know what you think!

nessita:
Deprecation timeline was corrected in #18321.

avallbona:
@nessita Thanks for your support along my first code contribution. The result of the fix I think it's satisfactory and I would have liked to had the chance to fix what @felixxm has fixed.

nessita:
> @nessita Thanks for your support along my first code contribution. The result of the fix I think it's satisfactory and I would have liked to had the chance to fix what @felixxm has fixed.

Right, this was my mistake: I should have noticed the deprecation number mismatch and I should have let you know so you could correct it. I'm sorry!

(In my head, I was still on `5.1` due to the recent beta release so I mixed the target deprecation version.)

PR #18260, opened 2024-06-09
Fixed #32915 -- Refactored runserver command to restore traceback

pi-sigma:
# Trac ticket number

ticket-32915

# Branch description

The traceback for `runserver --nostatic` with broken/missing settings was hidden because the custom runserver command which added `--nostatic` to the command args was only instantiated if the settings had no errors.

Previous PRs: [#1637](https://github.com/django/django/pull/16337), [14622](https://github.com/django/django/pull/14622)

# Checklist


django/core/management/commands/runserver.py
```diff
@@ -6,6 +6,7 @@
 from datetime import datetime
 
 from django.conf import settings
+from django.contrib.staticfiles.handlers import StaticFilesHandler
```
charettes, line 9: Isn't this going or be problem if `django.contrib.staticfiles` is not part of `INSTALLED_APPS`?

I think that changes should be made to the `staticfiles` provided command instead if any.

https://github.com/django/django/blob/main/django/contrib/staticfiles/management/commands/runserver.py

The moment `django.contrib.staticfiles` is added to `INSTALLED_APPS` calling `runserver` should resolve to the override in `staticfiles`.

django/core/management/commands/runserver.py
```diff
@@ -65,6 +66,18 @@ def add_arguments(self, parser):
...
+            action="store_false",
+            dest="use_static_handler",
+            help="Tells Django to NOT automatically serve static files at STATIC_URL.",
+        )
+        parser.add_argument(
+            "--insecure",
+            action="store_true",
+            dest="insecure_serving",
+            help="Allows serving static files even if DEBUG is False.",
+        )
```
sarahboyce, line 80: Moving these options here is a change to the documented behavior that these options are only available when staticfiles is installed (see https://docs.djangoproject.com/en/5.1/ref/contrib/staticfiles/#cmdoption-runserver-nostatic)
Although this works, I don't think this is the right solution

jacobtylerwalls:
Closing in favor of the other apporach.

PR #18261, opened 2024-06-09
Fixed #35537 -- Use a namedtuple for email attachments and alternatives

RealOrangeOne:
This makes it more descriptive to pull out the named fields

# Trac ticket number

ticket-XXXXX

# Branch description

This change replaces the opaque tuples used for email attachments and alternatives with named tuples. This makes unpacking the specific indexes more descriptive, whilst avoiding breaking any existing uses (since the field order hasn't changed).

```python
email.alternatives[0][0]  # before
email.alternatives[0].content  #after
```

Unpacking the alternatives doesn't seem to be documented, so there are no docs to update. Similarly, adding attachments / alternatives is done using documented methods, whose public API is unaffected by this change. Manual insertion into the `.alternatives` and `.attachments` may now fail, however users should be using the documented API methods.

# Checklist

### DIFF
--- a/django/core/mail/message.py
+++ b/django/core/mail/message.py
@@ -1,4 +1,5 @@
 import mimetypes
+from collections import namedtuple
 from email import charset as Charset
 from email import encoders as Encoders
 from email import generator, message_from_string
@@ -190,6 +191,10 @@ class SafeMIMEMultipart(MIMEMixin, MIMEMultipart):
         MIMEMultipart.__setitem__(self, name, val)
 
 
+Alternative = namedtuple("Alternative", ["content", "mimetype"])
+EmailAttachment = namedtuple("Attachment", ["filename", "content", "mimetype"])
+
+
 class EmailMessage:
     """A container for email information."""
 
@@ -338,7 +343,7 @@ class EmailMessage:
                         # actually binary, read() raises a UnicodeDecodeError.
                         mimetype = DEFAULT_ATTACHMENT_MIME_TYPE
 
-            self.attachments.append((filename, content, mimetype))
+            self.attachments.append(EmailAttachment(filename, content, mimetype))
 
     def attach_file(self, path, mimetype=None):
         """
@@ -471,13 +476,15 @@ class EmailMultiAlternatives(EmailMessage):
             cc,
             reply_to,
         )
-        self.alternatives = alternatives or []
+        self.alternatives = [
+            Alternative(*alternative) for alternative in (alternatives or [])
+        ]
 
     def attach_alternative(self, content, mimetype):
         """Attach an alternative content representation."""
         if content is None or mimetype is None:
             raise ValueError("Both content and mimetype must be provided.")
-        self.alternatives.append((content, mimetype))
+        self.alternatives.append(Alternative(content, mimetype))
 
     def _create_message(self, msg):
         return self._create_attachments(self._create_alternatives(msg))
@@ -492,5 +499,9 @@ class EmailMultiAlternatives(EmailMessage):
             if self.body:
                 msg.attach(body_msg)
             for alternative in self.alternatives:
-                msg.attach(self._create_mime_attachment(*alternative))
+                msg.attach(
+                    self._create_mime_attachment(
+                        alternative.content, alternative.mimetype
+                    )
+                )
         return msg
--- a/docs/topics/email.txt
+++ b/docs/topics/email.txt
@@ -282,8 +282,13 @@ All parameters are optional and can be set at any time prior to calling the
   new connection is created when ``send()`` is called.
 
 * ``attachments``: A list of attachments to put on the message. These can
-  be either :class:`~email.mime.base.MIMEBase` instances, or ``(filename,
-  content, mimetype)`` triples.
+  be either :class:`~email.mime.base.MIMEBase` instances, or a named tuple
+  with attributes ``(filename, content, mimetype)``.
+
+  .. versionchanged:: 5.2
+
+    In older versions, tuple items of ``attachments`` were regular tuples,
+    as opposed to named tuples.
 
 * ``headers``: A dictionary of extra headers to put on the message. The
   keys are the header name, values are the header values. It's up to the
@@ -392,10 +397,10 @@ Django's email library, you can do this using the
 
 .. class:: EmailMultiAlternatives
 
-    A subclass of :class:`~django.core.mail.EmailMessage` that has an
-    additional ``attach_alternative()`` method for including extra versions of
-    the message body in the email. All the other methods (including the class
-    initialization) are inherited directly from
+    A subclass of :class:`~django.core.mail.EmailMessage` that allows
+    additional versions of the message body in the email via the
+    ``attach_alternative()`` method. This directly inherits all methods
+    (including the class initialization) from
     :class:`~django.core.mail.EmailMessage`.
 
     .. method:: attach_alternative(content, mimetype)
@@ -414,6 +419,24 @@ Django's email library, you can do this using the
             msg = EmailMultiAlternatives(subject, text_content, from_email, [to])
             msg.attach_alternative(html_content, "text/html")
             msg.send()
+
+    .. attribute:: alternatives
+
+        A list of named tuples with attributes ``(content, mimetype)``. This is
+        particularly useful in tests::
+
+            self.assertEqual(len(msg.alternatives), 1)
+            self.assertEqual(msg.alternatives[0].content, html_content)
+            self.assertEqual(msg.alternatives[0].mimetype, "text/html")
+
+        Alternatives should only be added using the
+        :meth:`~django.core.mail.EmailMultiAlternatives.attach_alternative`
+        method.
+
+        .. versionchanged:: 5.2
+
+            In older versions, ``alternatives`` was a list of regular tuples, as opposed
+            to named tuples.
 
 Updating the default content type
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/releases/5.2.txt
+++ b/docs/releases/5.2.txt
@@ -128,7 +128,15 @@ Decorators
 Email
 ~~~~~
 
-* ...
+* Tuple items of :class:`EmailMessage.attachments
+  <django.core.mail.EmailMessage>` and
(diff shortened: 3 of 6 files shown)


LilyFoote (approved):

EnriqueSoria (approved):

sarahboyce:
In https://github.com/django/django/pull/18279, I added some docs for `EmailMultiAlternatives`. 
I think it would be nice to add docs for the `alternatives` attribute here as part of this change 👍

docs/topics/email.txt
```diff
@@ -415,6 +415,16 @@ Django's email library, you can do this using the
...
 
+    .. attribute:: alternatives
+
+        A list of named tuples with attributes ``(content, mimetype)``. This is
+        particularly useful in tests::
+
+            self.assertEqual(len(msg.alternatives), 1)
+            self.assertEqual(msg.alternatives[0].content, html_content)
+
+        Alternatives should only be added using the ``attach_alternative()`` method.
```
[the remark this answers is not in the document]
RealOrangeOne: The change is backwards compatible, but I agree it'd be nice to flag seeing it's much nicer. I'll get it added.

docs/topics/email.txt
```diff
@@ -282,8 +282,12 @@ All parameters are optional and can be set at any time prior to calling the
...
 
 * ``attachments``: A list of attachments to put on the message. These can
-  be either :class:`~email.mime.base.MIMEBase` instances, or ``(filename,
-  content, mimetype)`` triples.
+  be either :class:`~email.mime.base.MIMEBase` instances, or a named tuple
+  with attributes ``(filename, content, mimetype)``.
+
+  .. versionchanged:: 5.2
+
+    ``attachments`` is now a list of named tuples as opposed to regular tuples.
```
sarahboyce, line 290:
```suggestion
    In older versions, tuple items of ``attachments`` were regular tuples, 
    as opposed to named tuples.
```
Maybe 🤔 
We might also need to move this to the bottom of the `EmailMessage` definition

docs/topics/email.txt
```diff
@@ -415,6 +419,20 @@ Django's email library, you can do this using the
...
+        particularly useful in tests::
+
+            self.assertEqual(len(msg.alternatives), 1)
+            self.assertEqual(msg.alternatives[0].content, html_content)
+
+        Alternatives should only be added using the ``attach_alternative()`` method.
+
+        .. versionchanged:: 5.2
+
+            ``alternatives`` is now a list of named tuples as opposed to regular tuples.
```
sarahboyce, line 434:
```suggestion
            In older versions, ``alternatives`` was a list of regular tuples, as opposed
            to named tuples.
```

docs/topics/email.txt
```diff
@@ -393,9 +397,9 @@ Django's email library, you can do this using the
 .. class:: EmailMultiAlternatives
 
     A subclass of :class:`~django.core.mail.EmailMessage` that has an
-    additional ``attach_alternative()`` method for including extra versions of
-    the message body in the email. All the other methods (including the class
-    initialization) are inherited directly from
+    additional ``attach_alternative()`` method and ``alternatives`` attribute
+    for including extra versions of the message body in the email. All the other
+    methods (including the class initialization) are inherited directly from
     :class:`~django.core.mail.EmailMessage`.
```
sarahboyce, line 403:
```suggestion
    A subclass of :class:`~django.core.mail.EmailMessage` that allows
    additional versions of the message body in the email via the
    ``attach_alternative()`` method. This directly inherits all methods
    (including the class initialization) from
    :class:`~django.core.mail.EmailMessage`.
```
This is based off [this discussion](https://github.com/django/django/pull/18278#discussion_r1644179202).

sarahboyce (commented):
Nice! Have a last couple thoughts

Also in the 5.2 release notes, let's add something like
```
Email
~~~~~

* Tuple items of ``attachments`` of :class:`~django.core.mail.EmailMessage`
  and :class:`~django.core.mail.EmailMultiAlternatives` are now named tuples,
  as opposed to regular tuples.

* :attr:`EmailMultiAlternatives.alternatives
  <django.core.mail.EmailMultiAlternatives.alternatives>` is now a list of
  named tuples, as opposed to regular tuples.

```

docs/topics/email.txt
```diff
@@ -415,6 +420,21 @@ Django's email library, you can do this using the
             msg.attach_alternative(html_content, "text/html")
             msg.send()
 
+    .. attribute:: alternatives
+
+        A list of named tuples with attributes ``(content, mimetype)``. This is
+        particularly useful in tests::
+
+            self.assertEqual(len(msg.alternatives), 1)
+            self.assertEqual(msg.alternatives[0].content, html_content)
```
sarahboyce, line 429:
```suggestion
            self.assertEqual(msg.alternatives[0].mime_type, "text/html")
```
Note that we've documented it to be `mimetype` rather than `mime_type`, I _think_ we prefer `mime_type` and should update the docs?

I also think adding tests doesn't hurt as at the very least it's easier to spot if we document the names wrong if we're using them in tests
```diff
--- a/tests/mail/tests.py
+++ b/tests/mail/tests.py
@@ -550,6 +550,29 @@ class MailTests(HeadersCheckMixin, SimpleTestCase):
         msg.attach("example.txt", "Text file content", "text/plain")
         self.assertIn(html_content, msg.message().as_string())

+    def test_attachments(self):
+        msg = EmailMessage()
+        file_name = "example.txt"
+        file_content = "Text file content"
+        mime_type = "text/plain"
+        msg.attach(file_name, file_content, mime_type)
+        self.assertEqual(msg.attachments[0][0], file_name)
+        self.assertEqual(msg.attachments[0].filename, file_name)
+        self.assertEqual(msg.attachments[0][1], file_content)
+        self.assertEqual(msg.attachments[0].content, file_content)
+        self.assertEqual(msg.attachments[0][1], mime_type)
+        self.assertEqual(msg.attachments[0].mime_type, mime_type)
+
+    def test_alternatives(self):
+        msg = EmailMultiAlternatives()
+        html_content = "<p>This is <strong>html</strong></p>"
+        mime_type = "text/html"
+        msg.attach_alternative(html_content, mime_type)
+        self.assertEqual(msg.alternatives[0][0], html_content)
+        self.assertEqual(msg.alternatives[0].content, html_content)
+        self.assertEqual(msg.alternatives[0][1], mime_type)
+        self.assertEqual(msg.alternatives[0].mime_type, mime_type)
+
     def test_none_body(self):
```

I also can't accept tickets I create, so feel free to accept this ticket: https://code.djangoproject.com/ticket/35537

RealOrangeOne: A quick search through the codebase suggests `mimetype` is more common than `mime_type`, although there isn't much consistency.

Definite +1 for adding tests though - I'll get those added!

sarahboyce: Let's go for `mimetype` then 👍

django/core/mail/message.py
```diff
@@ -190,6 +191,10 @@ def __setitem__(self, name, val):
         MIMEMultipart.__setitem__(self, name, val)
 
 
+Alternative = namedtuple("Alternative", ["content", "mime_type"])
```
sarahboyce, line 194:
```suggestion
Alternative = namedtuple("Alternative", ["content", "mimetype"])
```
This is currently documented as `mimetype`, test update is required

RealOrangeOne: Damn, I updated one and missed the other :sweat_smile:

sarahboyce (approved):
Thank you 🥳 this looks good to me

medmunds:
Was this intended to allow passing a namedtuple into the EmailMessage constructor as an attachment? The updated docs seem to state it's supported:

> class EmailMessage
> … All parameters are optional and can be set at any time prior to calling the send() method.
> ...
> * attachments: A list of attachments to put on the message. These can be either [MIMEBase](https://docs.python.org/3/library/email.mime.html#email.mime.base.MIMEBase) instances, or a named tuple with attributes (filename, content, mimetype).

If so, I don't see this covered in the tests.

```python
Attachment = namedtuple("Attachment", ("filename", "content", "mimetype"))

msg = EmailMessage(..., attachments=[Attachment("test.jpg", jpg_content, "image/jpeg")])
# you can also modify attachments before send():
msg.attachments.insert(0, Attachment("first.txt", "I'm the first attachment", "text/plain"))
# or overwrite it entirely: msg.attachments = [Attachment(...), Attachment(...)]
msg.send()
```

For backwards compatibility, this also needs to work with regular (unnamed) tuples. I think it still does, but it's not being tested. (Or this should be documented as a backwards incompatible change that works for now but might break in some future release.)

Also, it's a little obscure, but older versions supported similar modifications to `alternatives` using regular (unnamed) tuples at any time before send(). The new requirement to only call add_alternative() probably should be mentioned in backwards incompatible changes.

RealOrangeOne:
The documentation refers to the type of the attribute, not the constructor argument. As implemented, the constructor will allow either the named or unnamed variants, although that is undocumented (and untested) functionality.

@sarahboyce happy to submit a PR with added tests and docs if you agree?

sarahboyce:
> The documentation refers to the type of the attribute, not the constructor argument. As implemented, the constructor will allow either the named or unnamed variants, although that is undocumented (and untested) functionality.
> 
> @sarahboyce happy to submit a PR with added tests and docs if you agree?

Yes please

> The new requirement to only call add_alternative() probably should be mentioned in backwards incompatible changes.

Agreed 👍

RealOrangeOne:
Opened as https://github.com/django/django/pull/18318

medmunds:
> The documentation refers to the type of the attribute, not the constructor argument.

I've always read the [class EmailMessage docs](https://docs.djangoproject.com/en/5.1/topics/email/#emailmessage-objects) as referring to both (emphasis added):

> The `EmailMessage` class is **initialized with the following parameters** (in the given order, if positional arguments are used).   All parameters are optional and **can be set at any time** prior to calling the `send()` method.
> * `subject`: The subject line of the email.
> * …
> * `headers`: A dictionary of extra headers… The **corresponding attribute** is `extra_headers`.
> * …

\[BTW, thanks for this namedtuple improvement! I've been working with Django's email alternatives and attachments pretty frequently for over a decade now, and *still* can't remember the order.]

PR #18268, opened 2024-06-11
Fixed #35520 -- Used correct database for `ModelAdmin` transactions

RealOrangeOne:
# Trac ticket number

ticket-35520

# Branch description

In `ModelAdmin.delete_view` and `ModelAdmin.changelist_view`, the transaction is hardcoded to use `db_for_write`, even if the request is read-only (eg `GET`). This potentially results in the wrong database being selected (or errors if `db_for_write` is used to prevent writing to models).

This PR uses `db_for_read` for any read requests, based on the method. Most users won't notice the difference, as the same database will be selected. But those who are using a custom router will get the database they expect.

# Checklist

### DIFF
--- a/django/contrib/admin/options.py
+++ b/django/contrib/admin/options.py
@@ -1814,6 +1814,9 @@ class ModelAdmin(BaseModelAdmin):
 
     @csrf_protect_m
     def changeform_view(self, request, object_id=None, form_url="", extra_context=None):
+        if request.method in ("GET", "HEAD", "OPTIONS", "TRACE"):
+            return self._changeform_view(request, object_id, form_url, extra_context)
+
         with transaction.atomic(using=router.db_for_write(self.model)):
             return self._changeform_view(request, object_id, form_url, extra_context)
 
@@ -2175,6 +2178,9 @@ class ModelAdmin(BaseModelAdmin):
 
     @csrf_protect_m
     def delete_view(self, request, object_id, extra_context=None):
+        if request.method in ("GET", "HEAD", "OPTIONS", "TRACE"):
+            return self._delete_view(request, object_id, extra_context)
+
         with transaction.atomic(using=router.db_for_write(self.model)):
             return self._delete_view(request, object_id, extra_context)
 
--- a/tests/admin_views/test_multidb.py
+++ b/tests/admin_views/test_multidb.py
@@ -40,6 +40,7 @@ urlpatterns = [
 @override_settings(ROOT_URLCONF=__name__, DATABASE_ROUTERS=["%s.Router" % __name__])
 class MultiDatabaseTests(TestCase):
     databases = {"default", "other"}
+    READ_ONLY_METHODS = {"get", "options", "head", "trace"}
 
     @classmethod
     def setUpTestData(cls):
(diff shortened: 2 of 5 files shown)


django/contrib/admin/options.py
```diff
@@ -1814,7 +1814,12 @@ def _get_obj_does_not_exist_redirect(self, request, opts, object_id):
 
     @csrf_protect_m
     def changeform_view(self, request, object_id=None, form_url="", extra_context=None):
-        with transaction.atomic(using=router.db_for_write(self.model)):
+        db = (
+            router.db_for_read(self.model)
+            if request.method in ("GET", "HEAD", "OPTIONS", "TRACE")
```
RealOrangeOne, line 1819: There isn't currently a util for whether the request's method is read-only, so I've had to duplicate this set of methods a fair bit. Not sure this is the point we want to extract them into a util?

django/contrib/admin/options.py
```diff
@@ -1814,7 +1814,12 @@ def _get_obj_does_not_exist_redirect(self, request, opts, object_id):
 
     @csrf_protect_m
     def changeform_view(self, request, object_id=None, form_url="", extra_context=None):
-        with transaction.atomic(using=router.db_for_write(self.model)):
+        db = (
+            router.db_for_read(self.model)
```
charettes, line 1818: I'm not sure of the value of creating a transaction here as Django recommends running in `READ COMMITTED` which means that creating a transaction is moot as there won't be any snapshot isolation for reads.

RealOrangeOne: It ought to be unnecessary to create a transaction - I'll remove it.

RealOrangeOne: Fixed in [fefc3b7](https://github.com/django/django/pull/18268/commits/fefc3b7e02b3da670fbc9ab8855251c4e034eb4e)

tests/admin_views/test_multidb.py
```diff
@@ -68,6 +72,17 @@ def test_add_view(self, mock):
...
 
+    @mock.patch("django.contrib.admin.options.transaction")
+    def test_get_add_view(self, mock):
+        for db in self.databases:
+            with self.subTest(db=db):
+                Router.target_db = db
+                self.client.force_login(self.superusers[db])
+                self.client.get(
+                    reverse("test_adminsite:admin_views_book_add"),
+                )
```
sarahboyce, line 83:
```suggestion
    def test_read_only_methods_add_view(self, mock):
        for db in self.databases:
            with self.subTest(db=db):
                Router.target_db = db
                self.client.force_login(self.superusers[db])
                add_url = reverse("test_adminsite:admin_views_book_add")
                self.client.get(add_url)
                self.client.options(add_url)
                self.client.head(add_url)
                self.client.trace(add_url)
```
I think for each test we can add all of the methods

tests/admin_views/test_multidb.py
```diff
@@ -56,6 +56,10 @@ def setUpTestData(cls):
             b.save(using=db)
             cls.test_book_ids[db] = b.id
 
+    def setUp(self):
```
sarahboyce, line 59: Would this be better as a `tearDown`?

sarahboyce (commented):
Thank you for this @RealOrangeOne 👍 
I have minor comments

tests/auth_tests/test_admin_multidb.py
```diff
@@ -57,3 +61,16 @@ def test_add_view(self, mock):
...
+    def test_read_only_methods_add_view(self, mock):
+        for db in self.databases:
+            for method in {"GET", "OPTIONS", "HEAD", "TRACE"}:
+                with self.subTest(db_connection=db, method=method):
+                    mock.mock_reset()
+                    Router.target_db = db
+                    self.client.force_login(self.superusers[db])
+                    self.client.request(
+                        url=reverse("test_adminsite:auth_user_add"), method=method
+                    )
```
sarahboyce, line 75: I don't think the test refactor is working as expected. If I revert the patch, these tests still pass as the response here is a 404

RealOrangeOne: Adding assertions to the other tests I added, the same is true there, too. I'll have a look into what's happening, and put some assertions on the `.status_code` in too for future.

sarahboyce (approved):
Thank you for this @RealOrangeOne ⭐

PR #18270, opened 2024-06-12
Fixed #35518 -- Used simpler string operations for converter-less routes

RealOrangeOne:
# Trac ticket number

ticket-35518

# Branch description

The `RoutePattern` assumes all routes provided include some form of converter, and so needs to change it into a regex for matching. However, if no converters are included in the string, the additional overhead of using a regex vs simpler string operations is unnecessary.

Replacing this with a simpler string comparison results in between a 50 and 75% reduction in match time, which stacks up quickly as an application generally has numerous URLs, because `RoutePattern.match` is called once per defined route per request.

**Before**

```python
In [2]: endpoint_pattern = RoutePattern("foo/", "name", is_endpoint=True)

In [3]: pattern = RoutePattern("foo/", "name", is_endpoint=False)

In [4]: %timeit pattern.match("foo/")
441 ns ± 2.68 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [5]: %timeit endpoint_pattern.match("foo/")
435 ns ± 0.974 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

**After**

```python
In [4]: %timeit pattern.match("foo/")
187 ns ± 1.84 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [5]: %timeit endpoint_pattern.match("foo/")
103 ns ± 0.192 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

(Benchmarked on Python 3.12, running `main`)

Theoretically, these improvements get better based on the length of the route pattern (although at this scale, not notably).

This optimisation could be done by adding a different kind of pattern (eg `LiteralPattern`), but the added complexity to a project probably isn't necessary, not to mention the migration effort to take advantage of this.

# Checklist

### DIFF
--- a/django/urls/resolvers.py
+++ b/django/urls/resolvers.py
@@ -322,17 +322,25 @@ class RoutePattern(CheckURLMixin):
         self.name = name
 
     def match(self, path):
-        match = self.regex.search(path)
-        if match:
-            # RoutePattern doesn't allow non-named groups so args are ignored.
-            kwargs = match.groupdict()
-            for key, value in kwargs.items():
-                converter = self.converters[key]
-                try:
-                    kwargs[key] = converter.to_python(value)
-                except ValueError:
-                    return None
-            return path[match.end() :], (), kwargs
+        # Only use regex overhead if there are converters.
+        if self.converters:
+            if match := self.regex.search(path):
+                # RoutePattern doesn't allow non-named groups so args are ignored.
+                kwargs = match.groupdict()
+                for key, value in kwargs.items():
+                    converter = self.converters[key]
+                    try:
+                        kwargs[key] = converter.to_python(value)
+                    except ValueError:
+                        return None
+                return path[match.end() :], (), kwargs
+        # If this is an endpoint, the path should be exactly the same as the route.
+        elif self._is_endpoint:
+            if self._route == path:
+                return "", (), {}
+        # If this isn't an endpoint, the path should start with the route.
+        elif path.startswith(self._route):
+            return path.removeprefix(self._route), (), {}
         return None
 
     def check(self):
--- a/tests/urlpatterns/test_resolvers.py
+++ b/tests/urlpatterns/test_resolvers.py
@@ -13,6 +13,12 @@ class RoutePatternTests(SimpleTestCase):
     def test_str(self):
         self.assertEqual(str(RoutePattern(_("translated/"))), "translated/")
 
+    def test_has_converters(self):
+        self.assertEqual(len(RoutePattern("translated/").converters), 0)
+        self.assertEqual(len(RoutePattern(_("translated/")).converters), 0)
+        self.assertEqual(len(RoutePattern("translated/<int:foo>").converters), 1)
+        self.assertEqual(len(RoutePattern(_("translated/<int:foo>")).converters), 1)
+
 
 class ResolverCacheTests(SimpleTestCase):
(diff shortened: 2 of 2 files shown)


django/urls/resolvers.py
```diff
@@ -315,24 +315,38 @@ class RoutePattern(CheckURLMixin):
     regex = LocaleRegexRouteDescriptor()
 
     def __init__(self, route, name=None, is_endpoint=False):
-        self._route = route
-        self._regex, self.converters = _route_to_regex(str(route), is_endpoint)
+        self._route = str(route)
```
adamchainz, line 318: I think this is not safe for translations. The descriptor at `self.regex` goes through translations. Also, we would still want to apply translations in the new fast path. The cast to string can be done there.

django/urls/resolvers.py
```diff
@@ -315,24 +315,38 @@ class RoutePattern(CheckURLMixin):
...
+        self._regex, self.converters = _route_to_regex(self._route, is_endpoint)
         self._regex_dict = {}
         self._is_endpoint = is_endpoint
+        self._has_converters = bool(self.converters)
```
adamchainz, line 322: no need, `if self.converters` should be just as fast

adamchainz (changes requested):

django/urls/resolvers.py
```diff
@@ -322,17 +322,31 @@ def __init__(self, route, name=None, is_endpoint=False):
...
+        # If this is an endpoint, the path should be
+        # exactly the same as the route
+        elif self._is_endpoint:
+            if self._route == path:
+                return "", (), {}
+
+        # If this isn't an endpoint, the path should start with the route
+        elif path.startswith(self._route):
+            return path.removeprefix(self._route), (), {}
+
```
pauloxnet, line 349:
```suggestion
        # If there are no converters, skip the regex overhead and use string operations
        # If this is an endpoint, the path should be exactly the same as the route
        elif self._is_endpoint:
            if self._route == path:
                return "", (), {}
        # If this isn't an endpoint, the path should start with the route
        elif path.startswith(self._route):
            return path.removeprefix(self._route), (), {}
```

tests/urlpatterns/test_resolvers.py
```diff
@@ -13,6 +13,13 @@ class RoutePatternTests(SimpleTestCase):
     def test_str(self):
         self.assertEqual(str(RoutePattern(_("translated/"))), "translated/")
 
+    def test_has_converters(self):
+        self.assertEqual(len(RoutePattern("translated/").converters), 0)
+        self.assertEqual(len(RoutePattern(_("translated/")).converters), 0)
+
```
pauloxnet, line 19:
```suggestion
```

pauloxnet (approved):
LGTM

I added code suggestions to compact a bit the code.

django/urls/resolvers.py
RealOrangeOne: I disagree - the extra newlines break up each condition, and make it clearer what the comment is for.

RealOrangeOne:
I've added some more benchmarks to [the ticket](https://code.djangoproject.com/ticket/35518#comment:8) which make the gain clearer.

adamchainz (approved):

sarahboyce (approved):
Thank you!

PR #18276, opened 2024-06-15
Fixed #35508: Add an --ignore-deps flag to squashmigrations (WIP)

pygeek:
# Trac ticket number

ticket-XXXXX

# Branch description

# Checklist

### DIFF
--- a/django/core/management/commands/squashmigrations.py
+++ b/django/core/management/commands/squashmigrations.py
@@ -5,12 +5,13 @@ from django.apps import apps
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
 from django.core.management.utils import run_formatters
-from django.db import DEFAULT_DB_ALIAS, connections, migrations
+from django.db import DEFAULT_DB_ALIAS, connections, migrations, models
 from django.db.migrations.loader import AmbiguityError, MigrationLoader
 from django.db.migrations.migration import SwappableTuple
 from django.db.migrations.optimizer import MigrationOptimizer
 from django.db.migrations.writer import MigrationWriter
 from django.utils.version import get_docs_version
+from django.db.migrations.operations.base import OperationCategory
 
 
 class Command(BaseCommand):
@@ -35,6 +36,14 @@ class Command(BaseCommand):
         parser.add_argument(
             "migration_name",
             help="Migrations will be squashed until and including this migration.",
+        )
+        parser.add_argument(
+            "--ignore-dependencies",
+            "--ignore-deps",
+            action="store_false",
+            dest="ignore_dependencies",
+            help="Ignore dependencies, except for those included in the"
+            " initial migration.",
         )
         parser.add_argument(
             "--no-optimize",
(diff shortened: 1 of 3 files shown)


django/core/management/commands/squashmigrations.py
```diff
@@ -36,6 +37,14 @@ def add_arguments(self, parser):
             "migration_name",
             help="Migrations will be squashed until and including this migration.",
         )
+        parser.add_argument(
+            "--ignore-dependencies",
+            "--ignore-deps",
+            action="store_false",
```
shaib, line 43:
```suggestion
            action="store_true",
```
(store_false is when the sense of the variable is reversed from the sense of the flag, e.g. below where the flag is called "no-input" but the variable is called "interactive")

django/core/management/commands/squashmigrations.py
```diff
@@ -140,14 +150,37 @@ def handle(self, **options):
...
+                    and isinstance(operation.field, models.fields.related.RelatedField)):
+                        #and operation field app_label != app_label:
+                        continue
+
+                    filtered_operations.append(operation)
+
+                operations.extend(filtered_operations)
+                continue
+
             operations.extend(smigration.operations)
```
shaib, line 182: This should probably be put under `else:`

shaib (commented):
I see that this is just initial, so no detailed review, I just raised points which I suspect will save you some time in further work

django/core/management/commands/squashmigrations.py
pygeek: I was working through the logic at that point, prior to changing the existing code. I just committed a more complete solution. Perhaps, you could clarify something else for me, however.
1. How do I clean up the squash file my test creates?
2. Should I exclude multi-app `ModelOperation`s altogether, or mutate them to exclude fields that reference other apps?
3. What would you recommend doing in the case of swappable tuple? Is there a predictable way to check if it's referencing an external app? Let me know if the logic I currently have makes sense.

I'm still writing tests. Also, let me know if I missed anything.

PS: I may decide to check the `dependencies` for multi-app dependencies, before combing through all of the operations to find references to other apps.

shaib: Hi, sorry, I've somehow missed your questions.
I'm currently at the DjangoCon Europe sprints, and am picking this up.
I've rebased your work on the current main branch, that is now at https://github.com/shaib/django/tree/squashmig-no-deps-clinton

Let me know if you want to keep working on it, although it's been almost two years.

shaib: In any case, to answer your questions (for myself too...):

1. To write tests for migration-file-generating commands, use the `temporary_migration_module` context manager (defined in `MigrationTestBase`, so available as an attribute of `self` in test classes).
2. Many `ModelOperation`s are, in fact, irrelevant, because they do not operate on fields; so it is better to focus at the more-specific operation level. and even then, it changes by case. In the case of `CreateModel`, it is better to modify the operation, because it is likely that migrations down the line will rely on the model existing -- even if they don't rely on the relation field. In the case of Index or Constraint operations (some of which, by the way, do not inherit ModelOperation), it is better to remove the whole operation, because the constraint or index will be entirely incorrect if it is missing a field.
3. Swappable targets mean that the same app, in different projects, will be using different models. In that case, one must assume that they are not from our own app -- as app authors, we have explicitly given control of this to someone else.

PR #18280, opened 2024-06-18
Fixed #27452 - Added SerialField.

csirmazbendeguz:
# Trac ticket number
ticket-27452

# Branch description
A basic implementation of serial fields without support for altering field types.

It adds SerialFields without the complexity of supporting altering between different types (see previous PR).

[Previous PR](https://github.com/django/django/pull/18123)

# Checklist
- [X] I have attached screenshots in both light and dark modes for any UI changes.

### DIFF
--- a/django/contrib/postgres/fields/serial.py
+++ b/django/contrib/postgres/fields/serial.py
@@ -1,0 +1,76 @@
+from django.db import models
+from django.db.models import NOT_PROVIDED
+from django.db.models.expressions import DatabaseDefault
+from django.utils.translation import gettext_lazy as _
+
+__all__ = ("BigSerialField", "SmallSerialField", "SerialField")
+
+
+class SerialFieldMixin:
+    db_returning = True
+
+    def __init__(self, *args, **kwargs):
+        default = DatabaseDefault()
+
+        if not kwargs.setdefault("blank", True):
+            raise ValueError(f"{self.__class__.__name__} must be blank.")
+        if kwargs.setdefault("null", False):
+            raise ValueError(f"{self.__class__.__name__} must not be null.")
+        if kwargs.setdefault("default", default) is not default:
+            raise ValueError(f"{self.__class__.__name__} cannot have a default.")
+        if kwargs.get("db_default", NOT_PROVIDED) is not NOT_PROVIDED:
+            raise ValueError(
+                f"{self.__class__.__name__} cannot have a database default."
+            )
+
+        super().__init__(*args, **kwargs)
+
+    def deconstruct(self):
+        name, path, args, kwargs = super().deconstruct()
+        del kwargs["default"]
+        return name, path, args, kwargs
+
+    def get_prep_value(self, value):
+        value = super().get_prep_value(value)
+        if value is None:
+            value = self.get_default()
+        return value
+
+
+class BigSerialField(SerialFieldMixin, models.BigIntegerField):
+    description = _("Big serial")
+
+    def get_internal_type(self):
+        return "BigSerialField"
+
+    def db_type(self, connection):
+        return "bigserial"
+
+    def rel_db_type(self, connection):
+        return "bigint"
+
+
+class SmallSerialField(SerialFieldMixin, models.SmallIntegerField):
+    description = _("Small serial")
+
+    def get_internal_type(self):
+        return "SmallSerialField"
+
+    def db_type(self, connection):
+        return "smallserial"
+
+    def rel_db_type(self, connection):
+        return "smallint"
+
+
+class SerialField(SerialFieldMixin, models.IntegerField):
+    description = _("Serial")
+
+    def get_internal_type(self):
+        return "SerialField"
+
+    def db_type(self, connection):
+        return "serial"
+
+    def rel_db_type(self, connection):
+        return "integer"
--- a/django/db/backends/postgresql/operations.py
+++ b/django/db/backends/postgresql/operations.py
@@ -2,6 +2,7 @@ import json
 from functools import lru_cache, partial
 
 from django.conf import settings
+from django.contrib.postgres.fields.serial import SerialFieldMixin
 from django.db.backends.base.operations import BaseDatabaseOperations
 from django.db.backends.postgresql.psycopg_any import (
     Inet,
@@ -43,6 +44,15 @@ class DatabaseOperations(BaseDatabaseOperations):
         "AutoField": "integer",
         "BigAutoField": "bigint",
         "SmallAutoField": "smallint",
+        "SerialField": "integer",
+        "BigSerialField": "bigint",
+        "SmallSerialField": "smallint",
+    }
+    integer_field_ranges = {
+        **BaseDatabaseOperations.integer_field_ranges,
+        "SmallSerialField": (-32768, 32767),
+        "SerialField": (-2147483648, 2147483647),
+        "BigSerialField": (-9223372036854775808, 9223372036854775807),
     }
 
     if is_psycopg3:
@@ -55,6 +65,9 @@ class DatabaseOperations(BaseDatabaseOperations):
             "PositiveSmallIntegerField": numeric.Int2,
             "PositiveIntegerField": numeric.Int4,
             "PositiveBigIntegerField": numeric.Int8,
+            "SmallSerialField": numeric.Int2,
+            "SerialField": numeric.Int4,
+            "BigSerialField": numeric.Int8,
         }
 
     def unification_cast_sql(self, output_field):
@@ -250,7 +263,7 @@ class DatabaseOperations(BaseDatabaseOperations):
             # underlying sequence name from the table name and column name.
 
             for f in model._meta.local_fields:
-                if isinstance(f, models.AutoField):
+                if isinstance(f, (models.AutoField, SerialFieldMixin)):
                     output.append(
                         "%s setval(pg_get_serial_sequence('%s','%s'), "
                         "coalesce(max(%s), 1), max(%s) %s null) %s %s;"
@@ -265,9 +278,6 @@ class DatabaseOperations(BaseDatabaseOperations):
                             style.SQL_TABLE(qn(model._meta.db_table)),
                         )
                     )
-                    # Only one AutoField is allowed per model, so don't bother
-                    # continuing.
-                    break
         return output
 
     def prep_for_iexact_query(self, x):
--- a/docs/ref/contrib/postgres/fields.txt
+++ b/docs/ref/contrib/postgres/fields.txt
@@ -920,3 +920,28 @@ define :class:`~django.contrib.postgres.constraints.ExclusionConstraint`. See
 docs/current/rangetypes.html#RANGETYPES-INCLUSIVITY>`_.
 
 .. _different bounds: https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-IO
+
+``Serial`` fields
+=================
+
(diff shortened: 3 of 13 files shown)


LilyFoote (approved):
Looks good to me!

docs/ref/contrib/postgres/fields.txt
```diff
@@ -920,3 +920,24 @@ define :class:`~django.contrib.postgres.constraints.ExclusionConstraint`. See
...
+``Serial`` fields
+=================
+
+.. versionadded:: 5.2
+
+.. class:: SerialField(**options)
+
+    An automatically incrementing :class:`~django.db.models.IntegerField`,
+    much like :class:`~django.db.models.AutoField` except that it does not have
+    to be a primary key, and is not limited to a single instance per model.
```
charettes, line 933: Maybe we should mention that it's backed by the Postgres `serial` type?

charettes (commented):
I think we need to override `rel_db_type` as well. Do we have any tests that covers foreign keys to a unique `SerialField`?

charettes:
Apart from the `get_prep_value` question and the possible issues with foreign keys to a unique `SerialField` attempted to be created as serials this is looking great to me!

docs/ref/contrib/postgres/fields.txt
csirmazbendeguz: ok

django/contrib/postgres/fields/serial.py
```diff
@@ -0,0 +1,67 @@
...
+    def deconstruct(self):
+        name, path, args, kwargs = super().deconstruct()
+        del kwargs["default"]
+        return name, path, args, kwargs
+
+    def get_prep_value(self, value):
+        value = super().get_prep_value(value)
+        if value is None:
+            value = self.get_default()
+        return value
```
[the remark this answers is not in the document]
csirmazbendeguz: It's to allow `SerialModel.objects.create(serial=None)`. If we want to disable this, it can be removed.

The reason it's not in `AutoFieldMixin` is because primary keys are a special case, if it's set to `None`, it's omitted from the `INSERT` query.

csirmazbendeguz:
Thanks a lot @charettes , you were right ! I wasn't aware. I pushed the foreign key fix.

LilyFoote:
Looking at the differences between this PR and the other, I'm wondering if we should move the `inspectdb` changes into this one too. If I understand correctly, they're independent of the migration story that's causing difficulty.

csirmazbendeguz:
> Looking at the differences between this PR and the other, I'm wondering if we should move the `inspectdb` changes into this one too. If I understand correctly, they're independent of the migration story that's causing difficulty.

@LilyFoote , I haven't added the introspection changes here because I tried to keep this PR minimal. If you agree, I would open a separate PR to add that once this is merged. It could be added separately, especially because I feel like the smaller I keep my PRs the faster they can be reviewed and merged by the maintainers.

It's also questionable if we want to add that at all, since we don't have introspection for any other Postgres field either. The only reason I added it to the original PR was that I was using introspection in the tests.

docs/releases/5.2.txt
```diff
@@ -70,7 +70,10 @@ Minor features
 :mod:`django.contrib.postgres`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-* ...
+* :class:`~django.contrib.postgres.fields.SerialField`,
+  :class:`~django.contrib.postgres.fields.BigSerialField` and
+  :class:`~django.contrib.postgres.fields.SmallSerialField` have been added
+  corresponding to their respective PostgreSQL database types.
```
sarahboyce, line 76: @nessita do you think we should promote this a headline feature?

(nit: we will change this to present tense but want to check where we should put it first)

tests/postgres_tests/test_serial.py
```diff
@@ -0,0 +1,344 @@
...
+        self.assertEqual(obj_1.small_serial, 1)
+        self.assertEqual(obj_1.serial, 1)
+        self.assertEqual(obj_1.big_serial, 1)
+
+        obj_2 = SerialModel.objects.create()
+        self.assertEqual(obj_2.small_serial, 2)
+        self.assertEqual(obj_2.serial, 2)
+        self.assertEqual(obj_2.big_serial, 2)
+
+        obj_3 = SerialModel.objects.create(
```
sarahboyce, line 127: I think we should either break this test up into smaller cases with descriptive test names for what we're checking or add one or two comments. I was trying to see if any of the object creation is unnessary

This is maybe a suggestion, we can probably do better
```python
        # When setting to None, will continue the sequence.
        obj_3 = SerialModel.objects.create(
            small_serial=None,
            serial=None,
            big_serial=None,
        )
        self.assertEqual(obj_3.small_serial, 3)
        self.assertEqual(obj_3.serial, 3)
        self.assertEqual(obj_3.big_serial, 3)

        # Serial fields do not need to be unique.
        manual_obj_1 = SerialModel.objects.create(
            small_serial=1,
            serial=1,
            big_serial=1,
        )
        self.assertEqual(manual_obj_1.small_serial, 1)
        self.assertEqual(manual_obj_1.serial, 1)
        self.assertEqual(manual_obj_1.big_serial, 1)

        # The sequence is not interrupted by other objects.
        manual_obj_5 = SerialModel.objects.create(
            small_serial=5,
            serial=5,
            big_serial=5,
        )
        self.assertEqual(manual_obj_5.small_serial, 5)
        self.assertEqual(manual_obj_5.serial, 5)
        self.assertEqual(manual_obj_5.big_serial, 5)

        obj_4 = SerialModel.objects.create()
        self.assertEqual(obj_4.small_serial, 4)
        self.assertEqual(obj_4.serial, 4)
        self.assertEqual(obj_4.big_serial, 4)

        obj_5 = SerialModel.objects.create()
        self.assertEqual(obj_5.small_serial, 5)
        self.assertEqual(obj_5.serial, 5)
        self.assertEqual(obj_5.big_serial, 5)
```

tests/postgres_tests/test_serial.py
```diff
@@ -0,0 +1,344 @@
...
+            {"small_serial": small_hi},
+            {"serial": lo},
+            {"serial": hi},
+            {"big_serial": big_lo},
+            {"big_serial": big_hi},
+        )
+
+        for kwargs in test_cases:
+            with self.subTest(kwargs=kwargs):
+                SerialModel(**kwargs).full_clean()
```
sarahboyce, line 193:
```suggestion
                self.assertIsNone(SerialModel(**kwargs).full_clean())
```
Optional

tests/postgres_tests/test_serial.py
```diff
@@ -0,0 +1,344 @@
...
+
+class SerialPKModelTests(SerialModelBaseTests):
+    def test_create(self):
+        self.reset_sequences_to_1(SerialPKModel)
+
+        obj_1 = SerialPKModel.objects.create()
+        self.assertEqual(obj_1.id, 1)
+
+        obj_2 = SerialPKModel.objects.create()
+        self.assertEqual(obj_2.id, 2)
```
sarahboyce, line 320:
```suggestion
```
I think we do basically this inside `SerialFKModelTests.test_create` so we probably don't need this

docs/ref/contrib/postgres/fields.txt
```diff
@@ -920,3 +920,28 @@ define :class:`~django.contrib.postgres.constraints.ExclusionConstraint`. See
...
+
+    It's backed by PostgreSQL's ``serial`` type. See `the PostgreSQL
+    documentation for the full details <https://www.postgresql.org/
+    docs/current/datatype-numeric.html#DATATYPE-SERIAL>`__.
```
sarahboyce, line 937:
```suggestion
    This uses the PostgreSQL ``serial`` type. See `the PostgreSQL documentation
    for the full details <https://www.postgresql.org/docs/current/
    datatype-numeric.html#DATATYPE-SERIAL>`__.
```

sarahboyce (commented):
🥳 looks really good
I also checked loading fixtures and in the admin, all working nicely 👍

tests/postgres_tests/test_serial.py
csirmazbendeguz: `full_clean` never returns anything so I'm not sure what you're suggesting?

tests/postgres_tests/test_serial.py
csirmazbendeguz: I added some comments, let me know if that's acceptable?

csirmazbendeguz:
Thanks a lot @sarahboyce ! I addressed your comments.

tests/postgres_tests/test_serial.py
sarahboyce: Yes, I like it thank you 👍

tests/postgres_tests/test_serial.py
sarahboyce: `.full_clean()` will either raise an error or return None, so in some of the existing tests around `.full_clean()` it has an `assertIsNone` so that the test has an assert
It's not ideal to have tests without asserts, another option I've seen is saving afterwards and asserting the values

csirmazbendeguz: Okay, if it's a matter of consistency then of course let's do it this way, I just personally disagree.

docs/releases/5.2.txt
```diff
@@ -70,7 +70,10 @@ Minor features
 :mod:`django.contrib.postgres`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-* ...
+* :class:`~django.contrib.postgres.fields.SerialField`,
+  :class:`~django.contrib.postgres.fields.BigSerialField` and
+  :class:`~django.contrib.postgres.fields.SmallSerialField` are added
+  corresponding to their respective PostgreSQL database types.
```
sarahboyce, line 76:
```suggestion
* The new fields :class:`~django.contrib.postgres.fields.SerialField`,
  :class:`~django.contrib.postgres.fields.BigSerialField` and
  :class:`~django.contrib.postgres.fields.SmallSerialField` are automatically
  incrementing integer fields that use the PostgreSQL ``serial`` type. These
  are similar to :class:`~django.db.models.AutoField` except that they do not
  have to be a primary key, and are not limited to a single instance per model.
```

sarahboyce:
(You are very welcome to add yourself to the [AUTHORS](https://github.com/django/django/blob/main/AUTHORS) if you wish to be listed 🥇)

docs/releases/5.2.txt
sarahboyce: On reflection, let's leave it for now. I know you have bigger things planned and we can always promote it later 👍

csirmazbendeguz:
Thanks @sarahboyce , I added all suggestions.

sarahboyce (approved):
Awesome work 🎊

django/db/backends/postgresql/operations.py
```diff
@@ -2,6 +2,7 @@
 from functools import lru_cache, partial
 
 from django.conf import settings
+from django.contrib.postgres.fields.serial import SerialFieldMixin
```
adamchainz, line 5: Ideally, contrib apps would not be imported by core Django, nor require any knowledge within the core. They should act like third-party packages, with no privileged treatment...

csirmazbendeguz:
[forum discussion](https://forum.djangoproject.com/t/serialfields-yay-or-nay/32541)

PR #18283, opened 2024-06-19
Document the unusability metadata of password

Flimm:
This is a small patch to documentation.

It isn't clear how `set_unusable_password` marks a user's password as unusable. This commit clarifies that the `password` field itself contains metadata marking a password as unusable.

# Trac ticket number

N/A

### DIFF
--- a/docs/ref/contrib/auth.txt
+++ b/docs/ref/contrib/auth.txt
@@ -54,7 +54,8 @@ Fields
 
         Required. A hash of, and metadata about, the password. (Django doesn't
         store the raw password.) Raw passwords can be arbitrarily long and can
-        contain any character. See the :doc:`password documentation
+        contain any character. The metadata in this field may mark the password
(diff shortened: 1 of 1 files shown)


jaap3:
I agree that this is an important distinction. 

`has_usable_password` can also use some clarification. It strictly looks at the `UNUSABLE_PASSWORD_PREFIX` (and `None`) to determine that a password is usable. It considers a blank string to be usable. This is currently implied, but requires the reader to understand wat `set_unusable_password` does in the first place.

matthiask (approved):
I wondered if "metadata" should be explained. But then I saw that it's already used in the docs without any explanation, so this is a strict improvement to the status quo.

nessita (approved):
Thank you @Flimm for this work and thank you @jaap3 and @matthiask for reviews :star2: 

I pushed a small tweaked and will land when CI is green again.

PR #18284, opened 2024-06-19
Update validators.py

majid-vaghari:
Fix unnecessary lazy evaluation in BaseValidator

# Trac ticket number

ticket-N/A

# Branch description
if you run this:

```python
from django.utils.translation import gettext_lazy
from django.core.validators import MinValidator

MinValidator(0, message=gettext_lazy('hi'))
```

This will raise:

```
django.core.exceptions.AppRegistryNotReady: The translation infrastructure cannot be initialized before the apps registry is ready. Check that you don't make non-lazy gettext calls at import time.
```

And the problem is this line because it tries to evaluate message instead of checking if it is None. This PR will solve this issue

# Checklist

### DIFF
--- a/django/core/validators.py
+++ b/django/core/validators.py
@@ -401,7 +401,7 @@ class BaseValidator:
 
     def __init__(self, limit_value, message=None):
         self.limit_value = limit_value
-        if message:
+        if message is not None:
             self.message = message
 
     def __call__(self, value):


majid-vaghari:
Alternatively the `__bool__` method on lazy text object could be overridden but this seemed easier. 🤷🏻‍♂️

jaap3:
This would be a change in behaviour, for example:

```python
MinValidator(0, message="")
```

Will now override the message, while before it wouldn't.

It appears there's nothing depending on that behaviour in Django's test suite, and it seems like reasonable behaviour, but someone might be depending on the old behaviour.

Adding tests will help clear up the behaviour and prevent future regressions.

majid-vaghari:
> This would be a change in behaviour, for example:
> 
> ```python
> MinValidator(0, message="")
> ```
> 
> Will now override the message, while before it wouldn't.
> 
> It appears there's nothing depending on that behaviour in Django's test suite, and it seems like reasonable behaviour, but someone might be depending on the old behaviour.
> 
> Adding tests will help clear up the behaviour and prevent future regressions.

I see your point but if you examine the file mentioned, all other validator classes use `message is not None` when they want to check the message. Also, this call to evaluate lazy text objects is not necessary. So it feels like this is the intended behavior. Nevertheless I also agree that adding tests for these behaviors is beneficial and more verbose. I'll try to add them in the following day

majid-vaghari:
@jaap3 I tried adding test cases in `tests/validators/tests.py` But it seems in here django is already setup. For example if I add this test:
```python
class TestGetTextLazy(SimpleTestCase):
    def test_min_value(self):
        print('hi')
        print(_('hello'))
```
It runs with no problems whatsoever. Can you help me where I can test django so it is not already setup and apps are not loaded?

majid-vaghari:
Can anyone please check this pull request?

majid-vaghari:
@jaap3 Can you please check this?

jaap3:
Sorry, I don't know how to get a testcase working for this.

PR #18285, opened 2024-06-19
Fixed #35461 -- Added detail in django tutorial step 8 so that django…

RosanaRufer:
One solution for https://code.djangoproject.com/ticket/35461

# Trac ticket number
[ticket-35461](https://code.djangoproject.com/ticket/35461)

# Branch description
- Following the tutorial 'as is' won't lead to the toolbar working in /polls because the tutorial omits usage of complete HTML templates.
- This solution proposal implements an option of the list of options discussed in the thread. Reasoning available [here](https://www.notion.so/Tutorial-instructions-for-adding-django-debug-toolbar-are-misleading-1f1429081df4473cb5bad80790cc4176).

Screenshot 2024-06-19 at 21 16 09

# Checklist

### DIFF
--- a/docs/intro/tutorial08.txt
+++ b/docs/intro/tutorial08.txt
@@ -48,11 +48,11 @@ The steps are not duplicated in this tutorial, because as a third-party
 package, it may change separately to Django's schedule.
 
 Once installed, you should be able to see the DjDT "handle" on the right side
-of the browser window when you refresh the polls application. Click it to open
-the debug toolbar and use the tools in each panel. See the `panels
-documentation page
-<https://django-debug-toolbar.readthedocs.io/en/latest/panels.html>`__ for more
-information on what the panels show.
+of the browser window when you browse to ``http://localhost:8000/admin/``.
+Click it to open the debug toolbar and use the tools in each panel. See the
+`panels documentation page`__ for more information on what the panels show.
+
+__ https://django-debug-toolbar.readthedocs.io/en/latest/panels.html
 
 Getting help from others
 ========================


docs/intro/tutorial08.txt
```diff
@@ -48,12 +48,16 @@ The steps are not duplicated in this tutorial, because as a third-party
 package, it may change separately to Django's schedule.
 
 Once installed, you should be able to see the DjDT "handle" on the right side
-of the browser window when you refresh the polls application. Click it to open
+of the browser window when you refresh the admin of your project. Click it to open
```
tim-schilling, line 51:
```suggestion
of the browser window when you browse to the admin of your project (http://localhost:8000/admin). Click it to open
```
I'm not quite sure the user will be in the polls area of their web application. It might be more appropriate to tell them to browse to the admin. I'm not sure about including the URL.

docs/intro/tutorial08.txt
```diff
@@ -48,12 +48,16 @@ The steps are not duplicated in this tutorial, because as a third-party
...
 the debug toolbar and use the tools in each panel. See the `panels
 documentation page
 <https://django-debug-toolbar.readthedocs.io/en/latest/panels.html>`__ for more
 information on what the panels show.
 
+To see the toolbar in the polls application in addition to the admin, make sure 
+all your templates are wrapped in complete HTML and their content is 
+inside the ``<body></body>`` tags.
```
tim-schilling, line 59: This is a pretty long sentence and complete HTML is something that probably needs further explanation for beginners.

```suggestion
To see the toolbar in the polls application make sure all your templates
are using `complete HTML documents <https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/Getting_started#anatomy_of_an_html_document>`_.
The toolbar will only render in responses that contain the ``<body></body>`` tags.
```

Note: I didn't cover the line-lengths / formatting well.

RosanaRufer: Good one, applied in https://github.com/django/django/pull/18285/commits/4e774ca18c9ae93eece57907c85567c14454f7bb

docs/intro/tutorial08.txt
RosanaRufer: 👌 https://github.com/django/django/pull/18285/commits/4e774ca18c9ae93eece57907c85567c14454f7bb

tim-schilling (approved):

docs/intro/tutorial08.txt
```diff
@@ -48,12 +48,15 @@ The steps are not duplicated in this tutorial, because as a third-party
...
-<https://django-debug-toolbar.readthedocs.io/en/latest/panels.html>`__ for more
+of the browser window when you browse to the admin of your project (http://localhost:8000/admin). 
+Click it to open the debug toolbar and use the tools in each panel. 
+See the `panels documentation page <https://django-debug-toolbar.readthedocs.io/en/latest/panels.html>`__ for more
 information on what the panels show.
 
+To see the toolbar in the polls application make sure all your templates
+are using `complete HTML documents <https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/Getting_started#anatomy_of_an_html_document>`_.
+The toolbar will only render in responses that contain the ``<body></body>`` tags.
+
```
sarahboyce, line 59:
```suggestion
of the browser window when you browse to ``http://localhost:8000/admin/``.
Click it to open the debug toolbar and use the tools in each panel. See the
`panels documentation page`__ for more information on what the panels show.

To see the DjDT "handle" in the polls application make sure all your templates
are using `complete HTML documents`__. The "handle" will only render in
responses that contain ``<body></body>`` tags.

__ https://django-debug-toolbar.readthedocs.io/en/latest/panels.html
__ https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/Getting_started#anatomy_of_an_html_document

```

- Applied some [line length wrapping](https://docs.djangoproject.com/en/5.0/internals/contributing/writing-documentation/#guidelines-for-restructuredtext-files)
- Previously we call it the `DjDT "handle"` and then `the toolbar` (tweaked)

docs/intro/tutorial08.txt
```diff
@@ -48,12 +48,15 @@ The steps are not duplicated in this tutorial, because as a third-party
...
-documentation page
-<https://django-debug-toolbar.readthedocs.io/en/latest/panels.html>`__ for more
+of the browser window when you browse to the admin of your project (http://localhost:8000/admin). 
+Click it to open the debug toolbar and use the tools in each panel. 
+See the `panels documentation page <https://django-debug-toolbar.readthedocs.io/en/latest/panels.html>`__ for more
 information on what the panels show.
 
+To see the toolbar in the polls application make sure all your templates
+are using `complete HTML documents <https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/Getting_started#anatomy_of_an_html_document>`_.
+The toolbar will only render in responses that contain the ``<body></body>`` tags.
```
sarahboyce, line 58: We can have this in an `... admonition::` or we can remove this paragraph entirely?
In your reasoning docs, you mention
> The main goal of this section is to learn how to add a package in the first place and a bit of django-debug-toolbar in second place.

It's a good point, after telling them they can see this on the polls app we don't encourage them to do anything with this information (such as checking a query).
This information is certainly secondary (they don't have to do this) and possibly not worth the headspace? What do you think?

sarahboyce (commented):
Thank you for this @RosanaRufer 👍 
Really liked how you linked your reasoning doc 😁 I have a couple of thoughts

docs/intro/tutorial08.txt
RosanaRufer: Thanks @sarahboyce. Did you use one of the commands in the `makefile` to automatically format these or is it done manually? Thanks

docs/intro/tutorial08.txt
RosanaRufer: Done in https://github.com/django/django/pull/18285/commits/d5133759f662f0c6b6a90ed3c904fa16c4c5ed70

sarahboyce: > I think it's a reasonable question for people to ask when going through the tutorial of "Why does the toolbar not show up for the application I've built?". The toolbar doesn't make it super obvious that you need a </body> tag since you need to go to the reference documentation to find that.

I agree it's a reasonable question. 
I like to think of it this way... for this tutorial, which is often the first time someone uses Django, we are designing the journey for the reader. In this designed journey, they don't need to worry. The less information/worries the better

Curious readers can google and explore on their own

tim-schilling: I understand the reasoning to keep the tutorials/documentation as slim as possible and people can eventually figure out why. I don't agree with it though because it's not a clear reason that can be quickly searched for on the internet.

sarahboyce: Should be an admonition for sure 

My first google of "why isn't django-debug-toolbar showing" had the relevant docs from django-debug-toolbar appear 🤷‍♀️ but I know the issue and what the fix is. Maybe other people would google differently 

Saw someone on reddit ask this question of the tutorial and there's the ticket raiser too, but we previously told them to go to the polls application and that it would be visible there. Hard to know if anyone would do that without us telling them to

tim-schilling: If you'd like to leave it off here, I'll add it to the toolbar's docs to explain what's going on.

sarahboyce: Ok let's proceed with this now, an admonition can be added at a later date if required 👍

sarahboyce (approved):
Thank you @RosanaRufer ⭐ welcome onboard ⛵

PR #18286, opened 2024-06-19
Initial application of last patch from #13376 and made tests pass

nanorepublica:
# Trac ticket number

ticket-13376

# Branch description

This PR adds the ability to control when a message is hidden from view. 

Currently there is restrictions for the number of times a message is iterated over and for an expiry time.

# Checklist

### DIFF
--- a/django/contrib/messages/restrictions.py
+++ b/django/contrib/messages/restrictions.py
@@ -1,0 +1,105 @@
+import time as time_provider
+
+
+class Restriction:
+    JSON_SEPARATOR = '@'
+
+    def __init__(self, name):
+        self.type = name
+
+    def is_expired(self):
+        """
+        indicates whether given message should be removed
+        """
+        raise NotImplementedError ()
+
+    def on_display(self):
+        """
+        called when iterated - does nothing by default
+        """
+        pass
+
+    def to_json(self):
+        """
+        returns json representation of restriction
+        """
+        raise NotImplementedError
+
+    @classmethod
+    def from_json_param(cls, *args):
+        """
+        returns restriction on the basis of data encoded in json
+        """
+        raise NotImplementedError
+
+    def __cmp__(self, other):
+        if self.__eq__(other): return 0
+        return -1
+
+
+class TimeRestriction(Restriction):
+    JSON_TYPE_CODE = 't'
+
+    def __init__(self, seconds):
+        """
+        seconds - expiration time since now
+        """
+        Restriction.__init__(self, 'time')
+        created = time_provider.time()
+        self.expires = created + int(seconds)
+
+    def set_expirity_time(self, expiration_time):
+        """
+        Sets expilcity expiration time
+        """
+        self.expires = int(expiration_time)
+
+    def is_expired(self):
+        return self.expires < time_provider.time()
+
+    def __eq__(self, other):
+        return self.type == other.type and not bool(self.expires ^ other.expires)
+
+    def __hash__(self):
+        return self.expires
+
+    def to_json(self):
+        return '%s%s%s' % (self.JSON_TYPE_CODE, self.JSON_SEPARATOR, self.expires)
+
+    @classmethod
+    def from_json_param(cls, expirity_time):
+        ret = TimeRestriction(0)
+        ret.set_expirity_time(expirity_time)
+        return ret
+
+
+
+class AmountRestriction(Restriction):
+    JSON_TYPE_CODE = 'a'
+
+    def __init__(self, amount):
+        assert int(amount) >= 0
+        Restriction.__init__(self, 'amount')
+        self.can_be_shown = int(amount)
+
+    def on_display(self):
+        self.can_be_shown -= 1
+
+    def is_expired(self):
+        return int(self.can_be_shown) <= 0
+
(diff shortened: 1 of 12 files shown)


nanorepublica:
This is my initial application of the last patch attached to the Trac ticket. 

There is more to do to apply the full patch, which I will be adding soon

django/contrib/messages/storage/base.py
```diff
@@ -28,11 +56,18 @@ def _prepare(self):
     def __eq__(self, other):
         if not isinstance(other, Message):
             return NotImplemented
-        return self.level == other.level and self.message == other.message
+        return (
+            self.level == other.level
+            and self.message == other.message
+            and self.restrictions == other.restrictions
```
matthiask, line 62: If restrictions were made hashable `self.restrictions` could be a `set()`, which would mean that this comparison isn't depending upon the ordering of restrictions in the messages themselves.

django/contrib/messages/storage/cookie.py
```diff
@@ -31,12 +36,19 @@ class MessageDecoder(json.JSONDecoder):
     Decode JSON that includes serialized ``Message`` instances.
     """
 
+    def create_message(self, *vars):
+        """creates message on the basis of encoded data"""
+        args = vars[:-1]
+        restrictions = vars[-1]
```
matthiask, line 42: I appreciate keyword-only arguments...! Makes sense, but it's not nice to look at.

matthiask (commented):
Interesting, thanks. I want to pay attention to this because of https://pypi.org/project/django_user_messages/

nanorepublica:
@matthiask Thanks for the notes. This PR is a simple application of the patch that was originally on the ticket with most conflicts resolved.

I am pausing on this PR until I get this smaller [PR](https://github.com/django/django/pull/18332) merged with does less in terms of code changes but allows third-party packages like yours to extend the messages framework in a more useful way which would include the restrictions mentioned here. Once that PR is merged I will reconsider this PR in the context of the ticket.

django/contrib/messages/restrictions.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def __init__(self, seconds):
+        """
+        seconds - expiration time since now
+        """
+        Restriction.__init__(self, "time")
+        created = time_provider.time()
+        self.expires = created + int(seconds)
+
+    def set_expirity_time(self, expiration_time):
```
anorthall, line 52:
```suggestion
    def set_expiry_time(self, expiration_time):
```

django/contrib/messages/restrictions.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        Sets expilcity expiration time
```
anorthall, line 54:
```suggestion
        Set expiration time to the given absolute value
```

django/contrib/messages/restrictions.py
```diff
@@ -0,0 +1,107 @@
...
+    def __hash__(self):
+        return self.expires
+
+    def to_json(self):
+        return "%s%s%s" % (self.JSON_TYPE_CODE, self.JSON_SEPARATOR, self.expires)
+
+    @classmethod
+    def from_json_param(cls, expirity_time):
+        ret = TimeRestriction(0)
+        ret.set_expirity_time(expirity_time)
```
anorthall, line 73:
```suggestion
    def from_json_param(cls, expiry_time):
        ret = TimeRestriction(0)
        ret.set_expiry_time(expiry_time)
```

django/contrib/messages/restrictions.py
```diff
@@ -0,0 +1,107 @@
...
+        return ret
+
+
+class AmountRestriction(Restriction):
+    JSON_TYPE_CODE = "a"
+
+    def __init__(self, amount):
+        assert int(amount) >= 0
+        Restriction.__init__(self, "amount")
```
anorthall, line 82: Should we use a call to `super()` instead of calling `Restriction` explicitly?

django/contrib/messages/storage/base.py
```diff
@@ -1,21 +1,49 @@
...
     storage classes (typically session- or cookie-based) and rendered in a view
     or template.
     """
 
-    def __init__(self, level, message, extra_tags=None):
+    def __init__(self, level, message, extra_tags=None, restrictions=[]):
         self.level = int(level)
         self.message = message
         self.extra_tags = extra_tags
+        self.restrictions = restrictions or list([res.AmountRestriction(1)])
```
anorthall, line 43: Is the `list()` call necessary?

django/contrib/messages/storage/base.py
```diff
@@ -1,21 +1,49 @@
...
+
 class Message:
     """
     Represent an actual message that can be stored in any of the supported
     storage classes (typically session- or cookie-based) and rendered in a view
     or template.
     """
 
-    def __init__(self, level, message, extra_tags=None):
+    def __init__(self, level, message, extra_tags=None, restrictions=[]):
```
anorthall, line 39: Would it be better to keep away from putting a mutable list as a default, and use `None`, in case it opens up to mutation bugs later?

PR #18290, opened 2024-06-20
Updated screenshots in tutorial.

sarahboyce:
Want to update the tutorial images.

This cause caused me to raise two tickets:
- https://code.djangoproject.com/ticket/35540 in this ticket `docs/intro/_images/admin01.png` will be updated
- https://code.djangoproject.com/ticket/35541 in this ticket `docs/intro/_images/admin04t.png` and `docs/intro/_images/admin06t.png` will be updated

Open questions:
- ~I have kept the sidebar in the images (not sure if they existed back when these images were last updated). I can remove the sidebar, I kept it as this is "likely what they will see". What do you think?~ - removed the top and left sidebar to limit the image. Added advantage is this will be less sensitive to css updates
- ~I updated the inlines to show 2 rather than 3, that's only because it was two big for a nice screenshot and I would need to zoom out. I can change this back to 3. What do you think?~ reverted

Ran the minimizer commands 👍 

These screenshots appear on these tutorial pages:
- https://docs.djangoproject.com/en/5.0/intro/tutorial02/#start-the-development-server (1-6)
- https://docs.djangoproject.com/en/5.0/intro/tutorial07/#customize-the-admin-form (7-14)


nessita:
>     * I have kept the sidebar in the images (not sure if they existed back when these images were last updated). I can remove the sidebar, I kept it as this is "likely what they will see". What do you think?

I reviewed the before/after screenshots, and I think we should consider removing or hiding the sidebar. My main concern is that these images are part of the tutorial. Showing only the elements relevant to the tutorial flow would align better with our guidance of keeping the tutorial focused and avoiding any distractions or confusion for the user/student.

>     * I updated the inlines to show 2 rather than 3, that's only because it was two big for a nice screenshot and I would need to zoom out. I can change this back to 3. What do you think?

I'm on the fence for this one because it will trigger new translations for these pages. I would suggest we keep the 3 extra.

> Other:
> 
>     * I think I need to run the [image minimizer commands](https://docs.djangoproject.com/en/5.0/internals/contributing/writing-documentation/#minimizing-images). I don't have the required software installed

Did you try to install these packages in WSL? they may be available there and usable! I think you only need to run:
```
sudo apt update -y
sudo apt install -y optipng
```

sarahboyce:
> Did you try to install these packages in WSL?

This worked thank you 🥳

docs/intro/_images/admin10t.png
sarahboyce, line 1: Potentially a css issue, the headers of `Choice #1` used to be more visually distinguished than they are now, stronger bottom border and background colour

docs/intro/_images/admin11t.png
nessita, line 1: I'm wondering why the "Votes" field, which is an integer field, does not seem to be showing as an `input type="number"` (the "up" and "down" arrows are not shown). Could this be a browser issue?

Also the buttons that used to be included are now removed, was this intentional?

docs/intro/_images/admin10t.png
nessita: Could we perhaps create a separated ticket to trac/triage this?

docs/intro/_images/admin07.png
nessita, line 1: This one is adding the submit row to the screenshot where and it wasn't showing before. I wonder if we should be consistent on always showing or never showing unless the tutorial specifically talks about this row. What do you think?

docs/intro/_images/admin05t.png
nessita, line 1: These and other screenshots are "loosing" the breadcrumbs, is that intentional? I wonder whether the breadcrumbs help or confuse students...
@evildmp Would you have a suggestion regarding how minimal/constrained should the screenshots be in the Django Tutorial? (please see other similar questions/comments in this PR)

nessita (commented):
Overall looks good! :star2:
I have added some "honest" questions, I think for most of those we can go either way...
Did you try, by any chance, to automate this on top of/next to the WIP PR I started for screenshot automation?

docs/intro/_images/admin05t.png
evildmp: For a user who is not already familiar with an interface, a screenshot of it presents a very heavy cognitive load to process. There is a surprising amount of information, and it's not obvious where or what it all is. 

There is no very good way to guide a reader's eye through a screenshot to whatever we need them to see. There is no natural order of viewing. 

The fewer and less detailed the screenshots the better, in general.

The main problem though is that what the image captures is not necessarily the meaning of what we want to convey. Say for example that an interface uses a dark mode; now what the screenshot shows will be completely different from what the user sees. Or, if the interface is responsive and the user is viewing the screen on a different kind of device. Or, if the user is seeing things in a different language.

In such cases the screenshot can actually distract the user from what we want them to see and notice. Instead of providing reassurance, the screenshot can cause doubt.

That's because we want to draw the user's attention to the *meaning* of what's on-screen, not what it happens to look like.

Also - as we can see here - screenshots can involve quite a bit of maintenance to keep up.

That's not to say they shouldn't be included, and sometimes they are the best thing, or even necessary to include.

docs/intro/_images/admin07.png
evildmp: I don't believe the user will even notice the difference. The user will notice what changes on their own screen, in response to what they have done, and when that change is connected to an action of their, it will be locked into a meaning that will stay with them.

docs/intro/_images/admin11t.png
evildmp: > Also the buttons that used to be included are now removed, was this intentional?

As a pointed example of how difficult it is to make sense of screenshot images, I have spent about five minutes trying to understand which removed buttons you are referring to. 

I still can't identify them. But I spent five minutes with a feeling of doubt and uncertainty (similar to that expressed in your own *I'm wondering why the "Votes" field [...] Could this be a browser issue?*).

sarahboyce: > I'm wondering why the "Votes" field, which is an integer field, does not seem to be showing as an input type="number" (the "up" and "down" arrows are not shown). Could this be a browser issue?

From the browser I took the screenshots from, these are not visible unless the input is in focus. 
We don't currently have standards on screenshots including which browser to take it from.

> Also the buttons that used to be included are now removed, was this intentional?

It was an accident but happened because I had taken the screenshot for `docs/intro/_images/admin14t.png` and this aligns with that screenshot.

> I have spent about five minutes trying to understand which removed buttons you are referring to.

These are the "Save add another" etc buttons which were at the bottom of the old version of this image

docs/intro/_images/admin05t.png
sarahboyce: > These and other screenshots are "loosing" the breadcrumbs, is that intentional?

Yes. Previously I did a screenshot to include the breadcrumbs. This meant that for laptop screens, you see the sidebar. We then decided to not take a screenshot of the sidebar which in turn makes the breadcrumb not fit the screenshot.
We can "hack" this in. One way is showing the screen on tablet view. However, considering we will probably want to show the most likely/similar image to the user, this screenshot covers both laptop and tablet views.

Some other notes...

> Say for example that an interface uses a dark mode; now what the screenshot shows will be completely different from what the user sees.

It is possible in docs to have the screenshot "change" from light mode/dark mode depending on what mode your browser is viewing the docs. This would mean we need to take the screenshot twice, once in dark mode and once in light mode. See [this Sphinx theme](https://github.com/pradyunsg/furo) with the image change as a feature.

> Or, if the interface is responsive and the user is viewing the screen on a different kind of device.

Currently an issue, assuming laptop screen

> Or, if the user is seeing things in a different language.

As this is the intro tutorial, the default settings mean this will display in English. This however would be an issue for our other screenshots

docs/intro/_images/admin07.png
sarahboyce: If we want the screenshot to be very focused, we can take it of the fields without the headers and bottom submit row.
When I was taking a series of screenshots, I tried to make them roughly consistent which meant almost always including the header and submit row
I don't mind too much

sarahboyce:
> Did you try, by any chance, to automate this on top of/next to the WIP PR I started for screenshot automation?

No, I thought it would be quicker to do this manually for now
In the process of looking at the screenshots, it seems we are defining what the screenshots should be. We should also probably check with each one "is it really needed" for the tutorial to reduce the required maintenance. I also think we are missing good helpers for cropping screenshots to a particular section.

docs/intro/_images/admin10t.png
sarahboyce: https://code.djangoproject.com/ticket/35558 :+1:

sarahboyce:
- only resized images to a certain width if they were larger, if they were smaller I have not resized them because I don't like stretching images
- images 10 and 14 are updated in https://github.com/django/django/pull/18308
- 481af5b148c5dc822880a5ed439867eb95059766 is from running `advpng -z4 `find . -type f -not -path "./_build/*" -name "*.png"``

nessita (approved):

PR #18291, opened 2024-06-20
Exclude pylibmc on MacOS and Windows

jaap3:
# Trac ticket number

N/A

# Branch description

Following the [quickstart](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/unit-tests/#quickstart) I ran into the following issue:

```
$ python -m pip install -r requirements/py3.txt
...
Building wheels for collected packages: pylibmc
  Building wheel for pylibmc (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building wheel for pylibmc (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [6 lines of output]
      In file included from src/_pylibmcmodule.c:34:
      src/_pylibmcmodule.h:42:10: fatal error: 'libmemcached/memcached.h' file not found
      #include <libmemcached/memcached.h>
               ^~~~~~~~~~~~~~~~~~~~~~~~~~
      1 error generated.
      error: command '/usr/bin/clang' failed with exit code 1
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for pylibmc
Failed to build pylibmc
```

I'm on MacOS, libmemcached is not available and I'm not sure how to get it.

Seeing as Windows is already exempt from this requirement, I've added MacOS to be excluded as well.

Not sure if this requires a track ticket?

# Checklist
- [x] The commit message is written in past tense, ~~mentions the ticket number~~, and ends with a period.

### DIFF
--- a/tests/requirements/py3.txt
+++ b/tests/requirements/py3.txt
@@ -8,8 +8,8 @@ geoip2; python_version < '3.13'
 jinja2 >= 2.11.0
 numpy; sys_platform != 'win32' or python_version < '3.13'
 Pillow >= 6.2.1; sys_platform != 'win32' or python_version < '3.13'
-# pylibmc/libmemcached can't be built on Windows.
-pylibmc; sys_platform != 'win32'
+# pylibmc/libmemcached can't be built on Windows and MacOS.
+pylibmc; sys_platform != 'win32' and sys_platform != 'darwin'
 pymemcache >= 3.4.0
 pywatchman; sys_platform != 'win32'
 PyYAML


nessita:
Hello @jaap3, thank you for this PR!

From some chats at the DjangoCon Europe sprints, I *think* that the right solution is to pass a special env var when installing the dependencies. A few search results point to [this solution](https://sendapatch.se/projects/pylibmc/install.html#from-pypi):
```
brew install libmemcached
pip install pylibmc --install-option="--with-libmemcached=/usr/local/Cellar/libmemcached"
```
(the path may be different in your specific Mac, perhaps something inside `/opt/homebrew`? see for example https://github.com/lericson/pylibmc/issues/101)

Could you confirm this works? If so we may need to tweak the docs to mention this (and we could re-purpose this PR for that).

jaap3:
@nessita Thanks for the suggestion. I'd rather avoid installing homebrew, I've had terrible experiences with it messing up my system in the past. 

Having to install homebrew and compiling code just to run Django's test suite (as part of an exploratory change that is unrelated to memcache) is a bit much. Seeing as the test suite does pass without it installed, it can perhaps be skipped until really desired?

ajosephau:
I'm on a Mac m2 (16-inch, 2023) running macOS Sequoia 15.0.1 (24A348) and I was able to get Django running on `main` branch working by running:

```
brew install libmemcached
LIBMEMCACHED=/opt/homebrew/Cellar/libmemcached/1.0.18_2 pip install pylibmc
```

(noting the path to `libmemcached` was based wherever homebrew installed it.) I know you would like to avoid homebrew, but here is a [macports](https://ports.macports.org/port/libmemcached/) package if that helps?

nessita:
Thank you everyone for your input. I think that with the provided information, this PR should be closed since we don't want to exclude pylibmc in MacOS.

PR #18293, opened 2024-06-21
Fixed #35413 -- Fixed lookup error message to be clearer.

wookkl:
# Trac ticket number
[ticket-35413](https://code.djangoproject.com/ticket/35413)

# Branch description
Fixed lookup error message to be clearer.

It seems that the only way to output a clear message is to pass the `lookups` to the `try_transform` method.

If you suggest another way, I'll try it!

# Checklist

### DIFF
--- a/django/db/models/sql/query.py
+++ b/django/db/models/sql/query.py
@@ -1369,7 +1369,7 @@ class Query(BaseExpression):
         # __exact is the default lookup if one isn't given.
         *transforms, lookup_name = lookups or ["exact"]
         for name in transforms:
-            lhs = self.try_transform(lhs, name)
+            lhs = self.try_transform(lhs, name, lookups)
         # First try get_lookup() so that the lookup takes precedence if the lhs
         # supports both transform and lookup for the name.
         lookup_class = lhs.get_lookup(lookup_name)
@@ -1403,7 +1403,7 @@ class Query(BaseExpression):
 
         return lookup
 
-    def try_transform(self, lhs, name):
+    def try_transform(self, lhs, name, lookups=None):
         """
         Helper method for build_lookup(). Try to fetch and initialize
         a transform for name parameter from lhs.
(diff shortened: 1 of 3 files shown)


AMK9978 (approved):

django/db/models/sql/query.py
```diff
@@ -1369,14 +1369,14 @@ def build_lookup(self, lookups, lhs, rhs):
         # __exact is the default lookup if one isn't given.
         *transforms, lookup_name = lookups or ["exact"]
         for name in transforms:
-            lhs = self.try_transform(lhs, name)
+            lhs = self.try_transform(lhs, name, lookups)
```
sarahboyce, line 1372: This one appears to be key and is covered by tests 👍 

All the others I can update from `.try_transform(lhs, lookup_name, lookups)` to `.try_transform(lhs, lookup_name, [lookup_name])` without any test failures

This makes me think, are we missing some tests?
If the answer to that question is "no, it can only fail here" then maybe we shouldn't force three arguments to `try_transform` and instead have something like...
```python
    def try_transform(self, lhs, name, lookups=None):
         # ....
            if lookups is not None:
                name_index = lookups.index(name)
                unsupported_lookup = LOOKUP_SEP.join(lookups[name_index:])
            else:
                unsupported_lookup = name
            raise FieldError(
                "Unsupported lookup '%s' for %s or join on the field not "
                "permitted%s" % (unsupported_lookup, output_field.__name__, suggestion)
            )
```
so passing in lookups is optional 🤔

sarahboyce (commented):
Thank you @wookkl - this looks really good

I added some thoughts 👍

django/db/models/sql/query.py
wookkl: Hi @sarahboyce. Thank you for your feedback 🙏🏻

I have confirmed that the Lookup error occurs only here.
Therefore, based on your feedback, I have modified the `try_transform` method to make passing `lookups` optional.
Also added the missing tests.

sarahboyce (approved):
Thank you @wookkl ⭐ 
Welcome onboard ⛵

PR #18294, opened 2024-06-21
Fixed #35545, Refs #32833 -- Fixed ContentTypeManager.get_for_models() crash in CreateModel migrations.

sarahboyce:
# Trac ticket number

ticket-35545

# Branch description

# Checklist

### DIFF
--- a/django/contrib/contenttypes/models.py
+++ b/django/contrib/contenttypes/models.py
@@ -75,7 +75,7 @@ class ContentTypeManager(models.Manager):
                 ct = self._get_from_cache(opts)
             except KeyError:
                 needed_models[opts.app_label].add(opts.model_name)
-                needed_opts[opts].append(model)
+                needed_opts[(opts.app_label, opts.model_name)].append(model)
             else:
                 results[model] = ct
         if needed_opts:
@@ -89,18 +89,13 @@ class ContentTypeManager(models.Manager):
             )
             cts = self.filter(condition)
             for ct in cts:
-                opts_models = needed_opts.pop(
-                    ct._meta.apps.get_model(ct.app_label, ct.model)._meta, []
-                )
+                opts_models = needed_opts.pop((ct.app_label, ct.model), [])
                 for model in opts_models:
                     results[model] = ct
                 self._add_to_cache(self.db, ct)
         # Create content types that weren't in the cache or DB.
-        for opts, opts_models in needed_opts.items():
-            ct = self.create(
-                app_label=opts.app_label,
-                model=opts.model_name,
-            )
+        for (app_label, model_name), opts_models in needed_opts.items():
+            ct = self.create(app_label=app_label, model=model_name)
             self._add_to_cache(self.db, ct)
             for model in opts_models:
                 results[model] = ct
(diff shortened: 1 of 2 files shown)


sarahboyce:
@csirmazbendeguz does this work for you?

tests/contenttypes_tests/test_models.py
```diff
@@ -99,6 +99,24 @@ def test_get_for_models_migrations(self):
             cts, {ContentType: ContentType.objects.get_for_model(ContentType)}
         )
 
+    def test_get_for_models_migrations_create_model(self):
```
charettes, line 108: You'll want to use `@isolate_apps("contenttypes_tests")` here to avoid having `Foo` pollute the global registry and leak through the suite.

django/contrib/contenttypes/models.py
```diff
@@ -89,18 +89,13 @@ def get_for_models(self, *models, for_concrete_models=True):
             )
             cts = self.filter(condition)
             for ct in cts:
-                opts_models = needed_opts.pop(
-                    ct._meta.apps.get_model(ct.app_label, ct.model)._meta, []
-                )
+                opts_models = needed_opts.pop((ct.app_label, ct.model), [])
```
charettes, line 92: I assume this is the crux of the issue.

charettes (commented):
I wonder if this is the right approach as it will allow models that are not registered in the registry associated with the `ContentType` model to have content types created for them.

Maybe that's a sign of a large issue?

tests/contenttypes_tests/test_models.py
sarahboyce: Thank you ⭐

django/contrib/contenttypes/models.py
sarahboyce: Yes this is the crux of it 👍

sarahboyce:
> Maybe that's a sign of a larger issue in the original patch?

Which patch are you referring to here? :)

csirmazbendeguz:
@sarahboyce , your branch is working for me

charettes:
> Which patch are you referring to here? 🙂

The one that introduced the regression https://github.com/django/django/pull/18105

charettes (approved):
> I wonder if this is the right approach as it will allow models that are not registered in the registry associated with the ContentType model to have content types created for them.

It appears that `get_for_model` already has this peculiarity which explains why the code was working before #18105 was merged and switched to `get_for_models`.

Since it makes `get_for_models` more coherent with the canonical `get_for_model` I think this is fine.

tests/contenttypes_tests/test_models.py
```diff
@@ -2,14 +2,20 @@
...
-from django.test import TestCase, override_settings
+from django.db.migrations.state import ModelState, ProjectState
+from django.test import TestCase, TransactionTestCase, override_settings
 from django.test.utils import isolate_apps
 
 from .models import Author, ConcreteModel, FooWithUrl, ProxyModel
 
 
-class ContentTypesTests(TestCase):
+class ContentTypesTests(TransactionTestCase):
```
charettes, line 12: I'm surprised `TransactionTestCase` is required here, what's the reason for it? It makes the test way slower as it cannot use transactions to isolate database changes and must flush untouched tables.

sarahboyce: There was a test failure in the CI and I panicked 😁 I was missing the `@isolate_apps("contenttypes_tests")` 
I will revert 👍

PR #18297, opened 2024-06-22
Fixed #25782 – update docstring on CacheMiddleware and cache_page.

wassef911:
…cache_page.

# Trac ticket number
ticket-25782

# Branch description
Mention a warning against a drawback of using **CacheMiddleware** and **cache_page**, if cache keys are different this might be storing the same values twice and results in doubling the cache size. 

# Checklist


sarahboyce:
Thank you for this PR @wassef911 🎉 

I think I would expect to update the documentation of our docs rather than this docstring, most likely adding a warning in here: https://docs.djangoproject.com/en/5.0/topics/cache/#django.views.decorators.cache.cache_page?

Having something like 
```
.. warning:: Avoiding setting ``cache`` when  ``UpdateCacheMiddleware`` is
    installed

    If ``cache`` is set to anything other than the
    :setting:`CACHE_MIDDLEWARE_ALIAS` setting and
    :class:`~django.middleware.cache.UpdateCacheMiddleware` is installed, the 
    response will be saved to both caches.
```
But probably worded a bit nicer 

Have you had a look at what it would take to resolve the issue? This might be preferable than documenting it 🤔

wassef911:
thanks for the pointer @sarahboyce, 
I'm still a bit lost on how things work TBH. (my first pr)

I'm starting to look into this and wrote a test case under **cache.tests.CacheMiddlewareTest**. But that issue seems fixed? and I just rushed in without trying I guess? 

```python
    def test_view_decorator_and_middleware_not_duplicating_values(self):
        default_view = cache_page(3, cache="other")(hello_world_view)
        middleware = CacheMiddleware(default_view, cache_alias=None)
        request = self.factory.get("/view/")

        response = default_view(request, "1")
        result = middleware.process_request(request)

        # Process the response through the middleware
        processed_response = middleware.process_response(request, response)

        self.assertEqual(response.content, b"Hello World 1")
        self.assertEqual(middleware.cache_alias, "default")
        self.assertIsNone(result)
        self.assertEqual(processed_response.content, b"Hello World 1")
```

any idea on what todo next? close the ticket or bring it up on the forum?

sarahboyce:
> thanks for the pointer @sarahboyce, I'm still a bit lost on how things work TBH. (my first pr)

No problem 👍 

> I'm starting to look into this and wrote a test case under **cache.tests.CacheMiddlewareTest**. But that issue seems fixed? and I just rushed in without trying I guess?

I can see that the response is cached twice so I think it's not fixed

```python
@override_settings(
    CACHES={
        "default": {"BACKEND": "django.core.cache.backends.locmem.LocMemCache"},
        "other": {
            "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
            "LOCATION": "other",
        },
    },
)
class DuplicateCacheKey(TestCase):
    def test_cache_page_conflict(self):
        @cache_page(3000, cache="other")
        def test_view(request):
            return HttpResponse(b"Hello, world")

        request = RequestFactory().get("/view/")
        cache_key = "views.decorators.cache.cache_page..GET.5b30bdb36b35025daf3570b18c3bb916.d41d8cd98f00b204e9800998ecf8427e.en"

        self.assertEqual(caches["default"].has_key(cache_key), False)
        self.assertEqual(caches["other"].has_key(cache_key), False)

        # Process the request through the middleware.
        UpdateCacheMiddleware(test_view)(request)
        # This has now been saved in both caches.
        self.assertEqual(caches["default"].has_key(cache_key), True)
        self.assertEqual(caches["other"].has_key(cache_key), True)
```

wassef911:
to address this,

added a cache control key called "x-successive-cache" on the response. (no strong opinions about the name. it seems unlikely it will clash with anything.)

and the overall approach is for the **cache_page** to set that using **@cache_control** internally, then the middleware will short-circuit if it exists... as it does for "private". 

the approach was suggested in the ticket description and sounds sane. 

quick side reference:
[MDN Web Docs - Cache-Control](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control).

tests/cache/tests.py
```diff
@@ -2984,3 +2984,41 @@ def test_all(self):
...
+        UpdateCacheMiddleware(test_view)(request)
+        self.assertEqual(caches["other"].has_key(cache_key), True)
+        self.assertEqual(caches["default"].has_key(cache_key), False)
+
+        CacheMiddleware(test_view)(request)
+        self.assertEqual(caches["other"].has_key(cache_key), True)
+        self.assertEqual(caches["default"].has_key(cache_key), False)
+
+        response = test_view(request)
+        self.assertEqual(response.content, b"Hello, world")
```
sarahboyce, line 3024: I want to try and make it clearer in the test that this is retreived from the cache, this is my idea but you might have a better one 👍 
```diff
--- a/tests/cache/tests.py
+++ b/tests/cache/tests.py
@@ -2996,11 +2996,13 @@ class CacheHandlerTest(SimpleTestCase):
     },
 )
 class DuplicateCacheKey(TestCase):
-    def test_cache_page_conflict(self):
-        """cache_page isn't causing duplicates with middlewares"""
+
+    @mock.patch("time.sleep")
+    def test_cache_page_conflict(self, mock_sleep):

         @cache_page(3000, cache="other")
         def test_view(request):
+            time.sleep(1)
             return HttpResponse(b"Hello, world")

         request = RequestFactory().get("/view/")
@@ -3015,10 +3017,13 @@ class DuplicateCacheKey(TestCase):
         UpdateCacheMiddleware(test_view)(request)
         self.assertEqual(caches["other"].has_key(cache_key), True)
         self.assertEqual(caches["default"].has_key(cache_key), False)
+        mock_sleep.assert_called_once()

         CacheMiddleware(test_view)(request)
         self.assertEqual(caches["other"].has_key(cache_key), True)
         self.assertEqual(caches["default"].has_key(cache_key), False)
+        mock_sleep.assert_called_once()

         response = test_view(request)
         self.assertEqual(response.content, b"Hello, world")
+        mock_sleep.assert_called_once()
```

django/middleware/cache.py
```diff
@@ -121,6 +121,12 @@ def process_response(self, request, response):
             cache_key = learn_cache_key(
                 request, response, timeout, self.key_prefix, cache=self.cache
             )
+            # check header for "x-successive-cache" to avoid multiple cache sets
+            # if not found, set it as key used
+            if response.has_header("x-successive-cache"):
+                return response
+            response["x-successive-cache"] = cache_key
```
sarahboyce, line 128: I think maybe `X-Cache-Success` would be better (I think headers tend to be title cased), I have no idea how to tell whether it is unlikely to collide with other headers, I agree it _feels_ unlikely

We probably want to set the header as something short and constant (as we currently don't care about the value) maybe something like `"true"`

sarahboyce (commented):
Thank you @wassef911 this looks great ⭐ 
I'm not 100% sure on header best practises so at some point it might be worth us checking with another person but I like where this is going 👍

tests/cache/tests.py
wassef911: yeah I think this is better, although one call on the last line is enough ?

sarahboyce: Yes or maybe we reset the mock and assert that it isn't called 🤔

django/middleware/cache.py
wassef911: after a bit more digging found that "Last-Modified" header will be present on cached responses after processing... If that exists we can confidently skip.

tests/cache/tests.py
wassef911: okay yeah sure

sarahboyce (approved):
Thank you @wassef911 this looks great 🥳 

Before I merge this, I saw in the ticket that it mentions how ticket-15855 should be resolved before this change
That ticket has quite a mature patch which would need cleaning up, applying to main, and then the following [open suggestions](https://github.com/django/django/pull/7229#pullrequestreview-7330859)
- Adding a release note to 5.2 within the "Miscellaneous" section and maybe adding some `.. versionchanged:: 5.2` notes
- Confirming whether it is still necessary to have UpdateCacheMiddleware at the top of the middlewares list after applying that change (and if not [updating the docs](https://docs.djangoproject.com/en/5.0/topics/cache/#order-of-middleware))

wassef911:
now, with that issue in mind, let's loop back to square 0 ? 😂  

idea is:
* if we skip ".set" call in the middleware like I did above, we solve duplication but we lose accounting for varys headers set by other successive middlewares in the "cache_key". 

* if we skip ".set" call in the decorators, we also solve duplication but lose granular control allowed by the decorators in projects having any subclass of "UpdateCacheMiddleware".


so to keep both and still solve initial problem, we should catch when these "two ways of doing things" are conjunction and let the decorator only mark the response for caching, not actually doing it, while forwarding whatever kwargs it has to the middleware to override it's end behaviour.

---- 

to maintain the same implementation we'd ideally create a new middleware that does nothing, only marking by a temporary header carrying that behaviour. ("Cache-Marker" : {**things})

for testing we'll have a middleware actually installed instead of just processing then keep the same assertions... ?

wassef911:
another implementation for the same idea might go like: 

rewrite the cache decorator to either instantiate a middleware and use or append the key to headers...  

but it will alot typing to get async/sync views to both work.  
so I'd say we stick to use of "decorator_from_middleware_with_args" ??

wassef911:
@sarahboyce, update is that I backed out from pushing what's mentioned above. 

I guess [ticket-15855](https://code.djangoproject.com/ticket/15855) will have to be worked on in a separate PR. 

The implementation I had in my draft wasn't adequate. 

--- 

for now, I updated release notes for the approved change. 
and, Yes I confirm, UpdateCacheMiddleware still needs to be at the top for it to handle undecorated views etc...

django/middleware/cache.py
```diff
@@ -121,6 +121,11 @@ def process_response(self, request, response):
             cache_key = learn_cache_key(
                 request, response, timeout, self.key_prefix, cache=self.cache
             )
+            # avoid multiple cache sets
+            if response.has_header("X-Cache-Success"):
+                return response
+            response["X-Cache-Success"] = "true"
```
sarahboyce, line 127:
```suggestion
            # Avoid duplicating cache.
            response["X-Cache-Success"] = "true"
```
And have the if statement above be 
```python
        if (
            timeout
            and response.status_code == 200
            and not response.has_header("X-Cache-Success")
        ):
```

docs/releases/5.2.txt
```diff
@@ -287,3 +287,6 @@ Miscellaneous
 
 * The ``all`` argument for the ``django.contrib.staticfiles.finders.find()``
   function is deprecated in favor of the ``find_all`` argument.
+
+* Subclasses of :class:`~django.middleware.cache.UpdateCacheMiddleware`
+  will no longer cause duplication when used with cache decorator.
```
sarahboyce, line 292:
```suggestion
```
We don't add release notes for bug fixes

wassef911:
> * Confirming whether it is still necessary to have UpdateCacheMiddleware at the top of the middlewares list after applying that change (and if not [updating the docs](https://docs.djangoproject.com/en/5.0/topics/cache/#order-of-middleware))

Its not necessary to have UpdateCacheMiddleware on top, it's only a workaround suggested by #15855, but let's question it?
as without FetchFromCacheMiddleware, the UpdateCacheMiddleware will be just setting the value to some cache with the correct headers but not actually serving them as the value by this decorator will be the one served... so the workaround just causes this problem and fixes nothing.

with this PR, will avoid that unnecessary extra write in the middleware, and continue having “Vary headers“ missing.  

on the other hand I think the "vary header" problem is smaller to have compared to doubling cache size so we should proceed with this until someone gathers the courage to deprecate the decorator?

django/middleware/cache.py
apollo13: We should not set a header that leaks back to the client!

apollo13 (changes requested):
Returning a header to the client is no option, the suggested heaader is not a standard header and even if it were we do not want to expose such an information to the requesting client.

sarahboyce (changes requested):
Thank you for the review @apollo13 

@wassef911 there is a `login_not_required` decorator and `LoginRequiredMiddleware` where one skips the other which might be worth looking into :+1:

PR #18298, opened 2024-06-22
Fixed #35511 -- Documentation uses the py binary in Windows, which may not always be available

alexgmin:
# Trac ticket number

[ticket-35511](https://code.djangoproject.com/ticket/35511)

# Branch description

# Checklist

### DIFF
--- a/docs/howto/windows.txt
+++ b/docs/howto/windows.txt
@@ -32,6 +32,14 @@ matches the version you installed by executing:
 
     ...\> py --version
 
+.. admonition:: ``py`` is not recognized or found
+
+    Depending on how you've installed Python (such as via the Microsoft Store),
+    ``py`` may not be available in the command prompt.
+
+    You will then need to use ``python`` instead of ``py`` when entering
+    commands.
+
 .. seealso::
 
     For more details, see :doc:`python:using/windows` documentation.


docs/howto/windows.txt
```diff
@@ -25,6 +25,8 @@ website should offer you a download button for the latest Python version.
 Download the executable installer and run it. Check the boxes next to "Install
 launcher for all users (recommended)" then click "Install Now".
 
+If you installed Python via the Microsoft Store, ``py`` won't be available, you will need to use ``python3`` instead.
```
sarahboyce, line 28: Thank you for pushing this forward @alexgmin ⭐ 

How do you feel about something like...
```
.. admonition:: ``py`` is not recognized

    Depending on how you've installed Python, ``py`` may not be available. You
    will need to use ``python`` or ``python3`` instead.
```
To be after the `...\> py --version` which is where they would first encounter the error.

Maybe let's not mention the Microsoft Store specifically as it may not be the only way that you can install Python without `py` being available

Note: all docs need to be [wrapped after 80 characters](https://docs.djangoproject.com/en/5.0/internals/contributing/writing-documentation/#guidelines-for-restructuredtext-files)

smithdc1: I'm sceptical about including `python3` here. Just `python` these days? 🤔

alexgmin: I'll be happy to change the position of the comment and wrapping the text (maybe there should be a CI check for that). However, on this comment:

> Maybe let's not mention the Microsoft Store specifically as it may not be the only way that you can install Python without `py` being available

Based on the [official docs](https://docs.python.org/3/using/windows.html), there are 2 other ways to install Python on Windows, but I wouldn't consider them as possible options for someone going through the tutorial.
I think is better to mention the Microsoft Store because if you run `python` or `python3` without having Python installed it [will open the store](https://learn.microsoft.com/en-us/windows/python/faqs#why-does-running-python-exe-open-the-microsoft-store-)

> I'm sceptical about including `python3` here. Just `python` these days? 🤔
Both are still available, and the [Python docs](https://docs.python.org/3/using/windows.html#windows-store) mention both, and that you can have `python3.11` and `python3.12` at the same time.

sarahboyce: `py` is also not available if you  opted out of this with the installer wizard. So it is not only limited to installations via Microsoft Store 

I guess this whole note is for users when "you are aware that you already have Python installed but you do not have access to `py`". We should perhaps change the title of the admonition to something of that affect?

alexgmin: I'm not sure if the opting out in the installer wizard is an option that should be considered that much for a tutorial page. 
You would have to click the checkbox in the installer to change it, which I doubt any person going through the tutorial at a basic level would do.

Said that, what do you think of this:
```
.. admonition:: ``py`` is not recognized

    Depending on how you've installed Python (via the Microsoft Store, for example), ``py`` may not be available. You
    will need to use ``python`` instead.
```
I also removed the `python3` suggestion and I think the other two mentions of the `python3` binary should be changed, but that's a different PR.

Not sure if it is right now, but I'd also suggest changing the title so it's the same or close to the same to the Windows error given when you execute a not found binary. I guess Powershell and CMD would return different wordings.

alexgmin: > Not sure if it is right now, but I'd also suggest changing the title so it's the same or close to the same to the Windows error given when you execute a not found binary. I guess Powershell and CMD would return different wordings.

CMD returns this error:
> py : The term 'py' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.

Powershell returns this one:
> + ~~
>     + CategoryInfo          : ObjectNotFound: (py:String) [], CommandNotFoundException
>     + FullyQualifiedErrorId : CommandNotFoundException

Since I assume most users will be using CMD I think the admonition title should stay as suggested:
`admonition:: ``py`` is not recognized`

I'm happy with the first title above but another option is to to change it to this:
`admonition:: ``py`` is not recognized or found`

sarahboyce: Both sound like good options to me

sarahboyce (approved):
Thank you @alexgmin ⭐

PR #18304, opened 2024-06-24
Fixed #35554, Refs #35060 -- Improved *args parsing in Model.save()/asave().

adamchainz:
# Trac ticket number

ticket-35554

# Branch description

Change the parsing introduced in 3915d4c70d0d7673abe675525b58117a5099afd3 / PR #17667 to handle some edge cases, with new tests.

# Checklist

### DIFF
--- a/django/db/models/base.py
+++ b/django/db/models/base.py
@@ -803,18 +803,20 @@ class Model(AltersData, metaclass=ModelBase):
                 RemovedInDjango60Warning,
                 stacklevel=2,
             )
-            for arg, attr in zip(
-                args, ["force_insert", "force_update", "using", "update_fields"]
-            ):
-                if arg:
-                    if attr == "force_insert":
-                        force_insert = arg
-                    elif attr == "force_update":
-                        force_update = arg
-                    elif attr == "using":
-                        using = arg
-                    else:
-                        update_fields = arg
+            total_len_args = len(args) + 1  # include self
+            if total_len_args > 5:
+                # Recreate the proper TypeError message from Python.
+                raise TypeError(
+                    "Model.save() takes from 1 to 5 positional arguments but "
+                    f"{total_len_args} were given"
+                )
+            force_insert = args[0]
+            try:
+                force_update = args[1]
+                using = args[2]
+                update_fields = args[3]
+            except IndexError:
+                pass
 
         self._prepare_related_fields_for_save(operation_name="save")
 
@@ -888,18 +890,20 @@ class Model(AltersData, metaclass=ModelBase):
                 RemovedInDjango60Warning,
                 stacklevel=2,
             )
-            for arg, attr in zip(
-                args, ["force_insert", "force_update", "using", "update_fields"]
-            ):
-                if arg:
-                    if attr == "force_insert":
-                        force_insert = arg
-                    elif attr == "force_update":
-                        force_update = arg
-                    elif attr == "using":
-                        using = arg
-                    else:
-                        update_fields = arg
+            total_len_args = len(args) + 1  # include self
+            if total_len_args > 5:
+                # Recreate the proper TypeError message from Python.
+                raise TypeError(
+                    "Model.asave() takes from 1 to 5 positional arguments but "
+                    f"{total_len_args} were given"
+                )
+            force_insert = args[0]
+            try:
+                force_update = args[1]
+                using = args[2]
+                update_fields = args[3]
+            except IndexError:
+                pass
 
         return await sync_to_async(self.save)(
             force_insert=force_insert,
--- a/tests/update_only_fields/tests.py
+++ b/tests/update_only_fields/tests.py
@@ -1,5 +1,6 @@
 from django.db.models.signals import post_save, pre_save
 from django.test import TestCase
+from django.utils.deprecation import RemovedInDjango60Warning
 
 from .models import Account, Employee, Person, Profile, ProxyEmployee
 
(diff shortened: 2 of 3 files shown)


adamchainz:
@salvo-polizzi @jacobtylerwalls @felixxm : as you were involved in the original PR, would you like to take a look?

django/db/models/base.py
```diff
@@ -803,18 +803,20 @@ def save(
...
+            if len_args >= 3:
+                using = args[2]
+            if len_args >= 4:
+                update_fields = args[3]
+            if len_args > 4:
+                # Recreate the message from Python
+                raise TypeError(
+                    "Model.save() takes from 1 to 5 positional arguments but "
+                    f"{len_args + 1} were given"
+                )
```
adamchainz, line 819: Just thought of this shorter, maybe faster, maybe less clear alternative:

```python
if len(args) > 4:
                # Recreate the message from Python
                raise TypeError(
                    "Model.save() takes from 1 to 5 positional arguments but "
                    f"{len(args) + 1} were given"
                )
            force_insert = args[0]
            try:
                force_update = args[1]
                using = args[2]
                update_fields = args[3]
            except IndexError:
                pass
```

what do you think?

nessita: My personal preference, and also in order to comply with our documented f-string format style (_f-strings should use only plain variable and property access_):
```python
            total_len_args = len(args) + 1  # include self
            if total_len_args > 5:
                # Recreate the proper TypeError message from Python.
                raise TypeError(
                    "Model.save() takes from 1 to 5 positional arguments but "
                    f"{total_len_args} were given"
                )
            force_insert = args[0]
            try:
                force_update = args[1]
                using = args[2]
                update_fields = args[3]
            except IndexError:
                pass
```
I believe a few more tests are needed because if we comment lines from the above, no test fails. Specifically, removing/commenting out all these lines:
```python
            # force_insert = args[0]
            try:
                # force_update = args[1]
                # using = args[2]
```
produces no test failures.

nessita: Something like this would work as a test:
```python
    def test_save_deprecation_positional_arguments_used(self):
        a = Article()
        fields = ["headline"]
        with (
            self.assertWarns(RemovedInDjango60Warning),
            mock.patch.object(a, "save_base") as mock_save_base,
        ):
            a.save(None, 1, 2, fields)
        mock_save_base.assert_called_once_with(
            using=2, force_insert=None, force_update=1, update_fields=frozenset(fields))
```

nessita (commented):
Looks good, I have suggested the code change that I think is best.

I also think we should fail if both args and kwargs were given for the same option, but perhaps this needs to be a separated ticket. Thoughts?

django/db/models/base.py
adamchainz: Thanks, that’s great. Added to my commit. I changed the test to assert on the `mock_calls` attribute, which I think is clearer and less error-prone, since there’s no chance to mistype `assert_called_onec_with`.

adamchainz:
> I also think we should fail if both args and kwargs were given for the same option, but perhaps this needs to be a separated ticket. Thoughts?

Yes, it would be good to protect against that case too. Maybe another ticket good just to keep this change focused.

The first pattern that comes to my mind is to check each variable against its default before picking from `args`:

```python
if force_insert is not False:
    # uh oh
force_insert = args[0]
if total_len_args >= 3:
    if force_update is not False:
        # uh oh
    force_update = args[1]
...
```

A bit involved.

tests/update_only_fields/tests.py
```diff
@@ -256,6 +257,16 @@ def post_save_receiver(**kwargs):
         pre_save.disconnect(pre_save_receiver)
         post_save.disconnect(post_save_receiver)
 
+    def test_empty_update_fields_positional(self):
```
nessita, line 260: While I don't object to this new test, I'm not sure what's the goal of adding it. Is there any logic/code path that is covered in this test but in the other tests?

adamchainz: The test fails on `main` because empty `update_fields` is falsey - the second issue I listed on the ticket.

```
$ ./runtests.py --parallel 1 update_only_fields
Testing against Django installed in '/Users/chainz/Documents/Projects/django/django'
Found 17 test(s).
.F...............
FAIL: test_empty_update_fields_positional (update_only_fields.tests.UpdateOnlyFieldsTests.test_empty_update_fields_positional)
  File "/Users/chainz/Documents/Projects/django/tests/update_only_fields/tests.py", line 264, in test_empty_update_fields_positional
    with (
  File "/Users/chainz/Documents/Projects/django/django/test/testcases.py", line 128, in __exit__
    self.test_case.assertEqual(
AssertionError: 1 != 0 : 1 queries executed, 0 expected
Captured queries were:
1. UPDATE "update_only_fields_person" SET "name" = 'Sara', "gender" = 'F', "pid" = NULL WHERE "update_only_fields_person"."id" = 1

```

jacobtylerwalls: (This regression was the reason I marked the ticket as a release blocker when I saw it.)

nessita: Thank you both! I have added the `async` version for this issue :rocket:

nessita:
I think the branch is ready for landing, will mark it as so and merge when CI is green.

Thank you @adamchainz and @jacobtylerwalls! :star2

tests/update_only_fields/tests.py
```diff
@@ -256,6 +257,29 @@ def post_save_receiver(**kwargs):
...
+        ):
+            s.save(False, False, None, [])
+
+    async def test_empty_update_fields_positional_asave(self):
+        s = await Person.objects.acreate(name="Sara", gender="F")
+        # Workaround for a lack of async assertNumQueries.
+        s.name = "Other"
+
+        msg = "Passing positional arguments to asave() is deprecated"
+        with (self.assertWarnsMessage(RemovedInDjango60Warning, msg),):
```
jacobtylerwalls, line 276: This is almost not worth mentioning, but would you prefer to un-parenthesize this or was this intentional?

nessita: Thanks for raising this! Is a leftover when I was fighting assertNumQueries not working with async. Fixing ASAP! :bow:

adamchainz:
Thanks for the extra test and fixups @nessita.

> Would you like to create the ticket and I'll triage it?

Done: https://code.djangoproject.com/ticket/35561

nessita:
> > I also think we should fail if both args and kwargs were given for the same option, but perhaps this needs to be a separated ticket. Thoughts?
> 
> Yes, it would be good to protect against that case too. Maybe another ticket good just to keep this change focused.
> 
> The first pattern that comes to my mind is to check each variable against its default before picking from `args`:
> 
> ```python
> if force_insert is not False:
>     # uh oh
> force_insert = args[0]
> if total_len_args >= 3:
>     if force_update is not False:
>         # uh oh
>     force_update = args[1]
> ...
> ```
> 
> A bit involved.

@adamchainz I have proposed #18313 to make the *args and **kwargs more strict and less repetitive. Please let me know what you think! :bow:

PR #18308, opened 2024-06-25
Fixed #35558 -- Applied inline header styling to h3 headers.

sarahboyce:
# Trac ticket number

ticket-35558

# Branch description

# Checklist


nessita (approved):
Looks great, thank you so much! :star2: 

I'm planning to merge asap, but I wanted to share something for future consideration: I find it a bit odd that the inline-related H3 is more prominent (#000) than the page's H2 (#333) in the light theme. This isn't an issue in the dark theme where both are white. (cc/ @knyghty)

knyghty:
> I find it a bit odd that the inline-related H3 is more prominent (#000) than the page's H2 (https://github.com/django/django/pull/333) in the light theme. This isn't an issue in the dark theme where both are white.

Seems like a good idea to me, we should take the dark theme's example here.

nessita:
> > I find it a bit odd that the inline-related H3 is more prominent (#000) than the page's H2 (#333) in the light theme. This isn't an issue in the dark theme where both are white.
> 
> Seems like a good idea to me, we should take the dark theme's example here.

Thanks for the feedback! Question though: would you go with black for both H2 and inline H3?

knyghty:
I think so yes, just with the h2 being a bit bigger than the h3.

nessita:
@knyghty I have followed up on this in ticket-35645 (PR #18432).

PR #18309, opened 2024-06-25
#35560: Model.full_clean() errors with GeneratedField and UniqueConstraint or CheckConstraint

sdolemelipone:
ticket-35560

Calling `full_clean()` on an unsaved model instance which has either a `UniqueConstraint` or `CheckConstraint` raises the error:

`AttributeError: Cannot read a generated field from an unsaved model.`

### DIFF
--- a/tests/model_fields/models.py
+++ b/tests/model_fields/models.py
@@ -609,3 +609,75 @@ class GeneratedModelNullVirtual(models.Model):
 
     class Meta:
         required_db_features = {"supports_virtual_generated_columns"}
+
+
+class GeneratedModelUniqueConstraint(models.Model):
+    name = models.CharField(max_length=10)
+    lower_name = models.GeneratedField(
+        expression=Lower("name"),
+        output_field=models.CharField(max_length=10),
+        db_persist=True,
+    )
+
+    class Meta:
+        required_db_features = {"supports_stored_generated_columns"}
+        constraints = [
+            models.UniqueConstraint(Lower("name"), name="name_uniq_generated")
+        ]
+
+
+class GeneratedModelUniqueConstraintVirtual(models.Model):
+    name = models.CharField(max_length=10)
+    lower_name = models.GeneratedField(
+        expression=Lower("name"),
+        output_field=models.CharField(max_length=10),
+        db_persist=False,
+    )
+
+    class Meta:
+        required_db_features = {"supports_virtual_generated_columns"}
+        constraints = [
+            models.UniqueConstraint(Lower("name"), name="name_uniq_generated_virtual")
+        ]
+
+
+class GeneratedModelCheckConstraint(models.Model):
+    a = models.IntegerField()
+    b = models.IntegerField()
+    field = models.GeneratedField(
+        expression=F("a") + F("b"),
+        output_field=models.IntegerField(),
+        db_persist=True,
+    )
+
+    class Meta:
+        required_db_features = {
+            "supports_stored_generated_columns",
+            "supports_table_check_constraints",
+        }
+        constraints = [
+            models.CheckConstraint(
+                condition=models.Q(a__gt=0), name="check_a_gt_0_generated"
+            )
+        ]
+
+
+class GeneratedModelCheckConstraintVirtual(models.Model):
+    a = models.IntegerField()
+    b = models.IntegerField()
+    field = models.GeneratedField(
+        expression=F("a") + F("b"),
+        output_field=models.IntegerField(),
+        db_persist=False,
+    )
+
+    class Meta:
+        required_db_features = {
+            "supports_stored_generated_columns",
+            "supports_table_check_constraints",
+        }
+        constraints = [
+            models.CheckConstraint(
+                condition=models.Q(a__gt=0), name="check_a_gt_0_generated_virtual"
+            )
+        ]
--- a/tests/model_fields/test_generatedfield.py
+++ b/tests/model_fields/test_generatedfield.py
@@ -18,6 +18,8 @@ from django.test.utils import isolate_apps
 from .models import (
     Foo,
     GeneratedModel,
+    GeneratedModelCheckConstraint,
+    GeneratedModelCheckConstraintVirtual,
     GeneratedModelFieldWithConverters,
     GeneratedModelNull,
     GeneratedModelNullVirtual,
@@ -25,6 +27,8 @@ from .models import (
     GeneratedModelOutputFieldDbCollationVirtual,
     GeneratedModelParams,
     GeneratedModelParamsVirtual,
+    GeneratedModelUniqueConstraint,
+    GeneratedModelUniqueConstraintVirtual,
     GeneratedModelVirtual,
 )
 
@@ -186,6 +190,23 @@ class GeneratedFieldTestMixin:
         m = self._refresh_if_needed(m)
         self.assertEqual(m.field, 3)
 
+    def test_full_clean_with_unique_constraint(self):
+        m = self.unique_constraint_model(name="ABC")
+        # full_clean() ignores GeneratedFields.
+        m.full_clean()
+        m.save()
+        m = self._refresh_if_needed(m)
+        self.assertEqual(m.lower_name, "abc")
+
+    @skipUnlessDBFeature("supports_table_check_constraints")
+    def test_full_clean_with_check_constraint(self):
+        m = self.check_constraint_model(a=1, b=2)
+        # full_clean() ignores GeneratedFields.
+        m.full_clean()
+        m.save()
+        m = self._refresh_if_needed(m)
+        self.assertEqual(m.field, 3)
+
     def test_create(self):
         m = self.base_model.objects.create(a=1, b=2)
         m = self._refresh_if_needed(m)
@@ -305,6 +326,8 @@ class GeneratedFieldTestMixin:
 class StoredGeneratedFieldTests(GeneratedFieldTestMixin, TestCase):
     base_model = GeneratedModel
     nullable_model = GeneratedModelNull
+    unique_constraint_model = GeneratedModelUniqueConstraint
+    check_constraint_model = GeneratedModelCheckConstraint
     output_field_db_collation_model = GeneratedModelOutputFieldDbCollation
     params_model = GeneratedModelParams
 
@@ -318,5 +341,7 @@ class StoredGeneratedFieldTests(GeneratedFieldTestMixin, TestCase):
 class VirtualGeneratedFieldTests(GeneratedFieldTestMixin, TestCase):
     base_model = GeneratedModelVirtual
     nullable_model = GeneratedModelNullVirtual
+    unique_constraint_model = GeneratedModelUniqueConstraintVirtual
+    check_constraint_model = GeneratedModelCheckConstraintVirtual
     output_field_db_collation_model = GeneratedModelOutputFieldDbCollationVirtual
     params_model = GeneratedModelParamsVirtual


sdolemelipone:
A fix for this would be to ignore generated fields in `model._get_field_value_map()` as below. But would that mean that generated fields would be ignored when validating unique and check constraints? If so is that something which should be added to the documentation?

```python
    def _get_field_value_map(self, meta, exclude=None):
        if exclude is None:
            exclude = set()
        meta = meta or self._meta
        field_map = {
            field.name: Value(getattr(self, field.attname), field)
            for field in meta.local_concrete_fields
            if field.name not in exclude and not field.generated
        }
        if "pk" not in exclude:
            field_map["pk"] = Value(self.pk, meta.pk)
        return field_map
```

nessita:
Hello @sdolemelipone, thank you for your contribution!

When you get a chance, could you please restore the PR template checklist in the first comment to ensure that the necessary items are completed in order to put this in the pending review queue?

sdolemelipone:
Hi @nessita, apologies this is my first non-trivial PR and I wasn't familiar with the process. I've restored the format of the ticket.

Yes I'm happy to work on the ticket, I will add my proposed solution as a further commit (but will need feedback/discussion from others on whether it is the correct approach).

Thanks

tests/model_fields/models.py
```diff
@@ -609,3 +609,75 @@ class GeneratedModelNullVirtual(models.Model):
...
+    lower_name = models.GeneratedField(
+        expression=Lower("name"),
+        output_field=models.CharField(max_length=10),
+        db_persist=True,
+    )
+
+    class Meta:
+        required_db_features = {"supports_stored_generated_columns"}
+        constraints = [
+            models.UniqueConstraint(Lower("name"), name="name_uniq_generated")
```
sarahboyce, line 625:
```suggestion
            models.UniqueConstraint(fields=["name"], name="name_uniq_generated"),
            models.UniqueConstraint(fields=["lower_name"], name="name_lower_uniq"),
```
If I update to have a unique constraint on `lower_name`, `StoredGeneratedFieldTests.test_full_clean_with_unique_constraint` still fails with an attribute error `AttributeError: Cannot read a generated field from an unsaved model.`

On your earlier question which was roughly around whether `GeneratedField`s fields should be ignored from `Model.validate_constraints(exclude=None)` - I _think_ so
Agree we should document it maybe as an admonition to `Model.validate_constraints(exclude=None)` (we also need a release note for 5.0.7 for this change).
Not allowing to create constraints on `GeneratedField`s seems harsh to me when it "works"

sarahboyce (commented):
Thank you for picking this up @sdolemelipone ⭐ tried to add hopefully helpful thoughts

tests/model_fields/models.py
sdolemelipone: Hi @sarahboyce, thanks for the review. I'll take a look into why using the `GeneratedField` in `UniqueConstraint.fields` still fails and suggest a change.

> Not allowing to create constraints on `GeneratedFields` seems harsh to me when it "works"
Yes other than `.full_clean()` they appear to work correctly. I'll investigate what would be required to make `full_clean()` work correctly (rather than forcing the exclusion of generated fields from constraints) and suggest an alternative solution.

sdolemelipone: I have added the tests as you suggested, and applied a change to `Constraint.validate()` which was also causing the `GeneratedField` attribute to be accessed before a model instance had been saved. This was because constraints initialised as  `UniqueConstraint(fields=[], ...)` are hanlded differently to constraints initialised via `UniqueConstraint(<expression>, ...)`.

I'll add my thoughts on the bigger picture in a comment  on the main PR.

sdolemelipone:
At present, `Model.validate_unique()` and `Model.validate_constraints()` do not use the modified attribute value for a model instance's `GeneratedField`s which are included in the constraint's `fields` or `expressions`. 

If the `GeneratedField` has `db_persist=True`, the model instance's `GeneratedField` attribute will return the value which corresponds to the most recent refresh of the instance from the database.  Those conditions will be checked at the database level when `Model.save()` is called (and correctly raise an `IntegrityError` if the constraints are violated), but `Model.full_clean()` will not raise a `ValidationError`.

I expect the best way to fix this bug (and resolve this ticket) would be the following:

Update the documentation to warn users that if they use `GeneratedField`s within constraints, those constraints will only work at a database level and not within form validation (so only then if they are non-virtual, i.e. if the constraint has `db_persist=True`).

Then a further ticket could be raised to resolve the situation in one of two ways:
1. Add checks to disallow `GeneratedField`s from being included in any constraints, _or_
2. Add functionality to update the values of all relevant `GeneratedField`s when checking constraints. This will require modification of `UniqueConstraint.validate()` and `CheckConstraint.validate()`, around the same lines of code which have been modified as part of this ticket.

Let me know if you agree and I'll add documentation to this PR as above.

sarahboyce:
> At present, `Model.validate_unique()` and `Model.validate_constraints()` do not use the modified attribute value for a model instance's `GeneratedField`s which are included in the constraint's `fields` or `expressions`.
> 
> The model instance's `GeneratedField` attribute will return the value which corresponds to the most recent refresh of the instance from the database. Those constraints will be checked at the database level when `Model.save()` is called (and correctly raise an `IntegrityError` if the constraints are violated), but `Model.full_clean()` will not raise a `ValidationError`.

Thank you for your analysis, suggestions and patience @sdolemelipone 👍 been thinking about this a little while

I think you're right that we need a new ticket to resolve this, and therefore I think we shouldn't document anything around this issue as part of this PR (as I believe this is a distinct and separate "problem").
As you've done the work around learning what is going on here, would you like to raise that ticket? I'd happily accept it 😁 

These are the next steps in my head:
- drop 2fcef2e88bb243ef0a20401f2daf2048c3f3637b from this PR 
- raise a ticket around 2fcef2e88bb243ef0a20401f2daf2048c3f3637b (include your analysis, patch, ref this PR etc, let's not lose any work here 👍, if you like you can pull out the existing stuff to a new PR)

We also need a release note here for Django 5.0.7 (`docs/releases/5.0.7.txt`), maybe something like
```
* Fixed a bug in Django 5.0 that caused a crash of ``Model.full_clean()`` on
  models with a ``GeneratedField`` and defined :attr:`Meta.constraints 
  <django.db.models.Options.constraints>`  (:ticket:`35560`).
```
What do you think?

sdolemelipone:
> Thank you for your analysis, suggestions and patience @sdolemelipone 👍 been thinking about this a little while
>
You're welcome and no problem, it's an interesting one :-)
 
> I think you're right that we need a new ticket to resolve this, and therefore I think we shouldn't document anything around this issue as part of this PR (as I believe this is a distinct and separate "problem"). As you've done the work around learning what is going on here, would you like to raise that ticket? I'd happily accept it 😁
> 

Yes I'd be happy to raise the next ticket and PR and transfer the information over. I agree it makes sense to just fix the crashes with this ticket.

> These are the next steps in my head:
> 
>     * drop [2fcef2e](https://github.com/django/django/commit/2fcef2e88bb243ef0a20401f2daf2048c3f3637b) from this PR
> 
>     * raise a ticket around [2fcef2e](https://github.com/django/django/commit/2fcef2e88bb243ef0a20401f2daf2048c3f3637b) (include your analysis, patch, ref this PR etc, let's not lose any work here 👍, if you like you can pull out the existing stuff to a new PR)
> 
> 
> We also need a release note here for Django 5.0.7 (`docs/releases/5.0.7.txt`), maybe something like
> 
> ```
> * Fixed a bug in Django 5.0 that caused a crash of ``Model.full_clean()`` on
>   models with a ``GeneratedField`` and defined :attr:`Meta.constraints 
>   <django.db.models.Options.constraints>`  (:ticket:`35560`).
> ```
> 
> What do you think?

The changes made in [2fcef2e](https://github.com/django/django/commit/2fcef2e88bb243ef0a20401f2daf2048c3f3637b) do fix a crash of `Model.full_clean()` when a generated field is included in a Constraint's `fields` parameter...I think they should stay in this PR, unless we are happy to leave a crash occurring until the second PR is released?

The change to allow `GeneratedField`s to be checked as part of Constraints will be a lot more complicated, I would be happy to give it a try but will need a bit of time to work out how Constraint`.validate()` functions.

sarahboyce:
> The changes made in [2fcef2e](https://github.com/django/django/commit/2fcef2e88bb243ef0a20401f2daf2048c3f3637b) do fix a crash of Model.full_clean() when a generated field is included in a Constraint's fields parameter...I think they should stay in this PR, unless we are happy to leave a crash occurring until the second PR is released?

This is how I see it...

The original ticket is that if you have any constraint on your model, adding a `GeneratedField` to that model will break `Model.full_clean()` and makes it challenging to use `GeneratedField` in this case. If you upgrade to Django 5.0, you can only add a `GeneratedField` to a limited number of models. Therefore this is a release blocking bug 👍 

The next one, is given you added a  `GeneratedField` to a model, and decided to add a constraint on this field, then `Model.full_clean()` breaks. This can be phrased something like "Django does not support constraints on a GeneratedField in Model.full_clean() validation".
Given that, I actually prefer the error on `Model.full_clean()` as this makes that very clear that we do not support this yet (rather than silencing the error without doing the constraint check).
We could categorise this as a "Bug" but we could also categorise this as a "New feature". 
As this doesn't break anything existing (you are making a new constraint here on the GeneratedField), and also because we should be able to write a constraint that mimics having this on a GeneratedField (both UniqueConstraint and CheckConstraint accept expressions which can be the expression defined for your GeneratedField) and therefore there is a "work around", I would probably class this as a new feature.

> The change to allow GeneratedFields to be checked as part of Constraints will be a lot more complicated, I would be happy to give it a try but will need a bit of time to work out how Constraint.validate() functions.

That's completely fine. Your work on this is completely voluntary and I apologise if I implied that I expect you to work on this

sdolemelipone:
Thank you for explaining that so throughly! I understand what you're saying and see why that's the better way to approach this. I'll do as you've suggested and add the release note.

sdolemelipone:
All done! I have modified the tests of `UniqueConstraints` slightly as prior to [2fcef2e](https://github.com/django/django/commit/2fcef2e88bb243ef0a20401f2daf2048c3f3637b) the tests did not cover the situation when a `UniqueConstraint` was defined by a `fields=[...]` parameter.

sarahboyce (approved):
Thank you for the report, discussion and fix @sdolemelipone ⭐ 

I have pushed minor tweaks to the docs and tests as after further testing the bug and fix was  isolated to `CheckConstraint` and so I have removed the tests for `UniqueConstraint`.
This makes the commit slightly smaller/more focussed as we will backport this to 5.0 👍 

Thank you again

sdolemelipone:
You're welcome, thank you too :-) I'll get on that next ticket now.

sdolemelipone:
I've added ticket-35575 to deal with the next steps for this issue.

sarahboyce:
> I have pushed minor tweaks to the docs and tests as after further testing the bug and fix was isolated to CheckConstraint and so I have removed the tests for UniqueConstraint.
This makes the commit slightly smaller/more focussed as we will backport this to 5.0 👍

This was wrong and I have added tests back for `UniqueConstraint`, the issue for `UniqueConstraint` is only when there is an expression defined rather than `fields`. This got lost after playing around with the generated fields being in fields - my mistake sorry. Updated 👍

sdolemelipone:
> This was wrong and I have added tests back for `UniqueConstraint`. The issue for `UniqueConstraint` is only when there is an expression defined rather than `fields`. This got lost after playing around with the generated fields being in fields - my mistake sorry. Updated 👍

Mine too - I tried to make the tests more comprehensive by incorporating `fields`, but didn't notice that there was no bug in that case (unless a `generatedField` was included in `fields`). This looks concise now! :+1:

PR #18310, opened 2024-06-25
Removed superfluous request.user assignment in RemoteUserMiddleware.

jaap3:
# Trac ticket number

N/A

# Branch description

[`auth.login` sets `request.user`](https://github.com/django/django/blob/bcc327aa326093a39f01a9bc98198807444900f3/django/contrib/auth/__init__.py#L149-L150), there is no need to assign `request.user` before calling `auth.login`.

The conditional in `auth.login` before it sets `request.user` will always succeed. `RemoteUserMiddleware` raises `ImproperlyConfigured` if it doesn't and will never reach this code in that situation.

# Checklist
- [x] The commit message is written in past tense, ~~mentions the ticket number~~, and ends with a period.

### DIFF
--- a/django/contrib/auth/middleware.py
+++ b/django/contrib/auth/middleware.py
@@ -142,9 +142,8 @@ class RemoteUserMiddleware(MiddlewareMixin):
         # to authenticate the user.
         user = auth.authenticate(request, remote_user=username)
         if user:
-            # User is valid.  Set request.user and persist user in the session
+            # User is valid. Set request.user and persist user in the session
             # by logging the user in.
-            request.user = user
             auth.login(request, user)
 
     def clean_username(self, username, request):


django/contrib/auth/middleware.py
```diff
@@ -142,9 +142,8 @@ def process_request(self, request):
         # to authenticate the user.
         user = auth.authenticate(request, remote_user=username)
         if user:
-            # User is valid.  Set request.user and persist user in the session
+            # User is valid. Set request.user and persist user in the session
             # by logging the user in.
-            request.user = user
```
auvipy, line 147: can you please verify this with proper unit tests?

auvipy (changes requested):
also is this fix related to any open trac ticket?

jaap3:
> also is this fix related to any open trac ticket?

No, I don't think it is required since this is a trivial change

jacobtylerwalls:
Hi @jaap3 👋 

> [auth.login sets request.user](https://github.com/django/django/blob/bcc327aa326093a39f01a9bc98198807444900f3/django/contrib/auth/__init__.py#L149-L150)

Of possible interest: this morning's Django 6.0.4 fixed a regression where `alogin()` did not set `request.user`, so this isn't so trivial after all. Also, I don't see a change in `aprocess_request()`.

 I'm not sure this is worth micro-optimizing, given how sensitive this is. It is the responsibility of `RUMiddleware` to set the user.

PR #18313, opened 2024-06-25
Fixed 35561 -- Made *args and **kwargs parsing more strict in Model.save()/asave().

nessita:
# Trac ticket number

ticket-35561

# Branch description
Fixing a 5.1 release blocker to ensure that arguments passed to `Model.save()` and `Model.asave()` do not allow for unsupported or duplicated parameters.

# Checklist

### DIFF
--- a/django/db/models/base.py
+++ b/django/db/models/base.py
@@ -776,18 +776,59 @@ class Model(AltersData, metaclass=ModelBase):
             return getattr(self, field_name)
         return getattr(self, field.attname)
 
+    # RemovedInDjango60Warning: When the deprecation ends, remove completely.
+    def _parse_args_and_kwargs(self, *args, method_name, **kwargs):
+        defaults = {
+            "force_insert": False,
+            "force_update": False,
+            "using": None,
+            "update_fields": None,
+        }
+
+        invalid_kwargs = [k for k in kwargs.keys() if k not in defaults]
+        if invalid_kwargs:
+            name = invalid_kwargs[0]
+            raise TypeError(
+                f"Model.{method_name}() got an unexpected keyword argument '{name}'"
+            )
+
+        if args:
+            warnings.warn(
+                f"Passing positional arguments to {method_name}() is deprecated",
+                RemovedInDjango60Warning,
+                stacklevel=2,
+            )
+            total_len_args = len(args) + 1  # include self
+            max_len_args = len(defaults) + 1
+            if total_len_args > max_len_args:
+                # Recreate the proper TypeError message from Python.
+                raise TypeError(
+                    f"Model.{method_name}() takes from 1 to {max_len_args} positional "
+                    f"arguments but {total_len_args} were given"
+                )
+
+        sentinel = object()
+
+        def get_param(arg_name, arg_index):
+            param = kwargs.get(arg_name, sentinel)
+            if arg_index < len(args):
+                if param is not sentinel:
+                    # Recreate the proper TypeError message from Python.
+                    raise TypeError(
+                        f"Model.{method_name}() got multiple values for argument "
+                        f"'{arg_name}'"
+                    )
+                return args[arg_index]
+
+            return param if param is not sentinel else defaults[arg_name]
+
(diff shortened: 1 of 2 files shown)


adamchainz (commented):
Thanks for getting this started @nessita . A shared parsing method is a good idea at this point.

I’m not such a fan of calling it in all cases though - it’s adding a lot of extra code to the usual keyword-only path. If we can get back to only doing extra work if `args` is passed, that would be great. Something like:

```
def save(self, *args, force_insert=False, ...):
    if args:
        force_insert, ... = self._parse_args(args, force_insert=force_insert, ...)
```

and have `_parse_args` check `force_insert` etc. against their defaults in that case.

Also, I had the idea that we might be able to leverage `inspect.Signature.bind` to do the parsing for us. It has slightly different `TypeError` messages to Python’s actual argument parser (not sure why), but I think that’s fine.

```ipython
In [1]: def save(force_insert=False): pass

In [2]: import inspect

In [3]: inspect.signature(save)
Out[3]: <Signature (force_insert=False)>

In [4]: sig = inspect.signature(save)

In [5]: sig.bind(True)
Out[5]: <BoundArguments (force_insert=True)>

In [6]: sig.bind(force_insert=True)
Out[6]: <BoundArguments (force_insert=True)>

In [7]: sig.bind(True, force_insert=True)
...
TypeError: multiple values for argument 'force_insert'

In [8]: sig.bind(True, True)
...
TypeError: too many positional arguments
```

nessita:
> Thanks for getting this started @nessita . A shared parsing method is a good idea at this point.
> 
> I’m not such a fan of calling it in all cases though - it’s adding a lot of extra code to the usual keyword-only path. If we can get back to only doing extra work if `args` is passed, that would be great. Something like:
> 
> ```
> def save(self, *args, force_insert=False, ...):
>     if args:
>         force_insert, ... = self._parse_args(args, force_insert=force_insert, ...)
> ```
> 
> and have `_parse_args` check `force_insert` etc. against their defaults in that case.

Yeah I guess we could do something like this, but I'm not comfortable with allowing passing both the positional arg and the named arg in the same call (which, IMHO, the only way to detect this is checking against a sentinel and not against a boolean, or None).

> Also, I had the idea that we might be able to leverage `inspect.Signature.bind` to do the parsing for us.

TIL about signature, I like it. The thing I can't see (yet?) is how to construct the real given `*args` and `**kwargs` without using sentinels for the `kwargs` (because comparing the named params against their default does not guarantee that the callsite did not pass it explicitly with its default value).

Does that make sense? Am I getting too picky?

adamchainz:
Right, I wasn't thinking of calls that pass the defaults in keywords, like `Model.save(True, force_insert=False)`. You’re right the only way to detect that is to change the defaults to sentinel values.

A reason I just thought of to avoid sentinels, or generic `*args, **kwargs`, is editor autocomplete. To help users correctly fill in arguments, it’s best to keep the function signature accurate during this transition. (It would also help with typing in django-stubs.)

Since we will still warn if there are *any* positional arguments, maybe it is a bit “picky”, or at least unpragmatic, to try and detect the “passed default” case.

nessita:
> Since we will still warn if there are _any_ positional arguments, maybe it is a bit “picky”, or at least unpragmatic, to try and detect the “passed default” case.

Fair enough, thanks for convincing me :-)

I have pushed a simpler helper to be used in both `save` and `asave`, only when `args` is given. What do you think?

sarahboyce (approved):
I think this is a nice approach 👍

adamchainz:
LGTM! The only thing I would change is to name the method `_parse_save_params` so it’s clearer what it’s for, especially since `Model` has so many members. Also, this will help reduce the (already small) risk of colliding with method names in users’ subclasses.

nessita:
> LGTM! The only thing I would change is to name the method `_parse_save_params` so it’s clearer what it’s for, especially since `Model` has so many members. Also, this will help reduce the (already small) risk of colliding with method names in users’ subclasses.

Good point! I created #18319 for this rename.

PR #18314, opened 2024-06-26
Fixed #35559 -- Avoided unnecessary query on sliced union of empty queries.

charettes:
# Trac ticket number

ticket-35559

# Branch description

While refs ticket-34125 focused on the SQL correctness of slicing of union of potentially empty queries it missed an optimization opportunity to avoid performing a query at all when all queries are empty.

Thanks @Lucidiot for the report.

### DIFF
--- a/django/db/models/sql/compiler.py
+++ b/django/db/models/sql/compiler.py
@@ -583,50 +583,28 @@ class SQLCompiler:
                     raise DatabaseError(
                         "ORDER BY not allowed in subqueries of compound statements."
                     )
-        elif self.query.is_sliced and combinator == "union":
-            for compiler in compilers:
-                # A sliced union cannot have its parts elided as some of them
-                # might be sliced as well and in the event where only a single
-                # part produces a non-empty resultset it might be impossible to
-                # generate valid SQL.
-                compiler.elide_empty = False
-        parts = ()
-        selected = self.query.selected
+        parts = []
+        empty_compiler = None
         for compiler in compilers:
             try:
-                # If the columns list is limited, then all combined queries
-                # must have the same columns list. Set the selects defined on
-                # the query on all combined queries, if not already set.
-                if selected is not None and compiler.query.selected is None:
-                    compiler.query = compiler.query.clone()
-                    compiler.query.set_values(selected)
-                part_sql, part_args = compiler.as_sql(with_col_aliases=True)
-                if compiler.query.combinator:
-                    # Wrap in a subquery if wrapping in parentheses isn't
-                    # supported.
-                    if not features.supports_parentheses_in_compound:
-                        part_sql = "SELECT * FROM ({})".format(part_sql)
-                    # Add parentheses when combining with compound query if not
-                    # already added for all compound queries.
-                    elif (
-                        self.query.subquery
-                        or not features.supports_slicing_ordering_in_compound
-                    ):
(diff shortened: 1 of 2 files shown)


sarahboyce:
@charettes I plan to review this soon, do you mind resolving the recent conflicts?

django/db/models/sql/compiler.py
```diff
@@ -583,50 +583,28 @@ def get_combinator_sql(self, combinator, all):
...
             except EmptyResultSet:
                 # Omit the empty queryset with UNION and with DIFFERENCE if the
                 # first queryset is nonempty.
                 if combinator == "union" or (combinator == "difference" and parts):
+                    empty_compiler = compiler
                     continue
                 raise
         if not parts:
             raise EmptyResultSet
+        elif len(parts) == 1 and combinator == "union" and self.query.is_sliced:
```
charettes, line 600: The check could be even more rigid by only taking that branch if the single non-empty compiled query was also sliced but that didn't seem worth the complexity.

charettes (commented):
@sarahboyce done!

The approach taken here is to avoid systematically disabling the empty query eliding optimization when dealing with sliced unions (`(SELECT ... ) UNION (SELECT ...) LIMIT N`) but instead if there is a single non-empty component which would result in a query of the form `(SELECT ...) LIMIT N` where the `(SELECT ...)` could include a `LIMIT` force the compilation of an empty query without eliding so queries of the form 

```sql
(SELECT .... LIMIT N) LIMIT M
```

Which is invalid, are turned into

```sql
(SELECT .... LIMIT N) UNION (SELECT ... WHERE 1=0) LIMIT M
```

Which is valid.

Note that
1. `empty_compiler` will always be available as combined queries ensures that `len(combined_queries) >=2` and if `len(parts) == 1` that means at least one query was elided.
2. The logic is enabled even when the combined singleton is not sliced because it reduced complexity but it could be done if you judge it necessary. It seems like an okay tradeoff given we previously completely disabled all empty query eliding when the union was sliced.

django/db/models/sql/compiler.py
sarahboyce: I can remove this elif block (and the `empty_compiler` logic) without something failing (on SQLite)
>          # in the event the later is also sliced it might result in invalid
>          # SQL due to the usage of multiple LIMIT clauses.

Can we add a test that covers this scenario?

charettes: This is a Postgres specific issues as detailed in ticket-34125 which added tests in b42bd1a193a49f4e7ede207d2c40898a3171fd41.

Try running `queries.test_qs_combinators.QuerySetSetOperationTests` tests against Postgres

```
  File "/Users/charettes/Workspace/django/django/db/backends/utils.py", line 105, in _execute
  File "/Users/charettes/Workspace/django/.venv/lib/python3.12/site-packages/psycopg/cursor.py", line 97, in execute
    raise ex.with_traceback(None)
psycopg.errors.SyntaxError: multiple LIMIT clauses not allowed
LINE 1: ...s_number" WHERE "queries_number"."num" <= 0 LIMIT 1) LIMIT 1
                                                                      ^
```

sarahboyce: Got it 🥳 Sorry I missed that, thank you 👍

sarahboyce (approved):
Thank you Simon ⭐

PR #18316, opened 2024-06-26
Fixed #35564 -- Improved readability of subclass identification in admin and auth checks

jaap3:
# Trac ticket number

ticket-35564

# Branch description

`_subclass_index` and `_contains_subclass` behave the same, apart from the return value. However, because of the error handling, and the delegation thereof in `_contains_subclass` it's somewhat hard to confirm at first glance.

This rewrite utilises [`contextlib.supress`](https://docs.python.org/3/library/contextlib.html#contextlib.suppress) to make the flow of the code easier to understand.

# Checklist

### DIFF
--- a/django/contrib/admin/checks.py
+++ b/django/contrib/admin/checks.py
@@ -1,4 +1,5 @@
 import collections
+import contextlib
 from itertools import chain
 
 from django.apps import apps
@@ -21,10 +22,9 @@ def _issubclass(cls, classinfo):
     issubclass() variant that doesn't raise an exception if cls isn't a
     class.
     """
-    try:
+    with contextlib.suppress(TypeError):
         return issubclass(cls, classinfo)
-    except TypeError:
-        return False
+    return False
 
 
 def _contains_subclass(class_path, candidate_paths):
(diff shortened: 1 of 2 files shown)


sarahboyce (approved):
Thank you ⭐

ngnpope:
We've explored using `contextlib.suppress` in the past and reverted due to performance concerns in #9038. Given this is in the checks framework that @adamchainz has spent effort speeding up recently, it's probably worth checking this isn't detrimental.

adamchainz:
Yeah, I would prefer if we revert it. Repeating my profiling [from the past ticket](https://code.djangoproject.com/ticket/27818#comment:7), I still see a 10x overhead from `suppress()`:

```ipython
In [1]: import contextlib

In [2]: def foo():
   ...:    try:
   ...:       pass
   ...:    except AttributeError:
   ...:       pass
   ...:

In [3]: def foo2():
   ...:    with contextlib.suppress(AttributeError):
   ...:       pass
   ...:

In [4]: %timeit foo()
25 ns ± 0.373 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [5]: %timeit foo2()
231 ns ± 1.91 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

Yes, it’s still a small part of the overall operation, but let’s stick to the fast pattern. Python will basically never be able to optimize this away.

sarahboyce:
TIL 😁 
No problem: https://github.com/django/django/pull/18322

PR #18326, opened 2024-06-30
Corrected syntax of CSS file in staticfiles_tests.

knyghty:
# Trac ticket number
N/A

# Branch description
This fixes an issue I noticed when working on Prettier formatting for CSS. The syntax in the file is invalid. Firstly, semi-colons are used where commas should be, and the `font-face` at-rule requires `font-family` to be present.

I also had to update the test with the new hash (which will likely change again in the formatting PR).


felixxm (approved):
@knyghty Thanks :+1:

sarahboyce (approved):

PR #18330, opened 2024-07-01
Fixed #35566 -- Added get_FOO_display() support for GeneratedFields.

sarahboyce:
_get_FIELD_display() did not handle when Field.choices was defined on output_field. Thank you to Ronie Martinez for the report.

# Trac ticket number

ticket-35566

# Branch description

# Checklist

### DIFF
--- a/django/db/models/fields/generated.py
+++ b/django/db/models/fields/generated.py
@@ -30,6 +30,9 @@ class GeneratedField(Field):
         self.expression = expression
         self.output_field = output_field
         self.db_persist = db_persist
+
+        if kwargs.get("choices", None) is None and output_field.choices is not None:
+            kwargs["choices"] = output_field.choices
         super().__init__(**kwargs)
 
     @cached_property
--- a/docs/ref/models/instances.txt
+++ b/docs/ref/models/instances.txt
@@ -924,6 +924,14 @@ For example::
     >>> p.get_shirt_size_display()
     'Large'
 
+A :class:`~django.db.models.GeneratedField` with ``choices`` set on
+:attr:`.GeneratedField.output_field` also has a ``get_FOO_display()`` method.
+
+.. versionchanged:: 5.2
+
+     Support for ``choices`` defined on :attr:`.GeneratedField.output_field`
+     was added.
+
 .. method:: Model.get_next_by_FOO(**kwargs)
 .. method:: Model.get_previous_by_FOO(**kwargs)
 
--- a/tests/model_fields/models.py
+++ b/tests/model_fields/models.py
@@ -687,3 +687,27 @@ class GeneratedModelUniqueConstraintVirtual(GeneratedModelVirtualBase):
                 F("a"), name="Generated model unique constraint virtual a"
             ),
         ]
+
+
+class GeneratedModelChoices(models.Model):
+    c = models.IntegerField(choices=Whiz.CHOICES, null=True)
+    c_squared = models.GeneratedField(
+        expression=F("c") * F("c"),
+        output_field=models.IntegerField(choices=Whiz.CHOICES, null=True),
+        db_persist=True,
+    )
+
+    class Meta:
+        required_db_features = {"supports_stored_generated_columns"}
+
+
+class GeneratedModelChoicesVirtual(models.Model):
+    c = models.IntegerField(choices=Whiz.CHOICES, null=True)
+    c_squared = models.GeneratedField(
+        expression=F("c") * F("c"),
+        output_field=models.IntegerField(choices=Whiz.CHOICES, null=True),
+        db_persist=True,
+    )
+
+    class Meta:
+        required_db_features = {"supports_virtual_generated_columns"}
--- a/docs/releases/5.2.txt
+++ b/docs/releases/5.2.txt
@@ -223,6 +223,9 @@ Models
   expression contains a set-returning function, enforcing subquery evaluation.
   This is necessary for many Postgres set-returning functions.
 
+* :meth:`~django.db.models.Model.get_FOO_display` now supports when ``choices``
+  is defined on :attr:`.GeneratedField.output_field`.
+
 Requests and Responses
 ~~~~~~~~~~~~~~~~~~~~~~
 
--- a/tests/model_fields/test_generatedfield.py
+++ b/tests/model_fields/test_generatedfield.py
@@ -21,6 +21,8 @@ from .models import (
     GeneratedModel,
     GeneratedModelCheckConstraint,
     GeneratedModelCheckConstraintVirtual,
+    GeneratedModelChoices,
+    GeneratedModelChoicesVirtual,
     GeneratedModelFieldWithConverters,
     GeneratedModelNull,
     GeneratedModelNullVirtual,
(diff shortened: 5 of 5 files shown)


tests/model_fields/tests.py
```diff
@@ -297,6 +298,16 @@ def test_empty_iterator_choices(self):
         self.assertEqual(WhizIterEmpty(c="").c, "")  # Empty value
 
 
+class GetFieldDisplayGeneratedTests(TestCase):
```
nessita, line 301: I believe this test would be best placed in test_generatedfield.py?

django/db/models/base.py
```diff
@@ -1283,7 +1284,13 @@ async def adelete(self, using=None, keep_parents=False):
 
     def _get_FIELD_display(self, field):
         value = getattr(self, field.attname)
-        choices_dict = dict(make_hashable(field.flatchoices))
```
nessita, line 1286: Could we instead have a code similar to this inside `GeneratedField` model definition? something like (untested):

```diff
diff --git a/django/db/models/fields/generated.py b/django/db/models/fields/generated.py
index 5b6b188df0..ffbed99f24 100644
--- a/django/db/models/fields/generated.py
+++ b/django/db/models/fields/generated.py
@@ -30,7 +30,7 @@ class GeneratedField(Field):
         self.expression = expression
         self.output_field = output_field
         self.db_persist = db_persist
-        super().__init__(**kwargs)
+        super().__init__(choices=getattr(output_field, "choices", None), **kwargs)
 
     @cached_property
     def cached_col(self):
```

nessita (commented):
Added some (very) initial comments, LMKWYT!

docs/releases/5.0.7.txt
```diff
@@ -9,4 +9,6 @@ Django 5.0.7 fixes several bugs in 5.0.6.
 Bugfixes
 ========
 
-* ...
+* Fixed a bug when calling :meth:`~django.db.models.Model.get_FOO_display` for
+  a ``GeneratedField`` with defined ``choices`` on
+  :attr:`~django.db.models.GeneratedField.output_field` (:ticket:`35566`).
```
felixxm, line 14: This is no niche that I would treat this as a new feature (ticket suggests the same), and we don't backport new features.

tests/model_fields/tests.py
sarahboyce: Now the logic has been moved within the GeneratedField definition, I agree this is a better place for it

docs/releases/5.0.7.txt
sarahboyce: I know we don't backport new features, I wasn't sure whether my original assessment that this is a new feature and not a bug was accurate. 
So I appreciate having another opinion here, thank you - updated 👍

django/db/models/fields/generated.py
```diff
@@ -30,6 +30,9 @@ def __init__(self, *, expression, output_field, db_persist=None, **kwargs):
         self.expression = expression
         self.output_field = output_field
         self.db_persist = db_persist
+
+        if kwargs.get("choices", None) is None and output_field.choices is not None:
+            kwargs["choices"] = output_field.choices
```
felixxm, line 35: I wonder if `choices` should be an acceptable argument for `GeneratedField`s :thinking: Can we check how forms and `clean()` behave in such configurations?

docs/ref/models/instances.txt
```diff
@@ -924,6 +924,16 @@ For example::
...
     'Large'
 
+A :class:`~django.db.models.GeneratedField` with ``choices`` set on
+:attr:`GeneratedField.output_field <django.db.models.GeneratedField.output_field>`
+also has a ``get_FOO_display()`` method.
+
+.. versionchanged:: 5.2
+
+     Support for ``choices`` defined on
+     :attr:`GeneratedField.output_field <django.db.models.GeneratedField.output_field>`
```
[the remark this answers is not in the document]
sarahboyce: Made my day

django/db/models/fields/generated.py
sarahboyce: `GeneratedField`s are ignored in `full_clean()`
I added to the test to show that the `GeneratedField` can be something that isn't one of the choices (as the choices go up to 5) and played around with it as a readonly field in the admin, which looks ok to me

So I think it's ok - is there a specific concern you want me to check for?

<details><summary>Admin screenshots</summary>

image

Screenshot from 2024-09-01 16-11-57

</details>

tim-schilling, line 35: This would allow two different ways to do one thing. Ideally we should only allow one. Would it make more sense to limiting this to just `output_field.choices` as the way to set this? There are other parameters to `GeneratedField` that aren't allowed already, such as `default`. Maybe `choices` should be disallowed too.

However, if we do allow multiple sources for choices, then should we include that in the documentation? Otherwise the developer would have to read the code to understand how one of these are being selected.

If multiple sources are provided, we may want to use the `NOT_PROVIDED` pattern instead of relying on `None`. This prevents fewer misunderstandings in what the developer wants to happen versus what actually happens.

tim-schilling (changes requested):
@raffaellasuardini and I took a look at this and think that we should consider limiting this to having only one way to do things. But if we do end up with supporting both, we should consider using the `NOT_PROVIDED` sentinel pattern when deciding the value of `self.choices` to avoid a weird overriding case.

tests/model_fields/models.py
```diff
@@ -687,3 +687,27 @@ class Meta:
...
         ]
+
+
+class GeneratedModelChoices(models.Model):
+    c = models.IntegerField(choices=Whiz.CHOICES, null=True)
+    c_squared = models.GeneratedField(
+        expression=F("c") * F("c"),
+        output_field=models.IntegerField(choices=Whiz.CHOICES, null=True),
+        db_persist=True,
+    )
```
raffaellasuardini, line 698: If we support multiple sources of choices, we should test the cases of GeneratedField(choices=FOO, output_field=Field(choices=BAR)) where the choices are different so there's a test confirming which is taking precedence.

tests/model_fields/models.py
```diff
@@ -687,3 +687,27 @@ class Meta:
...
+    class Meta:
+        required_db_features = {"supports_stored_generated_columns"}
+
+
+class GeneratedModelChoicesVirtual(models.Model):
+    c = models.IntegerField(choices=Whiz.CHOICES, null=True)
+    c_squared = models.GeneratedField(
+        expression=F("c") * F("c"),
+        output_field=models.IntegerField(choices=Whiz.CHOICES, null=True),
+        db_persist=True,
```
tim-schilling, line 709: Sorry, I'm not super familiar with `GeneratedField` and tests, but did you intend to have this as the following because `required_db_features = {"supports_virtual_generated_columns"}` is set?
```suggestion
        db_persist=False,
```

sarahboyce:
Thank you @raffaellasuardini and @tim-schilling :star: 
When triaging the ticket, I thought `choices` was not allowed on a `GeneratedField` and defining this on `output_field` was the only way to define it - but that's not the case! Thank you - I have closed the ticket with a comment, closing this too

PR #18331, opened 2024-07-01
Fixed #35571 -- Store template_names as string

guserav:
# Trac ticket number
<!-- Replace XXXXX with the corresponding Trac ticket number, or delete the line and write "N/A" if this is a trivial PR. -->

ticket-35571

# Branch description
When passing pathlib.Path or pathlib.PosixPath as template names (because templates are in subdirectories, so pathlib seems reasonable to use) the assertTemplateUsed breaks as it performs a str.join on all template names without sanitizing them.

```
File "<stuff>/.local/lib/python3.12/site-packages/django/test/testcases.py", line 681, in _assert_template_used                                                                                                                      
    % (template_name, ", ".join(template_names)),                                                                                                                                                                                              
                      ^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                
TypeError: sequence item 0: expected str instance, PosixPath found
```

This causes a lot of confusion as it is quite hard to track down the Template that introduced the PosixPath.

This is one proposed solution for fixing this. I.e. trying to sanitize all template names when they are passed into the template engine, as this seems easier then trying to find all cases where the stringiness of the template name is assumed. Another approach would be to have an assert that the template name is a string, but this could possible break code that doesn't utilize the test framework.

# Checklist
- [x] This PR targets the `main` branch. <!-- Backports will be evaluated and done by mergers, when necessary. -->

### DIFF
--- a/django/template/engine.py
+++ b/django/template/engine.py
@@ -156,11 +156,11 @@ class Engine:
         tried = []
         for loader in self.template_loaders:
             try:
-                template = loader.get_template(name, skip=skip)
+                template = loader.get_template(str(name), skip=skip)
                 return template, template.origin
             except TemplateDoesNotExist as e:
                 tried.extend(e.tried)
-        raise TemplateDoesNotExist(name, tried=tried)
+        raise TemplateDoesNotExist(str(name), tried=tried)
 
     def from_string(self, template_code):
         """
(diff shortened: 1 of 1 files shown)


nessita:
Hello @guserav, thank you for your first contribution! :star2: 

As I mentioned in the ticket, we would need to first fully understand the report by having a minimal reproducer or a Django test case, for then to be able to accept the ticket and thus properly reviewing this PR.
With the information at hand, I have to close this since the proposed solution feels like it could impact performance for no clear reason/gain. This would also need tests and some extra docs changes, but first we need to understand the actual problem to decide on the right fix.

PR #18332, opened 2024-07-02
Fixed #13376: Added kwargs argument to messages API to allow extension of functionality to the Storage & Message class.

nanorepublica:
# Trac ticket number

ticket-13376

# Branch description

This branch is a middle road for the mentioned Trac ticket. While not implementing the full requirements of the ticket it modifies the existing API in a backwards compatiable manner to either allow the ticket to be implemented in core or as a third-party package.
This also opens up the messages API to extensibility without further changes to core. 

# Checklist

### DIFF
--- a/django/contrib/messages/api.py
+++ b/django/contrib/messages/api.py
@@ -19,7 +19,7 @@ class MessageFailure(Exception):
     pass
 
 
-def add_message(request, level, message, extra_tags="", fail_silently=False):
+def add_message(request, level, message, extra_tags="", fail_silently=False, **kwargs):
     """
     Attempt to add a message to the request using the 'messages' app.
     """
@@ -37,7 +37,7 @@ def add_message(request, level, message, extra_tags="", fail_silently=False):
                 "django.contrib.messages.middleware.MessageMiddleware"
             )
     else:
-        return messages.add(level, message, extra_tags)
+        return messages.add(level, message, extra_tags, **kwargs)
 
 
 def get_messages(request):
@@ -72,7 +72,7 @@ def set_level(request, level):
     return True
 
 
-def debug(request, message, extra_tags="", fail_silently=False):
+def debug(request, message, extra_tags="", fail_silently=False, **kwargs):
     """Add a message with the ``DEBUG`` level."""
     add_message(
         request,
@@ -80,10 +80,11 @@ def debug(request, message, extra_tags="", fail_silently=False):
         message,
         extra_tags=extra_tags,
         fail_silently=fail_silently,
+        **kwargs,
     )
 
 
-def info(request, message, extra_tags="", fail_silently=False):
+def info(request, message, extra_tags="", fail_silently=False, **kwargs):
     """Add a message with the ``INFO`` level."""
     add_message(
         request,
@@ -91,10 +92,11 @@ def info(request, message, extra_tags="", fail_silently=False):
         message,
         extra_tags=extra_tags,
         fail_silently=fail_silently,
+        **kwargs,
     )
 
 
-def success(request, message, extra_tags="", fail_silently=False):
+def success(request, message, extra_tags="", fail_silently=False, **kwargs):
     """Add a message with the ``SUCCESS`` level."""
     add_message(
         request,
@@ -102,10 +104,11 @@ def success(request, message, extra_tags="", fail_silently=False):
         message,
         extra_tags=extra_tags,
         fail_silently=fail_silently,
+        **kwargs,
     )
 
 
-def warning(request, message, extra_tags="", fail_silently=False):
+def warning(request, message, extra_tags="", fail_silently=False, **kwargs):
     """Add a message with the ``WARNING`` level."""
     add_message(
         request,
@@ -113,10 +116,11 @@ def warning(request, message, extra_tags="", fail_silently=False):
         message,
         extra_tags=extra_tags,
         fail_silently=fail_silently,
+        **kwargs,
     )
 
 
-def error(request, message, extra_tags="", fail_silently=False):
+def error(request, message, extra_tags="", fail_silently=False, **kwargs):
     """Add a message with the ``ERROR`` level."""
     add_message(
         request,
@@ -124,4 +128,5 @@ def error(request, message, extra_tags="", fail_silently=False):
         message,
         extra_tags=extra_tags,
         fail_silently=fail_silently,
+        **kwargs,
     )
--- a/django/contrib/messages/api.py
+++ b/django/contrib/messages/api.py
@@ -19,7 +19,9 @@ class MessageFailure(Exception):
     pass
 
 
-def add_message(request, level, message, extra_tags="", fail_silently=False, **kwargs):
+def add_message(
+    request, level, message, extra_tags="", fail_silently=False, extra_kwargs=None
+):
     """
     Attempt to add a message to the request using the 'messages' app.
     """
@@ -37,7 +39,7 @@ def add_message(request, level, message, extra_tags="", fail_silently=False, **k
                 "django.contrib.messages.middleware.MessageMiddleware"
             )
     else:
-        return messages.add(level, message, extra_tags, **kwargs)
+        return messages.add(level, message, extra_tags, extra_kwargs=extra_kwargs)
 
 
 def get_messages(request):
@@ -72,7 +74,7 @@ def set_level(request, level):
     return True
 
 
-def debug(request, message, extra_tags="", fail_silently=False, **kwargs):
+def debug(request, message, extra_tags="", fail_silently=False, extra_kwargs=None):
     """Add a message with the ``DEBUG`` level."""
     add_message(
         request,
@@ -80,11 +82,11 @@ def debug(request, message, extra_tags="", fail_silently=False, **kwargs):
         message,
         extra_tags=extra_tags,
         fail_silently=fail_silently,
-        **kwargs,
+        extra_kwargs=extra_kwargs,
     )
 
 
-def info(request, message, extra_tags="", fail_silently=False, **kwargs):
+def info(request, message, extra_tags="", fail_silently=False, extra_kwargs=None):
     """Add a message with the ``INFO`` level."""
     add_message(
         request,
@@ -92,11 +94,11 @@ def info(request, message, extra_tags="", fail_silently=False, **kwargs):
         message,
         extra_tags=extra_tags,
         fail_silently=fail_silently,
-        **kwargs,
+        extra_kwargs=extra_kwargs,
     )
 
 
-def success(request, message, extra_tags="", fail_silently=False, **kwargs):
+def success(request, message, extra_tags="", fail_silently=False, extra_kwargs=None):
     """Add a message with the ``SUCCESS`` level."""
     add_message(
         request,
@@ -104,11 +106,11 @@ def success(request, message, extra_tags="", fail_silently=False, **kwargs):
         message,
         extra_tags=extra_tags,
         fail_silently=fail_silently,
-        **kwargs,
+        extra_kwargs=extra_kwargs,
     )
 
 
-def warning(request, message, extra_tags="", fail_silently=False, **kwargs):
+def warning(request, message, extra_tags="", fail_silently=False, extra_kwargs=None):
     """Add a message with the ``WARNING`` level."""
     add_message(
         request,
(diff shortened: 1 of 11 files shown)


docs/ref/contrib/messages.txt
```diff
@@ -258,6 +258,11 @@ The ``Message`` class
       the lowercase version of the name of the associated constant, but this
       can be changed if you need by using the :setting:`MESSAGE_TAGS` setting.
 
+    * ``extra_kwargs``: This variable accepts all further keyword arguments
```
nanorepublica, line 261: I'm not entirely happy with this, but believe it should be documented. Perhaps as a new section titled something like: `Custom Storage & Message classes`

nessita:
Hi @nanorepublica! Thank you for keeping up with contributing! :star2: 

I see that you have, intentionally perhaps, not checked the "has patch" flag in the ticket. I just wanted to be explicit saying that in order for this PR to be listed as "needing review" in the [Django Development Dashboard](https://code.djangoproject.com/query?status=!closed&needs_better_patch=0&needs_tests=0&needs_docs=0&has_patch=1&stage=Accepted&desc=1&order=changetime), the flag needs to be set when this is ready for review. No rush though!

nanorepublica:
@nessita Thanks for this! I will be checking that flag now!

django/contrib/messages/api.py
```diff
@@ -19,7 +19,7 @@ class MessageFailure(Exception):
     pass
 
 
-def add_message(request, level, message, extra_tags="", fail_silently=False):
+def add_message(request, level, message, extra_tags="", fail_silently=False, **kwargs):
```
sarahboyce, line 22: [add_message](https://docs.djangoproject.com/en/5.0/ref/contrib/messages/#django.contrib.messages.add_message) is documented and we will need to update the signature here.
You will need to add a `.. versionchanged:: 5.2` note about the signature change and a release note in 5.2
You can also want to update the `add_message` docs to be more specific about the signatures of `debug`, `info` etc

django/contrib/messages/storage/base.py
```diff
@@ -12,10 +12,11 @@ class Message:
     or template.
     """
 
-    def __init__(self, level, message, extra_tags=None):
+    def __init__(self, level, message, extra_tags=None, **kwargs):
         self.level = int(level)
         self.message = message
         self.extra_tags = extra_tags
+        self.extra_kwargs = kwargs
```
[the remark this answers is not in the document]
sarahboyce: `ResolverMatch` has [extra_kwargs](https://docs.djangoproject.com/en/5.0/ref/urlresolvers/#django.urls.ResolverMatch.extra_kwargs) which might inspire you
You might prefer `extra_kwargs=None` in the signature than **kwargs (not sure)

I see you have added some docs (which is great), this needs a test

docs/ref/contrib/messages.txt
sarahboyce: Also needs a `..versionchanged` note and reference in the release notes

Good point, we have a bit on BaseStorage and how you theoretically can make a custom storage (https://docs.djangoproject.com/en/5.0/ref/contrib/messages/#django.contrib.messages.storage.base.BaseStorage) adding something in a section there demonstrating the value of this would really help

If you make a really simple example, that might also be useful for writing a test

sarahboyce (commented):
Thank you for this @nanorepublica - I have initial comments 👍

django/contrib/messages/storage/base.py
```diff
@@ -125,6 +148,12 @@ def _prepare_messages(self, messages):
         for message in messages:
             message._prepare()
 
+    def get_message_class(self):
+        return self.message_class
+
+    def get_message(self, *args, **kwargs):
```
anorthall: The function name feels slightly misleading to me on a class which is a storage wrapper - it implies to me that an existing message is being fetched from the storage rather than a new message created (for later storage).

nanorepublica: Yep, your right. This represents a possibly refactoring to allow custom Message classes, but in order to do that we would need to use these classes throughout the application which isn't the case right now. I'll get this removed for this PR.

django/contrib/messages/storage/cookie.py
```diff
@@ -36,6 +38,9 @@ def process_messages(self, obj):
             if obj[0] == MessageEncoder.message_key:
                 if obj[1]:
                     obj[3] = mark_safe(obj[3])
+                # this conditional looks for extra_kwargs
+                if isinstance(obj[-1], str) and len(obj) > 5:
+                    obj[-1] = json.loads(obj[-1])
                 return Message(*obj[2:])
```
blighj, line 48: If we rely on the list length as a test for extra_kwargs, we'll hit a problem if there are no extra_tags, they won't be added to the list, so it will only have 5 elements, the 4 main ones and the new kwargs, but the code will assume there are no kwargs, in fact I assume it will see the kwargs as extra_tags. Perhaps we can always add extra_tags and or extra_kwargs to the list and use None as the way to check? The cost is adding two null's to the json of each cookie payload, which is small but I see the boolean is_safedata was already converted to 0|1 to save space so maybe it's a concern.

Here is a test that would fail currently if added to test_cookie.py>CookieTests
```py
    def test_extra_kwargs_without_extra_tags(self):
        msg = self.encode_decode("message", extra_kwargs={"permanent": True})
        self.assertIsNone(msg.extra_tags)
        self.assertEqual(msg.extra_kwargs, {"permanent": True})
```

tests/messages_tests/tests.py
```diff
@@ -156,6 +156,23 @@ def test_with_tags(self):
...
+            extra_kwargs={"permanent": True},
+        )
+        self.assertMessages(
+            response,
+            [
+                Message(
+                    constants.INFO, "INFO message.", extra_kwargs={"permanent": True}
+                ),
+            ],
+        )
```
blighj, line 175: I think this is going to fall through to `Message.__eq__`, which isn't aware of the extra_kwargs, so you aren't really testing if the extra_kwargs is there. I confirmed this by changing it locally to `{"permanent": False}` and it still passed, it might be worth adding an explicit check
matbe add something like
```py
    msgs = list(response.wsgi_request._messages)
    self.assertEqual(msgs[0].extra_kwargs, {"permanent": True})
```

django/contrib/messages/storage/base.py
```diff
@@ -65,6 +76,16 @@ def __len__(self):
...
+        """
+        Iterate over all messages to attempt to display them.
+
+        Any new messages (via _queued_messages) are added to any previously
+        stored messages (via _loaded_messaged). All messages are then returned
+        to be iterated over externally (typically via a template).
+
+        This method can be altered by a subclass to alter the logic of when
+        a message would be cleared from storage.
+        """
```
blighj: `_loaded_messaged` i think this should be `_loaded_messages`

But maybe you only need the last paragraph in any case, the first two describe the code which is fairly clear code so possibly doesn't need the description.

blighj (commented):
I've had a first look at this, and found things I think are issues, but maybe I'm misunderstanding things.

The other thing I don't understand is the need to json.dumps the extra_kwargs in base.py? It seems it would work fine without that.

nanorepublica:
Thanks for the review. I'll dig deeper into the comments later this week. 

IIRC the reason for the json.loads and dumps is to allow extra_kwargs to use python primitives (ie bools, lists, kwargs) and then to be stored and retrieved correctly from the message storage.

blighj:
> Thanks for the review. I'll dig deeper into the comments later this week.
> 
> IIRC the reason for the json.loads and dumps is to allow extra_kwargs to use python primitives (ie bools, lists, kwargs) and then to be stored and retrieved correctly from the message storage.

Does MessageEncoder not already cover that, since it is a subclass of json.JSONEncoder? I may be missing something, hopefully I'm not waisting your time with these questions.

django/contrib/messages/storage/base.py
sarahboyce: I would also remove this. We don't make any updates to `__iter__` here and adding more documentation to the internal functions feels out of scope to this ticket

docs/ref/contrib/messages.txt
```diff
@@ -91,6 +91,49 @@ To write your own storage class, subclass the ``BaseStorage`` class in
...
 ``_store`` methods.
 
+.. _custom-storage-class:
+
+Custom Storage Classes
+----------------------
+
+Here is an example of a custom storage class which uses ``extra_kwargs``
+passed to the message to specify whether a message should be permanent and
+therefore never expire after a single request::
```
sarahboyce:
```suggestion
.. _custom-storage-backends:

Custom storage backends
~~~~~~~~~~~~~~~~~~~~~~~

.. versionadded:: 6.1

You can use ``extra_kwargs`` to extend the behavior of existing storage
backends. In this example, ``SessionStorage`` is extended to allow specifying
whether a message should be permanent, and therefore never expire after a
single request::
```
Maybe

sarahboyce (commented):
Thank you for the updates @nanorepublica and the review @blighj ⭐ 
I also think `ModelAdmin.message_user` may need updating to accept the new kwarg

django/contrib/messages/storage/base.py
nanorepublica: In a later refactor this may change to add a public API for iterating over messages, but something for a later ticket/PR.

nanorepublica:
> Does MessageEncoder not already cover that, since it is a subclass of json.JSONEncoder? I may be missing something, hopefully I'm not waisting your time with these questions.

Currently MessageEncoder is limited to concrete implementations of Storage classes, where as the `json` usage in base.py is purely for the Message itself when updating the list of stored messages. 

We could perhaps MessageEncoder in time, I can see some refactor's for this app over time, of which this might be one.

django/contrib/messages/storage/cookie.py
```diff
@@ -22,6 +22,10 @@ def default(self, obj):
             message = [self.message_key, is_safedata, obj.level, obj.message]
             if obj.extra_tags is not None:
                 message.append(obj.extra_tags)
+            else:
+                message.append(None)
```
blighj: Minor point.
Could this just be `message.append(obj.extra_tags)` without the need for if/else

blighj (approved):
Thanks for addressing those changes I asked for, I've run the tests locally and everything looks good. One minor comment from me, but doesn't stop me approving as is. I still think I'd design it with the json handling at the MessageEncoder level, but not a strong opinion.

codingjoe (commented):
Hi @nanorepublica,

as discussed on Discord:

1. The tags vs. kwargs redundancy needs to be addressed. Either in deprecation or documentation.
2. I would suggest using `**extra_kwargs` not `extra_kwargs=None`. This would make `extra_kwargs` type consistent and remove the `is None` type checks.
3. Django's task framework introduced `django.utils.json.normalize_json`, which might be helpful for my second comment.

Let me know, when I should have another look.

Cheers!
Joe

sarahboyce:
> The tags vs. kwargs redundancy needs to be addressed. Either in deprecation or documentation.

This is a good point I hadn't appreciated. I think a deprecation to replace `extra_tags` with `extra_kwargs` would be my preference

codingjoe:
> > The tags vs. kwargs redundancy needs to be addressed. Either in deprecation or documentation.
> 
> This is a good point I hadn't appreciated. I think a deprecation to replace `extra_tags` with `extra_kwargs` would be my preference

I would second that. IMHO the `extra_tags` have always been somewhat awkward. And JSON-serializable kwargs are the superior replacement.

nanorepublica:
I have been pondering the tags/kwargs duplication. Kwargs is definitely a super set in terms of possibly data to pass to the framework. However extra_tags are combined with level tags (which involves the MESSAGE_TAGS setting for customisation of them.)

I think the tags property ought to remain, but the deprecation of extra_tags into extra_kwargs seems reasonable and we can mirror the functionality by looking for a key called extra_kwargs inside the extra_kwargs

Whether all that happens in the PR? I don't know. To note I have proposed a new feature issue to further improve the messages API here: https://github.com/django/new-features/issues/155 which this deprecation can definitely be included if not completed on the PR. 

As to the other 2 points, I will look into them.

codingjoe:
@nanorepublica if you intend to refactor the message API anyway, maybe let's hold new features unless they are part of the refactoring strategy. I'd focus my efforts on other PRs for 6.1; is that OK with you?

nanorepublica:
@codingjoe, This PR is the first step in the refactoring effort (it's the ticket/PR that opened my eyes to the need for it)

My point was questioning whether the extra tags depreciation ought to be handled as a separate ticket or bundled into this ticket as it's only tangential related to what this PR is trying to achieve.

PR #18333, opened 2024-07-02
Added contributor guidelines for performance optimizations.

sarahboyce:
# Branch description

This is following from a discussion within https://github.com/django/django/pull/18322

- Probably want to mention "trade-offs" and "discretion"
- I wouldn't describe myself as a performance person, if we want to give specific performance benchmarking recommendations beyond django-asv, would appreciate help in wording this
- Unsure if I should separate the advice between the performance regression and "suggested performance optimization" (a performance regression could perhaps be treated as a "bug" for example)
- Might make sense to say that test coverage of the code being optimized is a pre-requisite to accepting a performance optimization 🤔

### DIFF
--- a/docs/internals/contributing/bugs-and-features.txt
+++ b/docs/internals/contributing/bugs-and-features.txt
@@ -117,6 +117,15 @@ branch, and show us your work!
 
 See also: :ref:`documenting-new-features`.
 
+Requesting performance optimizations
+====================================
+
+Reports of a performance regression, or suggested performance optimizations,
+should provide benchmarks and commands for the ticket triager to reproduce.
+
+See the :ref:`django-asv-benchmarks` for more details of Django's existing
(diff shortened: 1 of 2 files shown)


smithdc1:
Hi Sarah -- I think this makes sense and is just the right balance. 

I was thinking originally that `django-asv` is just one way and that there are less onerous approaches that we could point to. However, there are so many that we couldn't document them so the generic "show us your steps" guidance seems the best approach.

sarahboyce:
> Hi Sarah -- I think this makes sense and is just the right balance.

In that case, marking it ready for review 👍

claudep (approved):
Thanks, looks good!

hramezani (approved):

sarahboyce:
Thank you all ⭐

PR #18335, opened 2024-07-02
Fixed #29522 -- Refactored the Deserializer functions to classes.

AMK9978:
# Trac ticket number

ticket-29522

# Branch description
Updated the [previous PR](https://github.com/django/django/pull/11218).
I applied the latest comments and also added a test to show that a developer can extend the Deserializer class and override its methods, although there is not a lot to be overridden. Extensibility is its main point.

# Checklist


docs/topics/serialization.txt
```diff
@@ -148,6 +148,12 @@ argument is passed in as ``True``::
 
     serializers.deserialize("xml", data, ignorenonexistent=True)
 
+In addition, you can extend the ``Deserializer`` class and extend it::
```
AMK9978, line 151: I know this may be lame but waiting for your comments. The same for the release note.

django/core/serializers/python.py
```diff
@@ -96,100 +96,120 @@ def getvalue(self):
         return self.objects
 
 
-def Deserializer(
-    object_list, *, using=DEFAULT_DB_ALIAS, ignorenonexistent=False, **options
-):
+class Deserializer:
```
sarahboyce, line 99: We already have `django.core.serializers.base.Deserializer` which the xml deserializer inherits
Then `json`, `jsonl` and `pyaml` each have `Deserializer` functions still, it would be great if this could be adapted so that these are all classes that inherit from `python.Deserializer` and the python `Deserializer` inherits from `base.Deserializer`
The current state is a bit messy which I think makes it harder to document :thinking:

django/core/serializers/python.py
```diff
@@ -96,100 +96,120 @@ def getvalue(self):
...
-    except (LookupError, TypeError):
-        raise base.DeserializationError(
-            "Invalid model identifier: '%s'" % model_identifier
+    def _handle_foreign_key_field(self, field, field_value):
+        return base.deserialize_fk_value(
+            field, field_value, self.using, self.handle_forward_references
         )
+
+    @staticmethod
+    def _get_model(model_identifier):
```
sarahboyce, line 208: If we look at `xml.Deserializer` this has `_get_model_from_node`, `_handle_m2m_field_node`, `_handle_fk_field_node` and `__next__` instead of `__iter__` is defined - it would be good to align this

sarahboyce (commented):
Thank you for picking this up and being patient @AMK9978  :star: 
I have a few thoughts on the design here :+1:

django/core/serializers/python.py
AMK9978: Fair enough. So, like what happens in the `xml_serializer`, creating the object list in `json`, `jsonl`, and `pyaml` which are now created using their `Deserializer` helper functions in `django/core/serializers/json.py`, `django/core/serializers/jsonl.py`, and `django/core/serializers/pyaml.py`, respectively, should become a method in the `Deserializer` class (the equivalent of `pulldom.parse(self.stream, self._make_parser())` in the `xml_deserializer`). For each three of them, there must be a class to implement that object creation🤔

AMK9978:
@sarahboyce 
I tried to make the deserializer classes look more similar with the least changes and still readable code. Waiting for your comments!

django/core/serializers/python.py
```diff
@@ -96,45 +96,64 @@ def getvalue(self):
...
+        self._iterator = None
+
+    def __iter__(self):
+        for obj in self.stream:
+            yield from self._handle_object(obj)
+
+    def __next__(self):
+        if self._iterator is None:
+            self._iterator = iter(self)
+        return next(self._iterator)
```
sarahboyce, line 124: This is untested/missing coverage

sarahboyce (commented):
Thank you for the updates @AMK9978 :star: this is shaping up
I have a few more suggestions/comments

AMK9978:
@sarahboyce Thanks for your comments. While I was keeping the classes alike, your simplification removed boilerplates. Loved it!
Btw, your suggestions needed some tweaks. For example, possible errors from `_handle_object` should be handled by the subclasses as they used to and tests require that. Plus, I added some tests for the functionality of the deserializer in Django.
Waiting for the next round of your comments :) (or approval!).

django/core/serializers/jsonl.py
```diff
@@ -39,19 +39,30 @@ def getvalue(self):
...
+
+    def _handle_object(self, obj):
         try:
-            yield from PythonDeserializer([json.loads(line)], **options)
+            yield from super()._handle_object(obj)
         except (GeneratorExit, DeserializationError):
             raise
         except Exception as exc:
-            raise DeserializationError() from exc
+            raise DeserializationError(f"Error deserializing object: {exc}") from exc
```
sarahboyce, line 58: I can remove this `_handle_object` without any test failures

django/core/serializers/jsonl.py
```diff
@@ -39,19 +39,30 @@ def getvalue(self):
...
+
+    @staticmethod
+    def _get_lines(stream):
+        for line in stream:
+            if not line.strip():
+                continue
```
sarahboyce, line 64: This appears not to be covered anymore, maybe due to the change to `.splitlines()` :thinking:

django/core/serializers/pyyaml.py
```diff
@@ -66,17 +65,23 @@ def getvalue(self):
...
+            raise DeserializationError() from exc
+        super().__init__(objects, **options)
+
+    def _handle_object(self, obj):
+        try:
+            yield from super()._handle_object(obj)
+        except (GeneratorExit, DeserializationError):
+            raise
+        except Exception as exc:
+            raise DeserializationError(f"Error deserializing object: {exc}") from exc
```
sarahboyce, line 87: I can remove this without any test failures

django/core/serializers/jsonl.py
```diff
@@ -39,19 +39,30 @@ def getvalue(self):
...
-        stream_or_string = stream_or_string.split("\n")
-
-    for line in stream_or_string:
-        if not line.strip():
-            continue
+
+    def __init__(self, stream_or_string, **options):
+        if isinstance(stream_or_string, bytes):
+            stream_or_string = stream_or_string.decode()
+        if isinstance(stream_or_string, str):
```
sarahboyce, line 48:
```suggestion
        if isinstance(stream_or_string, (bytes, str)):
```
This was also ran for bytes previously, was this a conscious change? Might be worth adding a test for bytes input (there isn't one in the existing test suite) to check everything is ok or this makes sense
The pyyaml also is missing coverage for bytes input
Both of these are existing coverage issues but it would be nice to resolve them considering we refactoring this area

docs/topics/serialization.txt
sarahboyce: I think maybe we should mention how folks can add their own serialization format, so for example we should mention the `SERIALIZATION_MODULES` setting and how they can extend extisting `Deserializer`s

I think this should be within the `Serialization formats` section, maybe at the bottom after `YAML`, as `Custom  serialization formats` (or something). This can have a `.. _custom-serialization-formats:` reference link we can use for the release notes.

It will also need a `.. versionchanged: 5.2` note

sarahboyce (commented):
Thank you for the updates @AMK9978 :star:
I think this is shaping up nicely :1st_place_medal:

django/core/serializers/jsonl.py
AMK9978: Because of insufficient tests. For example, if you remove `_handle_object` in the json.py and run `python runtests.py fixtures_regress`, you'll get ` Animal has no field named 'extra_name'`. I will add corresponding tests for these scenartios for `jsonl.py` and `pyyaml.py`.

django/core/serializers/jsonl.py
AMK9978: I didn't quite get what you meant because that check for bytes was unnecessary. If there's a `bytes` object, it gets converted into `str` in its above `if`. Anyway, I have added tests for bytes input and the code works well.

django/core/serializers/jsonl.py
AMK9978: I don't think changing `split('\n')` to `splintlines` changed anything else but I added a test for a jsonl input with empty lines.

docs/topics/serialization.txt
sarahboyce: I have tried to rewrite the docs a bit, can you review the updates? Would be good to get more eyes on them for sure

sarahboyce (commented):
Thank you for the tests @AMK9978 :star: this looks really good 

I have squashed and tweaked slightly, I would like another pair of eyes on the docs and then I think we're good to go :+1:

docs/topics/serialization.txt
```diff
@@ -347,6 +347,37 @@ again a mapping with the key being name of the field and the value the value:
...
+First, define a ``Serializer`` and a ``Deserializer`` class. These can override
+existing serialization format classes::
+
+    from django.core import serializers
+
+
+    class Serializer(serializers.python.Serializer): ...
+
+
+    class Deserializer(serializers.python.Deserializer): ...
```
AMK9978, line 366: @sarahboyce 
Awesome job! 
IMHO, I don't think custom deserialization quite fits into the  `Custom serialization formats`. Plus, adding a custom serializer to the `SERIALIZATION_MODULES` is a convenient way that only applies to serializers but not such a thing for a custom deserializer. So it doesn't seem cohesive to me.

sarahboyce: I think this is how folks are supposed to do it? `get_deserializer` loads the serializers defined by this setting, and then gets the `Deserializer` from it. So this is how you define both a custom serializer and deserializer

pauloxnet:
How Django performances change with this change?

docs/topics/serialization.txt
AMK9978: Yes, you're right about the usage and how `get_deserializer` returns and I was confused. I wanted to suggest a direct mention of custom deserialization in the section title. Plus, regarding 
`First, define a ``Serializer`` and a ``Deserializer`` class. `
having a custom Deserialization doesn't raise any error in the custom module unless deserialization for that custom format gets called. Maybe it indicates it's necessary (Not a big deal!).

sarahboyce: > having a custom Deserialization doesn't raise any error in the custom module unless deserialization for that custom format gets called. Maybe it indicates it's necessary (Not a big deal!).

I'm not quite sure what the suggestion is, do you think this needs a docs warning? Feel free to push up updates :+1:

AMK9978: Got it. I'll wait for other reviewers' comments. If the docs get no comment, it means it's clear enough.

sarahboyce:
> How Django performances change with this PR?

- Ran the benchmarks and the output is "no significant change" ( see [here](https://github.com/django/django/actions/runs/10848956962/))
- Ran some tests on main vs this branch and don't see a significant impact to performance

Is there a specific area that concerns you @pauloxnet ?

Whilst we have you, can you check if you're happy with the docs?

docs/topics/serialization.txt
```diff
@@ -347,6 +347,37 @@ again a mapping with the key being name of the field and the value the value:
 
 Referential fields are again represented by the PK or sequence of PKs.
 
+.. _custom-serialization-formats:
+
+Custom serialization formats
+----------------------------
+
+In addition to the above formats, you can create your own serialization format.
```
pauloxnet, line 355:
```suggestion
In addition to the default formats, you can create a custom serialization 
format.
```
I tried to rephrase, but feel free to discard it.

docs/topics/serialization.txt
pauloxnet: I think we should put an example of Serializer and Deserializer with minimal but working code, and not empty.

docs/topics/serialization.txt
```diff
@@ -347,6 +347,37 @@ again a mapping with the key being name of the field and the value the value:
...
+    class Serializer(serializers.python.Serializer): ...
+
+
+    class Deserializer(serializers.python.Deserializer): ...
+
+Then add the module containing the serializer definitions to your
+:setting:`SERIALIZATION_MODULES` setting::
+
+    SERIALIZATION_MODULES = {
+        "custom": "path.to.custom_serializer",
```
pauloxnet, line 372: This path here might be confusing if it is not defined above what it is. Maybe we could define above that the code goes in a python file with a certain name and then call it here to help understand that this is the path to the python file (or module) that contains the serializers.

pauloxnet:
> > How Django performances change with this PR?
> 
>     * Ran the benchmarks and the output is "no significant change" ( see [here](https://github.com/django/django/actions/runs/10848956962/))
> 
>     * Ran some tests on main vs this branch and don't see a significant impact to performance
> 
> 
> Is there a specific area that concerns you @pauloxnet ?

nothing in particular, I just thought it was a good idea to verify that switching from functions to classes didn't degrade performance.

> Whilst we have you, can you check if you're happy with the docs?

Overall it all seems ok to me, I just tried to give some suggestions, but feel free to discard them if they don't make sense to you.

docs/topics/serialization.txt
sarahboyce: Yes! @AMK9978 something like
```
.. code-block:: python
    :caption: ``path/to/custom_serializer.py``
```
should work

docs/topics/serialization.txt
sarahboyce: Yeah I agree, a more realistic example would be beneficial. @AMK9978 do you mind pulling something together?

docs/topics/serialization.txt
pauloxnet: Yes, exactly.

docs/topics/serialization.txt
AMK9978: @pauloxnet @sarahboyce 
I added a minimal functioning csv serializer and deserializer. Although due to the flat nature of csv, it requires a bit more code, I think it is quite useful as an example.

docs/topics/serialization.txt
```diff
@@ -347,6 +347,101 @@ again a mapping with the key being name of the field and the value the value:
...
+            header = ",".join(fields)
+            self.content.append(f"model,{header}\n")
+
+        def end_object(self, obj):
+            dumped_object_str = self.get_dump_object(obj)
+            self.content.append(f"{dumped_object_str}\n")
+
+        def end_serialization(self):
+            for line in self.content:
+                self.stream.write(line)
```
sarahboyce, line 407: I haven't tested but I am hoping we can do something like...
```suggestion
    class Serializer(serializers.python.Serializer):
        def get_dump_object(self, obj):
            dumped_object = super().get_dump_object(obj)
            row = [dumped_object["model"], str(dumped_object["pk"])]
            row += [str(value) for value in dumped_object["fields"].values()]
            return ",".join(row), dumped_object["model"]

        def end_object(self, obj):
            dumped_object_str, model = self.get_dump_object(obj)
            if self.first:
                fields = [field.name for field in apps.get_model(model)._meta.fields]
                header = ",".join(fields)
                self.stream.write(f"model,{header}\n")
            self.stream.write(f"{dumped_object_str}\n")

        def getvalue(self):
            return super(serializers.python.Serializer, self).getvalue()
```

docs/topics/serialization.txt
```diff
@@ -347,6 +347,101 @@ again a mapping with the key being name of the field and the value the value:
...
+                        model_name = row.get("model")
+                        model = apps.get_model(model_name)
+                    field_data = {
+                        field.name: row.get(field.name, "") for field in model._meta.fields
+                    }
+                    row["fields"] = field_data
+
+                    yield from super()._handle_object(row)
+                except Exception as e:
+                    raise DeserializationError(f"Error deserializing row {row}: {e}")
```
sarahboyce, line 430: Ideally we overwrite existing methods (not sure if this works)
```suggestion
        def __init__(self, stream_or_string, **options):
            if isinstance(stream_or_string, bytes):
                stream_or_string = stream_or_string.decode()
            try:
                objects = csv.DictReader(stream_or_string)
            except Exception as exc:
                raise DeserializationError() from exc
            super().__init__(objects, **options)

        def _handle_object(self, obj):
            try:
                model_fields = apps.get_model(obj["model"])._meta.fields
                obj["fields"] = {
                    field.name: obj[field.name]
                    for field in model_fields
                    if field.name in obj
                }
                yield from super()._handle_object(obj)
            except (GeneratorExit, DeserializationError):
                raise
            except Exception as exc:
               raise DeserializationError(f"Error deserializing object: {exc}") from exc
```

AMK9978:
@sarahboyce 
Spot on! And sorry for the inconvenience. I should've tweaked my code before pushing! 
I just added
```
if isinstance(stream_or_string, str):

                stream_or_string = stream_or_string.splitlines()
```
to the deserializer to enable it to work with `str` input, along with file input and bytes.
Btw, I was thinking of adding some tests to cover this line of json deserializer:
```
        if not isinstance(stream_or_string, (bytes, str)):
            stream_or_string = stream_or_string.read()
```
I don't know when Django wants to support type declaration but for now input types to the deserializers don't seem crystal clear (Not a big deal).

sarahboyce:
Just tested serializing and deserializing our Joe and Jane with the csv serializer and it worked for me :tada: 

> I don't know when Django wants to support type declaration but for now input types to the deserializers don't seem crystal clear (Not a big deal).

I don't think types will be added any time soon (it's quite a topic)

sarahboyce (approved):
Thank you for your work on this @AMK9978 :star:

PR #18337, opened 2024-07-03
common-passwords.txt sorted to use less space with compressed file

tanaydin:
# Branch description
This pull request has sorted the common-passwords.txt file to use less space with the compressed file common-passwords.txt.gzip.
It was 80K before, now it's 56K.


RealOrangeOne:
If post-processing is being done on this file, it might be worth writing a script to take the [raw data](https://gist.github.com/roycewilliams/226886fd01572964e1431ac8afc999ce) and create the export? That way the source is still auditable and verifiable.

tanaydin:
something line in the scripts folder you mean?

RealOrangeOne:
That probably makes sense as a place to put it, at least for now.

RealOrangeOne:
Looks like there's a lot of unrelated changes in this PR now. Perhaps a merge went wrong?

PR #18343, opened 2024-07-05
Fixed #35535 -- Add simple block tag

RealOrangeOne:
# Trac ticket number

ticket-35535

See also related forum thread: https://forum.djangoproject.com/t/feature-proposal-simple-block-tag/32229

# Branch description

Django's `simple_tag` is difficult to work with if you need to capture more complex content as one of the arguments. 

This PR adds the ability to define custom "block" tags, which capture their contents as the `content` argument.

There are probably some built-in tags which could take advantage of this, but I've left those out-of-scope of this PR. The same is true for extra niceties like a built-in `capture` tag, passing the un-rendered content into the tag instead etc.

# Checklist


docs/howto/custom-template-tags.txt
```diff
@@ -498,6 +498,127 @@ you see fit:
...
+
+    from django import template
+    from myapp.charts import render_chart
+
+    register = template.Library()
+
+
+    @register.simple_block_tag
+    def chart(content):
+        return render_chart(source=content)
```
RealOrangeOne, line 525: Something which occured to me, is that the use-cases of a tag like this are fairly limited. However, a variant which accepts a template and processes context (akin to `inclusion_tag`) could be incredibly helpful. Most of the building blocks are here to make that happen, but not sure if that warrants a separate PR, or it's worth trying to squeeze in here? I'm not sure what the diff size would be, which might be the deciding factor.

RealOrangeOne: I've tried this out, and the diff is quite small. `InclusionNode` can be subclassed in much the same way as was done with `SimpleNode`, with similar duplication of `inclusion_tag`. 

With that said, there's little overlap between the 2 feature, so it probably does make sense they be separate. I can contribute those once this is merged.

docs/howto/custom-template-tags.txt
```diff
@@ -498,6 +498,127 @@ you see fit:
...
+-----------------
+
+.. method:: django.template.Library.simple_block_tag()
+
+Passing arguments may not be enough for some usages -- you may need to pass chunks of
+rendered template into your custom tags. This can be achieved using the
+``simple_block_tag`` helper function of ``django.template.Library``.
+``simple_block_tag`` functions similar to :meth:`~django.template.Library.simple_tag()`,
+however additionally passes a ``content`` argument which contains its rendered template
+content.
```
sarahboyce, line 513:
```suggestion
If you need to pass a section of rendered template into your custom tag, this
can be achieved using the ``simple_block_tag`` helper function of
``django.template.Library``. ``simple_block_tag`` functions similar to
:meth:`~django.template.Library.simple_tag()`, but also passes a ``content``
argument which contains the rendered template content block contained by the
tag.
```
So I guess with this, the content itself is an argument and other arguments are not required (if that makes sense)

docs/howto/custom-template-tags.txt
```diff
@@ -498,6 +498,127 @@ you see fit:
...
+    register = template.Library()
+
+
+    @register.simple_block_tag
+    def chart(content):
+        return render_chart(source=content)
+
+The ``content`` argument contains everything in between the ``{% chart %}`` and
+``{% endchart %}`` tags. If there are other template tags or variables, they will be
+rendered before being passed in.
```
sarahboyce, line 529:
```suggestion
The ``content`` argument contains everything in between the ``{% chart %}`` and
``{% endchart %}`` tags. If there are other template tags or variables, they
will be rendered before being passed in.
```
Also in other places, docs need some wrapping adjustments

docs/howto/custom-template-tags.txt
```diff
@@ -498,6 +498,127 @@ you see fit:
...
+A few things to note about the ``simple_block_tag`` helper function:
+
+* The first argument must be called ``content``, and will contain the contents of the
+  template tag as a rendered string.
+* Checking for the required number of arguments, etc., has already been
+  done by the time our function is called, so we don't need to do that.
+* The quotes around the argument (if any) have already been stripped away,
+  so we receive a plain string.
+* If the argument was a template variable, our function is passed the
+  current value of the variable, not the variable itself.
```
sarahboyce, line 549:
```suggestion
* If an argument was a template variable, our function is passed the current
  value of the variable, not the variable itself.
```
As we already have the argument "content", all other arguments are essentially additional and optional I think

docs/howto/custom-template-tags.txt
```diff
@@ -498,6 +498,127 @@ you see fit:
...
+directly outputting it. This is done by using the ``as`` argument followed by
+the variable name. Doing so enables you to output the content yourself where
+you see fit:
+
+.. code-block:: html+django
+
+    {% chart as my_chart %}
+        A -> {B C}
+    {% endchart %}
+    <p>The chart is {{ the_time|length }} characters.</p>
```
sarahboyce, line 620:
```suggestion
    <p>The chart is {{ my_chart|length }} characters.</p>
```

django/template/library.py
```diff
@@ -153,6 +153,86 @@ def compile_func(parser, token):
...
+                _,
+            ) = getfullargspec(unwrap(func))
+            function_name = name or func.__name__
+
+            @wraps(func)
+            def compile_func(parser, token):
+                tag_params = params.copy()
+
+                if takes_context:
+                    if len(tag_params) >= 2 and tag_params[1] == "content":
```
sarahboyce, line 182: We are missing tests for when the `content` argument is misnamed
i.e. I can change this line to `if len(tag_params) >= 2:` and tests pass

sarahboyce (commented):
Thank you for this @RealOrangeOne, added initial comments
We need a release note in 5.2 :+1:

docs/howto/custom-template-tags.txt
```diff
@@ -498,6 +498,128 @@ you see fit:
...
+
+.. versionadded:: 5.2
+
+.. method:: django.template.Library.simple_block_tag()
+
+If you need to pass a section of rendered template into your custom tag, the
+``simple_block_tag`` helper function of ``django.template.Library`` functions
+similar to :meth:`~django.template.Library.simple_tag()`, but also passes a
+``content`` argument which contains its rendered template content contained by
+the tag.
```
sarahboyce, line 514:
```suggestion
If you need to pass a section of rendered template into your custom tag, the
``simple_block_tag`` helper function of ``django.template.Library`` passes a
``content`` argument, containing the rendered template content enclosed in the
tag. This functions similar to :meth:`~django.template.Library.simple_tag()`,
but with an additional ``content`` argument.
```
This sentence is quite long. I wonder if we can split it a bit

sarahboyce (commented):
Thank you for the updates @RealOrangeOne :+1: 

We still need a release note adding to 5.2

tests/template_tests/templatetags/custom.py
```diff
@@ -51,6 +56,12 @@ def one_param(arg):
 one_param.anything = "Expected one_param __dict__"
 
 
+@register.simple_block_tag
+def oneparamblock(content, arg):
```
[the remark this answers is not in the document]
RealOrangeOne: It's consistent with the cases for `simple_tag`, but not with the naming of built-in tags. Personally, I prefer the look of tags without underscores in, but happy to change them if that's the better way to go.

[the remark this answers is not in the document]
sarahboyce: I see what you mean :thinking: we can leave it @nessita do you have a preference here?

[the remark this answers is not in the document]
nessita: Thanks for summon me, could you please allow me for a few days to do a full review and comment?

docs/howto/custom-template-tags.txt
```diff
@@ -498,6 +498,128 @@ you see fit:
...
+the variable name. Doing so enables you to output the content yourself where
+you see fit:
+
+.. code-block:: html+django
+
+    {% chart as my_chart %}
+        A -> {B C}
+    {% endchart %}
+    <p>The chart is {{ my_chart|length }} characters.</p>
+
```
nessita, line 622: I think that there are too many bits that are duplicated from the "simple tags" explanation. I think those should be framed under a summary, helping reducing the duplication. How about something like:

```suggestion
* Variables passed to the tag are not included in the rendering context of the
  content, as would be when using the ``{% with %}`` tag.

Just like :ref:`simple_tag<howto-custom-template-tags-simple-tags>`,
``simple_block_tag``:

* Validates the quantity and quality of the arguments.
* Quotes are stripped from arguments if necessary.
* Output is escaped accordingly.
* Context can be accessed by passing ``takes_context=True`` at registration
  time. Note that in this case, the first argument *must* be called
  ``context``, and ``content`` must follows.
* The tag can be renamed using passing the ``name`` argument when registering.
* ``simple_block_tag`` functions may accept any number of positional or keyword
  arguments.
* The tag result can be stored in a template variable using the ``as`` variant.

.. admonition::  Content Escaping

    ``simple_block_tag`` behaves similarly to ``simple_tag`` regarding
    auto-escaping. For details on escaping and safety, refer to ``simple_tag``.

    Because the ``content`` argument has already been rendered by Django, it is
    already escaped. Therefore, it may be necessary to use
    :func:`~django.utils.safestring.mark_safe` to ensure it's not
    double-escaped.
```

docs/releases/5.2.txt
```diff
@@ -220,7 +220,9 @@ Signals
 Templates
 ~~~~~~~~~
 
-* ...
+* The new :meth:`~django.template.Library.simple_block_tag` method allows
+  creation of simple block tags, which support being passed a section
+  of rendered template.
```
nessita, line 225: Perhaps?
```suggestion
* The new :meth:`~django.template.Library.simple_block_tag` method enables the
  creation of simple block tags that can accept and process a section of
  rendered template content.
```

tests/template_tests/templatetags/custom.py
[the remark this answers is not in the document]
nessita: After a careful review, let's **please** add underscores in these new tags since that's the style of the custom.py file. Otherwise the inconsistency kind of hurts my eyes :rofl: [MY EYES, MY EYES](https://tenor.com/view/my-eyes-friends-phoebe-phoebe-buffay-meme-gif-25833919)!

django/template/library.py
```diff
@@ -153,6 +153,86 @@ def compile_func(parser, token):
         else:
             raise ValueError("Invalid arguments provided to simple_tag")
 
+    def simple_block_tag(self, func=None, takes_context=None, name=None):
```
nessita, line 156: I can't help noticing the amount of duplicated code with `simple_tag`. I tried to give it a go at refactoring this, my solution still has an issue somewhere which I can't find in the time frame I have, but I would love if you could give it a go and share your thoughts. Thanks!

nessita (commented):
Added some comments, thank you @RealOrangeOne for this work!

django/template/library.py
RealOrangeOne: I suspect reducing duplication is going to require the changes deeper down the stack, making them more brittle and higher impact. But with a fresh pair of eyes I'll give them a go and see where we get! It might be possible abstract them out at the very least.

tests/template_tests/templatetags/custom.py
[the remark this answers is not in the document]
RealOrangeOne: It works on my ~machine~ eyes :wink: 

(Makes sense - I'll make the change).

django/template/library.py
nessita: > I suspect reducing duplication is going to require the changes deeper down the stack, making them more brittle and higher impact.

Yeah, maybe...

> But with a fresh pair of eyes I'll give them a go and see where we get! It might be possible abstract them out at the very least.

Perhaps take a "quick" look and let me know what you find. I gave this a "quick" try yesterday and I ended up in a 2 hours rabbit hole. What I tried is (but is not working, when running the tests, `self.func` is unexpectedly `None`:

```python
    def _build_tag(
        self,
        build_node,
        node_name,
        func=None,
        takes_context=None,
        name=None,
        validate_params=None,
    ):
        """Register a callable as a compiled template tag."""

        def dec(func):
            (
                params,
                varargs,
                varkw,
                defaults,
                kwonly,
                kwonly_defaults,
                _,
            ) = getfullargspec(unwrap(func))
            function_name = name or func.__name__

            @wraps(func)
            def compile_func(parser, token):
                if validate_params is not None:
                    validated_params = validate_params(function_name, params)
                else:
                    validated_params = params
                bits = token.split_contents()[1:]
                target_var = None
                if len(bits) >= 2 and bits[-2] == "as":
                    target_var = bits[-1]
                    bits = bits[:-2]
                return build_node(
                    parser,
                    bits,
                    validated_params,
                    varargs,
                    varkw,
                    defaults,
                    kwonly,
                    kwonly_defaults,
                    function_name=function_name,
                    target_var=target_var,
                )

            self.tag(function_name, compile_func)
            return func

        if func is None:
            # @register.<node_name>(...)
            return dec
        elif callable(func):
            # @register.<node_name>
            return dec(func)
        else:
            raise ValueError(f"Invalid arguments provided to {node_name}")


    def simple_tag(self, func=None, takes_context=None, name=None):
        """
        Register a callable as a compiled template tag. Example:

        @register.simple_tag
        def hello(*args, **kwargs):
            return 'world'
        """

        def build_node(*args, function_name, target_var):
            args, kwargs = parse_bits(
                *args,
                takes_context=takes_context,
                name=function_name,
            )
            return SimpleNode(func, takes_context, args, kwargs, target_var)

        return self._build_tag(build_node, "simple_tag", func, takes_context, name)

    def simple_block_tag(self, func=None, takes_context=None, name=None):
        """
        Register a callable as a compiled block template tag. Example:

        @register.simple_block_tag
        def hello(content):
            return 'world'
        """

        def validate_params(function_name, params):
            tag_params = params.copy()
            if takes_context:
                if len(tag_params) >= 2 and tag_params[1] == "content":
                    del tag_params[1]
                else:
                    raise TemplateSyntaxError(
                        f"'{function_name}' is decorated with takes_context=True so"
                        " it must have a first argument of 'context' and a second "
                        "argument of 'content'"
                    )
            else:
                if tag_params and tag_params[0] == "content":
                    del tag_params[0]
                else:
                    raise TemplateSyntaxError(
                        f"'{function_name}' must have a first argument of 'content'"
                    )
            return tag_params

        def build_node(parser, *args, function_name, target_var):
            nodelist = parser.parse((f"end{function_name}",))
            parser.delete_first_token()
            args, kwargs = parse_bits(
                parser,
                *args,
                takes_context=takes_context,
                name=function_name,
            )
            return SimpleBlockNode(
                nodelist, func, takes_context, args, kwargs, target_var
            )

        return self._build_tag(
            build_node, "simple_block_tag", func, takes_context, name, validate_params
        )
```

RealOrangeOne: Whilst there is duplication, I'm not sure how valuable it is splitting it out. Other than the small skeleton of decorating a function and defining it as a tag, the only duplicated parts are:

- Handling a trailing "as"
- Calling `parse_bits`
- Returning a `Node` instance

Trying to remove duplication by separating them out will just make things harder to follow and maintain.

I'd strongly suggest we leave it as-is, and potentially improve it later if a third similar-looking tag is defined (which it may when I write an `include_block` tag).

nessita: Fair enough! I agree with your reasoning. :100: 

I'm out of bandwidth this week to complete the review but I'll do so early next week.

docs/howto/custom-template-tags.txt
```diff
@@ -498,6 +498,75 @@ you see fit:
...
+.. admonition::  Content Escaping
+
+    ``simple_block_tag`` behaves similarly to ``simple_tag`` regarding
+    auto-escaping. For details on escaping and safety, refer to ``simple_tag``.
+
+    Because the ``content`` argument has already been rendered by Django, it is
+    already escaped. Therefore, it may be necessary to use
+    :func:`~django.utils.safestring.mark_safe` to ensure it's not
+    double-escaped.
```
nessita, line 569: I think this is missing an empty line after the paragraphs ends:
```suggestion
    double-escaped.

```

tests/template_tests/templatetags/custom.py
```diff
@@ -51,6 +56,12 @@ def one_param(arg):
 one_param.anything = "Expected one_param __dict__"
 
 
+@register.simple_block_tag
+def one_param_block(content, arg):
+    """Expected oneparam __doc__"""
```
nessita, line 61:
```suggestion
    """Expected one_param_block __doc__"""
```

tests/template_tests/templatetags/custom.py
```diff
@@ -168,24 +309,52 @@ def simple_tag_takes_context_without_params():
...
+
 @register.simple_tag(takes_context=True)
 def escape_naive(context):
     """A tag that doesn't even think about escaping issues"""
     return "Hello {}!".format(context["name"])
 
 
+@register.simple_block_tag(takes_context=True)
+def escape_naive_block(context, content):
+    """A tag that doesn't even think about escaping issues"""
```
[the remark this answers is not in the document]
nessita: If yes, we should apply to the ones below.

nessita (commented):
Thank you @RealOrangeOne! I gave this another top-bottom pass. Once the minor nitpicks are fixed, I need to do a in-depth review in terms of test coverage and functionality. That would be after DjangoCon US.

Amazing work so far! :star2:

docs/howto/custom-template-tags.txt
```diff
@@ -498,8 +498,78 @@ you see fit:
...
```
balazs-endresz, line 569: Does this refer to the following from the `simple_tag` docs?

> Unlike other tag utilities, simple_tag passes its output through conditional_escape() if the template context is in autoescape mode, to ensure correct HTML and protect you from XSS vulnerabilities.
> If additional escaping is not desired, you will need to use mark_safe() if you are absolutely sure that your code does not contain XSS vulnerabilities. For building small HTML snippets, use of format_html() instead of mark_safe() is strongly recommended.

If so, I think it would be worth replacing the sentence above with something similar, because it's much less ambiguous. 

Otherwise, at least I'd suggest using only `format_html` instead of `mark_safe`.

Also, adding some html markup in the inner tag contents to all tests that rely on returning `content` would also help clarify what needs to be escaped and how exactly.

RealOrangeOne: No, this is slightly different. The `simple_tag` docs are specifically about the _output_ of the tag, which is still true for `simple_block_tag`.

This point is about the provided `content` string, which is different.

balazs-endresz: So, for example if I add some html markup in the block contents then `end_one_param_block` doesn't render it as html tags because the interpolation is done incorrectly in the tag definition.

But `escape_format_html_block` still works because it uses `format_html`, which calls `conditional_escape` on `content` and returns a safe string. If I do `return content` as is that works too.

These all seems expected to me, so I'm not sure what this note refers to exactly.

(Also, if there's html in the contents then that's NOT "escaped" in `content` but `content` is a safe string that renders as html, so I don't think that's a correct statement either.)

django/template/library.py
```diff
@@ -153,6 +153,92 @@ def compile_func(parser, token):
...
+                if len(bits) >= 2 and bits[-2] == "as":
+                    target_var = bits[-1]
+                    bits = bits[:-2]
+
+                # Use underscore to separate "end" to match user styling
+                end_tag_name = (
+                    f"end_{function_name}"
+                    if "_" in function_name
+                    else f"end{function_name}"
+                )
```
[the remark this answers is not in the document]
RealOrangeOne: In hindsight, it would be surprising. I think personally I prefer the naming consistency. I've made it customizable now, which lets people style it however they need (with tests, of course).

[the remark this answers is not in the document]
nessita: Does this "end tag name" customization is necessary? I would suggest we remove it and add it if eventually some asks for it...

docs/howto/custom-template-tags.txt
balazs-endresz: I might be still missing something but I think even the whole admonition could be removed because I don't see anything surprising about how this works. The `content` argument is a `SafeString` (as I'd expect it to be, but that might be still worth pointing out) and all the standard django rules about handling those apply here too.

nessita (approved):
I have added a new subsection in the docs with a more complete/complex example. I built it from real use case I had to solve a while ago. From my view, this is ready to be merged.

@sarahboyce @adamchainz @carltongibson I'm keen to know what you think of the latest version of the docs. Ideally I'll merge in about a week, depending on your feedback. Thanks!

docs/howto/custom-template-tags.txt
nessita: @balazs-endresz Thanks. I have pushed a few changes where I add a more complete example that shows how `format_html` is used. Let me know what you think!

balazs-endresz: Thank you, the docs change and the examples look good to me.

docs/howto/custom-template-tags.txt
```diff
@@ -498,6 +498,195 @@ you see fit:
...
+    <div class="msgbox info">
+      <details>
+        <summary>
+          <strong>Info</strong>: Please read for <i>Important Site</i>
+        </summary>
+        <p>
+          More information at: <a href="http://othersite.com">Other Site</a>
+        </p>
+      </details>
+    </div>
```
sarahboyce, line 688: I like that the example shows how this is useful in a more real way, I think it's too complete and I wouldn't put it in it's own section with a header. I would propose:
```suggestion
For a more complete example, a message box that supports different message
levels and content beyond a simple phrase could be implemented via a tag as
follows::

    from django import template
    from django.utils.html import format_html


    register = template.Library()


    @register.simple_block_tag()
    def msgbox(context, content, level, title):
        format_kwargs = {
            "level": level,
            "level_title": level.capitalize(),
            "content": content,
            "open": " open" if level == "error" else "",
            "title": title,
        }
        result = """
        <details{open} class="msgbox {level}">
          <summary>
            <strong>{level_title}!</strong>: {title}</i>
          </summary>
            <p>
              {content}
            </p>
          </details>
        """
        return format_html(result, **format_kwargs)


Then, the following template snippet:

.. code-block:: html+django

    {% msgbox level="error" title="You did something wrong!" %}
      Please fix all errors. Further documentation can be found at
      <a href="http://example.com">Docs</a>.
    {% endmsgbox %}

    {% msgbox level="info" title="Here is some help" %}
      More information at: <a href="http://othersite.com">Other Site</a>/
    {% endmsgbox %}

Would result in HTML as follows:

.. code-block:: html

    <details open class="msgbox error">
      <summary>
        <strong>Error!</strong>: You did something wrong!
      </summary>
      <p>
         Please fix all errors. Further documentation can be found at
         <a href="http://example.com">Docs</a>.
      </p>
    </details>

    <details open class="msgbox error">
      <summary>
        <strong>Info!</strong>: Here is some help
      </summary>
      <p>
        More information at: <a href="http://othersite.com">Other Site</a>
      </p>
    </details>

.. admonition::  Content Escaping

    ``simple_block_tag`` behaves similarly to ``simple_tag`` regarding
    auto-escaping. For details on escaping and safety, refer to ``simple_tag``.
    Because the ``content`` argument has already been rendered by Django, it is
    already escaped.
```

nessita: Thank you @sarahboyce ! Following our conversation where we realized this is part of howto docs (with a mix of reference docs which prompted the initial comment), I pushed some wording tweaks to the example but otherwise left it as a more hand-holding description.

PR #18348, opened 2024-07-07
Add form query parameter documentation

massover:
# Trac ticket number

N/A

# Branch description
From the conversation in this [thread](https://forum.djangoproject.com/t/add-getbool-to-request-query-params-querydict/30985), I thought it was a good idea to add specific form documentation for processing query paramters with a form.

# Checklist


benjaoming (commented):
Have you seen this at the top of the section?

https://github.com/django/django/blob/f287444b03a0910efdecb8fae9cb2d1f27bec629/docs/topics/forms/index.txt#L56-L88

I think the example could perhaps be shortened and reference the discussion section above, using ``:ref:`ref-get-and-post` `` and adding ``
.. _ref-get-and-post:`` above the "GET and POST" section?

(I.e. "With the shortcomings of GET-based form submission in mind (...)")

massover:
> Have you seen this at the top of the section?

Yes, that's section I mention in the [thread](https://forum.djangoproject.com/t/add-getbool-to-request-query-params-querydict/30985/6)

>  I think the example could perhaps be shortened

Do you mind just suggesting what you think we should remove? The intent of the documentation change in this pr is so when a new user looks in the documentation for "how do I use query params in a view?", there is an example for them to see. I understand that the code is pretty much the same as a POST request, but I still think it's valuable. 

> using :ref:`ref-get-and-post` and adding  .. _ref-get-and-post: above the "GET and POST" section?

I agree, good idea. Pretty much all of the words provide context specific to the example, so once you suggest a change to the example I can change the words, and include your `With the shortcomings of GET-based form submission in mind...`

PR #18351, opened 2024-07-08
Fixed #35462 -- Added support for JSON_ArrayAgg aggregate function.

lufafajoshua:
# Trac ticket number

ticket-35462

# Branch description
This tickect adds support for the `JSON_ARRAYAGG` and `JSON_GROUP_ARRAY` for SQLite , these  functions aggregate the contents of a specified column or `SQL` expressions converting each expression into a `JSON` value and returns a single `JSON` array containing those `JSON` values.

# Checklist

### DIFF
--- a/django/db/models/aggregates.py
+++ b/django/db/models/aggregates.py
@@ -3,8 +3,9 @@ Classes to represent the definitions of aggregate functions.
 """
 
 from django.core.exceptions import FieldError, FullResultSet
-from django.db.models.expressions import Case, Func, Star, Value, When
+from django.db.models.expressions import Case, Func, Star, Value, When, connection
 from django.db.models.fields import IntegerField
+from django.db.models.fields.json import JSONField
 from django.db.models.functions.comparison import Coalesce
 from django.db.models.functions.mixins import (
     FixDurationInputMixin,
@@ -20,6 +21,7 @@ __all__ = [
     "StdDev",
     "Sum",
     "Variance",
+    "JSONArrayAgg",
 ]
 
 
@@ -211,3 +213,12 @@ class Variance(NumericOutputFieldMixin, Aggregate):
 
     def _get_repr_options(self):
         return {**super()._get_repr_options(), "sample": self.function == "VAR_SAMP"}
+
+class JSONArrayAgg(Aggregate):
+    name = "JSONArrayAgg"
+    output_field = JSONField()
+    arity = 1
+
+    def __init__(self, expression, **extra):
+        self.function = "JSON_GROUP_ARRAY" if connection.vendor == "sqlite" else "JSON_ARRAYAGG"
+        super().__init__(expression, **extra)
--- a/tests/aggregation/tests.py
+++ b/tests/aggregation/tests.py
@@ -30,6 +30,7 @@ from django.db.models import (
     Variance,
     When,
     Window,
+    JSONArrayAgg,
 )
 from django.db.models.expressions import Func, RawSQL
 from django.db.models.functions import (
@@ -2157,6 +2158,10 @@ class AggregateTestCase(TestCase):
             .values_list("sum", flat=True)
         )
         self.assertEqual(list(author_qs), [337])
+
+    def test_JSONArrayAgg(self):
+        vals = Author.objects.filter(age__gt=29).aggregate(age_array=JSONArrayAgg("age"))
+        self.assertEqual(vals, {'age_array': [34, 35, 45, 37, 57, 46]})
 
 
 class AggregateAnnotationPruningTests(TestCase):


nessita:
> I think postgres tests fail because json_arrayagg is not supported on postgres < 16.

Could you point us to which postgres tests failures are you referring to? I only see lint errors reported by the automatic checks.

django/db/models/aggregates.py
```diff
@@ -211,3 +213,15 @@ def __init__(self, expression, sample=False, **extra):
...
+
+class JSONArrayAgg(Aggregate):
+    name = "JSONArrayAgg"
+    output_field = JSONField()
+    arity = 1
+
+    def __init__(self, expression, **extra):
+        self.function = (
+            "JSON_GROUP_ARRAY" if connection.vendor == "sqlite" else "JSON_ARRAYAGG"
+        )
```
charettes, line 226: 👋 @lufafajoshua!

You can't use `connection` for that since some projects might have `DATABASES` entries for different backends.

Refer to [the `as_sql` documentation](https://docs.djangoproject.com/en/5.0/ref/models/expressions/#django.db.models.Func.as_sql) particularly the part about `as_vendor` and refer to [examples such as `Greatest`](https://github.com/django/django/blob/dfd63ff43408e7901cc214b0482a7f844244d439/django/db/models/functions/comparison.py#L134-L143) to get an idea of how to achieve what you're after.

lufafajoshua:
Hey @charettes , how about that?

django/db/models/aggregates.py
```diff
@@ -211,3 +213,16 @@ def __init__(self, expression, sample=False, **extra):
...
+
+
+class JSONArrayAgg(Aggregate):
+    function = "JSON_ARRAYAGG"
+    name = "JSONArrayAgg"
+    output_field = JSONField()
+    arity = 1
+
+    def as_sqlite(self, compiler, connection, **extra_context):
+        "Use JSON_GROUP_ARRAY function on SQLite."
```
charettes, line 225: I don't think this comment is worth it, the code is self-explanatory.

django/db/models/aggregates.py
```diff
@@ -211,3 +213,16 @@ def __init__(self, expression, sample=False, **extra):
...
+        return super().as_sqlite(
+            compiler, connection, function="JSON_GROUP_ARRAY", **extra_context
+        )
```
charettes, line 228: Don't we also need to use `json_array` on Postgres?

tests/aggregation/tests.py
```diff
@@ -2158,6 +2160,15 @@ def test_having_with_no_group_by(self):
         )
         self.assertEqual(list(author_qs), [337])
 
+    def test_JSONArrayAgg(self):
+        try:
+            vals = Author.objects.filter(age__gt=29).aggregate(
+                age_array=JSONArrayAgg("age")
```
charettes, line 2166: It'd be good to have tests for the aggregation of non-JSON compatible types such a `DateTimeField` and such.

charettes (commented):
@lufafajoshua the implementation is looking good, great job!

You definitely want to extend the tests you have though so errors are not silenced to demonstrate your solution works on all backends though. The fact you silence `ProgrammingError` is a red flag to me.

django/db/models/aggregates.py
Hisham-Pak: Why do we need to use `json_array` when Postgres has added support for [json_array](https://www.postgresql.org/docs/16/functions-json.html) and [json_arrayagg](https://www.postgresql.org/docs/16/functions-aggregate.html) both in version 16? 🤔

charettes: [Django supports Postgres 13+](https://docs.djangoproject.com/en/5.1/ref/databases/#postgresql-notes) so we [can't use a function only available in Postgres 16+](https://www.postgresql.org/docs/current/release-16.html#RELEASE-16-FUNCTIONS).

I suspect we'll have to emulate `json_arrayagg` using `array_agg` on Postgres < 16.

Hisham-Pak: Thanks emulating using `array_agg` makes sense.

lufafajoshua: Hey friends, i would love to have just your thoughts  and feedback on this before i update my PR shortly, so far this is my emulated version of `JSON_ARRAYAGG` using `ARRAY_AGG`.


   
```
    def as_postgresql(self, compiler, connection, **extra_context):
        if not connection.features.is_postgresql_16:
            return super().as_sql(
                compiler,
                connection,
                function="ARRAY_AGG",
                template="ARRAY_TO_JSON(%(function)s(%(expressions)s))::text",
                **extra_context
            )
```

lufafajoshua:
Hey @charettes and @Hisham-Pak , what would be your feedback on the emulated json_arrayagg?

django/db/models/aggregates.py
```diff
@@ -211,3 +213,30 @@ def __init__(self, expression, sample=False, **extra):
...
+                compiler,
+                connection,
+                function="ARRAY_AGG",
+                template="TO_JSON(%(function)s(%(distinct)s%(expressions)s))",
+                **extra_context,
+            )
+        else:
+            sql, params = super().as_sql(compiler, connection, **extra_context)
+        if self.output_field.get_internal_type() == "JSONField":
+            sql = "CAST(%s AS JSONB)" % sql
```
charettes, line 241: Pretty sure this branch will always be true. Is there any reason not to always use `TO_JSONB` and `{sql}RETURNING jsonb` [instead](https://www.postgresql.org/docs/current/functions-aggregate.html#id-1.5.8.27.5.2.4.18.1.1.1)?

django/db/models/aggregates.py
```diff
@@ -211,3 +213,30 @@ def __init__(self, expression, sample=False, **extra):
 
     def _get_repr_options(self):
         return {**super()._get_repr_options(), "sample": self.function == "VAR_SAMP"}
+
+
+class JSONArrayAgg(Aggregate):
+    function = "JSON_ARRAYAGG"
+    name = "JSONArrayAgg"
```
charettes, line 220: Why is `name` required?

django/db/models/aggregates.py
```diff
@@ -211,3 +213,30 @@ def __init__(self, expression, sample=False, **extra):
...
+            compiler, connection, function="JSON_GROUP_ARRAY", **extra_context
+        )
+
+    def as_postgresql(self, compiler, connection, **extra_context):
+        if not connection.features.is_postgresql_16:
+            sql, params = super().as_sql(
+                compiler,
+                connection,
+                function="ARRAY_AGG",
+                template="TO_JSON(%(function)s(%(distinct)s%(expressions)s))",
```
charettes, line 235:
```suggestion
                template="TO_JSONB(ARRAY_AGG(%(distinct)s%(expressions)s))",
```

tests/aggregation/tests.py
```diff
@@ -2158,6 +2159,45 @@ def test_having_with_no_group_by(self):
...
+            vals, {"jsonarrayagg": [30.0, 23.09, 29.69, 29.69, 82.8, 75.0]}
+        )
+
+    def test_JSONArrayAgg_integerfield(self):
+        vals = Book.objects.aggregate(jsonarrayagg=JSONArrayAgg("pages"))
+        self.assertEqual(vals, {"jsonarrayagg": [447, 528, 300, 350, 1132, 946]})
+
+    def test_JSONArrayAgg_filter(self):
+        vals = Author.objects.filter(age__gt=29).aggregate(
+            jsonarrayagg=JSONArrayAgg("age")
```
charettes, line 2197: I don't think we need to test `filter.aggregate` as it has little to do with `JSONArrayAgg` itself.

We should cover the [different properties](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#aggregate-filter) that it inherits from `Aggregate` instead such as 

1. `distinct=True`
2. `filter=Q(...)`
3. `default`

Another area that should be covered are null handling.

charettes (commented):
The emulation through `TO_JSON(ARRAY_AGG(...))` seems to work great 🎉 

I left some comments about redundant and missing tests for `distinct`, `filter`, `default` and we should cover SQL `NULL` handling as well.

lufafajoshua:
Hey @charettes , requesting some review here.

tests/aggregation/tests.py
```diff
@@ -2158,6 +2159,79 @@ def test_having_with_no_group_by(self):
...
+        vals = Book.objects.aggregate(jsonarrayagg=JSONArrayAgg("pages"))
+        self.assertEqual(vals, {"jsonarrayagg": [447, 528, 300, 350, 1132, 946]})
+
+    def test_JSONArrayAgg_filter(self):
+        vals = Author.objects.aggregate(
+            jsonarrayagg=JSONArrayAgg("age", filter=Q(age__gt=29))
+        )
+        if connection.vendor == "mysql":
+            self.assertEqual(
+                vals, {"jsonarrayagg": [34, 35, 45, None, 37, None, None, 57, 46]}
```
charettes, line 2201: This is artifact of MySQl not supporting the `FILTER` clause for aggregate function and requiring its emulation with `CASE` and `WHEN`. This one will be hard to solve as we'll likely need to express the aggregation as a subquery.

tests/aggregation/tests.py
```diff
@@ -2158,6 +2159,79 @@ def test_having_with_no_group_by(self):
...
+            self.assertEqual(vals, {"jsonarrayagg": [34, 35, 45, 37, 57, 46]})
+
+    def test_JSONArrayAgg_empty_result_set(self):
+        Author.objects.all().delete()
+
+        val = Author.objects.aggregate(jsonarrayagg=JSONArrayAgg("age"))
+        if connection.vendor == "sqlite":
+            self.assertEqual(val, {"jsonarrayagg": []})
+        else:
+            self.assertEqual(val, {"jsonarrayagg": None})
```
charettes, line 2213: The same value should be returned in both cases. In this case I assume it should be `None`?

tests/aggregation/tests.py
```diff
@@ -2158,6 +2159,79 @@ def test_having_with_no_group_by(self):
...
+    def test_JSONArrayAgg_default_set(self):
+        Author.objects.all().delete()
+
+        val = Author.objects.aggregate(
+            jsonarrayagg=JSONArrayAgg("name", default=["<empty>"])
+        )
+        if connection.vendor == "sqlite":
+            self.assertEqual(val, {"jsonarrayagg": []})
+        else:
+            self.assertEqual(val, {"jsonarrayagg": ["<empty>"]})
```
charettes, line 2224: Ditto it should be `["<empty>"]` in both cases.

charettes (commented):
👋 @Hisham-Pak this is great but we'll need to find a way for the function to return the same thing on all backend.

Do you know what the spec says about `JSON_ARRAYAGG` against an empty set? Should it return `None` or an empty JSON array?

The MySQL one will be tricky to solve because because using `filter` results in

```sql
JSON_ARRAYAGG(CASE WHEN age > 29 THEN age END)
```

And it produces `NULL` artifacts on non-match. It's not an issue on other backends because they support

```sql
JSON_ARRAYAGG(age) FILTER (WHERE age > 29)
```

I suspect we'll need to perform something along the lines of

```sql
JSON_ARRAYAGG(
    SELECT age
    FROM author a0
    WHERE
        a0.id = author.id
        AND a0.age > 29
)
```

Hisham-Pak:
Hi @charettes and @lufafajoshua,

> Do you know what the spec says about `JSON_ARRAYAGG` against an empty set? Should it return None or an empty JSON array?

It seems that `JSON_ARRAYAGG` returns `null` for an empty set on all supported backends. However, `JSON_GROUP_ARRAY` on SQLite returns an empty JSON array. Therefore, we see that the `default` argument does not work with `JSON_GROUP_ARRAY` on SQLite. I think we should change this behavior for SQLite?

charettes:
> Therefore, we see that the default argument does not work with JSON_GROUP_ARRAY on SQLite. I think we should change this behavior for SQLite?

Do you an idea how to achieve it on SQLite? The only solution I could think of was to do

```sql
CASE WHEN COUNT(column) > 0 THEN JSON_GROUP_ARRAY(column) (ELSE {default})? END
```

I guess something like the following could work in `as_sqlite`

```python
def as_sqlite(compiler, connection, **extra_context):
    sql, params = self.as_sql(
        compiler, connection, function="JSON_GROUP_ARRAY", **extra_context
    )
    # JSON_GROUP_ARRAY defaults to returning an empty array on an empty set.
    if (default := self.default ) == []:
        return sql, params
    # Ensure Count() uses the same filter, default, and distinct.
    count = self.clone()
    count.__class__ = Count
    count_sql, count_params = compiler.compile(count)
    default_sql, default_params = "", ()
    if default is not None:
        default_sql, default_params = compiler.compile(default)
        default_sql = f" ELSE {default_sql}"
    sql = f"(CASE WHEN {count_sql} > 0 THEN {sql}{default_sql} END)"
    return sql, count_params + params + default_params
```

lufafajoshua:
Hey @charettes , i have applied your suggestion and it looks to be working great on `sqlite` with null being returned instead of `[]` and also using the default provided by the user incase of null results.

lufafajoshua:
Hey folks, I added the ``name`` attribute just to be able to provide a `default alias` as manually providing one may be inconvenient to some(IMO).

django/db/models/aggregates.py
```diff
@@ -211,3 +213,39 @@ def __init__(self, expression, sample=False, **extra):
 
     def _get_repr_options(self):
         return {**super()._get_repr_options(), "sample": self.function == "VAR_SAMP"}
+
+
+class JSONArrayAgg(Aggregate):
+    function = "JSON_ARRAYAGG"
+    output_field = JSONField()
+    name = "JSONArrayAgg"
```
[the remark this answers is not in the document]
lufafajoshua: I will be removing, perhaps will be added when someone requests for it in future.

django/db/models/aggregates.py
```diff
@@ -211,3 +213,39 @@ def __init__(self, expression, sample=False, **extra):
...
+    def as_postgresql(self, compiler, connection, **extra_context):
+        if not connection.features.is_postgresql_16:
+            sql, params = super().as_sql(
+                compiler,
+                connection,
+                function="ARRAY_AGG",
+                **extra_context,
+            )
+            return f"TO_JSONB({sql})", params
+        self.template = "%(function)s(%(distinct)s%(expressions)s RETURNING JSONB)"
```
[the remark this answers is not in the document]
lufafajoshua: Hey @charettes , after applying the above suggestion i get an  error `TypeError:django.db.models.expressions.Func.as_sql() got multiple values for keyword argument 'template'`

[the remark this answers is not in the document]
charettes: @lufafajoshua hard to tell for sure because you didn't provide the full traceback but I assume these lines are to blame

https://github.com/django/django/blob/c4f3006aec5d3d9cebcce6d410d1121e5c0a1003/django/db/models/aggregates.py#L128-L137

Looks like there is a bug in there we should do the following instead

```python
extra_context = {
    **extra_context,
    "template": self.filter_template % extra_context.get(
        "template", self.template
    )
}
sql, params = super().as_sql(
    compiler,
    connection,
    filter=filter_sql,
    **extra_context,
)
```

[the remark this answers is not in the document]
lufafajoshua: This is the full traceback.

`
>File "<console>", line 1, in <module>
  File "/home/lufafa/Django-Core-Codebase/django/django/db/models/manager.py", line 87, in manager_method
  return getattr(self.get_queryset(), name)(*args, **kwargs).    
  File "/home/lufafa/Django-Core-Codebase/django/django/db/models/query.py", line 604, in aggregate
    return self.query.chain().get_aggregation(self.db, kwargs)        
  File "/home/lufafa/Django-Core-Codebase/django/django/db/models/sql/query.py", line 616, in get_aggregation
    result = compiler.execute_sql(SINGLE)            
  File "/home/lufafa/Django-Core-Codebase/django/django/db/models/sql/compiler.py", line 1556, in execute_sql
    sql, params = self.as_sql()                 
  File "/home/lufafa/Django-Core-Codebase/django/django/db/models/sql/compiler.py", line 742, in as_sql
    extra_select, order_by, group_by = self.pre_sql_setup(                                      
  File "/home/lufafa/Django-Core-Codebase/django/django/db/models/sql/compiler.py", line 84, in pre_sql_setup
    self.setup_query(with_col_aliases=with_col_aliases)
  File "/home/lufafa/Django-Core-Codebase/django/django/db/models/sql/compiler.py", line 73, in setup_query
    self.select, self.klass_info, self.annotation_col_map = self.get_select(                                                      
  File "/home/lufafa/Django-Core-Codebase/django/django/db/models/sql/compiler.py", line 296, in get_select
    sql, params = self.compile(col)                
  File "/home/lufafa/Django-Core-Codebase/django/django/db/models/sql/compiler.py", line 550, in compile
    sql, params = vendor_impl(self, self.connection)             
  File "/home/lufafa/Django-Core-Codebase/django/django/db/models/aggregates.py", line 296, in as_postgresql
    return super().as_sql(compiler, connection, template=template, **extra_context)           
  File "/home/lufafa/Django-Core-Codebase/django/django/db/models/aggregates.py", line 134, in as_sql
    sql, params = super().as_sql(
TypeError: django.db.models.expressions.Func.as_sql() got multiple values for keyword argument 'template'`

lufafajoshua:
Requesting some review @charettes

docs/ref/models/querysets.txt
```diff
@@ -4047,6 +4047,17 @@ by the aggregate.
...
+~~~~~~~~~~~~~~~~
+
+.. class:: JSONArrayAgg(expression, output_field=None, sample=False, filter=None, default=None, **extra)
+
+    Converts each expression to a JSON value and returns a single JSON array
+    containing those JSON values.
+
+    * Return type: ``list`` containing an element for each value
+      in a set of JSON or SQL values.
+
```
charettes, line 4060: You'll need a `::versionaddded` mention here.

tests/aggregation/tests.py
```diff
@@ -2158,6 +2159,85 @@ def test_having_with_no_group_by(self):
...
+
+    @skipUnlessDBFeature("supports_aggregate_filter_clause")
+    def test_JSONArrayAgg_empty_result_set(self):
+        Author.objects.all().delete()
+
+        val = Author.objects.aggregate(jsonarrayagg=JSONArrayAgg("age"))
+
+        self.assertEqual(val, {"jsonarrayagg": None})
+
+    @skipUnlessDBFeature("supports_aggregate_filter_clause")
```
charettes, line 2215: Remove this skip; the `filter` clause is not used here and I suspect this is might be hiding some other failures (e.g. on Oracle)

tests/aggregation/tests.py
```diff
@@ -2158,6 +2159,85 @@ def test_having_with_no_group_by(self):
...
+    def test_JSONArrayAgg_default_set(self):
+        Author.objects.all().delete()
+
+        val = Author.objects.aggregate(
+            jsonarrayagg=JSONArrayAgg("name", default=["<empty>"])
+        )
+        self.assertEqual(val, {"jsonarrayagg": ["<empty>"]})
+
+    @skipUnlessDBFeature("supports_aggregate_filter_clause")
```
charettes, line 2224: Same thing `filter` is not used here so this should not be skipped.

tests/aggregation/tests.py
```diff
@@ -2158,6 +2159,85 @@ def test_having_with_no_group_by(self):
...
+    def test_JSONArrayAgg_distinct_false(self):
```
charettes, line 2225: Actually I'm not sure this test adds any value since `distinct=False` is the default so this behavior is implicitly tested in other instances.

tests/aggregation/tests.py
```diff
@@ -2158,6 +2159,85 @@ def test_having_with_no_group_by(self):
...
+        val = Author.objects.aggregate(jsonarrayagg=JSONArrayAgg("age", distinct=False))
+        self.assertEqual(val, {"jsonarrayagg": [34, 35, 45, 29, 37, 29, 25, 57, 46]})
+
+    @skipUnlessDBFeature("supports_aggregate_filter_clause")
```
charettes, line 2229: Remove as well, the `filter` kwarg is not used here.

tests/aggregation/tests.py
```diff
@@ -2158,6 +2159,85 @@ def test_having_with_no_group_by(self):
...
+    def test_JSONArrayAgg_distinct_true(self):
+        msg = "JSONArrayAgg does not allow distinct."
+        with self.assertRaisesMessage(TypeError, msg):
+            JSONArrayAgg("age", distinct=True)
+
+    @skipIfDBFeature("supports_aggregate_filter_clause")
```
charettes, line 2235: Remove as well, the `filter` kwarg is not used here.

tests/aggregation/tests.py
```diff
@@ -2158,6 +2159,85 @@ def test_having_with_no_group_by(self):
...
+
+    @skipUnlessDBFeature("supports_aggregate_filter_clause")
+    def test_JSONArrayAgg_filter(self):
+        vals = Author.objects.aggregate(
+            jsonarrayagg=JSONArrayAgg("age", filter=Q(age__gt=29))
+        )
+
+        self.assertEqual(vals, {"jsonarrayagg": [34, 35, 45, 37, 57, 46]})
+
+    @skipUnlessDBFeature("supports_aggregate_filter_clause")
```
charettes, line 2207: Remove as well, the `filter` kwarg is not used here.

tests/aggregation/tests.py
```diff
@@ -2158,6 +2159,85 @@ def test_having_with_no_group_by(self):
...
+                    "2008-03-03",
+                    "2008-06-23",
+                    "2008-11-03",
+                    "1995-01-15",
+                    "1991-10-15",
+                ]
+            },
+        )
+
+    @skipUnlessDBFeature("supports_aggregate_filter_clause")
```
charettes, line 2187: Remove as well, the `filter` kwarg is not used here.

django/db/models/aggregates.py
```diff
@@ -211,3 +215,47 @@ def __init__(self, expression, sample=False, **extra):
...
+        count = self.copy()
+        count.__class__ = Count
+        count_sql, count_params = compiler.compile(count)
+        default_sql = ""
+        default_params = () if self.filter is not None else []
+        if default is not None:
+            default_sql, default_params = compiler.compile(default)
+            default_sql = f" ELSE {default_sql}"
+        sql = f"(CASE WHEN {count_sql} > 0 THEN {sql}{default_sql} END)"
+        return sql, count_params + params + default_params
```
charettes, line 247: We might want to add a comment here explain why this is done for future readers as it's not explicit this is required to handle the fact that SQLite's `JSON_GROUP_ARRAY` doesn't return `NULL` when dealing with empty sets.

django/db/models/aggregates.py
```diff
@@ -211,3 +215,47 @@ def __init__(self, expression, sample=False, **extra):
...
         return {**super()._get_repr_options(), "sample": self.function == "VAR_SAMP"}
+
+
+class JSONArrayAgg(Aggregate):
+    function = "JSON_ARRAYAGG"
+    output_field = JSONField()
+    arity = 1
+
+    def as_sql(self, compiler, connection, **extra_context):
+        if not connection.features.supports_aggregate_filter_clause:
```
charettes, line 226: You should only raise if there's an actual filter; otherwise it's actually supported?

```suggestion
        if self.filter and not connection.features.supports_aggregate_filter_clause:
```

django/db/models/aggregates.py
```diff
@@ -211,3 +215,47 @@ def __init__(self, expression, sample=False, **extra):
...
+            raise NotSupportedError(
+                "JSONArrayAgg is not supported on this database backend."
```
charettes, line 228:
```suggestion
                "JSONArrayAgg(filter) is not supported on this database backend."
```

charettes (commented):
I think there might have been misunderstanding here; we only need to raise a `NotSupportedError` if *there is a filter* and it's not supported. The current code is systematically raising if it's not supported even if no filters are specified.

tests/aggregation/tests.py
```diff
@@ -2359,3 +2360,86 @@ def test_aggregate_combined_queries(self):
...
+            name="The Definitive Guide to Django: Web Development Done Right",
+            pages=447,
+            rating=4.5,
+            price=Decimal("30.00"),
+            contact=cls.a1,
+            publisher=cls.p1,
+            pubdate=datetime.date(2007, 12, 6),
+        )
+
+    def test_JSONArrayAgg(self):
```
charettes, line 2387: I think you can prune the `JSONArrayAgg` part of all these test names as they are already namespace by the `JSONArrayAggTests` test case name.

tests/aggregation/tests.py
```diff
@@ -2359,3 +2360,86 @@ def test_aggregate_combined_queries(self):
...
+        cls.b1 = Book.objects.create(
+            isbn="159059725",
+            name="The Definitive Guide to Django: Web Development Done Right",
+            pages=447,
+            rating=4.5,
+            price=Decimal("30.00"),
+            contact=cls.a1,
+            publisher=cls.p1,
+            pubdate=datetime.date(2007, 12, 6),
+        )
```
charettes, line 2385: Since aggregation usually happens over relationships we should have tests that cover aggregation over fields that span over relationships.

tests/aggregation/tests.py
```diff
@@ -2359,3 +2360,86 @@ def test_aggregate_combined_queries(self):
...
+
+    def test_JSONArrayAgg(self):
+        vals = Author.objects.aggregate(jsonarrayagg=JSONArrayAgg("name"))
```
charettes, line 2388: It'd be great if the tests covered relationships such as `books__name` as the way the tests are currently designed they never result in JOINs!

docs/ref/models/querysets.txt
```diff
@@ -4047,6 +4047,19 @@ by the aggregate.
         However, if ``sample=True``, the return value will be the sample
         variance.
 
+``JSONArrayAgg``
+~~~~~~~~~~~~~~~~
+
+.. versionadded:: 5.1
```
sarahboyce, line 4053:
```suggestion
.. versionadded:: 5.2
```

tests/aggregation/tests.py
```diff
@@ -2359,3 +2360,91 @@ def test_aggregate_combined_queries(self):
...
+
+    def test_decimalfield(self):
+        vals = Author.objects.aggregate(jsonarrayagg=JSONArrayAgg("book__price"))
+        self.assertEqual(vals, {"jsonarrayagg": [30.0, 23.09]})
+
+    def test_integerfield(self):
+        vals = Author.objects.aggregate(jsonarrayagg=JSONArrayAgg("book__pages"))
+        self.assertEqual(vals, {"jsonarrayagg": [447, 528]})
+
+    @skipUnlessDBFeature("supports_aggregate_filter_clause")
```
sarahboyce, line 2422: I feel `@skipUnlessDBFeature("supports_aggregate_filter_clause")` should be applied more consistently

django/db/models/aggregates.py
```diff
@@ -211,3 +215,49 @@ def __init__(self, expression, sample=False, **extra):
...
+                compiler,
+                connection,
+                function="ARRAY_AGG",
+                **extra_context,
+            )
+            return f"TO_JSONB({sql})", params
+        extra_context.setdefault(
+            "template", "%(function)s(%(distinct)s%(expressions)s RETURNING JSONB)"
+        )
+        return self.as_sql(compiler, connection, **extra_context)
```
sarahboyce, line 263:
```suggestion
        self.template = "%(function)s(%(distinct)s%(expressions)s RETURNING JSONB)"
        return self.as_sql(compiler, connection, **extra_context)
```

django/db/models/aggregates.py
```diff
@@ -123,13 +126,14 @@ def as_sql(self, compiler, connection, **extra_context):
...
-                    )
+                    extra_context = {
+                        **extra_context,
+                        "template": self.filter_template
+                        % extra_context.get("template", self.template),
+                    }
                     sql, params = super().as_sql(
                         compiler,
                         connection,
-                        template=template,
```
sarahboyce, line 132: I think we can revert this (made a suggestion below instead)

sarahboyce (commented):
Thank you @lufafajoshua - added initial comments

Can you also rebase onto main, resolve conflicts, squash to a single commit and add a release note to Django 5.2? :star:

django/db/models/aggregates.py
charettes: As a general rule of thumb you can't alter `self` during compilation as it alters the resolved object and could cause issues if the queryset object was reused in a different context. That explains the adjustments in `Aggregate.as_sql`.

django/db/models/aggregates.py
```diff
@@ -123,13 +126,14 @@ def as_sql(self, compiler, connection, **extra_context):
...
                     pass
                 else:
-                    template = self.filter_template % extra_context.get(
-                        "template", self.template
-                    )
+                    extra_context = {
+                        **extra_context,
+                        "template": self.filter_template
+                        % extra_context.get("template", self.template),
+                    }
```
charettes, line 133: Adding parentheses might make this this easier to grok while conforming to `black`

```suggestion
                    extra_context = {
                        **extra_context,
                        "template": (
                            self.filter_template % extra_context.get("template", self.template)
                         )
                    }
```

django/db/models/aggregates.py
charettes: An alternative could be to create a copy of `self` and assign the `.template` to it but that seems it would circumvent the problem that `Aggregate.as_sql` crash if you provide a `template` to it as it will attempt pass the same `kwarg` twice in its `super().as_sql` call.

tests/aggregation/tests.py
lufafajoshua: Hey @sarahboyce , what do you mean consistently?

sarahboyce: Sorry I believe I misinterpreted the db feature :+1:

django/db/models/aggregates.py
```diff
@@ -228,3 +233,49 @@ def __init__(self, expression, sample=False, **extra):
...
+    arity = 1
+
+    def as_sql(self, compiler, connection, **extra_context):
+        if self.filter and not connection.features.supports_aggregate_filter_clause:
+            raise NotSupportedError(
+                "JSONArrayAgg(filter) is not supported on this database backend."
+            )
+        return super().as_sql(compiler, connection, **extra_context)
+
+    def as_sqlite(self, compiler, connection, **extra_context):
```
sarahboyce, line 250:
```suggestion
    def as_sqlite(self, compiler, connection, **extra_context):
        def as_sql(self, compiler, connection, **extra_context):
            if not connection.features.supports_json_field:
                raise NotSupportedError(
                    "JSONFields are not supported on this database backend."
                )
```
Been reviewing https://github.com/django/django/pull/18541/ which has a top level check for json fields, I think this might want the same?

django/db/models/aggregates.py
```diff
@@ -228,3 +233,49 @@ def __init__(self, expression, sample=False, **extra):
...
     def _get_repr_options(self):
         return {**super()._get_repr_options(), "sample": self.function == "VAR_SAMP"}
+
+
+class JSONArrayAgg(Aggregate):
+    function = "JSON_ARRAYAGG"
+    output_field = JSONField()
+    arity = 1
+
+    def as_sql(self, compiler, connection, **extra_context):
```
sarahboyce, line 243:
```suggestion
    def as_sql(self, compiler, connection, **extra_context):
            if not connection.features.supports_json_field:
                raise NotSupportedError(
                    "JSONFields are not supported on this database backend."
                )
```
Been reviewing https://github.com/django/django/pull/18541/ which has a top level check for Json field support - I think this might want the same?

charettes: yeah I think it makes sense to add a similar check here.

tests/aggregation/tests.py
```diff
@@ -2380,3 +2381,91 @@ def test_aggregate_combined_queries(self):
...
+    def test_filter(self):
+        vals = Book.objects.aggregate(
+            jsonarrayagg=JSONArrayAgg("contact__age", filter=Q(contact__age__gt=35))
+        )
+        self.assertEqual(vals, {"jsonarrayagg": [45]})
+
+    def test_empty_result_set(self):
+        Author.objects.all().delete()
+        val = Author.objects.aggregate(jsonarrayagg=JSONArrayAgg("age"))
```
pauloxnet, line 2452:
```suggestion
        val = Author.objects.none().aggregate(jsonarrayagg=JSONArrayAgg("age"))
```
I suggest using the `none` method, instead of deleting all `Author` instances created in the `setUpTestData` method, which is discouraged because it may lead to tests not being well isolated from each other.

tests/aggregation/tests.py
```diff
@@ -2380,3 +2381,91 @@ def test_aggregate_combined_queries(self):
...
+        self.assertEqual(val, {"jsonarrayagg": None})
+
+    def test_default_set(self):
+        Author.objects.all().delete()
+        val = Author.objects.aggregate(
```
pauloxnet, line 2457:
```suggestion
        val = Author.objects.none().aggregate(
```
I suggest using the `none` method, instead of deleting all `Author` instances created in the `setUpTestData` method, which is discouraged because it may lead to tests not being well isolated from each other.

tests/aggregation/tests.py
```diff
@@ -2380,3 +2381,91 @@ def test_aggregate_combined_queries(self):
...
+    def test_datefield(self):
+        vals = Author.objects.aggregate(jsonarrayagg=JSONArrayAgg("book__pubdate"))
+        self.assertEqual(
+            vals,
+            {
+                "jsonarrayagg": [
+                    "2007-12-06",
+                    "2008-03-03",
+                ]
+            },
```
pauloxnet, line 2432:
```suggestion
            {"jsonarrayagg": ["2007-12-06", "2008-03-03"]},
```
Used similar formatting as in the other tests.

pauloxnet (approved):
The general PR seems ok for me.
I only added 3 small (not-blocking) code change suggestions.

docs/releases/5.2.txt
```diff
@@ -356,6 +360,7 @@ Models
   that the expression allows composite expressions, for example, to support
   :ref:`composite primary keys <cpk-and-database-functions>`.
 
+
```
pauloxnet, line 363:
```suggestion
```
Remove this extra empty row.

pauloxnet (approved):
I added a small code changes, but the PR seems ready to be merged.

sarahboyce:
buildbot, test on oracle.

john-parton:
Related: https://github.com/django/django/pull/19097

lufafajoshua:
buildbot, test on oracle.

lufafajoshua:
Can't surely tell why my changes aren't working here but everything seems okay locally, so i am closing this PR and am opening a new one.

PR #18356, opened 2024-07-08
Fixed #35575 -- Added support for constraint validation on GeneratedFields

sdolemelipone:
# Trac ticket number

ticket-35575

# Branch description

At present, `GeneratedField`s on a model instance are not recalculated when checking model constraints during `Model.full_clean()`. Instead, the model instance's `GeneratedField` attributes will return the values which correspond to the most recent refresh of the instance from the database (or raise an `AttributeError` if it has not yet been saved). The database does correctly raise an `IntegrityError` during `Model.save()` if any constraints involving `GeneratedField`s are violated.

This PR is to modify `BaseConstraint.validate()` to use up to date values for `GeneratedField` attributes which reflect unsaved changes on a model instance.

# Checklist
- [x] ~~I have attached screenshots in both light and dark modes for any UI changes.~~

### DIFF
--- a/tests/constraints/models.py
+++ b/tests/constraints/models.py
@@ -1,4 +1,5 @@
 from django.db import models
+from django.db.models.functions import Coalesce, Lower
 
 
 class Product(models.Model):
@@ -26,6 +27,46 @@ class Product(models.Model):
                 name="unicode_unit_list",
             ),
         ]
+
+
+class GeneratedFieldStoredProduct(models.Model):
+    name = models.CharField(max_length=255, null=True)
+    price = models.IntegerField(null=True)
+    discounted_price = models.IntegerField(null=True)
+    rebate = models.GeneratedField(
+        expression=Coalesce("price", 0)
+        - Coalesce("discounted_price", Coalesce("price", 0)),
+        output_field=models.IntegerField(),
+        db_persist=True,
+    )
+    lower_name = models.GeneratedField(
+        expression=Lower(models.F("name")),
+        output_field=models.CharField(max_length=255, null=True),
+        db_persist=True,
+    )
+
+    class Meta:
+        required_db_features = {"supports_stored_generated_columns"}
+
+
+class GeneratedFieldVirtualProduct(models.Model):
+    name = models.CharField(max_length=255, null=True)
+    price = models.IntegerField(null=True)
+    discounted_price = models.IntegerField(null=True)
+    rebate = models.GeneratedField(
+        expression=Coalesce("price", 0)
+        - Coalesce("discounted_price", Coalesce("price", 0)),
+        output_field=models.IntegerField(),
+        db_persist=False,
+    )
+    lower_name = models.GeneratedField(
+        expression=Lower(models.F("name")),
+        output_field=models.CharField(max_length=255, null=True),
+        db_persist=False,
+    )
+
+    class Meta:
+        required_db_features = {"supports_virtual_generated_columns"}
 
 
 class UniqueConstraintProduct(models.Model):
--- a/tests/constraints/tests.py
+++ b/tests/constraints/tests.py
@@ -4,7 +4,7 @@ from django.core.exceptions import ValidationError
 from django.db import IntegrityError, connection, models
 from django.db.models import F
 from django.db.models.constraints import BaseConstraint, UniqueConstraint
-from django.db.models.functions import Abs, Lower, Upper
+from django.db.models.functions import Abs, Lower, Sqrt, Upper
 from django.db.transaction import atomic
 from django.test import SimpleTestCase, TestCase, skipIfDBFeature, skipUnlessDBFeature
 from django.test.utils import ignore_warnings
@@ -13,6 +13,8 @@ from django.utils.deprecation import RemovedInDjango60Warning
 from .models import (
     ChildModel,
     ChildUniqueConstraintProduct,
+    GeneratedFieldStoredProduct,
+    GeneratedFieldVirtualProduct,
     JSONFieldModel,
     ModelWithDatabaseDefault,
     Product,
@@ -383,6 +385,29 @@ class CheckConstraintTests(TestCase):
         msg = f"Constraint “{json_exact_constraint.name}” is violated."
         with self.assertRaisesMessage(ValidationError, msg):
             json_exact_constraint.validate(JSONFieldModel, JSONFieldModel(data=data))
+
+    @skipUnlessDBFeature("supports_stored_generated_columns")
+    def test_validate_generated_field_stored(self):
+        self.assertGeneratedFieldIsValidated(model=GeneratedFieldStoredProduct)
+
+    @skipUnlessDBFeature("supports_virtual_generated_columns")
+    def test_validate_generated_field_virtual(self):
+        self.assertGeneratedFieldIsValidated(model=GeneratedFieldVirtualProduct)
+
+    def assertGeneratedFieldIsValidated(self, model):
+        constraint = models.CheckConstraint(
+            condition=models.Q(rebate__range=(0, 100)), name="bounded_rebate"
+        )
+        constraint.validate(model, model(price=50, discounted_price=20))
+
+        invalid_product = model(price=1200, discounted_price=500)
+        msg = f"Constraint “{constraint.name}” is violated."
+        with self.assertRaisesMessage(ValidationError, msg):
+            constraint.validate(model, invalid_product)
+
+        # Excluding referenced or generated fields should skip validation.
+        constraint.validate(model, invalid_product, exclude={"price"})
+        constraint.validate(model, invalid_product, exclude={"rebate"})
 
     def test_check_deprecation(self):
         msg = "CheckConstraint.check is deprecated in favor of `.condition`."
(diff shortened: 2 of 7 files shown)


sdolemelipone:
This PR is not yet complete - there are still a couple of outstanding issues (indicated by the failing test):

Problem 1
========

When calling `generated_field.expression.replace_expressions(replacements)`, if the `GeneratedField` implies `Q()` expressions such as
```python
models.GeneratedField(
    expression=models.Case(
        models.When(name__isnull=True, then=None),
        default=Lower("name"),
        output_field=models.CharField(max_length=10, null=True),
    ),
    output_field=models.CharField(max_length=10, null=True),
    db_persist=False,
)
```
then the following error is raised:
```
  File "/home/melipone/PycharmProjects/django/django/db/models/constraints.py", line 653, in validate
    for field_name, value in instance._get_field_value_map(
    ^^^^^^^^^^^^^^^^^
  File "/home/melipone/PycharmProjects/django/django/db/models/base.py", line 1360, in _get_field_value_map
    generated_field.expression.replace_expressions(replacements),
    ^^^^^^^^^^^^^^^^^
  File "/home/melipone/PycharmProjects/django/django/db/models/expressions.py", line 415, in replace_expressions
    expr.replace_expressions(replacements) if expr else None
    ^^^^^^^^^^^^^^^^^
  File "/home/melipone/PycharmProjects/django/django/db/models/expressions.py", line 415, in replace_expressions
    expr.replace_expressions(replacements) if expr else None
    ^^^^^^^^^^^^^^^^^
AttributeError: 'Q' object has no attribute 'replace_expressions'
```
I assume there is some further flattening of expressions that needs to take place as part of `_get_field_value_map()`?

Problem 2
========
At present, a special case is made when calling  `validate()` for a `UniqueConstraint` which uses the `fields` parameter which checks whether the constraint uses `nulls_distinct=True` or the database uses `interprets_empty_strings_as_nulls`:
```python
def validate(...):
    ...
    for field_name in self.fields:
        ...
        if field.generated:
            ...
        else:
            lookup_value = getattr(instance, field.attname)
            if (
                self.nulls_distinct is not False
                and lookup_value is None
                or (
                    lookup_value == ""
                    and connections[
                        using
                    ].features.interprets_empty_strings_as_nulls
                )
            ):
                # A composite constraint containing NULL value cannot cause
                # a violation since NULL != NULL in SQL.
                return
            lookup_kwargs[field.name] = lookup_value
```
A similar check needs to be made when checking null values of `GeneratedField`s. I haven't figured out how to do this yet.

sdolemelipone:
You're welcome, thanks for the quick feedback! The `'Q' object has no attribute 'replace_expressions'` issue does look thorny. But at least it's simpler for us here that this isn't related to `GeneratedField`.

> I think that solving both of them is something we should attempt but it seems like doing thing incrementally here, particularly in a context where some of these changes might be backported, will work better.

Agreed, adding this code in will improve the current situation even if it doesn't take care of everything. 

So next I'll remove the tests which deal with the `nulls_distinct` case and also fail with `replace_expressions` error (which surprisingly didn't fail here, they did on my local machine). Then this PR should be ready for review.

Assuming this PR is accepted I'll then raise a new ticket and add failing tests to that to deal with the `nulls_distinct` and `interprets_empty_strings_as_nulls` cases, then that can be worked on separately.

Let me know if there's anything else you'd suggest.

sdolemelipone:
I think this is ready for review with the following issues outstanding:

### Case() expressions
Any `GeneratedField` (or constraint) which includes a `Case` expression, such as the below, will generate an error during validation. This is not specific to `GeneratedField` and is being considered separately [here](https://code.djangoproject.com/ticket/34871#comment:10)
```python
models.Case(
    models.When(name__isnull=True, then=None),
    default=Lower("name"),
    output_field=models.CharField(max_length=10, null=True),
)
```
The error raised during `BaseConstraint.validate()`:
```
AttributeError: 'Q' object has no attribute 'replace_expressions'
```

### Null values in `UniqueConstraint`
Validation of a `UniqueConstraint` on a field which has a null value currently works (i.e. treats `nulls_distinct=True`) because of the following result:
```python
MyModel.objects.create(some_field=None)
MyModel.objects.filter(some_field=Value(None)).exists()  # False
# Even though
MyModel.objects.filter(some_field=None).exists() # True 
```
If this is working as intended then all is ok. If not then it might need further investigation. But this is not specific to `GeneratedField`.

### UniqueConstraints using `nulls_distinct=False` or `connection.interprets_empty_strings_as_nulls`
Further to the above point on null values, if a UniqueConstraint uses `nulls_distinct=False` or the database connection uses `interprets_empty_strings_as_nulls`, and a `GeneratedField` is included in the `fields` parameter, then a `ValidationError` is not raised by `UniqueConstraint.validate()`. Supporting them will be harder to solve. However, again this is not particularly specific to `GeneratedField`, as the same problem occurs for expressions passed in to `UniqueConstraint`.

If we're happy to move ahead given the above, let me know and I will add documentation/release notes as necessary.

charettes (commented):
I didn't look into the MySQL failures into details but I suspect something either needs to be rebased or one of the feature based model creation is misconfigured?

> Case() expressions

yep this is already contained in ticket-34871

> Null values in UniqueConstraint

This is known as well and due to the fact the lookup systems turns `foo=None` as `foo__isnull=True` which results in `"foo" IS NULL` while `foo=Value(None)` is too _opaque_ and gets turned in `"foo" = NULL` which is always `NULL` or _undefined_. I don't think that's a huge deal to be honest as I see little reason to use `Value(None)` (unless it is to specify a `Value(None, JSONField())` to mean JSON `null`) and it has little to do with constraint validation on generated field which is the focus of this ticket.

> UniqueConstraints using nulls_distinct=False or connection.interprets_empty_strings_as_nulls

I think this one is worth the creation of distinct ticket to at least keep track of it. I suspect we'll need to change the logic from immediately aborting to using a clever wrapping of `Coalesce` for generated expressions. This is not only a problem for `UniqueConstraints(fields=fields, nulls_distinct=False)` with generated fields though, a similar problem also exists for `UniqueConstraints(*expressions, nulls_distinct=False)` because the `expressions` are not wrapped in a `Coalesce` that ensures that `NULL` they might express are considered equal.

sdolemelipone:
Thanks for the feedback. I have rebased and pushed again. All the tests passed for the first push and nothing materially changed since, so hopefully this will do it.

> I don't think that's a huge deal to be honest as I see little reason to use. Value(None) (unless it is to specify a Value(None, JSONField()) to mean JSON null) and it has little to do with constraint validation on generated field which is the focus of this ticket.

The reason I mentioned this is that I noticed during `UniqueConstraint.validate()`, when `queryset(**lookup_kwargs).exists()` is called, any expressions in `lookup_kwargs` which evaluate as `None` actually evaluate as `Value(None)`, which is what causes the existence of other rows which evaluate to `NULL` for that expression not to cause a `ValidationError` to be raised. 

I am just finding my feet in contributing to the ORM and don't have a breadth of knowledge across the codebase to know whether this is an issue or not, but wanted to highlight it as it looked odd.

> UniqueConstraints using nulls_distinct=False or connection.interprets_empty_strings_as_nulls

That all makes sense, thanks. I'll raise a ticket for that issue shortly (but won't look at a PR until this one is resolved).

sdolemelipone:
I've raised ticket-35594 for the `nulls_distinct` issue.

charettes:
Thank you for your patience @sdolemelipone, it should be safe to rebase now 🎉

sdolemelipone:
This is ready for review now. I haven't updated any release notes as I wasn't sure if this is a 5.0 or 5.1 - let me know and I can add that!

After this, the only outstanding issue with `GeneratedField` and constraint validation will be those which use a `Case` statement. Should we add something to the documentation saying this is a known issue?

sdolemelipone:
MySQL is testing fine for me locally - can anyone point me in the direction of why they're failing on the server? I couldn't see anything obvious on Jenkins.

Just thinking about this some more, this PR is going to cause `Model.full_clean()` to break for models which have a `GeneratedField` which uses `Case` in it's `expression`, and **any** constraint defined. Prior to this PR, an error would only have occurred when calling `full_clean()` on **unsaved** model instances which had any `GeneratedField` defined.

Should we also add a change so that an error does not occur during `Model._get_field_value_map()`, until ticket-34871 is addressed?

So my instinct would be, this isn't a bugfix so should be in 5.1.0. Is that right?

tests/constraints/tests.py
```diff
@@ -1033,6 +1058,82 @@ def test_validate_expression_str(self):
...
+            nulls_distinct=False,
+        )
+        model.objects.create(name=None)
+        valid_product = model(name="Box")
+        constraint.validate(model, valid_product)
+
+        invalid_product = model(name=None)
+        msg = str(invalid_product.unique_error_message(model, ["lower_name"]))
+        with self.assertRaisesMessage(ValidationError, msg):
+            constraint.validate(model, invalid_product)
```
charettes, line 1135: Just a note to say that I really like how you broke these tests down, kudos!

charettes (approved):
> MySQL is testing fine for me locally - can anyone point me in the direction of why they're failing on the server? I couldn't see anything obvious on Jenkins.

CI has been acting weird for the past few days. I wouldn't worry too much about it as I pulled your branch and managed to run the full suite with success against MySQL locally as well.

> Just thinking about this some more, this PR is going to cause Model.full_clean() to break for models which have a GeneratedField which uses Case in it's expression, and any constraint defined. Prior to this PR, an error would only have occurred when calling full_clean() on unsaved model instances which had any GeneratedField defined.

I think it's fine considering that this is more of a problem with `Case` and constraint validation than a problem with `GeneratedField` itself. In the rare cases where this will be an issue (both features are relatively new) it should be possible to avoid the problem by passing `exclude` to `full_clean`.

The patch is looking good to me but since this is a bug in a newly released feature (`GeneratedField`) I think it will be backported to 5.0 and thus will require a release note.

tests/constraints/tests.py
sdolemelipone: :heart: thanks! :smile:

sdolemelipone:
> The patch is looking good to me but since this is a bug in a newly released feature (`GeneratedField`) I think it will be backported to 5.0 and thus will require a release note.

Thanks for the review and kind words! I've added the release note.

sdolemelipone:
@charettes one further question - I couldn't see an easy way to update the test migations for postgres so did that manually. Could you check that no further migrations would be required (and let me know what command you run to do it?).

charettes:
> I couldn't see an easy way to update the test migations for postgres so did that manually. 

That's what I usually end up doing as well. There's a way to get the framework to generate them for by defining a setting module configured for Postgres 

```
# tests/test_postgres.py
DATABASES = {
    "default": {
        "NAME": "django",
        "HOST": "localhost",
        "USER": "django",
        "PASSWORD": "django",
        "PORT": 5432,
    },
    "other": {
        "NAME": "django",
        "HOST": "localhost",
        "USER": "django",
        "PASSWORD": "django",
        "PORT": 5432,
    },
}

SECRET_KEY = "django_tests_secret_key"

# Use a fast hasher to speed up tests.
PASSWORD_HASHERS = [
    "django.contrib.auth.hashers.MD5PasswordHasher",
]

DEFAULT_AUTO_FIELD = "django.db.models.AutoField"

USE_TZ = False

INSTALLED_APPS = ['postgres_tests']
```

And then calling

```
PYTHONPATH=tests python -m django makemigrations postgres_tests --settings=test_postgres
```

sdolemelipone:
> PYTHONPATH=tests python -m django makemigrations postgres_tests --settings=test_postgres

Thanks, that and adding `INSTALLED_APPS` to my `test_postgres.py` file was exactly what I was looking for.

docs/releases/5.0.8.txt
```diff
@@ -11,3 +11,5 @@ Bugfixes
 
 * Added missing validation for ``UniqueConstraint(nulls_distinct=False)`` when
   using ``*expressions`` (:ticket:`35594`).
+* Added support for validation of model constraints which use a ``GeneratedField``
+  (:ticket:`35575`).
```
sarahboyce, line 15: As this ticket has been classed as a new feature, we won't backport to 5.0 (which only accepts bug fixes)
Can you instead add a release note to [Django 5.2 in the Models section](https://docs.djangoproject.com/en/dev/releases/5.2/#models)?

tests/postgres_tests/models.py
```diff
@@ -144,6 +144,15 @@ class RangesModel(PostgreSQLModel):
     dates_inner = DateRangeField(blank=True, null=True)
 
 
+class RangesModelGeneratedField(PostgreSQLModel):
+    ints = IntegerRangeField(blank=True, null=True)
+    ints_generated = models.GeneratedField(
+        expression=models.F("ints"),
+        output_field=IntegerRangeField(null=True),
+        db_persist=True,
+    )
```
sarahboyce, line 153: I think this is needed
```suggestion
    )

    class Meta:
        required_db_features = {"supports_stored_generated_columns"}
```

tests/postgres_tests/test_constraints.py
```diff
@@ -866,6 +873,25 @@ def test_validate_range_adjacent(self):
         constraint.validate(RangesModel, RangesModel(ints=(51, 60)))
         constraint.validate(RangesModel, RangesModel(ints=(10, 20)), exclude={"ints"})
 
+    def test_validate_generated_field_range_adjacent(self):
```
sarahboyce, line 876:
```suggestion
    @skipUnlessDBFeature("supports_stored_generated_columns")
    def test_validate_generated_field_range_adjacent(self):
```

tests/constraints/models.py
```diff
@@ -28,6 +29,46 @@ class Meta:
         ]
 
 
+class GeneratedFieldStoredProduct(models.Model):
```
sarahboyce, line 32: Just a thought, see https://github.com/django/django/pull/18407 on some tests where the model is defined in the test cases with `db_persist` as a parameter (might be worth seeing if you like it for your tests as it might save defining two very similar models

sarahboyce (commented):
Thank you for this @sdolemelipone :star: 
We merged in https://github.com/django/django/pull/18425 which was a release blocker and that's caused some conflicts for this PR (sorry about that) - would you mind taking a look and resolving? :pray:

sdolemelipone:
> Thank you for this @sdolemelipone ⭐ We merged in #18425 which was a release blocker and that's caused some conflicts for this PR (sorry about that) - would you mind taking a look and resolving? 🙏

Absolutely, I'll look through the changes you suggested and rebase.

tests/constraints/models.py
sdolemelipone: Thanks @sarahboyce, I tried this, but on saving an instance of a class defined within a test I receive the error below. I tried decorating the test method with `@isolate_apps("constraints")` but that didn't resolve the problem. Do you have any suggestions for how to work around this? If I can't resolve I'll just leave the tests as they were.



```
ERROR: test_validate_expression_generated_field_stored (constraints.tests.UniqueConstraintTests.test_validate_expression_generated_field_stored)
  File "/home/melipone/PycharmProjects/django/django/db/backends/utils.py", line 105, in _execute
    ^^^^^^^^^^^^^^^^^
  File "/home/melipone/PycharmProjects/django/.venv/lib/python3.12/site-packages/psycopg/cursor.py", line 97, in execute
    raise ex.with_traceback(None)
    ^^^^^^^^^^^^^^^^^
psycopg.errors.UndefinedTable: relation "constraints_testmodel" does not exist
LINE 1: INSERT INTO "constraints_testmodel" ("price", "discounted_pr...

```

sarahboyce: Leave it how it is :+1: I can have a look later but we may keep it how you have it anyway so don't worry too much

sdolemelipone: Will, do thanks :-)

tests/postgres_tests/models.py
sdolemelipone: Done, thanks!

tests/postgres_tests/test_constraints.py
sdolemelipone: Done, thanks :+1:

tests/postgres_tests/models.py
sdolemelipone: It appears that adding this causes the tests to fail in databases that don't support stored generated fields. None of the other models in postgres_tests/models.py have similar Meta flags, so I will leave this out.

sdolemelipone:
Ok, I think this is good to go, let me know if you need anything else @sarahboyce.

sarahboyce (approved):
Thank you @sdolemelipone :star: 
I pushed minor tweaks

PR #11841, opened 2019-09-27
28296 Adds as_subquery method to aggregates.

martsberger:
https://code.djangoproject.com/ticket/28296

Also leverages the new methods on Subquery to make the interface for Exists simpler.

PR #18361, opened 2024-07-11
Fixed #35444 -- Added generic support for Aggregate.order_by

camuthig:
# Trac ticket number

ticket-35444

# Branch description
There are three commits to this draft that represent the three phases of the change:

1. Deprecate the `ordering` argument on the PostgreSQL aggregate functions and replace it with `order_by` to create a consistent naming convention
2. Add generic support for `Aggregate.order_by` and migrate the Postgres-specific `StringAgg` class to the shared `aggregates` module, deprecating the Postgres version.
3. Deprecate the `OrderableAggMixin`

# Checklist

### DIFF
--- a/django/contrib/gis/db/models/aggregates.py
+++ b/django/contrib/gis/db/models/aggregates.py
@@ -33,15 +33,14 @@ class GeoAggregate(Aggregate):
         if not self.is_extent:
             tolerance = self.extra.get("tolerance") or getattr(self, "tolerance", 0.05)
             clone = self.copy()
-            source_expressions = self.get_source_expressions()
-            source_expressions.pop()  # Don't wrap filters with SDOAGGRTYPE().
+            *source_exprs, filter_expr, order_by_expr = self.get_source_expressions()
             spatial_type_expr = Func(
-                *source_expressions,
+                *source_exprs,
                 Value(tolerance),
                 function="SDOAGGRTYPE",
                 output_field=self.output_field,
             )
-            source_expressions = [spatial_type_expr, self.filter]
+            source_expressions = [spatial_type_expr, filter_expr, order_by_expr]
             clone.set_source_expressions(source_expressions)
             return clone.as_sql(compiler, connection, **extra_context)
         return self.as_sql(compiler, connection, **extra_context)
--- a/django/contrib/postgres/aggregates/general.py
+++ b/django/contrib/postgres/aggregates/general.py
@@ -1,7 +1,12 @@
-from django.contrib.postgres.fields import ArrayField
-from django.db.models import Aggregate, BooleanField, JSONField, TextField, Value
+import warnings
 
-from .mixins import OrderableAggMixin
+from django.contrib.postgres.fields import ArrayField
+from django.db.models import Aggregate, BooleanField, JSONField
+from django.db.models import StringAgg as _StringAgg
+from django.db.models import Value
+from django.utils.deprecation import RemovedInDjango70Warning
+
+from .mixins import _DeprecatedOrdering
 
 __all__ = [
     "ArrayAgg",
@@ -11,14 +16,16 @@ __all__ = [
     "BoolAnd",
     "BoolOr",
     "JSONBAgg",
-    "StringAgg",
+    "StringAgg",  # RemovedInDjango70Warning.
 ]
 
 
-class ArrayAgg(OrderableAggMixin, Aggregate):
+# RemovedInDjango61Warning: When the deprecation ends, replace with:
+# class ArrayAgg(Aggregate):
+class ArrayAgg(_DeprecatedOrdering, Aggregate):
     function = "ARRAY_AGG"
-    template = "%(function)s(%(distinct)s%(expressions)s %(order_by)s)"
     allow_distinct = True
+    allow_order_by = True
 
     @property
     def output_field(self):
@@ -47,19 +54,37 @@ class BoolOr(Aggregate):
     output_field = BooleanField()
 
 
-class JSONBAgg(OrderableAggMixin, Aggregate):
+# RemovedInDjango61Warning: When the deprecation ends, replace with:
+# class JSONBAgg(Aggregate):
+class JSONBAgg(_DeprecatedOrdering, Aggregate):
     function = "JSONB_AGG"
-    template = "%(function)s(%(distinct)s%(expressions)s %(order_by)s)"
     allow_distinct = True
+    allow_order_by = True
     output_field = JSONField()
 
 
-class StringAgg(OrderableAggMixin, Aggregate):
-    function = "STRING_AGG"
-    template = "%(function)s(%(distinct)s%(expressions)s %(order_by)s)"
-    allow_distinct = True
-    output_field = TextField()
+# RemovedInDjango61Warning: When the deprecation ends, replace with:
+# class StringAgg(_StringAgg):
+# RemovedInDjango70Warning: When the deprecation ends, remove completely.
+class StringAgg(_DeprecatedOrdering, _StringAgg):
 
     def __init__(self, expression, delimiter, **extra):
-        delimiter_expr = Value(str(delimiter))
-        super().__init__(expression, delimiter_expr, **extra)
+        if isinstance(delimiter, str):
+            warnings.warn(
+                "delimiter: str will be resolved as a field reference instead "
+                "of a string literal on Django 7.0. Pass "
+                f"`delimiter=Value({delimiter!r})` to preserve the previous behaviour.",
+                category=RemovedInDjango70Warning,
+                stacklevel=2,
+            )
+
+            delimiter = Value(delimiter)
+
+        warnings.warn(
+            "The PostgreSQL specific StringAgg function is deprecated. Use "
+            "django.db.models.aggregate.StringAgg instead.",
+            category=RemovedInDjango70Warning,
+            stacklevel=2,
+        )
+
+        super().__init__(expression, delimiter, **extra)
--- a/django/contrib/postgres/aggregates/mixins.py
+++ b/django/contrib/postgres/aggregates/mixins.py
@@ -1,6 +1,7 @@
+# RemovedInDjango70Warning: When the deprecation ends, remove completely.
 import warnings
 
-from django.utils.deprecation import RemovedInDjango61Warning
+from django.utils.deprecation import RemovedInDjango61Warning, RemovedInDjango70Warning
 
 
 # RemovedInDjango61Warning.
@@ -19,10 +20,17 @@ class _DeprecatedOrdering:
         super().__init__(*expressions, order_by=order_by, **extra)
 
 
+# RemovedInDjango70Warning.
 # RemovedInDjango61Warning: When the deprecation ends, replace with:
 # class OrderableAggMixin:
 class OrderableAggMixin(_DeprecatedOrdering):
     allow_order_by = True
 
     def __init_subclass__(cls, /, *args, **kwargs):
+        warnings.warn(
+            "OrderableAggMixin is deprecated. Use Aggregate and allow_order_by "
+            "instead.",
+            category=RemovedInDjango70Warning,
+            stacklevel=1,
+        )
         super().__init_subclass__(*args, **kwargs)
--- a/django/contrib/postgres/aggregates/mixins.py
+++ b/django/contrib/postgres/aggregates/mixins.py
@@ -1,15 +1,11 @@
 import warnings
 
-from django.core.exceptions import FullResultSet
-from django.db.models.expressions import OrderByList
 from django.utils.deprecation import RemovedInDjango61Warning
 
 
-class OrderableAggMixin:
-    # RemovedInDjango61Warning: When the deprecation ends, replace with:
-    # def __init__(self, *expressions, order_by=(), **extra):
+# RemovedInDjango61Warning.
+class _DeprecatedOrdering:
     def __init__(self, *expressions, ordering=(), order_by=(), **extra):
-        # RemovedInDjango61Warning.
         if ordering:
             warnings.warn(
                 "The ordering argument is deprecated. Use order_by instead.",
@@ -19,44 +15,14 @@ class OrderableAggMixin:
             if order_by:
                 raise TypeError("Cannot specify both order_by and ordering.")
             order_by = ordering
-        if not order_by:
-            self.order_by = None
-        elif isinstance(order_by, (list, tuple)):
-            self.order_by = OrderByList(*order_by)
-        else:
-            self.order_by = OrderByList(order_by)
-        super().__init__(*expressions, **extra)
 
-    def resolve_expression(self, *args, **kwargs):
-        if self.order_by is not None:
-            self.order_by = self.order_by.resolve_expression(*args, **kwargs)
-        return super().resolve_expression(*args, **kwargs)
+        super().__init__(*expressions, order_by=order_by, **extra)
 
-    def get_source_expressions(self):
-        return super().get_source_expressions() + [self.order_by]
 
-    def set_source_expressions(self, exprs):
-        *exprs, self.order_by = exprs
-        return super().set_source_expressions(exprs)
+# RemovedInDjango61Warning: When the deprecation ends, replace with:
+# class OrderableAggMixin:
+class OrderableAggMixin(_DeprecatedOrdering):
+    allow_order_by = True
 
-    def as_sql(self, compiler, connection):
-        *source_exprs, filtering_expr, order_by_expr = self.get_source_expressions()
-
-        order_by_sql = ""
-        order_by_params = []
-        if order_by_expr is not None:
-            order_by_sql, order_by_params = compiler.compile(order_by_expr)
-
-        filter_params = []
-        if filtering_expr is not None:
-            try:
-                _, filter_params = compiler.compile(filtering_expr)
-            except FullResultSet:
-                pass
-
-        source_params = []
-        for source_expr in source_exprs:
-            source_params += compiler.compile(source_expr)[1]
-
-        sql, _ = super().as_sql(compiler, connection, order_by=order_by_sql)
-        return sql, (*source_params, *order_by_params, *filter_params)
+    def __init_subclass__(cls, /, *args, **kwargs):
+        super().__init_subclass__(*args, **kwargs)
--- a/django/db/backends/base/features.py
+++ b/django/db/backends/base/features.py
@@ -257,6 +257,15 @@ class BaseDatabaseFeatures:
     # expressions?
     supports_aggregate_filter_clause = False
 
+    # Does the database support ORDER BY in aggregate expressions?
+    supports_aggregate_order_by_clause = False
+
+    # Does the database backend support DISTINCT when using multiple arguments in an
+    # aggregate expression? For example, Sqlite treats the "delimiter" argument of
+    # STRING_AGG/GROUP_CONCAT as an extra argument and does not allow using a custom
+    # delimiter along with DISTINCT.
+    supports_aggregate_distinct_multiple_argument = True
+
     # Does the backend support indexing a TextField?
     supports_index_on_text_field = True
 
--- a/django/db/models/aggregates.py
+++ b/django/db/models/aggregates.py
@@ -3,8 +3,17 @@ Classes to represent the definitions of aggregate functions.
 """
 
 from django.core.exceptions import FieldError, FullResultSet
-from django.db.models.expressions import Case, ColPairs, Func, Star, Value, When
-from django.db.models.fields import IntegerField
+from django.db import NotSupportedError
+from django.db.models.expressions import (
+    Case,
+    ColPairs,
+    Func,
+    OrderByList,
+    Star,
+    Value,
+    When,
+)
+from django.db.models.fields import IntegerField, TextField
 from django.db.models.functions import Coalesce
 from django.db.models.functions.mixins import (
     FixDurationInputMixin,
@@ -18,42 +27,91 @@ __all__ = [
     "Max",
     "Min",
     "StdDev",
+    "StringAgg",
     "Sum",
     "Variance",
 ]
 
 
+class AggregateFilter(Func):
+    arity = 1
+    template = " FILTER (WHERE %(expressions)s)"
+
+    def as_sql(self, compiler, connection, **extra_context):
+        if not connection.features.supports_aggregate_filter_clause:
+            raise NotSupportedError(
+                "Aggregate filter clauses are not supported on this database backend."
+            )
+        try:
+            return super().as_sql(compiler, connection, **extra_context)
+        except FullResultSet:
+            return "", ()
+
+    @property
+    def condition(self):
+        return self.source_expressions[0]
+
+    def __str__(self):
+        return self.arg_joiner.join(str(arg) for arg in self.source_expressions)
+
+
+class AggregateOrderBy(OrderByList):
+    template = " ORDER BY %(expressions)s"
+
+    def as_sql(self, compiler, connection, **extra_context):
+        if not connection.features.supports_aggregate_order_by_clause:
+            raise NotSupportedError(
+                "This database backend does not support specifying an order on "
+                "aggregates."
+            )
+
+        return super().as_sql(compiler, connection, **extra_context)
+
+
 class Aggregate(Func):
-    template = "%(function)s(%(distinct)s%(expressions)s)"
+    template = "%(function)s(%(distinct)s%(expressions)s%(order_by)s)%(filter)s"
     contains_aggregate = True
     name = None
-    filter_template = "%s FILTER (WHERE %%(filter)s)"
     window_compatible = True
     allow_distinct = False
+    allow_order_by = False
     empty_result_set_value = None
 
     def __init__(
-        self, *expressions, distinct=False, filter=None, default=None, **extra
+        self,
+        *expressions,
+        distinct=False,
+        filter=None,
+        default=None,
+        order_by=None,
+        **extra,
     ):
         if distinct and not self.allow_distinct:
             raise TypeError("%s does not allow distinct." % self.__class__.__name__)
+        if order_by and not self.allow_order_by:
+            raise TypeError("%s does not allow order_by." % self.__class__.__name__)
         if default is not None and self.empty_result_set_value is not None:
             raise TypeError(f"{self.__class__.__name__} does not allow default.")
+
         self.distinct = distinct
-        self.filter = filter
+        self.filter = filter and AggregateFilter(filter)
         self.default = default
+        self.order_by = AggregateOrderBy.from_param(
+            f"{self.__class__.__name__}.order_by", order_by
+        )
         super().__init__(*expressions, **extra)
 
     def get_source_fields(self):
-        # Don't return the filter expression since it's not a source field.
+        # Don't consider filter and order by expression as they have nothing
+        # to do with the output field resolution.
         return [e._output_field_or_none for e in super().get_source_expressions()]
 
     def get_source_expressions(self):
         source_expressions = super().get_source_expressions()
-        return source_expressions + [self.filter]
+        return source_expressions + [self.filter, self.order_by]
 
     def set_source_expressions(self, exprs):
-        *exprs, self.filter = exprs
+        *exprs, self.filter, self.order_by = exprs
         return super().set_source_expressions(exprs)
 
     def resolve_expression(
@@ -64,6 +122,11 @@ class Aggregate(Func):
         c.filter = (
             c.filter.resolve_expression(query, allow_joins, reuse, summarize)
             if c.filter
+            else None
+        )
+        c.order_by = (
+            c.order_by.resolve_expression(query, allow_joins, reuse, summarize)
+            if c.order_by
             else None
         )
         if summarize:
@@ -115,35 +178,45 @@ class Aggregate(Func):
         return []
 
     def as_sql(self, compiler, connection, **extra_context):
-        extra_context["distinct"] = "DISTINCT " if self.distinct else ""
-        if self.filter:
-            if connection.features.supports_aggregate_filter_clause:
-                try:
-                    filter_sql, filter_params = self.filter.as_sql(compiler, connection)
-                except FullResultSet:
-                    pass
-                else:
-                    template = self.filter_template % extra_context.get(
-                        "template", self.template
-                    )
-                    sql, params = super().as_sql(
-                        compiler,
-                        connection,
-                        template=template,
-                        filter=filter_sql,
-                        **extra_context,
-                    )
-                    return sql, (*params, *filter_params)
-            else:
+        if (
+            self.distinct
+            and not connection.features.supports_aggregate_distinct_multiple_argument
+            and len(super().get_source_expressions()) > 1
+        ):
+            raise NotSupportedError(
+                f"{self.name} does not support distinct with multiple expressions on "
+                f"this database backend."
+            )
+
+        distinct_sql = "DISTINCT " if self.distinct else ""
+        order_by_sql = ""
+        order_by_params = []
+        filter_sql = ""
+        filter_params = []
+
+        if (order_by := self.order_by) is not None:
+            order_by_sql, order_by_params = compiler.compile(order_by)
+
+        if self.filter is not None:
+            try:
+                filter_sql, filter_params = compiler.compile(self.filter)
+            except NotSupportedError:
+                # Fallback to a CASE statement on backends that don't support
+                # the FILTER clause.
                 copy = self.copy()
                 copy.filter = None
                 source_expressions = copy.get_source_expressions()
-                condition = When(self.filter, then=source_expressions[0])
+                condition = When(self.filter.condition, then=source_expressions[0])
                 copy.set_source_expressions([Case(condition)] + source_expressions[1:])
-                return super(Aggregate, copy).as_sql(
-                    compiler, connection, **extra_context
-                )
-        return super().as_sql(compiler, connection, **extra_context)
+                return copy.as_sql(compiler, connection, **extra_context)
+
+        extra_context.update(
+            distinct=distinct_sql,
+            filter=filter_sql,
+            order_by=order_by_sql,
+        )
+        sql, params = super().as_sql(compiler, connection, **extra_context)
+        return sql, (*params, *order_by_params, *filter_params)
 
     def _get_repr_options(self):
         options = super()._get_repr_options()
@@ -151,6 +224,8 @@ class Aggregate(Func):
             options["distinct"] = self.distinct
         if self.filter:
             options["filter"] = self.filter
+        if self.order_by:
+            options["order_by"] = self.order_by
         return options
 
 
@@ -179,17 +254,17 @@ class Count(Aggregate):
 
     def resolve_expression(self, *args, **kwargs):
         result = super().resolve_expression(*args, **kwargs)
-        expr = result.source_expressions[0]
+        source_expressions = result.get_source_expressions()
 
         # In case of composite primary keys, count the first column.
-        if isinstance(expr, ColPairs):
+        if isinstance(expr := source_expressions[0], ColPairs):
             if self.distinct:
                 raise ValueError(
                     "COUNT(DISTINCT) doesn't support composite primary keys"
                 )
 
-            cols = expr.get_cols()
-            return Count(cols[0], filter=result.filter)
+            source_expressions[0] = expr.get_cols()[0]
+            result.set_source_expressions(source_expressions)
 
         return result
 
@@ -216,6 +291,88 @@ class StdDev(NumericOutputFieldMixin, Aggregate):
 
     def _get_repr_options(self):
         return {**super()._get_repr_options(), "sample": self.function == "STDDEV_SAMP"}
+
+
+class StringAggDelimiter(Func):
+    arity = 1
+    template = "%(expressions)s"
+
+    def __init__(self, value):
+        self.value = value
+        super().__init__(value)
+
+    def as_mysql(self, compiler, connection, **extra_context):
+        template = " SEPARATOR %(expressions)s"
+
+        return self.as_sql(
+            compiler,
+            connection,
+            template=template,
+            **extra_context,
+        )
+
+
+class StringAgg(Aggregate):
+    template = "%(function)s(%(distinct)s%(expressions)s%(order_by)s)%(filter)s"
+    function = "STRING_AGG"
+    name = "StringAgg"
+    allow_distinct = True
+    allow_order_by = True
+    output_field = TextField()
+
+    def __init__(self, expression, delimiter, **extra):
+        self.delimiter = StringAggDelimiter(delimiter)
+        super().__init__(expression, self.delimiter, **extra)
+
+    def as_oracle(self, compiler, connection, **extra_context):
+        if self.order_by:
+            template = (
+                "%(function)s(%(distinct)s%(expressions)s) WITHIN GROUP (%(order_by)s)"
+                "%(filter)s"
+            )
+        else:
+            template = "%(function)s(%(distinct)s%(expressions)s)%(filter)s"
+
+        return self.as_sql(
+            compiler,
+            connection,
+            function="LISTAGG",
+            template=template,
+            **extra_context,
+        )
+
+    def as_mysql(self, compiler, connection, **extra_context):
+        extra_context["function"] = "GROUP_CONCAT"
+
+        template = "%(function)s(%(distinct)s%(expressions)s%(order_by)s%(delimiter)s)"
+        extra_context["template"] = template
+
+        c = self.copy()
+        # The creation of the delimiter SQL and the ordering of the parameters must be
+        # handled explicitly, as MySQL puts the delimiter at the end of the aggregate
+        # using the `SEPARATOR` declaration (rather than treating as an expression like
+        # other database backends).
+        delimiter_params = []
+        if c.delimiter:
+            delimiter_sql, delimiter_params = compiler.compile(c.delimiter)
+            # Drop the delimiter from the source expressions.
+            c.source_expressions = c.source_expressions[:-1]
+            extra_context["delimiter"] = delimiter_sql
+
+        sql, params = c.as_sql(compiler, connection, **extra_context)
+
+        return sql, (*params, *delimiter_params)
+
+    def as_sqlite(self, compiler, connection, **extra_context):
+        if connection.get_database_version() < (3, 44):
+            return self.as_sql(
+                compiler,
+                connection,
+                function="GROUP_CONCAT",
+                **extra_context,
+            )
+
+        return self.as_sql(compiler, connection, **extra_context)
 
 
 class Sum(FixDurationInputMixin, Aggregate):
--- a/docs/ref/models/expressions.txt
+++ b/docs/ref/models/expressions.txt
@@ -448,7 +448,7 @@ some complex computations::
 
 The ``Aggregate`` API is as follows:
 
-.. class:: Aggregate(*expressions, output_field=None, distinct=False, filter=None, default=None, **extra)
+.. class:: Aggregate(*expressions, output_field=None, distinct=False, filter=None, default=None, order_by=None, **extra)
 
     .. attribute:: template
 
@@ -473,6 +473,15 @@ The ``Aggregate`` API is as follows:
         allows passing a ``distinct`` keyword argument. If set to ``False``
         (default), ``TypeError`` is raised if ``distinct=True`` is passed.
 
+    .. attribute:: allow_order_by
+
+        .. versionadded:: 6.0
+
+        A class attribute determining whether or not this aggregate function
+        allows passing a ``order_by`` keyword argument. If set to ``False``
+        (default), ``TypeError`` is raised if ``order_by`` is passed as a value
+        other than ``None``.
+
     .. attribute:: empty_result_set_value
 
         Defaults to ``None`` since most aggregate functions result in ``NULL``
@@ -491,6 +500,12 @@ The ``filter`` argument takes a :class:`Q object <django.db.models.Q>` that's
 used to filter the rows that are aggregated. See :ref:`conditional-aggregation`
 and :ref:`filtering-on-annotations` for example usage.
 
+The ``order_by`` argument behaves similarly to the ``field_names`` input of the
+:meth:`~.QuerySet.order_by` function, accepting a field name (with an optional
+``"-"`` prefix which indicates descending order) or an expression (or a tuple
+or list of strings and/or expressions) that specifies the ordering of the
+elements in the result.
+
 The ``default`` argument takes a value that will be passed along with the
 aggregate to :class:`~django.db.models.functions.Coalesce`. This is useful for
 specifying a value to be returned other than ``None`` when the queryset (or
@@ -498,6 +513,10 @@ grouping) contains no entries.
 
 The ``**extra`` kwargs are ``key=value`` pairs that can be interpolated
 into the ``template`` attribute.
+
+.. versionchanged:: 6.0
+
+    The ``order_by`` argument was added.
 
 Creating your own Aggregate Functions
 -------------------------------------
--- a/tests/aggregation/models.py
+++ b/tests/aggregation/models.py
@@ -43,3 +43,7 @@ class Store(models.Model):
 
     def __str__(self):
         return self.name
+
+
+class Employee(models.Model):
+    work_day_preferences = models.JSONField()
--- a/tests/aggregation/tests.py
+++ b/tests/aggregation/tests.py
@@ -4,10 +4,11 @@ import re
 from decimal import Decimal
 
 from django.core.exceptions import FieldError
-from django.db import connection
+from django.db import NotSupportedError, connection
 from django.db.models import (
     Avg,
     Case,
+    CharField,
     Count,
     DateField,
     DateTimeField,
@@ -22,6 +23,7 @@ from django.db.models import (
     OuterRef,
     Q,
     StdDev,
+    StringAgg,
     Subquery,
     Sum,
     TimeField,
@@ -32,9 +34,11 @@ from django.db.models import (
     Window,
 )
 from django.db.models.expressions import Func, RawSQL
+from django.db.models.fields.json import KeyTextTransform
 from django.db.models.functions import (
     Cast,
     Coalesce,
+    Concat,
     Greatest,
     Least,
     Lower,
@@ -45,11 +49,11 @@ from django.db.models.functions import (
     TruncHour,
 )
 from django.test import TestCase
-from django.test.testcases import skipUnlessDBFeature
+from django.test.testcases import skipIfDBFeature, skipUnlessDBFeature
 from django.test.utils import Approximate, CaptureQueriesContext
 from django.utils import timezone
 
-from .models import Author, Book, Publisher, Store
+from .models import Author, Book, Employee, Publisher, Store
 
 
 class NowUTC(Now):
@@ -565,6 +569,28 @@ class AggregateTestCase(TestCase):
                     ratings=aggregate("rating", distinct=True)
                 )
                 self.assertEqual(books["ratings"], expected_result)
+
+    @skipUnlessDBFeature("supports_aggregate_distinct_multiple_argument")
+    def test_distinct_on_stringagg(self):
+        books = Book.objects.aggregate(
+            ratings=StringAgg(Cast(F("rating"), CharField()), Value(","), distinct=True)
+        )
+        self.assertEqual(books["ratings"], "3,4,4.5,5")
+
+    @skipIfDBFeature("supports_aggregate_distinct_multiple_argument")
+    def test_raises_error_on_multiple_argument_distinct(self):
+        message = (
+            "StringAgg does not support distinct with multiple expressions on this "
+            "database backend."
+        )
+        with self.assertRaisesMessage(NotSupportedError, message):
+            Book.objects.aggregate(
+                ratings=StringAgg(
+                    Cast(F("rating"), CharField()),
+                    Value(","),
+                    distinct=True,
+                )
+            )
 
     def test_non_grouped_annotation_not_in_group_by(self):
         """
@@ -1288,24 +1314,30 @@ class AggregateTestCase(TestCase):
             Book.objects.annotate(Max("id")).annotate(my_max=MyMax("id__max", "price"))
 
     def test_multi_arg_aggregate(self):
-        class MyMax(Max):
+        class MultiArgAgg(Max):
             output_field = DecimalField()
             arity = None
 
-            def as_sql(self, compiler, connection):
+            def as_sql(self, compiler, connection, **extra_context):
                 copy = self.copy()
-                copy.set_source_expressions(copy.get_source_expressions()[0:1] + [None])
-                return super(MyMax, copy).as_sql(compiler, connection)
+                # Most database backends do not support compiling multiple arguments on
+                # the Max aggregate, and that isn't what is being tested here anyway. To
+                # avoid errors, the extra argument is just dropped.
+                copy.set_source_expressions(
+                    copy.get_source_expressions()[0:1] + [None, None]
+                )
+
+                return super(MultiArgAgg, copy).as_sql(compiler, connection)
 
         with self.assertRaisesMessage(TypeError, "Complex aggregates require an alias"):
-            Book.objects.aggregate(MyMax("pages", "price"))
+            Book.objects.aggregate(MultiArgAgg("pages", "price"))
 
         with self.assertRaisesMessage(
             TypeError, "Complex annotations require an alias"
         ):
-            Book.objects.annotate(MyMax("pages", "price"))
+            Book.objects.annotate(MultiArgAgg("pages", "price"))
 
-        Book.objects.aggregate(max_field=MyMax("pages", "price"))
+        Book.objects.aggregate(max_field=MultiArgAgg("pages", "price"))
 
     def test_add_implementation(self):
         class MySum(Sum):
@@ -1318,6 +1350,8 @@ class AggregateTestCase(TestCase):
                 "function": self.function.lower(),
                 "expressions": sql,
                 "distinct": "",
+                "filter": "",
+                "order_by": "",
             }
             substitutions.update(self.extra)
             return self.template % substitutions, params
@@ -1351,7 +1385,13 @@ class AggregateTestCase(TestCase):
 
         # test overriding all parts of the template
         def be_evil(self, compiler, connection):
-            substitutions = {"function": "MAX", "expressions": "2", "distinct": ""}
+            substitutions = {
+                "function": "MAX",
+                "expressions": "2",
+                "distinct": "",
+                "filter": "",
+                "order_by": "",
+            }
             substitutions.update(self.extra)
             return self.template % substitutions, ()
 
@@ -1779,10 +1819,12 @@ class AggregateTestCase(TestCase):
                 Publisher.objects.none().aggregate(
                     sum_awards=Sum("num_awards"),
                     books_count=Count("book"),
+                    all_names=StringAgg("name", Value(",")),
                 ),
                 {
                     "sum_awards": None,
                     "books_count": 0,
+                    "all_names": None,
                 },
             )
         # Expression without empty_result_set_value forces queries to be
@@ -1873,6 +1915,12 @@ class AggregateTestCase(TestCase):
                     value=Aggregate("age", default=Value(5) * Value(7)),
                 )
                 self.assertEqual(result["value"], 35)
+
+    def test_stringagg_default_value(self):
+        result = Author.objects.filter(age__gt=100).aggregate(
+            value=StringAgg("name", delimiter=Value(";"), default=Value("<empty>")),
+        )
+        self.assertEqual(result["value"], "<empty>")
 
     def test_aggregation_default_group_by(self):
         qs = (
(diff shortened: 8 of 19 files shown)


camuthig:
@charettes I've made some progress on the suggestions. I have the three phases broken out into unique commits and have added some of the documentation and copied the tests from the postgres module into a new class that should run against all of the database backends.

I ran into two issues in the process that I wanted to get your opinion on.

1. Python is compiled against Sqlite3 version 3.37.0, which doesn't have support for the order by yet. So the ordering tests aren't running for that. I think I would need to compile Python from source and point it to a specific version of Sqlite3 to get past this, but I'm not sure if you have any thoughts.
2. Sqlite3 doesn't behave well when you combine `distinct` and a delimiter value. It throws an error of `SQLite doesn't support DISTINCT on aggregate functions accepting multiple arguments.` So it treats the delimiter as a second argument. In Sqlite if you don't provide a delimiter it will use a comma as a default. I was trying to build it such that if the user sets the delimiter to a comma and is using Sqlite as a backend, then the delimiter is ignored. However, I could figure out how to make that work. I tried a few things around custom SQL, using a class attribute instead of making it part of `expressions`, etc. but I couldn't quite get it right. If you can see a good way to treat that, I think it would be awesome to make distinct work in Sqlite for at least some scenarios, if not all.

As a side note, I named it `StringAgg` because that is what I am used to, and it feels like the more descriptive function name between that a `GROUP_CONCAT`, but I'm open to changing that to whatever you think is best. I guess would could always have an alias class `class GroupConcat(StringAgg): pass` if we really wanted to make it smooth for developers from either environment, but that doesn't seem to pass the smell test on "one way to do something"

charettes:
Hello @camuthig, thanks for spinning this up!

> Python is compiled against Sqlite3 version 3.37.0, which doesn't have support for the order by yet. So the ordering tests aren't running for that. I think I would need to compile Python from source and point it to a specific version of Sqlite3 to get past this, but I'm not sure if you have any thoughts.

It's possible CI might not be be setup here but if you're a *nix setup [you can use `LD_PRELOAD` to point at any SQLite version](https://til.simonwillison.net/sqlite/ld-preload#user-content-building-the-library). You don't have to build from source as [the SQLite project provides pre-built binaries](https://sqlite.org/download.html).

> Sqlite3 doesn't behave well when you combine distinct and a delimiter value. It throws an error of SQLite doesn't support DISTINCT on aggregate functions accepting multiple arguments. So it treats the delimiter as a second argument. In Sqlite if you don't provide a delimiter it will use a comma as a default. I was trying to build it such that if the user sets the delimiter to a comma and is using Sqlite as a backend, then the delimiter is ignored. However, I could figure out how to make that work. I tried a few things around custom SQL, using a class attribute instead of making it part of expressions, etc. but I couldn't quite get it right. If you can see a good way to treat that, I think it would be awesome to make distinct work in Sqlite for at least some scenarios, if not all.

Well [that's odd for sure](https://www.sqlite.org/lang_aggfunc.html#group_concat).

> In any aggregate function that takes a single argument, that argument can be preceded by the keyword DISTINCT. In such cases, duplicate elements are filtered before being passed into the aggregate function. For example, the function "count(distinct X)" will return the number of distinct values of column X instead of the total number of non-null values in column X. 

Usually when we run into these edge cases we [add a database feature](https://github.com/django/django/blob/dfd63ff43408e7901cc214b0482a7f844244d439/django/db/backends/base/features.py#L5) (e.g. `supports_aggregate_distinct_multiple_argument`) and use for two purpose.

1. Adjust `Aggregate.as_sql` to raise an error when `self.distinct and not connection.features.supports_aggregate_distinct_multiple_argument and len(super().get_expressions()) > 1`
2. Add tests for `@skipUnlessDBFeature("supports_aggregate_distinct_multiple_argument")` that cover the backends that support it and `@skipIfDBFeature` that makes sure the proper exception is raised.

> As a side note, I named it StringAgg because that is what I am used to, and it feels like the more descriptive function name between that a GROUP_CONCAT, but I'm open to changing that to whatever you think is best. 

I think `StringAgg` is fine as that's the name most backends use (Postgres, SQLite, SQLServer)

django/contrib/postgres/aggregates/general.py
```diff
@@ -49,17 +51,15 @@ class BoolOr(Aggregate):
...
-    output_field = TextField()
-
+class StringAgg(_StringAgg):
     def __init__(self, expression, delimiter, **extra):
-        delimiter_expr = Value(str(delimiter))
-        super().__init__(expression, delimiter_expr, **extra)
+        warnings.warn(
+            "The PostgreSQL specific StringAgg function is deprecated. Use "
+            "django.db.models.aggregate.StringAgg instead.",
+            category=RemovedInDjango60Warning,
```
charettes, line 63: We might have to adjust the `stacklevel` here so the warning points at the location of `StringAgg` instantiation. I think `stacklevel=1` is the right one?

django/contrib/postgres/aggregates/mixins.py
```diff
@@ -1,47 +1,24 @@
...
+        warnings.warn(
+            "The OrderableAggMixin is deprecated. Use Aggregate and allow_order_by "
+            "instead.",
+            category=RemovedInDjango60Warning,
+        )
+
+        if ordering:
+            warnings.warn(
+                "The ordering argument is deprecated. Use order_by instead.",
+                category=RemovedInDjango60Warning,
```
charettes, line 20: I think this will need `stacklevel` as well.

django/contrib/postgres/aggregates/mixins.py
```diff
@@ -1,47 +1,24 @@
...
-
-        source_params = []
-        for source_expr in source_exprs:
-            source_params += compiler.compile(source_expr)[1]
+    def __init__(self, *expressions, ordering=(), **extra):
+        warnings.warn(
+            "The OrderableAggMixin is deprecated. Use Aggregate and allow_order_by "
+            "instead.",
+            category=RemovedInDjango60Warning,
+        )
```
charettes, line 15: I wonder if we should move this [to `__init_subclass__` instead](https://docs.python.org/3/reference/datamodel.html#object.__init_subclass__) and do something like

```python
class _DeprecatedOrdering:
    def __init__(self, *expressions, ordering=(), **extra):
        if ordering:
            warnings.warn(
                "The ordering argument is deprecated. Use order_by instead.",
                category=RemovedInDjango60Warning,
            )

class OrderableAggMixin(_DeprecatedOrdering):
    allow_order_by = True

    def __init_subclass__(cls, /, *args, **kwargs):
        warnings.warn(
            "OrderableAggMixin is deprecated. Use Aggregate and allow_order_by "
            "instead.",
            category=RemovedInDjango60Warning,
        )
        super().__init_subclass__(*args, **kwargs)
```

And then in `postgres.aggregates` subclass `_DeprecatedOrdering` instead and explicitly set `allow_order_by = True` as well.

That would make any subclassing of `OrderableAggMixin` emit the deprecation warning while allowing `postgres.aggregates` to benefit from the deprecation period.

django/db/models/aggregates.py
```diff
@@ -196,6 +245,52 @@ def _get_repr_options(self):
...
+class StringAgg(Aggregate):
+    template = "%(function)s(%(distinct)s%(expressions)s%(order_by)s)%(filter)s"
+    function = "STRING_AGG"
+    name = "StringAgg"
+    allow_distinct = True
+    allow_order_by = True
+    output_field = TextField()
+
+    def __init__(self, expression, delimiter, **extra):
+        delimiter = Value(str(delimiter))
```
charettes, line 257: We'll want to force the delimiter to be specified as a `Value`. This current form disallows using another expression as the delimiter such as a column reference.

django/db/models/aggregates.py
```diff
@@ -196,6 +245,52 @@ def _get_repr_options(self):
...
+        super().__init__(expression, delimiter, **extra)
+
+    def as_sql(self, compiler, connection, **extra_context):
+        return super().as_sql(compiler, connection, **extra_context)
```
charettes, line 261: I think this one can go?

django/db/models/aggregates.py
```diff
@@ -196,6 +245,52 @@ def _get_repr_options(self):
...
+            connection,
+            function="GROUP_CONCAT",
+            **extra_context,
+        )
+
+    def as_sqlite(self, compiler, connection, **extra_context):
+        return self.as_sql(
+            compiler,
+            connection,
+            function="GROUP_CONCAT",
```
charettes, line 289: I _think_ [we can use `STRING_AGG` on SQLite as well](https://www.sqlite.org/lang_aggfunc.html#group_concat)?

tests/aggregation/tests.py
```diff
@@ -2158,6 +2177,18 @@ def test_having_with_no_group_by(self):
         )
         self.assertEqual(list(author_qs), [337])
 
+    def test_string_agg(self):
+        values = Publisher.objects.aggregate(
+            stringagg=StringAgg("name", delimiter=";")
```
charettes, line 2182:
```suggestion
            stringagg=StringAgg("name", delimiter=Value(";"))
```

It'd be great to add testing for expression delimiters as well!

tests/aggregation/tests.py
```diff
@@ -2359,3 +2390,256 @@ def test_aggregate_combined_queries(self):
...
+class StringAggTestModel(Model):
+    char_field = CharField(max_length=30, blank=True)
+    text_field = TextField(blank=True)
+    json_field = JSONField(null=True)
+
+
+class StatTestModel(Model):
+    int1 = IntegerField()
+    int2 = IntegerField()
+    related_field = ForeignKey(StringAggTestModel, SET_NULL, null=True)
```
charettes, line 2404: Try reuse the existing models defined in `aggregation/models.py` instead. We can't add new tables/models for each new features otherwise the suite slowly gets slower.

tests/expressions_window/tests.py
```diff
@@ -1706,14 +1706,14 @@ def test_fail_insert(self):
...
-            "Sum(Value(10000), order_by=OrderBy(F(pk), descending=False)) OVER ()"
+            "Sum(Value(10000)) OVER ()"
         )
         with self.assertRaisesMessage(FieldError, msg):
             Employee.objects.create(
                 name="Jameson",
                 department="Management",
                 hire_date=datetime.date(2007, 7, 1),
-                salary=Window(expression=Sum(Value(10000), order_by=F("pk").asc())),
+                salary=Window(expression=Sum(Value(10000))),
```
charettes, line 1716: Can you remind me why these changes are necessary?

tests/aggregation/tests.py
```diff
@@ -2359,3 +2390,256 @@ def test_aggregate_combined_queries(self):
...
+            StringAggTestModel.objects.filter(id__in=Subquery(subquery)),
+            [self.aggs[0]],
+        )
+
+    def test_string_agg_distinct(self):
+        StringAggTestModel.objects.all().delete()
+
+        StringAggTestModel.objects.create(char_field="Foo")
+        StringAggTestModel.objects.create(char_field="Foo")
+        StringAggTestModel.objects.create(char_field="Bar")
```
charettes, line 2626: Could use `bulk_create` here.

tests/aggregation/tests.py
```diff
@@ -2359,3 +2390,256 @@ def test_aggregate_combined_queries(self):
...
+
+        for distinct_value, expected_foo, expected_bar in [
+            (True, 1, 1),
+            (False, 2, 1),
+        ]:
+            with self.subTest(aggregate=f"distinct_{distinct_value}"):
+                # WIP I should brew install sqlite3 from my camuthig account in order to get
+                #   a more recent version that supports ordering group_concat
+                # WIP I'm getting the below, but only one argument is provided...
+                # SQLite doesn't support DISTINCT on aggregate functions accepting multiple arguments.
```
charettes, line 2636: Left some pointers about how to best approach this problem.

tests/aggregation/tests.py
```diff
@@ -2359,3 +2390,256 @@ def test_aggregate_combined_queries(self):
...
+            )
+            .filter(
+                char_field__in=["Foo1", "Foo2"],
+            )
+            .order_by("char_field")
+            .values_list("char_field", "agg")
+        )
+        self.assertEqual(list(values), expected_result)
+
+    def test_filter_in_subquery_with_exclude(self):
```
charettes, line 2604: I appreciate the testing but does this cover any particular edge case in subqueries?

tests/aggregation/tests.py
```diff
@@ -2359,3 +2390,256 @@ def test_aggregate_combined_queries(self):
...
+        with self.assertNumQueries(1):
+            values = StringAggTestModel.objects.aggregate(
+                aggregation=aggregation,
+            )
+            self.assertEqual(values, {"aggregation": None})
+
+    def test_default_argument(self):
+        StringAggTestModel.objects.all().delete()
+        tests = [
+            (StringAgg("char_field", delimiter=";", default="<empty>"), "<empty>"),
```
charettes, line 2454: I would expect this to crash as `default` must be passed as `Value`? Passing `str` to `Func` should normally attempt resolve to field references (e.g. `"<empty>" -> F("<empty>")") and crash if missing.

charettes (commented):
Thanks for the patch @camuthig!

I left a few comments and pointers. My main points of feedbacks are

1. `delimiter` should not be wrapped in a `Value` as that prevent references to columns and normally string arguments are considered to be `F`'ield references.
2. I appreciate the attention to testing but we should focus them on `StringAgg` behaviour and `order_by` support. Did you notice anything special when subqueries were involved?

django/db/models/aggregates.py
camuthig: I expected that as well, but I was getting errors that the function wasn't defined. It turns out, `STRING_AGG` was added in 3.44.0 at the same time as the `order_by` was added to the behaviors. 

https://www.sqlite.org/changes.html#version_3_44_0

Since my local sqlite versions have been 3.37, it was failing. Not sure if we have a preference on which to use for sqlite in this case. `GROUP_CONCAT` feels like it is a bit more flexible since it is older, but I don't know if the behaviors are exactly the same or not under the hood. I'll have to figure that out once I get a newer version of sqlite hooked up to my tests.

tests/aggregation/tests.py
camuthig: These tests are taken directly from the PostgreSQL specific of these tests. My thoughts were

1. We are deprecating that class in 6.0, so we will delete all of the StringAgg tests from that module at that time
2. Those tests exist for a reason, even if I am not familiar with them, so I didn't want to drop any and cause a regression

This is why I have some tests that may seem very specific, like the subqueries, and why I went about defining these models. It kept me from re-implementing the same behaviors in a way that might cause a regression. However, I am open to using the models already defined in `aggregation/models.py`, I'll just need to sit with the data a bit to come up with the right expectations. I think it might be best to keep the tests, like the subqueries, in place to make sure we maintain the coverage. If you think those behaviors are covered in other tests, though, then I'm open to cleaning them up.

charettes: That makes sense, if they already exist we should definitely just port them over.

> However, I am open to using the models already defined in aggregation/models.py

If you could port them while adapting them to use the existing models that would be awesome.

django/db/models/aggregates.py
charettes: I missed that, let's stick to `GROUP_CONCAT` then. I guess we could add a `connection.get_database_version() < (3, 44)` check so eventually when we drop support for < 3.44 we'll drop that code at the same time.

tests/aggregation/tests.py
charettes: Also, if we're porting them over we should delete the original ones from `postgres_tests` given `postgres.aggregates.StringAgg` will only a shim at this point.

tests/expressions_window/tests.py
camuthig: I need to remind myself as well. I don't exactly remember writing this line. I'll refresh my memory and get back to you.

tests/aggregation/tests.py
camuthig: > delimiter should not be wrapped in a Value as that prevent references to columns and normally string arguments are considered to be `F`'ield references.

The `Value` wrapper on a string comes right from the old implementation in the Postgres-specific version. Would it make more sense to allow for the string to come through and wrap it as a value for 5.2 with a deprecation warning in 6? That does break the ability to use a string as an alias for `F` for 5.2, though.

camuthig: I think I have something together that wraps the `__init__` from the Postgres version of the `StringAgg` class and does the `Value(str(delimiter))` and logs a deprecation warning. The main `StringAgg` can then expect an `Expression` of some sort.

charettes: > I think I have something together that wraps the __init__ from the Postgres version of the StringAgg class and does the Value(str(delimiter)) and logs a deprecation warning. The main StringAgg can then expect an Expression of some sort.

That's perfect, using the deprecation period for `postgres.StringAgg` to `models.StringAgg` seems like a great way to normalize this behavior 🎉

django/contrib/postgres/aggregates/general.py
```diff
@@ -49,17 +51,15 @@ class BoolOr(Aggregate):
...
+        )
+        super().__init__(expression, delimiter, **extra)
```
charettes, line 65: @camuthig to make sure I understand, this is where you would add the logic to pass `Value(delimiter_expr)` to `super()` with a deprecation warning?

camuthig: @charettes I just pushed a change with this logic. Yes, this is where I was planning to do it.

tests/aggregation/tests.py
camuthig: I did some work to merge the tests into the existing `AggregateTestCase`, keeping some of the more esoteric ones specific to the `StringAgg` behaviors but making them work with the existing data/models. I had to add a JSON field to the existing models to that end, but otherwise, I think it works pretty well.

tests/expressions_window/tests.py
camuthig: @charettes Oh, when I started validating `allow_order_by` in `Aggregate.__init__`, this test started throwing an error because `Sum` does not actually support order by. So I chose to remove the order by.

camuthig:
@charettes I think I have updates for all of your comments at this phase. I was able to build a custom python environment with pyenv and brew using guidance [here](https://github.com/pyenv/pyenv/issues/333#issuecomment-234707676) and test the group concat and ordering behaviors with a more recent version of sqlite, so that is cool.

I know there are some conflicts and I can work on cleaning those up. I'm also not totally sure what is wrong with the docs. I ran the same command on my own machine, and everything seems to build properly.

tests/aggregation/tests.py
```diff
@@ -2158,6 +2199,155 @@ def test_having_with_no_group_by(self):
...
+        )
+
+        expected_values = {
+            "stringagg": "Practical Django Projects;"
+                         "Python Web Development with Django;Paradigms of Artificial "
+                         "Intelligence Programming: Case Studies in Common Lisp",
+        }
+        self.assertEqual(values, expected_values)
+
+    @skipUnlessDBFeature("supports_aggregate_order_by_clause")
```
charettes, line 2264:
```suggestion
    @skipUnlessDBFeature("supports_json_field", "supports_aggregate_order_by_clause")
```

tests/aggregation/tests.py
```diff
@@ -2158,6 +2199,155 @@ def test_having_with_no_group_by(self):
...
+            ),
+        )
+        self.assertEqual(values, {"stringagg": "en,fr"})
+
+    def test_string_agg_filter_in_subquery(self):
+        """
+        This test is based on tests taken from existing PostgreSQL specific tests and
+        kept to avoid regressions as StringAgg is ported to the shared database module.
+        """
+
```
charettes, line 2281: Good to point out in the PR review tab but they don't need to live in the tests themselves.

tests/aggregation/tests.py
```diff
@@ -2158,6 +2199,155 @@ def test_having_with_no_group_by(self):
...
+            .annotate(agg=aggregate)
+            .values("agg")
+        )
+        values = list(
+            Book.objects.annotate(
+                agg=Subquery(subquery),
+            )
+            .exclude(agg__isnull=True)
+            # Different engines treat null STRING_AGG differently, so excluding it for
+            # consistency.
```
charettes, line 2300: Could you elaborate on that?

I seems that SQLite, Postgres, MySQL, and Oracle all return `NULL` which should translate to `None`?

1. [SQLite](https://dbfiddle.uk/fzC4JzJm)
2. [MySQL](https://dbfiddle.uk/jf5-JeSq)
3. [Postgres](https://dbfiddle.uk/62y1WY4y)
4. [Oracle](https://dbfiddle.uk/e0gP_WKJ)

Did you possibly run into issues because of a lack of ordering on the returned value and the fact some backends default to `NULLS LAST` instead of `NULLS FIRST`?

If that's the case you should use `assertQuerySetEqual(..., ordered=False)` or specify an `order_by(OrderBy('agg', nulls_last=True))`.

I think it'd be good to include at least one tests that ensures that `None` is returned on an empty not-NULL set.

tests/aggregation/tests.py
```diff
@@ -2158,6 +2199,155 @@ def test_having_with_no_group_by(self):
...
+        ]
+
+        self.assertEqual(expected_values, values)
+
+    @skipUnlessDBFeature("supports_aggregate_order_by_clause")
+    def test_order_by_in_subquery(self):
+        """
+        This test is based on tests taken from existing PostgreSQL specific tests and
+        kept to avoid regressions as StringAgg is ported to the shared database module.
+        """
```
charettes, line 2319:
```suggestion
```

tests/expressions_window/tests.py
charettes: That makes sense!

tests/postgres_tests/test_aggregates.py
```diff
@@ -151,12 +145,21 @@ def test_default_argument(self):
                     )
                     self.assertEqual(values, {"aggregation": expected_result})
 
+    def test_ordering_warns_of_deprecation(self):
+        msg = "The ordering argument is deprecated. Use order_by instead."
+        with self.assertWarnsMessage(RemovedInDjango60Warning, msg):
+            values = AggregateTestModel.objects.aggregate(
+                arrayagg=ArrayAgg("integer_field", ordering=F("integer_field").desc())
+            )
+            self.assertEqual(values, {"arrayagg": [2, 1, 0, 0]})
```
charettes, line 154: We'll need also need tests for the creation of `postgres.StringAgg`.

django/contrib/postgres/aggregates/mixins.py
```diff
@@ -1,9 +1,19 @@
...
+
 from django.core.exceptions import FullResultSet
 from django.db.models.expressions import OrderByList
+from django.utils.deprecation import RemovedInDjango60Warning
 
 
 class OrderableAggMixin:
-    def __init__(self, *expressions, ordering=(), **extra):
+    def __init__(self, *expressions, ordering=(), order_by=(), **extra):
+        if ordering:
```
charettes, line 9: I think we should raise a `TypeError` if both are specified.

charettes (commented):
This is looking great, I left a few comments regarding null handling and a few other things.

I like that you started to breakdown the commits I may suggest a different order and naming scheme.

> Refs #35444 -- Deprecated contrib.postgres aggregates ordering for order_by.
>
> (Explain the reasoning to better align with Window.order_by and plans to move support for Aggregate)

> Fixed #35444 -- Add generic support for Aggregate.order_by.
>
> (Explain that this now supersedes the contrib.postgres support and introduces StringAgg to exercise this support)

^^ this commit should make `OrderableAggMixin` unused.

> Refs #35444 -- Deprecated contrib.postgres.OrderableAggMixin
>
> (Explain that it serves no purposes now that Aggregate supports order_by but it is kept around for a deprecation cycle as some users might be relying on it)

tests/postgres_tests/test_aggregates.py
charettes, line 154: The following should ensure that the proper `stacklevel` is passed.

```suggestion
        with self.assertWarnsMessage(RemovedInDjango60Warning, msg) ctx:
            values = AggregateTestModel.objects.aggregate(
                arrayagg=ArrayAgg("integer_field", ordering=F("integer_field").desc())
            )
            self.assertEqual(values, {"arrayagg": [2, 1, 0, 0]})
        self.assertEqual(ctx.filename, __file__)
```

tests/aggregation/tests.py
camuthig: You are right, I was reading the test results wrong. I went with `assertQuerySetEqual(..., ordered=False)` to resolve this.

camuthig:
@charettes I have update the code to hit all of your comments and get the linters passing. I will take a dive into the failing tests shortly. At least some of them appear to be because we have moved the `connection.features.supports_aggregate_filter_clause` into the `AggregateFilter.as_sql` function, but have historically allowed for this check to fail and convert it to a `WHEN` statement. Now, though, when converting to the `WHEN`, we are throwing the `NotSupportedError` again. If you have thoughts on how you would like to see that flow, I'm open to suggestions.

charettes:
> At least some of them appear to be because we have moved the connection.features.supports_aggregate_filter_clause into the AggregateFilter.as_sql function, but have historically allowed for this check to fail and convert it to a WHEN statement. Now, though, when converting to the WHEN, we are throwing the NotSupportedError again. If you have thoughts on how you would like to see that flow, I'm open to suggestions.

The following should do

```diff
diff --git a/django/db/models/aggregates.py b/django/db/models/aggregates.py
index 6cf0bd9a60..d070944039 100644
--- a/django/db/models/aggregates.py
+++ b/django/db/models/aggregates.py
@@ -29,6 +29,10 @@ class AggregateFilter(Func):
     arity = 1
     template = " FILTER (WHERE %(expressions)s)"

+    @property
+    def condition(self):
+        return self.source_expressions[0]
+
     def as_sql(self, compiler, connection, **extra_context):
         if not connection.features.supports_aggregate_filter_clause:
             raise NotSupportedError
@@ -187,7 +191,7 @@ def as_sql(self, compiler, connection, **extra_context):
                 copy = self.copy()
                 copy.filter = None
                 source_expressions = copy.get_source_expressions()
-                condition = When(self.filter, then=source_expressions[0])
+                condition = When(self.filter.condition, then=source_expressions[0])
                 copy.set_source_expressions([Case(condition)] + source_expressions[1:])
                 return copy.as_sql(compiler, connection, **extra_context)
```

The problem was that we were building the fallback `When` with the `AggregateFilter` instead of its underlying `condition: Q`.

camuthig:
@charettes thanks for the recommendation. It works great. 

I spent time digging into the other failures in MySQL and determined that the MySQL implementation of `GROUP_CONCAT` doesn't actually play nicely with any of the other engine implementations. I have some code together that passed the aggregate tests on both MySQL and Postgres locally, but it is certainly messy and probably brittle.

First, it allows for multiple expressions, which we can ignore for now. The bigger issues are that MySQL doesn't support filtering on aggregates and it uses a different format order between the delimiter and ordering.

Sqlite: `GROUP_CONCAT(expression, delimiter ORDER BY order_by)`
MySQL: `GROUP_CONCAT(expressions* ORDER BY order_by SEPARATOR delimiter)`

So here, if you want to introduce a delimiter you MUST have a different template, which is solvable. The bigger issue is that the order matters between `ORDER BY` and `SEPARATOR`: `SEPARATOR delimiter ORDER BY order_by` is invalid syntax. This required me to manually build the parameters in `as_sql` and force the order of things to get the parameters into the right order depending on the engine.

This was further compounded by the fact MySQL doesn't support aggregate filtering and we are forcing a `CASE` statement. So if you have `filter`, `order_by`, and `delimiter` you need a third ordering of parameters, with the filter parameters coming before even the expressions. 😞 

Do you think there are some level of feature flags we can throw on MySQL to make this easier? Or maybe allow for engine-specific default separators to be ignored in our queries? I have a commit with some experimentation here: https://github.com/django/django/pull/18361/commits/34c8fe08cd15c62a5e74f1308babe01e0bd9dac0

charettes:
> I have a commit with some experimentation here: https://github.com/django/django/commit/cbe50126aa80ca8b5d79c48578aca1d29a673288

Thanks for pushing a commit demonstrating the scope of the problem.

Looking at it my immediate reaction would be to avoid over complicating `StringAgg.as_sql` and favor encapsulating the logic entirely in `as_mysql` instead. In there you'll know that the `FILTER` clause is not usable and will likely be able to simplify things quite a bit from some local testing locally.

> Or maybe allow for engine-specific default separators to be ignored in our queries?

I'm not sure I'm understanding what you mean here.

Now that the MySQL tests are passing I'll give Oracle a test run as well.

charettes:
buildbot, test on oracle.

camuthig:
@charettes do you have recommendations on how to test the Oracle engine myself? I am having trouble with the Oracle Docker images. Maybe cause I am on a M1 Mac? I'm not sure. It looks like a pretty common errors, so it seems some change I have made at a base level is causing the compiled queries to be incorrect but I can't really tell why without running against the database.

charettes:
@camuthig I usually try two approaches on my M1

1. I had success once I ran `brew install colima` and tried again
2. Low-tech approach is to attempt to generate the raw SQL locally and rely on tools such as [dbfiddle](https://dbfiddle.uk/Wh4zpMKJ).

I really wish CI ran with `--debug-sql` at least so we could peak at the generated SQL

Any suggestions @felixxm?

charettes:
I managed to get tests running on Oracle here's the non-localized error message

```
ERROR: test_annotate_values_list (aggregation.tests.AggregateTestCase)
  File "/django/source/django/db/backends/utils.py", line 105, in _execute
  File "/django/source/django/db/backends/oracle/base.py", line 577, in execute
    return self.cursor.execute(query, self._param_generator(params))
  File "/usr/local/lib/python3.10/site-packages/oracledb/cursor.py", line 710, in execute
    impl.execute(self)
  File "src/oracledb/impl/thin/cursor.pyx", line 196, in oracledb.thin_impl.ThinCursorImpl.execute
  File "src/oracledb/impl/thin/protocol.pyx", line 440, in oracledb.thin_impl.Protocol._process_single_message
  File "src/oracledb/impl/thin/protocol.pyx", line 441, in oracledb.thin_impl.Protocol._process_single_message
  File "src/oracledb/impl/thin/protocol.pyx", line 433, in oracledb.thin_impl.Protocol._process_message
  File "src/oracledb/impl/thin/messages.pyx", line 74, in oracledb.thin_impl.Message._check_and_raise_exception
oracledb.exceptions.DatabaseError: ORA-22848: cannot use NCLOB type as comparison key
Help: https://docs.oracle.com/error-help/db/ora-22848/

  File "/django/source/tests/aggregation/tests.py", line 939, in test_annotate_values_list
    self.assertEqual(list(books), [(self.b1.id, "159059725", 34.5)])
  File "/django/source/django/db/models/query.py", line 381, in __iter__
    self._fetch_all()
  File "/django/source/django/db/models/query.py", line 1909, in _fetch_all
    self._result_cache = list(self._iterable_class(self))
  File "/django/source/django/db/models/query.py", line 229, in __iter__
    return compiler.results_iter(
  File "/django/source/django/db/models/sql/compiler.py", line 1536, in results_iter
    results = self.execute_sql(
  File "/django/source/django/db/models/sql/compiler.py", line 1585, in execute_sql
    cursor.execute(sql, params)
  File "/django/source/django/db/backends/utils.py", line 122, in execute
    return super().execute(sql, params)
  File "/django/source/django/db/backends/utils.py", line 79, in execute
  File "/django/source/django/db/backends/utils.py", line 92, in _execute_with_wrappers
  File "/django/source/django/db/backends/utils.py", line 100, in _execute
  File "/django/source/django/db/utils.py", line 91, in __exit__
  File "/django/source/django/db/backends/utils.py", line 105, in _execute
  File "/django/source/django/db/backends/oracle/base.py", line 577, in execute
    return self.cursor.execute(query, self._param_generator(params))
  File "/usr/local/lib/python3.10/site-packages/oracledb/cursor.py", line 710, in execute
    impl.execute(self)
  File "src/oracledb/impl/thin/cursor.pyx", line 196, in oracledb.thin_impl.ThinCursorImpl.execute
  File "src/oracledb/impl/thin/protocol.pyx", line 440, in oracledb.thin_impl.Protocol._process_single_message
  File "src/oracledb/impl/thin/protocol.pyx", line 441, in oracledb.thin_impl.Protocol._process_single_message
  File "src/oracledb/impl/thin/protocol.pyx", line 433, in oracledb.thin_impl.Protocol._process_message
  File "src/oracledb/impl/thin/messages.pyx", line 74, in oracledb.thin_impl.Message._check_and_raise_exception
django.db.utils.DatabaseError: ORA-22848: cannot use NCLOB type as comparison key
Help: https://docs.oracle.com/error-help/db/ora-22848/
```

tests/aggregation/models.py
```diff
@@ -30,6 +30,7 @@ class Book(models.Model):
     contact = models.ForeignKey(Author, models.CASCADE, related_name="book_contact_set")
     publisher = models.ForeignKey(Publisher, models.CASCADE)
     pubdate = models.DateField()
+    print_info = models.JSONField(null=True, default=None)
```
charettes, line 33: @camuthig this is the source of all your Oracle test failures...

Django uses a `NCLOB` column to persists `JSONField` and Oracle doesn't allow to `GROUP BY` a `NCLOB` field.

I don't think the usage of a `JSONField` is mandatory to test this feature so I'd suggest using a different field type instead.

tests/aggregation/tests.py
```diff
@@ -2178,6 +2215,147 @@ def test_having_with_no_group_by(self):
...
+        }
+        self.assertEqual(values, expected_values)
+
+    @skipUnlessDBFeature("supports_json_field", "supports_aggregate_order_by_clause")
+    def test_string_agg_jsonfield_order_by(self):
+        values = Book.objects.aggregate(
+            stringagg=StringAgg(
+                KeyTextTransform("lang", "print_info"),
+                delimiter=Value(","),
+                order_by=KeyTextTransform("lang", "print_info"),
```
charettes, line 2287: I guess this was ported from the Postgres tests.

I suggest creating a distinct model for this particular case instead.

charettes (commented):
Oracle failures were due to your addition of the `print_info` field.

felixxm:
> @camuthig I usually try two approaches on my M1
> 
>     1. I had success once I ran `brew install colima` and tried again
> 
>     2. Low-tech approach is to attempt to generate the raw SQL locally and rely on tools such as [dbfiddle](https://dbfiddle.uk/Wh4zpMKJ).
> 
> 
> I really wish CI ran with `--debug-sql` at least so we could peak at the generated SQL
> 
> Any suggestions @felixxm?

The "easiest" way is to use [VM provided by Oracle](https://www.oracle.com/database/technologies/databaseappdev-vm.html).

camuthig:
@charettes I have updated the code with two things:

1. Swapped the JSON test behaviors to a new, JSON specific model to avoid adverse effects on other tests.
2. Changed up how I dealt with MySQL to make it a bit more simple, and generally focused on just MySQL.

I haven't had a chance to set up the Oracle VM, so I don't know that the tests are going to pass on that backend. Can we run it again, and if issues continue, then I will get that VM configured.

I kept the MySQL changes as a separate commit to make them easier to review at this phase: https://github.com/django/django/pull/18361/commits/43415289e723fbcb7fc772f4c17718f7b72d30e1 . Let me know what you think of that. If we are happy with it, I will squash it back down.

felixxm:
buildbot, test on oracle.

charettes:
@camuthig that's awesome, thanks for the summary!

Sure we can give the Oracle suite a run and see how it goes. I have the VM setup locally now so I should be able to provide guidance as well if we run into issues.

charettes:
buildbot, test on oracle.

camuthig:
@charettes I still can't get oracle running locally. Not sure what is wrong with my system but the VM keeps crashing on me. I do believe I saw an error in the template for the Oracle SQL that was causing the failures on the order by queries and fixed it. Question for you: if I add the buildbot message myself, will it kick off the Oracle tests? And is it okay if I do that as I iterate? I'm not sure how that buildbot works and what is okay versus an over step.

charettes:
@camuthig I believe it should yes or that we can add you to an allow list otherwise. Could you trying `buildbot, test on oracle.`

camuthig:
buildbot, test on oracle.

camuthig:
Looks like the buildbot responded @charettes ! Thanks.

charettes:
One reported failure with the associated query

```
FAIL: test_string_agg_filter_in_subquery (aggregation.tests.AggregateTestCase)
  File "/django/source/tests/aggregation/tests.py", line 2329, in test_string_agg_filter_in_subquery
    self.assertQuerySetEqual(expected_values, values, ordered=False)
  File "/django/source/django/test/testcases.py", line 1250, in assertQuerySetEqual
    return self.assertDictEqual(Counter(items), Counter(values), msg=msg)
AssertionError: Count[40 chars] 1, 'Paul Bissex;Wesley J. Chun': 1, 'Peter No[48 chars]: 1}) != Count[40 chars] 1, '': 1, 'Paul Bissex;Wesley J. Chun': 1, 'P[46 chars]: 1})
  Counter({'Adrian Holovaty': 1,
           'Brad Dayley': 1,
+          '': 1,
           'Paul Bissex;Wesley J. Chun': 1,
           'Peter Norvig;Stuart Russell': 1,
-          'Peter Norvig': 1,
?                           ^

+          'Peter Norvig': 1})
?                           ^^

-          None: 1})

(0.021)
SELECT
    (
        SELECT
            LISTAGG (
                CASE WHEN NOT (U2."NAME" LIKE TRANSLATE(J % USING NCHAR_CS) ESCAPE TRANSLATE('\' USING NCHAR_CS)
                    AND U2."NAME" IS NOT NULL) THEN
                            U2."NAME"
                ELSE
                    NULL
                        END,;

) AS "AGG"
FROM
    "AGGREGATION_BOOK" U0
    LEFT OUTER JOIN "AGGREGATION_BOOK_AUTHORS" U1 ON (U0."ID" = U1."BOOK_ID")
    LEFT OUTER JOIN "AGGREGATION_AUTHOR" U2 ON (U1."AUTHOR_ID" = U2."ID")
WHERE
    U0."ID" = ("AGGREGATION_BOOK"."ID")
GROUP BY
    U0."ID",
    U0."ISBN",
    U0."NAME",
    U0."PAGES",
    U0."RATING",
    U0."PRICE",
    U0."CONTACT_ID",
    U0."PUBLISHER_ID",
    U0."PUBDATE") AS "AGG"
FROM
    "AGGREGATION_BOOK";

 args=('J%', ';'); ALIAS=DEFAULT
```

charettes:
I think we'll have no choice but not to support `StringAgg(filter)` on Oracle (we could raise `NotSupportedError` in `as_oracle` given `self.filter`) because `LISTAGG` doesn't offer a way to filter out values and Oracle doesn't support the `FILTER (WHERE ...)` clause which forces the usage of `CASE WHEN` which yields `NULL` on mismatches.

As another PR ran into (https://github.com/django/django/pull/18351#pullrequestreview-2287172954) in order to support this properly aggregates would need to support being turned into subquery aggregates (https://github.com/django/django/pull/11841) which is way beyond the scope of this PR.

I'm not sure what's causing the GIS failures but the error message is

```
django.db.utils.DatabaseError: ORA-00907: missing right parenthesis
Help: https://docs.oracle.com/error-help/db/ora-00907/
```

And the generated SQL for `test_geoagg_subquery` is 

```
SELECT 
  SDO_UTIL.TO_WKBGEOMETRY(
    SDO_AGGR_UNION() SDOAGGRTYPE("__COL1", NULL, % s)
  ) 
FROM 
  (
    SELECT 
      DISTINCT "GEOAPP_CITY"."ID" AS "COL1", 
      "GEOAPP_CITY"."NAME" AS "COL2", 
      "GEOAPP_CITY"."POINT" AS "COL3", 
      "GEOAPP_CITY"."POINT" AS "__COL1" 
    FROM 
      "GEOAPP_CITY" 
    WHERE 
      SDO_INSIDE(
        "GEOAPP_CITY"."POINT", 
        SDO_GEOMETRY(% s, 4326)
      ) = 'TRUE'
  ) subquery
```

on your branch while it's

```sql
SELECT 
  SDO_UTIL.TO_WKBGEOMETRY(
    SDO_AGGR_UNION() SDOAGGRTYPE("__COL1", NULL, % s)
  ) 
FROM 
  (
    SELECT 
      DISTINCT "GEOAPP_CITY"."ID" AS "COL1", 
      "GEOAPP_CITY"."NAME" AS "COL2", 
      "GEOAPP_CITY"."POINT" AS "COL3", 
      "GEOAPP_CITY"."POINT" AS "__COL1" 
    FROM 
      "GEOAPP_CITY" 
    WHERE 
      SDO_INSIDE(
        "GEOAPP_CITY"."POINT", 
        SDO_GEOMETRY(MULTIPOLYGON(%s), 4326)
      ) = 'TRUE'
  ) subquery
```

on main

I highly suspect `GeoAggregate.as_oracle` needs adjustments with regards to `self.filter`.

charettes:
@camuthig I think the following should address the Oracle GIS failures

```diff
diff --git a/django/contrib/gis/db/models/aggregates.py b/django/contrib/gis/db/models/aggregates.py
index abff1a5206..e526a8d3ef 100644
--- a/django/contrib/gis/db/models/aggregates.py
+++ b/django/contrib/gis/db/models/aggregates.py
@@ -33,15 +33,14 @@ def as_oracle(self, compiler, connection, **extra_context):
         if not self.is_extent:
             tolerance = self.extra.get("tolerance") or getattr(self, "tolerance", 0.05)
             clone = self.copy()
-            source_expressions = self.get_source_expressions()
-            source_expressions.pop()  # Don't wrap filters with SDOAGGRTYPE().
+            *source_expressions, filter_expr, order_by_expr = self.get_source_expressions()
             spatial_type_expr = Func(
                 *source_expressions,
                 Value(tolerance),
                 function="SDOAGGRTYPE",
                 output_field=self.output_field,
             )
-            source_expressions = [spatial_type_expr, self.filter]
+            source_expressions = [spatial_type_expr, filter_expr, order_by_expr]
             clone.set_source_expressions(source_expressions)
             return clone.as_sql(compiler, connection, **extra_context)
         return self.as_sql(compiler, connection, **extra_context)
```

camuthig:
@charettes I pulled in your change and will run it against the Oracle here. I really appreciate your help on that one, I wasn't going to be able to triage that one quickly.

Regarding the `test_string_agg_filter_in_subquery` the challenge here is that Oracle is treating these case clauses differently than, for example, MySQL. MySQL is treating those values as `None` and we are getting the same results for MySQL as PostgreSQL. Oracle appears to be casting `null` to a string and getting `''` instead. I'm good with blocking the filter behavior for any backends that don't support filter natively. I think that check would be pretty straightforward to add in there and would cover MySQL and Oracle.

charettes:
> Oracle appears to be casting null to a string and getting '' instead. 

@camuthig the reason for this is this is that [for Oracle](https://github.com/django/django/blob/f4813211e2d8017b56b7447f56ad17df3fae9aa3/django/db/backends/oracle/features.py#L16) `NULL` and `""` (empty string) are the same thing which means that it doesn't provide facilities to differentiate between the two (it can't).

Thinking more about it I think that in order to account for that we should make the following adjustments

```diff
diff --git a/tests/aggregation/tests.py b/tests/aggregation/tests.py
index a23ae47487..fd535b1867 100644
--- a/tests/aggregation/tests.py
+++ b/tests/aggregation/tests.py
@@ -2323,7 +2323,7 @@ def test_string_agg_filter_in_subquery(self):
             "Paul Bissex;Wesley J. Chun",
             "Peter Norvig;Stuart Russell",
             "Peter Norvig",
-            None,
+            "" if connection.features.interprets_empty_strings_as_nulls else None,
         ]

         self.assertQuerySetEqual(expected_values, values, ordered=False)
```

Instead of completely disabling filtering or Oracle. The values are filtered, Oracle just doesn't offer a way to differentiate the kind of _emptyness_ its dealing with.

camuthig:
@charettes that is really interesting to know about Oracle. I like that strategy, so I've put it in and will get it all run. Hopefully we'll go green on this round!

charettes:
buildbot, test on oracle.

camuthig:
@charettes any thoughts on the one server side binding postgres error? It doesn't reproduce for me locally, and I'm not sure how the changes in this branch could have started impacting update/create behaviors in transactions at this phase. Is there chance it was a flakey failure or something?

django/db/models/aggregates.py
```diff
@@ -196,6 +256,93 @@ def _get_repr_options(self):
...
+    def as_oracle(self, compiler, connection, **extra_context):
+        if self.order_by:
+            template = (
+                "%(function)s(%(distinct)s%(expressions)s) WITHIN GROUP (%(order_by)s)"
+                "%(filter)s"
+            )
+        else:
+            template = "%(function)s(%(distinct)s%(expressions)s)%(filter)s"
+
+        return super().as_sql(
```
charettes, line 299: I think that both `super().as_sql` and `self.as_sql` resolve to the same since `StringAgg` doesn't override `as_sql`?

```suggestion
        return self.as_sql(
```

django/db/models/aggregates.py
```diff
@@ -196,6 +256,93 @@ def _get_repr_options(self):
...
+    def as_sqlite(self, compiler, connection, **extra_context):
+        if connection.get_database_version() < (3, 44):
+            return self.as_sql(
+                compiler,
+                connection,
+                function="GROUP_CONCAT",
+                **extra_context,
+            )
+
+        return super().as_sql(compiler, connection, **extra_context)
```
charettes, line 343:
```suggestion
        return self.as_sql(compiler, connection, **extra_context)
```

charettes (commented):
@camuthig I think it can be safely ignored yes it seems a flaky failure due to CI setup which means we've got all the tests passing 🎉 

So at this point the only parts missing are

1. Fold commits 4,5.
2. Document the addition of `Aggregate(order_by)` and `allow_order_by` [here](https://github.com/django/django/blob/16af0c60ed90813ed70ecef4e64a89301ed0f0a2/docs/ref/models/expressions.txt#L453) and add a release note about it [here](https://github.com/django/django/blob/16af0c60ed90813ed70ecef4e64a89301ed0f0a2/docs/releases/5.2.txt#L208-L224)
3. Document `StringAgg` [here](https://github.com/django/django/blob/16af0c60ed90813ed70ecef4e64a89301ed0f0a2/docs/ref/models/querysets.txt#L4047) and add a 5.2 release note about it as well.

camuthig:
buildbot, test on oracle.

camuthig:
@charettes is there anything I need to do to pull this forward in the review process? I know it was DjangoCon last week, so people were probably super busy. I just want to make sure I'm not forgetting part of the process.

camuthig:
Thanks @nessita ! I think I have everything configured properly to get it into the queue now.

django/db/models/aggregates.py
```diff
@@ -115,42 +166,51 @@ def get_group_by_cols(self):
...
-                        **extra_context,
-                    )
-                    return sql, (*params, *filter_params)
-            else:
+        if (
+            self.distinct
+            and not connection.features.supports_aggregate_distinct_multiple_argument
+            and len(super().get_source_expressions()) > 1
+        ):
+            raise NotSupportedError
```
charettes, line 174: Minor but I think we'll want to raise a more explicit message here as it might be user facing


```suggestion
            raise NotSupportedError(
                f"{self.name} doesn't support `distinct` with multiple expressions on {connection.vendor}"
            )
```

We should also add a `@skipIfDBFeature("supports_aggregate_distinct_multiple_argument")` test for it that `assertRaisesMessage`.

charettes (commented):
The patch is in a pretty good state @camuthig thanks for your efforts here. I left some comments about testing the lack of support for multiple expression distinct on SQLite but otherwise it seems ready for an eventual final review.

You could also change the PR title to match the commit message format (e.g. `Fixed #35444 -- Added generic support for Aggregate.order_by`) but that's very minor.

camuthig:
@nessita I wanted to follow up on this pull request. I was traveling for a few weeks, so it fell of the radar to check in on it. I believe the requested changes are all there, and I can clean up any doc conflicts as needed.

I don't know the 5.2 release schedule, but do you believe we can get this in there or is it going to be better to wait for 6.0?

django/contrib/gis/db/models/aggregates.py
```diff
@@ -33,15 +33,14 @@ def as_oracle(self, compiler, connection, **extra_context):
         if not self.is_extent:
             tolerance = self.extra.get("tolerance") or getattr(self, "tolerance", 0.05)
             clone = self.copy()
-            source_expressions = self.get_source_expressions()
-            source_expressions.pop()  # Don't wrap filters with SDOAGGRTYPE().
+            *source_expr, filter_expr, order_by_expr = self.get_source_expressions()
             spatial_type_expr = Func(
-                *source_expressions,
+                *source_expr,
```
charettes, line 38: Since it might contain more than one it'd switch to `source_exprs`.

charettes (approved):
> Am I reading you latest comment correctly in that this has an implicit/soft approval from your end?

@nessita that's a fair assessment I think this path is ready for a final review!

camuthig:
@charettes hopefully just one last commit based on failing tests from the composite PK changes (which I am _really_ excited about) that were brought in once I rebased from main.

The new `ColPairs` type used in aggregates, like `Count`, was essentially taking the filter that was originally passed into the aggregate class, which was then wrapped in `AggregateFilter`, and wrapping it again. So we have like `AggregateFilter(AggregateFilter(stuff))`. This results in the template generating twice and creating bad SQL.

https://github.com/django/django/pull/18361/commits/9027c537c6523c388a8ad9aa32b881c04d6760e5

I made some changes in the above commit to allow for receiving an instance of `AggregateFilter` and avoid the extra wrapping. I'm open for your thoughts on how this change is written to make them match the project. I'm hoping this will have the tests passing and once we get it write, I'll rebase it into the clean commits.

charettes:
@camuthig thanks for rebasing and looking into the composite primary key support conflicts, we're all excited about this feature landing!

I think that a more adequate solution would be

```diff
diff --git a/django/db/models/aggregates.py b/django/db/models/aggregates.py
index a9fbd86753..0efc17d57b 100644
--- a/django/db/models/aggregates.py
+++ b/django/db/models/aggregates.py
@@ -37,13 +37,6 @@ class AggregateFilter(Func):
     arity = 1
     template = " FILTER (WHERE %(expressions)s)"

-    @classmethod
-    def from_param(cls, param):
-        if isinstance(param, AggregateFilter):
-            return param
-        else:
-            return cls(param)
-
     def as_sql(self, compiler, connection, **extra_context):
         if not connection.features.supports_aggregate_filter_clause:
             raise NotSupportedError
@@ -96,7 +89,7 @@ def __init__(
             raise TypeError(f"{self.__class__.__name__} does not allow default.")

         self.distinct = distinct
-        self.filter = filter and AggregateFilter.from_param(filter)
+        self.filter = filter and AggregateFilter(filter)
         self.order_by = order_by
         self.default = default
         self.order_by = order_by and AggregateOrderBy.from_param(
@@ -254,17 +247,16 @@ def __init__(self, expression, filter=None, **extra):

     def resolve_expression(self, *args, **kwargs):
         result = super().resolve_expression(*args, **kwargs)
-        expr = result.source_expressions[0]
+        source_expressions = result.get_source_expressions()

         # In case of composite primary keys, count the first column.
-        if isinstance(expr, ColPairs):
+        if isinstance(expr := source_expressions[0], ColPairs):
             if self.distinct:
                 raise NotSupportedError(
                     "COUNT(DISTINCT) doesn't support composite primary keys"
                 )
-
-            cols = expr.get_cols()
-            return Count(cols[0], filter=result.filter)
+            source_expressions[0] = expr.get_cols()[0]
+            result.set_source_expressions(source_expressions)

         return result
```

which can be included in the `Fixed #35444 -- Added generic support for Aggregate.order_by.` commit

camuthig:
@charettes I appreciate the input. I've got it cleaned up and everything looks good to go now.

charettes:
Thanks for seeing this through @camuthig, the PR should be ready for a final review when Django fellows get to it. It might possible get in the 5.2 release window but if it's not the case it will surely make it in 6.0 given the advanced state of the work.

If that's something that means something to do don't hesitate to add yourself to the `AUTHORS` file, you're more than deserved your name in there!

django/db/backends/base/features.py
```diff
@@ -257,6 +257,15 @@ class BaseDatabaseFeatures:
...
     supports_aggregate_filter_clause = False
 
+    # Does the database support ORDER BY in aggregate expressions?
+    supports_aggregate_order_by_clause = False
+
+    # Does the database backend support DISTINCT when using multiple arguments in an
+    # aggregate expression? For example, Sqlite treats the "delimiter" argument of
+    # STRING_AGG/GROUP_CONCAT as an extra argument and does not allow using a custom
+    # delimiter along with DISTINCT.
+    supports_aggregate_distinct_multiple_argument = True
```
sarahboyce, line 267: I think these need release notes

docs/releases/5.2.txt
```diff
@@ -466,3 +471,11 @@ Miscellaneous
 * The fallback to ``request.user`` when ``user`` is ``None`` in
   ``django.contrib.auth.login()`` and ``django.contrib.auth.alogin()`` will be
   removed.
```
sarahboyce, line 473: The depredations need to be added to `docs/internals/deprecation.txt`

sarahboyce (commented):
Thank you for this work @camuthig :star: 

I've given an initial review mostly around the docs
Can you also resolve the conflicts?

django/db/backends/base/features.py
sarahboyce: I think I'm wrong, ignore this!

docs/releases/5.2.txt
camuthig: I have added some documentation around the items to be removed. Please let me know if it meets the needs/standards.

camuthig:
@sarahboyce I believe the requested tests are all added and the documentation should be good. The failing test doesn't mean anything to me. My guess is it is unrelated to this PR, but let me know if there is something else to work on.

sarahboyce (commented):
I have merged the first commit https://github.com/django/django/commit/d734f1651ccc0a74325f7b55f7eecc68edef6453
Resolved conflicts and forced pushed updates here

I need to do a review of the deprecation to check I understand what we're ending up with and hopefully the tests are still passing after the conflict resolution :crossed_fingers: 
Note that the feature freeze is Jan 15th, still optimistic for this one but there's a chance that this will have to be in 6.0

django/contrib/postgres/aggregates/mixins.py
```diff
@@ -1,15 +1,10 @@
 import warnings
```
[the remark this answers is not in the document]
charettes: So something like?

```suggestion
# RemovedInDjango61Warning: When the deprecation ends, remove completely.
import warnings
```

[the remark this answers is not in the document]
sarahboyce: Yes :+1: assuming I understand the plan correctly :grin:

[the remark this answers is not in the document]
charettes: yeah I think that adds up, one the deprecation period ends there will be no need for specialized Postgres aggregate mixins.

django/db/models/aggregates.py
```diff
@@ -18,42 +27,92 @@
...
 
+class AggregateFilter(Func):
+    arity = 1
+    template = " FILTER (WHERE %(expressions)s)"
+
+    def as_sql(self, compiler, connection, **extra_context):
+        if not connection.features.supports_aggregate_filter_clause:
+            raise NotSupportedError(
+                "Aggregate filter clauses are not supported on this database backend."
+            )
```
sarahboyce, line 44: I think this is missing a test

docs/ref/models/expressions.txt
```diff
@@ -493,6 +502,11 @@ The ``filter`` argument takes a :class:`Q object <django.db.models.Q>` that's
 used to filter the rows that are aggregated. See :ref:`conditional-aggregation`
 and :ref:`filtering-on-annotations` for example usage.
 
+The ``order_by`` argument takes a field name (with an optional ``"-"`` prefix
+which indicates descending order) or an expression (or a tuple or list of
+strings and/or expressions) that specifies the ordering of the elements in the
+result.
```
sarahboyce, line 508: I couldn't spot tests around a tuple/list of fields - can you check/add tests for these?

I also wonder if we can say that this is similar to what you can input to `order_by()`

sarahboyce:
buildbot, test on oracle.

docs/releases/5.2.txt
```diff
@@ -334,6 +334,15 @@ Models
...
   :ref:`composite primary keys <cpk-and-database-functions>`.
 
+* The new ``order_by`` argument for :class:`~django.db.models.Aggregate` allows
+  to specify the ordering of the elements in the result.
+
+* The new :attr:`.Aggregate.allow_order_by` class attribute determines whether
+  the aggregate function allows passing an ``order_by`` keyword argument.
+
+* The new :class:`~django.db.models.StringAgg` aggregate returns the input
+  values concatenated into a string, separated by the ``delimiter`` string.
```
sarahboyce, line 344: I wonder if we should mention that this used to only be supported on Postgres

sarahboyce, line 344: I wonder if we should mention that this aggregate used to only be supported on Postgres

sarahboyce (commented):
Pushed a couple of docs tweaks and some more deprecation comments

django/db/models/aggregates.py
camuthig: @sarahboyce we are actually catching this exception and using it in the `Aggregate.as_sql`, which is why there isn't a specific test case for it.

https://github.com/django/django/pull/18361/files#diff-c7d87b6e6b973bf87e5685442a6c1da9531c24bcaa86fe158c40ea0276986f32R204

charettes: Right, we used this approach as it makes the management of source expressions for backends that don't support filtered aggregates easier to reason about.

docs/ref/models/expressions.txt
charettes: > I also wonder if we can say that this is similar to what you can input to order_by()

It should effectively accept similar inputs.

django/db/models/aggregates.py
sarahboyce: Ah thank you!

charettes:
@camuthig thanks for your continued efforts here. The changes unfortunately didn't make it in time for 5.2 feature freeze so you'll need to repoint at 6.0 on top of the already provided comments. Let me know if I can be of any help in pushing this forward.

tests/postgres_tests/test_aggregates.py
```diff
@@ -158,8 +154,9 @@ def test_ordering_warns_of_deprecation(self):
             self.assertEqual(values, {"arrayagg": [2, 1, 0, 0]})
         self.assertEqual(ctx.filename, __file__)
 
+    # RemovedInDjango61Warning: Remove this test
     def test_ordering_and_order_by_causes_error(self):
-        with self.assertWarns(RemovedInDjango61Warning):
+        with warnings.catch_warnings(record=True, action="always") as wm:
```
camuthig, line 159: Things got a little weird here when I update the warnings to use 7.0. The issue being that interacting with the class this way triggers both a 6.1 warning (for the `order_by` usage) and 7.0 warnings (for using the PostgreSQL StringAgg). I couldn't get `assertWarns` to work with both getting thrown so dug in and tried using the underlying behaviors of it to get it down.

Let me know if there is a better way to write this, and I will refactor it.

docs/ref/models/expressions.txt
camuthig: I have updated the documentation to include something about this. Let me know if you would like any changes.

docs/releases/5.2.txt
camuthig: I have updated the documentation to include something about this. Let me know if you would like any changes. (also moved to 6.0.txt)

tests/postgres_tests/test_aggregates.py
charettes: I think the way you did it is great, it's a peculiar case given it triggers two distinct warnings of different categories 👍

camuthig:
I think everything should be up to date and marked as deprecated as of 6.0 and to be removed in 7.0. I updated the documentation as requested as well.

I ran into testing issues around the layers of deprecation warnings. I have a working solution for it, but I'm not sure it is the right pattern.

sarahboyce:
buildbot, test on oracle.

charettes:
I should be able to have a look at it shortly. Looks at the test failures I assume this is happening because some aggregate is stashing `tuple` instances in `source_expressions` which shouldn't happen (only `Resolvable | None` should be allowed).

The `is None` is important as otherwise if a resolvable with a `__bool__` causing side-effect is passed it will be evaluated. The classic example is `QuerySet` which evaluates its query.

django/db/models/aggregates.py
```diff
@@ -18,42 +27,92 @@
...
             raise TypeError(f"{self.__class__.__name__} does not allow default.")
+
         self.distinct = distinct
-        self.filter = filter
+        self.filter = filter and AggregateFilter(filter)
+        self.order_by = order_by
         self.default = default
+        self.order_by = order_by and AggregateOrderBy.from_param(
+            f"{self.__class__.__name__}.order_by", order_by
+        )
```
charettes, line 102: Something when wrong with the rebase here, we assign to `self.order_by` twice?

charettes:
@sarahboyce I haven't figured out in which commit these should be folded but here are the required changes to get the tests passing

```diff
diff --git a/django/contrib/postgres/aggregates/mixins.py b/django/contrib/postgres/aggregates/mixins.py
index a6849c3930..8d3e40177b 100644
--- a/django/contrib/postgres/aggregates/mixins.py
+++ b/django/contrib/postgres/aggregates/mixins.py
@@ -16,7 +16,7 @@ def __init__(self, *expressions, ordering=(), order_by=(), **extra):
             if order_by:
                 raise TypeError("Cannot specify both order_by and ordering.")
             order_by = ordering
-
+        order_by = order_by or None
         super().__init__(*expressions, order_by=order_by, **extra)


diff --git a/django/db/models/aggregates.py b/django/db/models/aggregates.py
index 8b644e3599..48e00aa946 100644
--- a/django/db/models/aggregates.py
+++ b/django/db/models/aggregates.py
@@ -95,7 +95,6 @@ def __init__(

         self.distinct = distinct
         self.filter = filter and AggregateFilter(filter)
-        self.order_by = order_by
         self.default = default
         self.order_by = order_by and AggregateOrderBy.from_param(
             f"{self.__class__.__name__}.order_by", order_by
```

charettes:
We might want to also consider having `OrderByList.from_params` return `None` when passed an empty `tuple | list` to prevent similar crashes when `Aggregate(order_by)` is used directly instead.

```patch
index 8b644e3599..a791593508 100644
@@ -95,9 +95,8 @@ def __init__(

         self.distinct = distinct
         self.filter = filter and AggregateFilter(filter)
-        self.order_by = order_by
         self.default = default
-        self.order_by = order_by and AggregateOrderBy.from_param(
+        self.order_by = AggregateOrderBy.from_param(
             f"{self.__class__.__name__}.order_by", order_by
         )
diff --git a/django/db/models/expressions.py b/django/db/models/expressions.py
index 68e5d2667e..ced5557732 100644
@@ -1466,7 +1466,11 @@ def __init__(self, *expressions, **extra):

     @classmethod
     def from_param(cls, context, param):
+        if param is None:
+            return None
         if isinstance(param, (list, tuple)):
+            if not param:
+                return None
             return cls(*param)
         elif isinstance(param, str) or hasattr(param, "resolve_expression"):
             return cls(param)
@@ -1937,8 +1941,7 @@ def __init__(
                 self.partition_by = (self.partition_by,)
             self.partition_by = ExpressionList(*self.partition_by)

-        if self.order_by is not None:
-            self.order_by = OrderByList.from_param("Window.order_by", self.order_by)
+        self.order_by = OrderByList.from_param("Window.order_by", self.order_by)
         super().__init__(output_field=output_field)
         self.source_expression = self._parse_expressions(expression)[0]
```

sarahboyce:
buildbot, test on oracle.

sarahboyce:
buildbot, test on oracle.

sarahboyce (approved):
Thank you! Great job on this :star:

charettes:
Thank you for sticking around and seeing this through @camuthig, that was quite the 10 months adventure. Hopefully it was a valuable experience for you!

camuthig:
Woo! We did it! 

Thanks for all the help along the way, @charettes . It's definitely been a good experience for me. I learned a lot about how the ORM is working behind the scenes, which was my own goal. Now that we have this part done, I look forward to moving some of the other Postgres-specific functions over to general support. At least `ArrayAgg` but maybe something of the JSON agg too.

tests/aggregation/tests.py
```diff
@@ -2202,6 +2250,167 @@ def test_aggregate_arity(self):
...
+                filter=Q(name__startswith="P"),
+            )
+        )
+
+        expected_values = {
+            "stringagg": "Practical Django Projects;"
+            "Python Web Development with Django;Paradigms of Artificial "
+            "Intelligence Programming: Case Studies in Common Lisp",
+        }
+        self.assertEqual(values, expected_values)
```
timgraham, line 2328: As far as I see, `STRING_AGG()` doesn't define an ordering in the absence of an ORDER BY, so many of these assertions could be relaxed regarding the order in which the values are concatenated with ";". Did I miss something?

charettes: @timgraham you're right, when ordering is not explicitly specified it is undefined so the test should be adjusted to compare the `;` split set of values.

timgraham: Thanks. I've proposed https://github.com/django/django/pull/21536.

PR #18362, opened 2024-07-11
Fixed #35594 -- Added unique nulls distinct validation for expressions.

charettes:
Thanks Mark Gensler for the report.

# Trac ticket number

ticket-35594

### DIFF
--- a/django/db/models/constraints.py
+++ b/django/db/models/constraints.py
@@ -653,19 +653,16 @@ class UniqueConstraint(BaseConstraint):
             queryset = queryset.exclude(pk=model_class_pk)
         if not self.condition:
             if queryset.exists():
-                if self.expressions:
+                if self.fields:
+                    # When fields are defined, use the unique_error_message() for
+                    # backward compatibility.
                     raise ValidationError(
-                        self.get_violation_error_message(),
-                        code=self.violation_error_code,
+                        instance.unique_error_message(model, self.fields),
                     )
-                # When fields are defined, use the unique_error_message() for
-                # backward compatibility.
-                for model, constraints in instance.get_constraints():
-                    for constraint in constraints:
-                        if constraint is self:
-                            raise ValidationError(
-                                instance.unique_error_message(model, self.fields),
-                            )
+                raise ValidationError(
+                    self.get_violation_error_message(),
+                    code=self.violation_error_code,
+                )
         else:
             against = instance._get_field_value_map(meta=model._meta, exclude=exclude)
             try:
--- a/django/db/models/constraints.py
+++ b/django/db/models/constraints.py
@@ -8,7 +8,7 @@ from django.db import connections
 from django.db.models.constants import LOOKUP_SEP
 from django.db.models.expressions import Exists, ExpressionList, F, RawSQL
 from django.db.models.indexes import IndexExpression
-from django.db.models.lookups import Exact
+from django.db.models.lookups import Exact, IsNull
 from django.db.models.query_utils import Q
 from django.db.models.sql.query import Query
 from django.db.utils import DEFAULT_DB_ALIAS
@@ -642,12 +642,16 @@ class UniqueConstraint(BaseConstraint):
                     meta=model._meta, exclude=exclude
                 ).items()
             }
-            expressions = []
+            filters = []
             for expr in self.expressions:
                 if hasattr(expr, "get_expression_for_validation"):
                     expr = expr.get_expression_for_validation()
-                expressions.append(Exact(expr, expr.replace_expressions(replacements)))
-            queryset = queryset.filter(*expressions)
+                rhs = expr.replace_expressions(replacements)
+                condition = Exact(expr, rhs)
+                if self.nulls_distinct is False:
+                    condition = Q(condition) | Q(IsNull(expr, True), IsNull(rhs, True))
+                filters.append(condition)
+            queryset = queryset.filter(*filters)
         model_class_pk = instance._get_pk_val(model._meta)
         if not instance._state.adding and model_class_pk is not None:
             queryset = queryset.exclude(pk=model_class_pk)
--- a/docs/releases/5.0.8.txt
+++ b/docs/releases/5.0.8.txt
@@ -9,4 +9,5 @@ Django 5.0.8 fixes several bugs in 5.0.7.
 Bugfixes
 ========
 
-* ...
+* Implemented missing validation for ``UniqueConstraint(nulls_distinct=False)``
+  when using ``*expressions`` (:ticket:`35594`).
--- a/tests/constraints/tests.py
+++ b/tests/constraints/tests.py
@@ -896,6 +896,13 @@ class UniqueConstraintTests(TestCase):
                 ChildUniqueConstraintProduct(name=self.p1.name, color=self.p1.color),
             )
 
+    def test_validate_fields_unattached(self):
+        Product.objects.create(price=42)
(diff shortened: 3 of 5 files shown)


django/db/models/constraints.py
```diff
@@ -666,6 +664,10 @@ def validate(self, model, instance, exclude=None, using=DEFAULT_DB_ALIAS):
                             raise ValidationError(
                                 instance.unique_error_message(model, self.fields),
                             )
+                raise ValidationError(
+                    self.get_violation_error_message(),
+                    code=self.violation_error_code,
+                )
```
charettes, line 670: This change was necessary to allow non-conditional unique constraints over fields validation to be tested without actually creating the constraint. The fact that no `ValidationError` could be raised under some conditions was definitely overlooked.

That likely explains why the initial `nulls_distinct` tests were added to `tests/validation` instead of `tests/constraints` in the first place https://github.com/django/django/pull/17058/files#diff-adae729932de755cd1b0e4a4413203256a3b17864523cba7ea646ae6d9951c0bR222

sdolemelipone:
Nice, I was playing around with this at the same time. I came up with roughly the same fix which is satisfying :-)  Does the test need a `@skipUnlessDBFeature("supports_nulls_distinct_unique_constraints")`?

charettes:
> Does the test need a @skipUnlessDBFeature("supports_nulls_distinct_unique_constraints")?

We could add one for the sake of it but given the validation logic testing doesn't require the creation of the constraint I figured there would be no harm in not skipping it.

django/db/models/constraints.py
```diff
@@ -642,30 +642,34 @@ def validate(self, model, instance, exclude=None, using=DEFAULT_DB_ALIAS):
...
+            conditions = []
             for expr in self.expressions:
                 if hasattr(expr, "get_expression_for_validation"):
                     expr = expr.get_expression_for_validation()
-                expressions.append(Exact(expr, expr.replace_expressions(replacements)))
-            queryset = queryset.filter(*expressions)
+                lhs, rhs = expr, expr.replace_expressions(replacements)
+                condition = Exact(lhs, rhs)
+                if self.nulls_distinct is False:
+                    condition = Q(condition) | Q(IsNull(lhs, True), IsNull(rhs, True))
```
charettes, line 652: I guess we could eventually have a `NotDistinctFrom(Lookup)` lookup that would use `{lhs} IS NOT DISTINCT FROM {rhs}` on backend that supports it and emulate it otherwise (e.g. `<=>` [on MySQL](https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html#operator_equal-to), fallback to `OR ...`). This should be good for now though since this is our first requirement for an internal usage.

django/db/models/constraints.py
sdolemelipone: I was wondering why the tests were spread across `tests/validation` and `tests/constraints`. Once all these `UniqueConstraint` tickets are resolved, is there a need to move most of the tests to `tests/constraints`?

django/db/models/constraints.py
sdolemelipone: Yes that sounds good. 
Just to expand on what I was trying to communicate [here](https://github.com/django/django/pull/18356#issuecomment-2217977675), I was expecting to find some code that would have looked like
```python
...
if self.nulls_distinct is not False:
    condition = Q(condition) & Q(IsNull(lhs, False))
```
and was surprised that it isn't necessary during the unique check. But I see everything works correctly, this is just for my understanding.

sdolemelipone: After re-reading your comment on the other ticket I understand this a little better now, thanks.

sdolemelipone:
I see this is still marked as patch needs improvement - anything I can help with? I think the use of `IsNull` takes care of `interprets_empty_strings_as_nulls`, is a test needed for that too?

charettes:
@sdolemelipone sorry for the delay, I think the PR should be ready for review now.

> I think the use of IsNull takes care of interprets_empty_strings_as_nulls, is a test needed for that too?

Good thinking but I don't think it's necessary as we already have tests to ensure `IsNull` and `interprets_empty_strings_as_nulls` interopability and there's no feature support overlap between the two (`interprets_empty_strings_as_nulls` is an Oracle artifact and `supports_nulls_distinct_unique_constraints` is only enabled on Postgres).

charettes:
buildbot, test on oracle.

sdolemelipone:
> @sdolemelipone sorry for the delay, I think the PR should be ready for review now.
Great. I'll review later today/tomorrow, if that's required/useful. :-)

django/db/models/constraints.py
sdolemelipone: The `if not self.expressions:` clause looks a bit too clever. Is it reasonable to assume that the `model` passed in to `UniqueConstraint.validate(model, ...)` is the model class which is relevant for this constraint instance? If we refactor the code to remove the call to  `instance.get_constraints()` as below then no tests fail in the suite (other than a difference of error message in  `test_validate_fields_unattached`).

From what I can see `constraint.validate()` is always called with the correct `model` class by `Model.validate_constraints()`. Would the below not be clearer?
```python
...
if queryset.exists():
    if self.fields:
        raise ValidationError(
            instance.unique_error_message(model, self.fields),
        )
    else:
        raise ValidationError(
            self.get_violation_error_message(),
            code=self.violation_error_code,
        )
```
I'm happy to raise a separate PR if you think this one isn't the place... or happy to be taught about the edge case which I'm missing :-)

sdolemelipone (commented):
Just one suggestion / question in the comments, other than that all looks great.

charettes:
> Is it reasonable to assume that the model passed in to UniqueConstraint.validate(model, ...) is the model class which is relevant for this constraint instance?

I would say yes. Under non-tests circumstances constraints are always attached to models and during testing we can control that.

I like your approach better so I'll adjust the first commit to use this approach instead.

This just made me realize that we also likely want to respect `violation_error_message` and `violation_error_code` when they are specified over using the backward compatibility fallback. Surely if someone explicitly use this option they want to override the one default. That's a distinct problem though and likely don't need to be backported.

> or happy to be taught about the edge case which I'm missing :-)

The only edge case that I know of is multi-table inheritance but thinking about it I can't see how this could be a problem.

sdolemelipone:
> I like your approach better so I'll adjust the first commit to use this approach instead.

Great, I'm glad it was useful!

> The only edge case that I know of is multi-table inheritance but thinking about it I can't see how this could be a problem.

That was my thinking too.

> This just made me realize that we also likely want to respect `violation_error_message` and `violation_error_code` when they are specified over using the backward compatibility fallback. Surely if someone explicitly use this option they want to override the one default. That's a distinct problem though and likely don't need to be backported.

Yes, I can't see why a user would pass those in and not want them to be used.

Just the mariadb test failing, which I can't get to work locally either but that's probably my test settings, and I see it was being addressed in #18366. Apart from that all looks good.

sdolemelipone:
Also, when this is merged I will rebase #18356 and then can address `nulls_distinct=False` for `GeneratedField` in `*expressions` in the same PR.

sarahboyce (approved):
Thank you for this both of you ⭐ looks great
I couldn't push changes here (I want to reorder the commits as I probably won't backport 04507cdbc8a9deed612abd8b30e48173d5419c35 as it is a test clean-up commit) and so I have opened #18376 which I plan to merge shortly

sarahboyce:
This is merged thank you

charettes:
Thank you Sarah!

PR #18365, opened 2024-07-13
[DRAFT] #32577 autofield allow non ints, add uuid auto

tomwojcik:
# Trac ticket number
ticket-32577 

# Branch description
PostgreSQL has introduced support for UUIDv7 (see [PostgreSQL Commitfest #47/4388](https://commitfest.postgresql.org/47/4388/)). It's an important enhancement, because it seems to perform much better, so one can assume a potentially fast adoption. At the same time, developers wanted to use UUIDv7 that are generated automatically (that are not part of the "global mixin") for a long time.

Generating primary keys within the database is a best practice, and PostgreSQL stands out as the (_to be confirmed_) sole database supported by Django that enables UUID primary key generation natively (using `random_uuid`), without requiring additional extensions.

With the release of Django 5.0, support for db_default has been added, which allows us to combine it with the feature from the above. (https://docs.djangoproject.com/en/5.0/ref/models/fields/#db-default).

In this PR:
- postgres db-side uuidv4 field is added
- the variation of the above for primary keys is added
- refactored autofield -> autofieldmixin

todo: address commit messages requirement
todo: add docs

# Checklist


django/contrib/postgres/fields/uuid.py
```diff
@@ -0,0 +1,21 @@
...
+from django.db.models import UUIDField
+from django.db.models.fields import NOT_PROVIDED, AutoFieldMixin
+
+
+class UUID4Field(UUIDField):
+    def __init__(self, *args, **kwargs):
+        kwargs["db_default"] = RandomUUID()
+        kwargs["default"] = NOT_PROVIDED
+        super().__init__(*args, **kwargs)
+
```
tomwojcik, line 11: @ngnpope [mentioned](https://code.djangoproject.com/ticket/32577#comment:11) that 

```
    def deconstruct(self):
        name, path, args, kwargs = super().deconstruct()
        del kwargs['db_default']
        return name, path, args, kwargs
``` 

would also be needed, but I'm not sure. Can you advise?

django/contrib/postgres/functions.py
```diff
@@ -2,6 +2,13 @@
 
 
 class RandomUUID(Func):
+    """
+    Without `pgcrypto` or `uuid-ossp` extensions enabled, available only
+    since PostgreSQL 13.
+
+    https://www.postgresql.org/docs/13/functions-uuid.html
+    """
+
```
felixxm, line 11: Django 5.2 (currently under development) only supports PostgreSQL 14+, so this comment is unnecessary.

docs/ref/contrib/postgres/fields.txt
```diff
@@ -920,3 +920,56 @@ define :class:`~django.contrib.postgres.constraints.ExclusionConstraint`. See
...
+Django includes two PostgreSQL-specific UUID field types for working with UUIDs: :class:`~django.contrib.postgres.fields.UUID4Field` and :class:`~django.contrib.postgres.fields.UUID4AutoField`. These fields provide convenient ways to use UUID v4 values generated by the database, specifically targeting `PostgreSQL 13`_ and above. Using database-generated UUIDs can be beneficial for ensuring the uniqueness and integrity of data, especially in distributed systems.
+
+
+``UUID4Field``
+--------------
+
+.. class:: UUID4Field(**options)
+
+
+.. versionadded:: 6.0
```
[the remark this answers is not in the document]
felixxm: Use `5.2` for now and add release notes. Let's be optimistic.

docs/ref/contrib/postgres/fields.txt
```diff
@@ -920,3 +920,57 @@ define :class:`~django.contrib.postgres.constraints.ExclusionConstraint`. See
...
+
+The :class:`~django.contrib.postgres.fields.UUID4Field` class is a subclass of Django's :class:`~django.db.models.UUIDField` that uses PostgreSQL's native ``gen_random_uuid()`` function to generate UUID v4 values. This is done by setting the ``db_default`` parameter to :class:`~django.contrib.postgres.functions.RandomUUID`.
+
+This field is particularly useful when you want the database to handle the generation of UUIDs, decreasing the risk of collisions. It's worth mentioning that using database generated UUID v4 **does not guarantee uniqueness**, but `the risk of a collision is extremely low`_.
+
+
```
felixxm, line 942: Avoid unnecessary blank lines in docs
```suggestion

```

django/contrib/postgres/fields/uuid.py
```diff
@@ -0,0 +1,21 @@
+from django.contrib.postgres.functions import RandomUUID
+from django.db.models import UUIDField
+from django.db.models.fields import NOT_PROVIDED, AutoFieldMixin
+
+
+class UUID4Field(UUIDField):
+    def __init__(self, *args, **kwargs):
+        kwargs["db_default"] = RandomUUID()
```
felixxm, line 8: Why we only define it on PostgreSQL? It should be possible to define it on other dbs, e.g. MySQL.

felixxm (commented):
@tomwojcik Thanks :+1: This is a good start.

JohnDoeKaiina:
Le mar. 27 août 2024 à 07:52, Mariusz Felisiak ***@***.***> a
écrit :

django/contrib/postgres/functions.py
tomwojcik: https://github.com/django/django/pull/18365/commits/51f3d91cf72bc889ecff86fee0b7ba0f4084a826

docs/ref/contrib/postgres/fields.txt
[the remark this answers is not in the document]
tomwojcik: https://github.com/django/django/pull/18365/commits/a38f0c0df71d78fc603e17ef432dbd2225453c38

https://github.com/django/django/pull/18365/commits/7892226ccb4c4824425cd6a4b81c1b6d9887cc12

django/contrib/postgres/fields/uuid.py
tomwojcik: For me the goal of this ticket/PR is to allow to define custom auto primary keys, such as UUIDs, generated in the DB, used in tandem with `DEFAULT_AUTO_FIELD`. This PR allows to do just that and uuid4 auto for psql just proves that it works.

If there was an already existing `RandomUUID` equivalent for MySQL, I'd be happy to use it, but there's none (in the latest Django version), and implementing it is IMO out of scope for this PR.

Let me know what you think.

pauloxnet: It seems like a good approach to me.

django/contrib/postgres/fields/uuid.py
tomwojcik: https://github.com/django/django/pull/18365/commits/68421fb006b286c19f5e3e9fc3e0bca1018f2c96

tomwojcik:
Are there any plans to proceed with this PR? What are the next steps?

ulgens:
@tomwojcik On Trac, issue has `Patch needs improvement` flag set. Removing it should push it to the review queue.

tomwojcik:
> @tomwojcik On Trac, issue has `Patch needs improvement` flag set. Removing it should push it to the review queue.

I didn't know I have permissions to do that. Thanks!

browniebroke (commented):
I was exploring implementing this as 3rd party package and got blocked by the `AutoField` subclasses check - glad to see this being worked on in core Django.

django/contrib/postgres/fields/uuid.py
nessita: @felixxm What did you have in mind for an implementation in other DBs? As far as I could find, neither MySQL, Oracle, nor SQLite provide a way to generate a UUIDv4 at the database level. MySQL does have a UUID() function, but it returns a UUIDv1, not UUIDv4.

felixxm: For example, MariaDB 11.7+ has `UUID_v4()`, on SQLite we could define a user function with ``uuid.uuid4()``

nessita (changes requested):
Hello @tomwojcik, thank you for this work, and for your patience :star2: 

I have started the study of the ticket/feature, and I'll be doing rounds of reviews as often as possible. Some initial comments:

* Every docs and code comments need to be wrapped at 79cols.
* In the ticket, @ngnpope mentioned that some of the steps involve "_Replace isinstance(..., AutoField) checks with the new flag._" and "_Deprecate isinstance (and issubclass?) checks on AutoField within AutoFieldMeta_.". I don't see this included in this work, could you please share your view around that?
* There is a very recent comment in the ticket, from @charettes, that needs studying and analysis, as it proposes a change of direction to solve this feature in a simpler manner.

It may be useful to have a Django test project that uses the proposed solution to showcase how the new feature could be used. In any case I'll setup a local project to experiment with this and other approaches, so no worries if you can't provide that yet.

I'll set the ticket flag "patch needs improvement" until a clearer implementation design arises from the recent comments in the ticket.

Thank you again!

django/contrib/postgres/fields/uuid.py
```diff
@@ -0,0 +1,23 @@
...
+    def deconstruct(self):
+        name, path, args, kwargs = super().deconstruct()
+        del kwargs["db_default"]
+        return name, path, args, kwargs
+
+    def get_internal_type(self):
+        return "UUID4Field"
+
+
+class UUID4AutoField(AutoFieldMixin, UUID4Field):
```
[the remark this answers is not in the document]
browniebroke: I have the same question. It feels like this name might limit potential future extensibility (e.g. add support for more UUID versions, by passing the UUID version as parameter)? All the UUID classes aren't specifying the UUID version currently, so would be more consistent.

docs/releases/5.2.txt
```diff
@@ -297,6 +309,11 @@ Models
 * Added support for validation of model constraints which use a
   :class:`~django.db.models.GeneratedField`.
 
+* Introduced ``AutoFieldMixin` to decouple the automatic primary key generation
+  behavior from the underlying data type. This allows creating
+  auto-incrementing primary keys of any data type, such as UUID fields,
+  without being limited to integer-based primary keys with ``AutoField``.
```
browniebroke, line 315: I feel like this is a useful change on its own that would unlock other kinds of autofields as 3rd party package. As this PR has stalled a bit, is it worth extracting it?

ulgens: +1 for this. I can try to help with extracting it if help is needed.

django/contrib/postgres/fields/uuid.py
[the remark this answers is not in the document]
tomwojcik: When I opened this PR, Postgresql just merged uuid v7, so implementing that in Django for pk should be straighforward after the changes in this PR. If so, i'd be nice to distinguish between v4 and v7. 

On the other hand, I agree. Losing consistency may not be worth it. I'll update the naming in a moment.

docs/releases/5.2.txt
tomwojcik: I will align with upstream main in a moment.

django/contrib/postgres/fields/uuid.py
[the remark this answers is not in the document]
ulgens: +1 for UUID4. UUIDv7 is coming hot and defaulting to v4 doesn't seem ideal.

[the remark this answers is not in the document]
tomwojcik: Consistency with the past implementation is very important IMO. Postgresql created `gen_random_uuid` themselves, there's no v4 in the name of this function. 

On the other hand I already do

```
class UUID4Field(UUIDField):
```

so 

```
class UUID4AutoField(AutoFieldMixin, UUID4Field):
```

feels natural. Also, explicit is better than implicit is a thing. 

I don't feel attached to any of the names. For now I will keep `UUID4AutoField` as it has +1 from me as well, if my vote counts. Let me know if you have a different opinion.

tomwojcik:
Hi @nessita, thanks for taking the time to review this PR! I appreciate the feedback.

> Every docs and code comments need to be wrapped at 79cols.
 
Sorry, I thought linting will catch this. I fixed it in `fields.txt`. Hope that's all, please let me know if not. Thank you.

> In the ticket, @ngnpope mentioned that some of the steps involve "Replace isinstance(..., AutoField) checks with the new flag." and "Deprecate isinstance (and issubclass?) checks on AutoField within AutoFieldMeta.". I don't see this included in this work, could you please share your view around that?

That's correct, those specific refactoring steps identified by Nick aren't included in this PR. I focused primarily on addressing the core goal of Trac ticket #32577: enabling custom, database-generated primary keys, using UUID4AutoField as the key example.

To achieve this, I concentrated on the prerequisites Nick mentioned:

> We could lift the restriction to allow other non-AutoField types, but the concept of an automatically generated id field needs to be decoupled from AutoField a bit.

This was done by introducing AutoFieldMixin. It separates the behavior of an auto-generating PK from the underlying field type.

The AutoFieldMixin approach felt more flexible for the goal of enabling custom PKs beyond just UUIDs. As @browniebroke https://github.com/django/django/pull/18365#pullrequestreview-2663932268 mentioned, it'd be best to offload the less popular PKs to 3rd party packages. UUID4AutoField is here just to prove that it works. Also, it happens to be the most popular non-int pk.

Replacing the legacy isinstance(..., AutoField) checks across the entire codebase and subsequently deprecating AutoFieldMeta represents a much larger refactoring task. It has significant backward-compatibility implications and the person refactoring it needs to understand a lot of Django core that's invisible to everyday users. I don't feel up to the challenge.

> There is a very recent comment in the ticket, from @charettes, that needs studying and analysis, as it proposes a change of direction to solve this feature in a simpler manner.

Yes, I saw @charettes's [alternative approach](https://github.com/nessita/django/pull/24). It looks promising and simpler for enabling UUID PKs specifically. However, my understanding is that the AutoFieldMixin approach implemented here enables a broader feature set, including support for other automatically generated PK types.

Of course, I'm open to feedback if the Django team prefers a different direction or implementation strategy.

> It may be useful to have a Django test project that uses the proposed solution...

Absolutely, I can create a small example project demonstrating the use of UUID4AutoField if this implementation approach gets the green light. Just let me know.

docs/ref/contrib/postgres/fields.txt
```diff
@@ -920,3 +920,55 @@ define :class:`~django.contrib.postgres.constraints.ExclusionConstraint`. See
 docs/current/rangetypes.html#RANGETYPES-INCLUSIVITY>`_.
 
 .. _different bounds: https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-IO
+
+UUID Fields
+===========
+
+Django includes two PostgreSQL-specific UUID field types for working with UUIDs: :class:`~django.contrib.postgres.fields.UUID4Field` and :class:`~django.contrib.postgres.fields.UUID4AutoField`. These fields provide convenient ways to use UUID v4 values generated by the database, specifically targeting `PostgreSQL 13`_ and above. Using database-generated UUIDs can be beneficial for decreasing the chance of collisions in distributed systems.
```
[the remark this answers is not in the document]
tomwojcik: done

docs/ref/contrib/postgres/fields.txt
```diff
@@ -920,3 +920,55 @@ define :class:`~django.contrib.postgres.constraints.ExclusionConstraint`. See
...
+
+
+``UUID4Field``
+--------------
+
+.. class:: UUID4Field(**options)
+
+
+.. versionadded:: 5.2
```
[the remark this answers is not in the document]
tomwojcik: done

charettes:
@tomwojcik 

> It looks promising and simpler for enabling UUID PKs specifically. However, my understanding is that the AutoFieldMixin approach implemented here enables a broader feature set, including support for other automatically generated PK types.

Could you go into more details about which _other automatically generated PK types_ kind of fields are not expressible today as either `Field(default)` or `Field(db_default)` and would require `AutoFieldMixin`?

AFAIK the only difference between `AutoField` derivative today and other db-generated fields is that on backends that don't support the `RETURNING` clause on `INSERT` (that is only MySQL) the [PEP-0249](https://peps.python.org/pep-0249/) related `last_insert_id` function of the driver must be called.

You can have a look for yourself here https://github.com/django/django/blob/2722cb61ccae84f593e6d2c28814e3c628743994/django/db/models/sql/compiler.py#L1922-L1958

surister:
> @tomwojcik
> 
> > It looks promising and simpler for enabling UUID PKs specifically. However, my understanding is that the AutoFieldMixin approach implemented here enables a broader feature set, including support for other automatically generated PK types.
> 
> Could you go into more details about which _other automatically generated PK types_ kind of fields are not expressible today as either `Field(default)` or `Field(db_default)` and would require `AutoFieldMixin`?
> 
> AFAIK the only difference between `AutoField` derivative today and other db-generated fields is that on backends that don't support the `RETURNING` clause on `INSERT` (that is only MySQL) the [PEP-0249](https://peps.python.org/pep-0249/) related `last_insert_id` function of the driver must be called.
> 
> You can have a look for yourself here
> 
> https://github.com/django/django/blob/2722cb61ccae84f593e6d2c28814e3c628743994/django/db/models/sql/compiler.py#L1922-L1958
> 
> In other words, it appears to me that the generalized support for db-generated fields made the initially intended approach of `AutoFieldMixin` and `AutoFieldMeta` obsolete as nowadays all the logic has been adapted to duck-type over `field.db_returning` and `isinstance(field, AutoField)` last specialization is to handle `last_insert_id` calls on inserts as described above.

With `AutoFieldMixin`  I can create my custom non-integer auto field and set it to `settings.DEFAULT_AUTO_FIELD`, otherwise I cannot have an auto added 'id' that is not integer, I can change the syntax at `BaseDatabaseWrapper` but then Django will complain when it tries to do integer operations on it.

charettes:
> With AutoFieldMixin I can create my custom non-integer auto field and set it to settings.DEFAULT_AUTO_FIELD, otherwise I cannot have an auto added 'id' that is not integer, I can change the syntax at BaseDatabaseWrapper but then Django will complain when it tries to do integer operations on it.

Let me rephrase my question.

In a world where the `DEFAULT_AUTO_FIELD` setting is supersedes (replaced) with a `DEFAULT_PK_FIELD` one that allows any field that matches the following criteria

```python
field.primary_key and (field.default or field.db_returning)
```

such that fields like

```python
class UUID4PrimaryKey(UUIDFIeld):
    def __init__(self, **kwargs):
        kwargs |= {
            "primary_key": True,
            "default": uuid.uuid4
        }
        super().__init__(**kwargs)

class UUID4GeneratedPrimaryKey(UUIDFIeld):
    def __init__(self, **kwargs):
        kwargs |= {
            "primary_key": True,
            "db_default": RandomUUID,
        }
        super().__init__(**kwargs)
```

are all reference able by `DEFAULT_PK_FIELD` *can you think of other automatically generated PK types that would require subclassing `AutoFieldMixin`*?

My point is that I believe that choosing `DEFAULT_AUTO_FIELD` as the setting name and limiting it's usage to subclasses of `AutoFieldMixin` was a solution that didn't account for `Field.db_default` support landing so _early_ (it had been in the making for years) and that allowing any fields that can generate a value by itself (either though `default` or `db_default`) and naming the setting `DEFAULT_PK_FIELD` instead (entirely decoupling it from the _auto_ naming which is inherited from the MySQL `AUTOINCREMENT` heritage) is a better choice.

surister:
> > With AutoFieldMixin I can create my custom non-integer auto field and set it to settings.DEFAULT_AUTO_FIELD, otherwise I cannot have an auto added 'id' that is not integer, I can change the syntax at BaseDatabaseWrapper but then Django will complain when it tries to do integer operations on it.
> 
> Let me rephrase my question.
> 
> In a world where the `DEFAULT_AUTO_FIELD` setting is superseded (replaced) with a `DEFAULT_PK_FIELD` one that allows any field matching the following criteria
> 
> ```python
> field.primary_key and (field.default or field.db_returning)
> ```
> 
> such that fields like
> 
> ```python
> class UUID4PrimaryKey(UUIDFIeld):
>     def __init__(self, **kwargs):
>         kwargs |= {
>             "primary_key": True,
>             "default": uuid.uuid4
>         }
>         super().__init__(**kwargs)
> 
> class UUID4GeneratedPrimaryKey(UUIDFIeld):
>     def __init__(self, **kwargs):
>         kwargs |= {
>             "primary_key": True,
>             "db_default": RandomUUID,
>         }
>         super().__init__(**kwargs)
> ```
> 
> are all reference able by `DEFAULT_PK_FIELD` _can you think of other automatically generated PK types that would require subclassing `AutoFieldMixin`_?
> 
> My point is that I believe that choosing `DEFAULT_AUTO_FIELD` as the setting name and limiting it's usage to subclasses of `AutoFieldMixin` was a solution that didn't account for `Field.db_default` support landing so _early_ (it had been in the making for years) and that allowing any fields that can generate a value by itself (either though `default` or `db_default`) and naming the setting `DEFAULT_PK_FIELD` instead (entirely decoupling it from the _auto_ naming which is inherited from the MySQL `AUTOINCREMENT` heritage) is a better choice.

I see your point and it makes sense, thanks.

PR #18367, opened 2024-07-15
Removed duplicate inline from tests.admin_inlines.admin.PhotographerAdmin.

sarahboyce:
Introduced in 01ed59f753139afb514170ee7f7384c155ecbc2d

Pulled out from https://github.com/django/django/pull/18301

### DIFF
--- a/tests/admin_inlines/admin.py
+++ b/tests/admin_inlines/admin.py
@@ -146,7 +146,6 @@ class PhotographerAdmin(admin.ModelAdmin):
     ]
     inlines = [
         PhotoTabularInline,
-        PhotoTabularInline,
         PhotoStackedExtra2Inline,
         PhotoStackedExtra3Inline,
         PhotoStackedCollapsibleInline,


felixxm (approved):

nessita:
Hey @sarahboyce, the duplicated inline was not a mistake, it was intentionally added. Could you please elaborate on why it needs to be removed?

sarahboyce:
> Hey @sarahboyce, the duplicated inline was not a mistake, it was intentionally added. Could you please elaborate on why it needs to be removed?

It made the following test simpler:
https://github.com/django/django/commit/65344f0e1ecfff3e9623c18c51eff8fe72f9a9df

As it didn't break the existing test suite and I couldn't see commentary as to why we need it, I removed it

nessita:
> > Hey @sarahboyce, the duplicated inline was not a mistake, it was intentionally added. Could you please elaborate on why it needs to be removed?
> 
> It made the following test simpler: [65344f0](https://github.com/django/django/commit/65344f0e1ecfff3e9623c18c51eff8fe72f9a9df)
> 
> As it didn't break the existing test suite and I couldn't see commentary as to why we need it, I removed it

Thank you for the extra rationale, I will re-check why no test failed :100:

PR #18368, opened 2024-07-15
Fixed #35529 -- Added support for positional arguments in querystring template tag.

g-nie:
Ticket: https://code.djangoproject.com/ticket/35529
Forum discussion: https://forum.djangoproject.com/t/adding-a-template-tag-to-generate-query-strings/24521/

I would rather use dictionary merging  `{**d1, **d2}` but this was not enough for `QueryDict`s that may contain list values.

### DIFF
--- a/django/template/defaulttags.py
+++ b/django/template/defaulttags.py
@@ -4,12 +4,13 @@ import re
 import sys
 import warnings
 from collections import namedtuple
-from collections.abc import Iterable
+from collections.abc import Iterable, Mapping
 from datetime import datetime
 from itertools import cycle as itertools_cycle
 from itertools import groupby
 
 from django.conf import settings
+from django.http import QueryDict
 from django.utils import timezone
 from django.utils.html import conditional_escape, escape, format_html
 from django.utils.lorem_ipsum import paragraphs, words
@@ -1173,16 +1174,22 @@ def now(parser, token):
 
 
 @register.simple_tag(name="querystring", takes_context=True)
-def querystring(context, query_dict=None, **kwargs):
+def querystring(context, *args, **kwargs):
     """
-    Build a query string using `query_dict` and `kwargs` arguments.
+    Build a query string using `args` and `kwargs` arguments.
 
     This tag constructs a new query string by adding, removing, or modifying
-    parameters, starting from the given `query_dict` (defaulting to
-    `request.GET`). Keyword arguments are processed sequentially, with later
-    arguments taking precedence.
+    parameters from the given positional and keyword arguments. Positional
+    arguments must be mappings (such as `QueryDict` or `dict`), and
+    `request.GET` is used as the starting point if `args` is empty.
+
+    Keyword arguments are treated as an extra, final mapping. These mappings
+    are processed sequentially, with later arguments taking precedence.
 
     A query string prefixed with `?` is returned.
+
+    Raise TemplateSyntaxError if a positional argument is not a mapping or if
+    keys are not strings.
 
     For example::
 
@@ -1197,18 +1204,31 @@ def querystring(context, query_dict=None, **kwargs):
 
         {# Use a custom ``QueryDict`` #}
         {% querystring my_query_dict foo=3 %}
+
+        {# Use multiple positional and keyword arguments #}
+        {% querystring my_query_dict my_dict foo=3 bar=None %}
     """
-    if query_dict is None:
-        query_dict = context.request.GET
-    params = query_dict.copy()
-    for key, value in kwargs.items():
-        if value is None:
-            if key in params:
-                del params[key]
-        elif isinstance(value, Iterable) and not isinstance(value, str):
-            params.setlist(key, value)
-        else:
-            params[key] = value
+    if not args:
+        args = [context.request.GET]
+    params = QueryDict(mutable=True)
+    for d in [*args, kwargs]:
+        if not isinstance(d, Mapping):
+            raise TemplateSyntaxError(
+                "querystring requires mappings for positional arguments (got "
+                "%r instead)." % d
+            )
+        for key, value in d.items():
+            if not isinstance(key, str):
+                raise TemplateSyntaxError(
+                    "querystring requires strings for mapping keys (got %r "
+                    "instead)." % key
+                )
+            if value is None:
+                params.pop(key, None)
+            elif isinstance(value, Iterable) and not isinstance(value, str):
+                params.setlist(key, value)
+            else:
+                params[key] = value
     query_string = params.urlencode() if params else ""
     return f"?{query_string}"
 
--- a/docs/ref/templates/builtins.txt
+++ b/docs/ref/templates/builtins.txt
@@ -964,8 +964,14 @@ output (as a string) inside a variable. This is useful if you want to use
 
 Outputs a URL-encoded formatted query string based on the provided parameters.
 
-This tag requires a :class:`~django.http.QueryDict` instance, which defaults to
-:attr:`request.GET <django.http.HttpRequest.GET>` if none is provided.
+This tag accepts positional arguments, which must be mappings (such as
+:class:`~django.http.QueryDict` or :class:`dict`). If no positional arguments
+are provided, :attr:`request.GET <django.http.HttpRequest.GET>` is used as the
+default to construct the query string.
+
+Positional arguments are processed sequentially, while keyword arguments are
+treated as key-value pairs, applied last. Later arguments take precedence over
+earlier ones, ensuring the most recent pairs are reflected in the final result.
 
 The result always includes a leading ``"?"`` since this tag is mainly used for
 links, and an empty result could prevent the page from reloading as expected.
@@ -1033,16 +1039,33 @@ Handling lists
 If ``my_list`` is ``["red", "blue"]``, the output will be
 ``?color=red&color=blue``, preserving the list structure in the query string.
 
-Custom QueryDict
-~~~~~~~~~~~~~~~~
+Customizing the base QueryDict
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can pass custom ``QueryDict`` or ``dict`` instances as positional arguments
+to replace ``request.GET``. When multiple arguments are provided, key-value
+pairs from later arguments take precedence over earlier ones.
+
+For example, if ``my_query_dict`` is ``<QueryDict: {'color': ['blue'], 'size':
+['S']}>`` and ``my_dict`` is ``{'color': 'orange', 'fabric': 'silk', 'type':
+'dress'}``, this outputs ``?color=orange&size=S&fabric=silk``.
 
 .. code-block:: html+django
 
-    {% querystring my_query_dict %}
+    {% querystring my_query_dict my_dict size="S" type=None %}
 
-You can provide a custom ``QueryDict`` to be used instead of ``request.GET``.
-So if ``my_query_dict`` is ``<QueryDict: {'color': ['blue']}>``, this outputs
-``?color=blue``. If ``my_query_dict`` is empty, the output will be ``?``.
+If all keys are removed by setting them to ``None``, this outputs ``?``:
+
+.. code-block:: html+django
+
+    {% querystring my_query_dict my_dict color=None size=None fabric=None type=None %}
+
+Similarly, if all positional arguments are empty and keyword arguments do not
+contribute any new params, the output will also be ``?``.
+
+.. versionchanged:: 6.0
+
+    Support for multiple positional mapping arguments was added.
 
 Dynamic usage
 ~~~~~~~~~~~~~
--- a/tests/template_tests/syntax_tests/test_querystring.py
+++ b/tests/template_tests/syntax_tests/test_querystring.py
@@ -1,5 +1,6 @@
 from django.http import QueryDict
 from django.template import RequestContext
+from django.template.base import TemplateSyntaxError
 from django.test import RequestFactory, SimpleTestCase
 
 from ..utils import setup
@@ -13,6 +14,10 @@ class QueryStringTagTests(SimpleTestCase):
         output = self.engine.render_to_string(template_name, context)
         self.assertEqual(output, expected)
 
+    def assertTemplateSyntaxError(self, template_name, context, expected):
+        with self.assertRaisesMessage(TemplateSyntaxError, expected):
+            self.engine.render_to_string(template_name, context)
+
     @setup({"querystring_empty_get_params": "{% querystring %}"})
     def test_querystring_empty_get_params(self):
         context = RequestContext(self.request_factory.get("/"))
@@ -25,6 +30,19 @@ class QueryStringTagTests(SimpleTestCase):
         for context in [non_empty_context, empty_context]:
             with self.subTest(context=context):
                 self.assertRenderEqual("querystring_remove_all_params", context, "?")
+
+    @setup(
+        {
+            "querystring_remove_all_params_custom_querydict": (
+                "{% querystring my_query_dict my_dict a=None %}"
+            )
+        }
+    )
+    def test_querystring_remove_all_params_custom_querydict(self):
+        context = {"my_query_dict": QueryDict("a=1&b=2"), "my_dict": {"b": None}}
+        self.assertRenderEqual(
+            "querystring_remove_all_params_custom_querydict", context, "?"
+        )
 
     @setup({"querystring_non_empty_get_params": "{% querystring %}"})
     def test_querystring_non_empty_get_params(self):
@@ -42,7 +60,7 @@ class QueryStringTagTests(SimpleTestCase):
 
     @setup({"querystring_empty_params": "{% querystring qd %}"})
     def test_querystring_empty_params(self):
-        cases = [None, {}, QueryDict()]
+        cases = [{}, QueryDict()]
         request = self.request_factory.get("/")
         qs = "?a=b"
         request_with_qs = self.request_factory.get(f"/{qs}")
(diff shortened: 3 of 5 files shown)


g-nie:
Sure thing @nessita, I just rebased.

sarahboyce:
Thank you for this @g-nie - this looks really good :star: 
The docs got reorganized in cf03aa4e94625971852a09e869f7ee7c328b573f which has caused some conflicts
Also from the forum discussion, this change was not classed as a release blocker and so we will add this to Django 5.2
Can you add a release note to Django 5.2 and a `.. versionchanged:: 5.2` note to the docs?

g-nie:
Sounds good @sarahboyce. I took the chance to add myself to `AUTHORS` as well 😄

tests/template_tests/syntax_tests/test_querystring.py
```diff
@@ -65,6 +74,65 @@ def test_querystring_remove_nonexistent(self):
...
+        context = {"somevar": 0}
+        msg = "'int' object has no attribute 'items'"
+        with self.assertRaisesMessage(AttributeError, msg):
+            self.engine.render_to_string("querystring_non_dict_args", context)
+
+    @setup({"querystring_multiple_args": "{% querystring my_query_dict x=0 %}"})
+    def test_querystring_multiple_args(self):
+        context = {"my_query_dict": QueryDict("a=1&b=2")}
+        output = self.engine.render_to_string("querystring_multiple_args", context)
+        self.assertEqual(output, "?a=1&amp;b=2&amp;x=0")
```
sarahboyce, line 120: I think this test is already covered by `test_querystring_with_explicit_query_dict_and_no_request` and can be removed

django/template/defaulttags.py
```diff
@@ -1188,21 +1189,36 @@ def querystring(context, query_dict=None, **kwargs):
...
+            if not isinstance(key, str):
+                raise TemplateSyntaxError(
+                    "querystring received non-string dict key: %s" % key
+                )
+            if value is None:
+                query_dict.pop(key, None)
+            elif isinstance(value, Iterable) and not isinstance(value, str):
+                query_dict.setlist(key, value)
+            else:
+                query_dict[key] = value
```
sarahboyce, line 1221: I am not sure if we want more validation on values but here but here is another possible dictionary input and it's output for a test
```python
    @setup({"querystring_dict_value": "{% querystring my_dict %}"})
    def test_querystring_dict_value(self):
        context = {"my_dict": {"a": {"a": 0}}}
        output = self.engine.render_to_string("querystring_dict_value", context)
        self.assertEqual(output, "?a=a")
```
Is that what we want? Or do we want it to raise an error do you think? :thinking:

sarahboyce (commented):
Thank you for the updates @g-nie :star:

docs/ref/templates/builtins.txt
```diff
@@ -961,20 +961,20 @@ output (as a string) inside a variable. This is useful if you want to use
...
-:attr:`request.GET <django.http.HttpRequest.GET>` if none is provided.
+This tag requires a :class:`~django.http.QueryDict` or :class:`dict` instance,
+which defaults to :attr:`request.GET <django.http.HttpRequest.GET>`
+if none is provided.
 
-If the :class:`~django.http.QueryDict` is empty and no additional parameters
-are provided, an empty string is returned. A non-empty result includes a
-leading ``"?"``.
+If the ``dict``s are empty and no additional parameters are provided,
+an empty string is returned. A non-empty result includes a leading ``"?"``.
```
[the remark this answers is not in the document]
g-nie: I'm kind of thinking that the fact that the tag results to a `QueryDict` (right before `urlencode()`) is an implementation detail, and as far as I can tell we don't mention this elsewhere in the docs. It could seem a bit confusing, what do you think?

docs/ref/templates/builtins.txt
```diff
@@ -961,20 +961,20 @@ output (as a string) inside a variable. This is useful if you want to use
...
 
 .. admonition:: Using ``request.GET`` as default
 
     To use ``request.GET`` as the default ``QueryDict`` instance, the
     ``django.template.context_processors.request`` context processor should be
     enabled. If it's not enabled, you must either explicitly pass the
-    ``request`` object into the template context, or provide a ``QueryDict``
-    instance to this tag.
+    ``request`` object into the template context, or provide a
+    ``QueryDict``/``dict`` instance to this tag.
```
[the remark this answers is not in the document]
g-nie: Shouldn't we mention here that users can also pass `dict`s?

tests/template_tests/syntax_tests/test_querystring.py
g-nie: Oh right

django/template/defaulttags.py
g-nie: That is interesting. In my opinion this specific case kind of makes sense, it's not really unexpected to me. The inner dict is treated as an iterable by the tag and `QueryDict.setlist()` is used.
What definitely is more problematic is passing arbitrary objects as dict values:

```python
@setup({"querystring_class_value": "{% querystring my_dict %}"})
def test_querystring_class_value(self):
    class MyClass:
        pass

    context = {"my_dict": {"a": MyClass()}}
    output = self.engine.render_to_string("querystring_class_value", context)
    print(output)
```
Which outputs `?a=%3Ctemplate_tests.syntax_tests.test_querystring.QueryStringTagTests.test_querystring_dict_value.%3Clocals%3E.MyClass+object+at+0x72a2e4f19990%3E`

On the other hand I'm not sure how we can limit this properly. Could we only allow string and numeric dict values?
And then what happens with Iterables, do we also check the types of their contents too?
Maybe it's "ok" to allow other values, given that in the cases that I can think of, the tag will either raise an exception, or the output will be obviously, noticeably wrong to the user.

docs/ref/templates/builtins.txt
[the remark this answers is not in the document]
sarahboyce: I feel like this bit is about having the `request.GET` `QueryDict` being used as the default and so that you can pass a dict is, to me, not relevant for this note (if that makes sense)

docs/ref/templates/builtins.txt
```diff
@@ -993,16 +993,16 @@ Outputs the current query string verbatim. So if the query string is
...
 
-    {% querystring my_query_dict %}
+    {% querystring my_dict %}
 
-You can provide a custom ``QueryDict`` to be used instead of ``request.GET``.
-So if ``my_query_dict`` is ``<QueryDict: {'color': ['blue']}>``, this outputs
-``?color=blue``.
+You can provide a custom ``QueryDict``/``dict`` to be used instead of
+``request.GET``. So if ``my_dict`` is ``<QueryDict: {'color': ['blue']}>``,
+this outputs ``?color=blue``.
```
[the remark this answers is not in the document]
sarahboyce: > "Key values from later arguments take precedence over previously seen keys"

That's nicer :star: 

>  Also, my previous section included a query_dict + dict + extra kwargs example, you think that's not necessary to have?

So in the order of the docs, after this section it talks about setting keys. So, not within this section, but we can have a think about if we want to have this example somewhere else (not sure)

docs/ref/templates/builtins.txt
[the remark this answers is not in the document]
g-nie: I like this

docs/ref/templates/builtins.txt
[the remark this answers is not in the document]
g-nie: Ok yes this makes sense, I'm changing it.

sarahboyce (approved):
Thank you for this @g-nie :star: this looks good to me!

@knyghty and @nessita have want to have a quick look before I land this, so will wait for a thumbs up from them :+1:

nessita:
> Thank you for this @g-nie ⭐ this looks good to me!
> 
> @knyghty and @nessita have want to have a quick look before I land this, so will wait for a thumbs up from them 👍

Thank you Sarah! I would like to review this, I would do so after the 5.1 final release on Wed.

nessita (commented):
Thank you @g-nie for this branch! Great work :star2: 

I added a few initial comments, I then got sidetracked when testing this new behavior because it crashed for some existing cases I had in a test project. My test view looks like this:

* views.py
```python
def qs_view(request):
    qss = [
        None, {}, request.GET,
        QueryDict.fromkeys(["start", "start", "end"], value="now"),
        QueryDict("color=red&size=S"),
        {i: i*2 for i in range(5)},
    ]
    empty_qs = QueryDict("")
    my_list = ["red", "blue"]
    context = {
        "qss": qss, "val": "something not simple",
        "my_list": my_list, "empty_qs": empty_qs,
    }
    return render(request, "qs.html", context)
```
* template `qs.html` (fragment):
```html
<ol>
  <li><pre>{% querystring %}</pre></li>
  <li><pre>{% querystring a='b' %}</pre></li>
  <li><pre>{% querystring a=b %}</pre></li>
  <li><pre>{% querystring a=val %}</pre></li>
  <li><pre>{% querystring foo=1 %}</pre></li>
  <li><pre>{% querystring color=my_list %}</pre></li>
  <li><pre>{% querystring request.GET a=1 b=2 z=None %}</pre></li>
  <li><pre>{% querystring empty_qs something="old" %}</pre></li>
  <li><pre>{% querystring empty_qs %}</pre></li>
</ol>

<h3>Usage of <tt>querystring</tt> (part II)</h3>

<p>
  {% for i in qss %}
  <pre>{{ i }}: {% querystring i %}</pre>
  {% endfor %}
</p>
```

For the context value being `None` I get `'NoneType' object has no attribute 'items'`, but in `main` (and `5.1`) this works, resulting in the empty string (which is consistent with the docs).
Also my dict `{i: i*2 for i in range(5)}` does not work either, it fails with `querystring received non-string dict key: 0`, I wonder if we should use `str()` of the keys, and only fails if `str()` fails?

To aid this, I have proposed PR #18487 adding extra tests for some of these cases. We should merge that, then rebase this PR, fix conflicts and continue from there.

django/template/defaulttags.py
```diff
@@ -1170,11 +1171,12 @@ def now(parser, token):
...
-def querystring(context, query_dict=None, **kwargs):
+def querystring(context, *args, **kwargs):
     """
-    Add, remove, and change parameters of a ``QueryDict`` and return the result
-    as a query string. If the ``query_dict`` argument is not provided, default
-    to ``request.GET``.
+    Add, remove, and change parameters of ``QueryDict`` or ``dict``
+    arguments and return the result as a query string. Later arguments
+    override any previously seen keys. If no argument is provided,
+    default to ``request.GET``.
```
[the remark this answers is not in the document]
g-nie: Yea that seems better to me 👍

g-nie:
@nessita I just rebased to take the new tests into account.
The only code change that I made was the addition of 
```python
if d is None:
    continue
```
to make the tests pass. Let me know if that makes sense.

django/template/defaulttags.py
nessita: There is one use case for using, for example, int values as keys which I added to my former review comment (reproduced below):

> Also my dict `{i: i*2 for i in range(5)}` does not work either, it fails with `querystring received non-string dict key: 0`, I wonder if we should use `str()` of the keys, and only fail if `str()` fails?

So resurfacing that here: Is there a strong objection to applying `str()` to every key?

nessita: On  a second look and thought, I agree we can tweak this after feature freeze following developer feedback. Will resolve for now,

tests/template_tests/syntax_tests/test_querystring.py
```diff
@@ -65,6 +70,14 @@ def test_querystring_remove(self):
         context = RequestContext(request)
         self.assertRenderEqual("querystring_remove", context, expected="?a=1")
 
+    @setup({"querystring_remove_dict": "{% querystring my_dict a=1 %}"})
+    def test_querystring_remove_from_dict(self):
+        request = self.request_factory.get("/", {"test": "value"})
+        template = self.engine.get_template("querystring_remove_dict")
+        context = RequestContext(request, {"my_dict": {"test": None}})
+        output = template.render(context)
+        self.assertEqual(output, "?a=1")
```
nessita, line 79: I noticed that the newly added tests would not use the helper I landed in PR #18487. Could you please adapt each new tests (the ones for non errors) to use `assertRenderEqual`?

django/template/defaulttags.py
```diff
@@ -1188,21 +1191,44 @@ def querystring(context, query_dict=None, **kwargs):
...
-        elif isinstance(value, Iterable) and not isinstance(value, str):
-            query_dict.setlist(key, value)
-        else:
-            query_dict[key] = value
+    try:
+        query_dict = context.request.GET.copy()
+    except AttributeError:
+        if not args:
+            raise
+        query_dict = QueryDict(mutable=True)
```
nessita, line 1211: I think this logic is changing the documented behavior. The docs says:

> This tag requires a QueryDict or dict instance, which defaults to request.GET if none is provided.

The docs also say:

> You can provide custom QueryDict or dict instances as positional arguments to be used instead of request.GET.

The way this code is written, it seems that `request.GET` is used as a starting point, if available, and then any argument is added on top of that. I think that is not correct, and we'd need:

* a test ensuring the documented behavior holds, and
* a change in this part to ensure `request.GET` is used only if not `args`.

nessita (changes requested):
Hey @g-nie, this is in great shape, thanks for your patience and for merging my test changes in.

I have added a couple of comments that I think are blockers to land this PR. Please let me know what you think!

django/template/defaulttags.py
g-nie: I believe the documented behavior is already covered by tests:

> This tag requires a QueryDict or dict instance, which defaults to request.GET if none is provided.

covered by `test_querystring_non_empty_get_params`

> You can provide custom QueryDict or dict instances as positional arguments to be used instead of request.GET.

covered by `test_querystring_replace`

I think the change that you suggest might be inaccurate, and not needed (?): 

> a change in this part to ensure request.GET is used only if not args.

We still might want to use **parts** of request.GET even when args is present, see `test_querystring_replace`

tests/template_tests/syntax_tests/test_querystring.py
g-nie: Yep of course

django/template/defaulttags.py
nessita: @g-nie Hello! I have finally re-reviewed this work. So, following your reasoning in these comments, we either have an error in the docs, or we have a missing test.

How *I* read the current docs, they say that *only* if no mappings are provided, *then* `request.GET` is used as starting point/default. I exemplify this with this test:

```python
    @setup({"querystring_request_get_ignored": "{% querystring my_mapping %}"})
    def test_querystring_request_get_ignored(self):
        cases = [({"y": "x"}, "y=x"), ({}, "")]
        request = self.request_factory.get("/", {"x": "y", "a": "b"})
        for param, expected in cases:
            with self.subTest(param=param):
                context = RequestContext(request, {"my_mapping": param})
                self.assertRenderEqual(
                    "querystring_request_get_ignored", context, expected=expected
                )
```

This of course fails with:
```
FAIL: test_querystring_request_get_ignored (template_tests.syntax_tests.test_querystring.QueryStringTagTests.test_querystring_request_get_ignored) (param={'y': 'x'})
  File "/home/nessita/fellowship/django/tests/template_tests/syntax_tests/test_querystring.py", line 56, in test_querystring_request_get_ignored
    self.assertRenderEqual(
    ~~~~~~~~~~~~~~~~~~~~~~^
        "querystring_request_get_ignored", context, expected=expected
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/nessita/fellowship/django/tests/template_tests/syntax_tests/test_querystring.py", line 15, in assertRenderEqual
    self.assertEqual(output, expected)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
AssertionError: '?x=y&amp;a=b&amp;y=x' != 'y=x'

FAIL: test_querystring_request_get_ignored (template_tests.syntax_tests.test_querystring.QueryStringTagTests.test_querystring_request_get_ignored) (param={})
  File "/home/nessita/fellowship/django/tests/template_tests/syntax_tests/test_querystring.py", line 56, in test_querystring_request_get_ignored
    self.assertRenderEqual(
    ~~~~~~~~~~~~~~~~~~~~~~^
        "querystring_request_get_ignored", context, expected=expected
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/nessita/fellowship/django/tests/template_tests/syntax_tests/test_querystring.py", line 15, in assertRenderEqual
    self.assertEqual(output, expected)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
AssertionError: '?x=y&amp;a=b' != ''
```

Personally, I think that the behavior shown in the test is the one we want: if we pass any mapping as a starting point, the code ignores `request.GET`. Otherwise, `kwargs` are used to tweak `request.GET` as desired/instructed. I believe this was the original intent for this template tag.

@sarahboyce @knyghty Do you have an opinion on the above?

I will be pushing some minor updates (release notes changed to ` 6.0`, etc) including the failing test, so please pull changes before pushing any further tweaks (assuming we agree on semantics).

nessita (commented):
Added a comment asking for clarification on tag semantics when a mapping is given in any of the positional arguments.

django/template/defaulttags.py
knyghty: I think the docs have the intended behaviour which I think lines up with your comment. It's more flexible to be able to omit `request.GET` for cases where you wish to ignore it. In the cases where you _do_ want it I would suggest using `{% querystring request.GET ... %}` explicity.

I do also think it's perhaps a bit confusing though if `request.GET` is sometimes used and sometimes not - hopefully the docs are enough to cover this.

Hope I'm understanding correctly, I've only had a cursory glance!

g-nie: Hello! So I read everything again and I think that what you suggest @nessita makes sense. I went ahead and updated the code to make `test_querystring_request_get_ignored` pass.

> I do also think it's perhaps a bit confusing though if request.GET is sometimes used and sometimes not - hopefully the docs are enough to cover this.

I think the docs should cover it, let me know if something is still unclear/confusing.

(I didn't clean up the commits just yet)

django/template/defaulttags.py
```diff
@@ -1170,11 +1171,19 @@ def now(parser, token):
...
-    as a query string. If the ``query_dict`` argument is not provided, default
-    to ``request.GET``.
+    Modify and return a query string with updated parameters.
+
+    Add, remove, or change parameters in a given `QueryDict` or `dict`,
+    returning the result as a query string. Any arguments provided will
+    override existing keys. If no arguments are given, `request.GET` is used.
+
+    Positional arguments must be mappings and, in order of appearance, will
+    override existing keys.
```
knyghty, line 1183: I wonder if this could be more explicit. One way to read this is that earlier arguments will override later ones, but I think the opposite is the case (and makes more sense). If I can get confused about this as a native speaker, I think others could as well.

Maybe we can say something more similar to the following paragraph. Maybe:

"Positional arguments must be mappings. These will override existing keys, with later arguments taking precedence."

I also prefer to split the sentences up to make them easier to read.

django/template/defaulttags.py
```diff
@@ -1170,11 +1171,19 @@ def now(parser, token):
...
+def querystring(context, *args, **kwargs):
     """
-    Add, remove, and change parameters of a ``QueryDict`` and return the result
-    as a query string. If the ``query_dict`` argument is not provided, default
-    to ``request.GET``.
+    Modify and return a query string with updated parameters.
+
+    Add, remove, or change parameters in a given `QueryDict` or `dict`,
+    returning the result as a query string. Any arguments provided will
+    override existing keys. If no arguments are given, `request.GET` is used.
```
knyghty, line 1180: Is this completely correct? If no arguments are given, yes, but it's also used if only keyword arguments are given.

docs/ref/templates/builtins.txt
```diff
@@ -959,20 +959,20 @@ output (as a string) inside a variable. This is useful if you want to use
 
 Outputs a URL-encoded formatted query string based on the provided parameters.
 
-This tag requires a :class:`~django.http.QueryDict` instance, which defaults to
-:attr:`request.GET <django.http.HttpRequest.GET>` if none is provided.
+This tag requires a :class:`~django.http.QueryDict` or :class:`dict` instance,
```
knyghty, line 962: Maybe it's more friendly for new users like this but perhaps we should say a mapping? Maybe we can say "a dict or other mapping"?

knyghty (commented):
These are mostly things to consider, don't take them too seriously, I am not sure myself :)

nessita:
@g-nie Hey there! Would you have some time to address Tom's comment and resolve the conflicts? Once that's done, please reset the ticket flags—I’ll do my best to provide a timely re-review.

django/template/defaulttags.py
g-nie: You're right. Actually, it's also used if both args and kwargs are not given. It's basically always used unless args is given, so I'm updating this to: `If no positional arguments are given, request.GET is used.`

so that it reflects the `if` condition:

```python
if args:
    query_dict = QueryDict(mutable=True)
else:
    query_dict = context.request.GET.copy()
```

django/template/defaulttags.py
g-nie: Yes I like this better 👍

docs/ref/templates/builtins.txt
g-nie: Sure, makes sense

g-nie:
> @g-nie Hey there! Would you have some time to address Tom's comment and resolve the conflicts? Once that's done, please reset the ticket flags—I’ll do my best to provide a timely re-review.

Hey @nessita! I pushed changes for @knyghty's comments, resolved conflicts and rebased.
Sorry I'm not sure which ticket flags you mean 😅 Is it the "Patch needs improvement" one?

nessita (approved):
@g-nie Thank you for the latest work :star2: 

I had to change the logic a bit, and restored a lost test, because otherwise this branch was reverting the work and fix done for ticket-36182. I also added one more test related to that ticket.

Lastly, I made some tweaks in the docs. This is now approved! I will land when CI is green again.

nessita:
This has been rebased onto https://github.com/django/django/pull/19289 and should not be merged until that PR is merged (and this branch properly rebased onto `main`).

nessita:
The pre-requisite PR has been merged so I have rebased this PR and fixed conflicts. I will merge tomorrow once CI is green.

PR #18369, opened 2024-07-15
Removed leftover KeyError handling after Query.tables attribute cleanup.

nessita:
Follow up from f7f5edd50d03e8482f8a6da5fb5202b895d68cd6.


charettes (approved):
Thanks for the cleanup 🎉

PR #18371, opened 2024-07-16
Fixed #35604, Refs #35326 -- Made FileSystemStorage.exists() behaviour independent from allow_overwrite.

sarahboyce:
# Trac ticket number

ticket-35604

# Branch description

Partially reverts 0b33a3abc2ca7d68a24f6d0772bc2b9fa603744e.

`Storage.exists(name)` was documented to "return False if the name is available for a new file." but return True if the file exists. This is ambiguous in the overwrite file case.

Overwriting `get_available_name()` should still ensure all file name validation is applied, ref CVE-2024-39330.

This is motivated from this discussion on django-storages: https://github.com/jschneier/django-storages/issues/1430

Note: this concern was already raised by @bcail [here](https://github.com/django/django/pull/18020#issuecomment-2107536342) (thank you). At the time, I believed the impact to be limited and the implementation to still be within the definition of `Storages.exists(name)`.

This is to be backported to 5.1

# Checklist

### DIFF
--- a/django/core/files/storage/base.py
+++ b/django/core/files/storage/base.py
@@ -51,6 +51,10 @@ class Storage:
         validate_file_name(name, allow_relative_path=True)
         return name
 
+    def is_name_available(self, name, max_length=None):
+        exceeds_max_length = max_length and len(name) > max_length
+        return not self.exists(name) and not exceeds_max_length
+
     # These methods are part of the public API, with default implementations.
 
     def get_valid_name(self, name):
@@ -82,11 +86,11 @@ class Storage:
         validate_file_name(file_name)
         file_ext = "".join(pathlib.PurePath(file_name).suffixes)
         file_root = file_name.removesuffix(file_ext)
-        # If the filename already exists, generate an alternative filename
-        # until it doesn't exist.
+        # If the filename is not available, generate an alternative
+        # filename until one is available.
         # Truncate original name if required, so the new filename does not
         # exceed the max_length.
-        while self.exists(name) or (max_length and len(name) > max_length):
+        while not self.is_name_available(name, max_length=max_length):
             # file_ext includes the dot.
             name = os.path.join(
                 dir_name, self.get_alternative_name(file_root, file_ext)
--- a/django/core/files/storage/filesystem.py
+++ b/django/core/files/storage/filesystem.py
@@ -4,7 +4,6 @@ from datetime import datetime, timezone
 from urllib.parse import urljoin
 
 from django.conf import settings
-from django.core.exceptions import SuspiciousFileOperation
 from django.core.files import File, locks
 from django.core.files.move import file_move_safe
 from django.core.signals import setting_changed
@@ -192,14 +191,18 @@ class FileSystemStorage(Storage, StorageSettingsMixin):
             # concurrently.
             pass
 
-    def exists(self, name):
-        try:
-            exists = os.path.lexists(self.path(name))
-        except SuspiciousFileOperation:
-            raise
+    def is_name_available(self, name, max_length=None):
         if self._allow_overwrite:
-            return False
-        return exists
+            return not (max_length and len(name) > max_length)
+        return super().is_name_available(name, max_length=max_length)
+
+    def get_alternative_name(self, file_root, file_ext):
+        if self._allow_overwrite:
+            return f"{file_root}{file_ext}"
+        return super().get_alternative_name(file_root, file_ext)
+
+    def exists(self, name):
+        return os.path.lexists(self.path(name))
 
     def listdir(self, path):
         path = self.path(path)
--- a/tests/file_storage/tests.py
+++ b/tests/file_storage/tests.py
@@ -95,18 +95,18 @@ class FileStorageTests(SimpleTestCase):
         """
         Standard file access options are available, and work as expected.
         """
-        self.assertFalse(os.path.exists(os.path.join(self.temp_dir, "storage_test")))
+        self.assertFalse(self.storage.exists("storage_test"))
         f = self.storage.open("storage_test", "w")
         f.write("storage contents")
         f.close()
-        self.assertTrue(os.path.exists(os.path.join(self.temp_dir, "storage_test")))
+        self.assertTrue(self.storage.exists("storage_test"))
 
         f = self.storage.open("storage_test", "r")
         self.assertEqual(f.read(), "storage contents")
         f.close()
 
         self.storage.delete("storage_test")
-        self.assertFalse(os.path.exists(os.path.join(self.temp_dir, "storage_test")))
+        self.assertFalse(self.storage.exists("storage_test"))
 
     def _test_file_time_getter(self, getter):
         # Check for correct behavior under both USE_TZ=True and USE_TZ=False.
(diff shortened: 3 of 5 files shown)


nessita (commented):
Thank you @sarahboyce for working on this! :star2: 

I think this looks good. @jschneier would you like to take a quick look?

nessita (approved):
I took a second look and I think this is ready for checkin.

jschneier:
The only issue I can see is if the name doesn't fit inside the length requirement we won't handle that.

django/core/files/storage/filesystem.py
```diff
@@ -192,14 +191,15 @@ def delete(self, name):
             # concurrently.
             pass
 
-    def exists(self, name):
-        try:
-            exists = os.path.lexists(self.path(name))
-        except SuspiciousFileOperation:
-            raise
+    def get_available_name(self, name, max_length=None):
+        # Ensure protection against insecure names.
```
jschneier, line 195: This shouldn't be necessary anymore with the CVE fix, right?

sarahboyce: You're right that with the CVE the same validation is applied elsewhere and so it is not strictly necessary

sarahboyce:
> The only issue I can see is if the name doesn't fit inside the length requirement we won't handle that.

Thank you! I had not taken that into account. I've changed the approach adding a new private method `_name_is_unavailable` which is overridden in the `FileSystemStorage` to handle the `allow_overwrite` case.
Added some tests around the filename truncation, needs a fresh review 👍

nessita (approved):
Thank you @sarahboyce for the update, this looks good! :star2: 

There are two things that I'd like to mention so we ponder on these (no immediate action is necessary, I think):
1. There is something about the backend that allows overwrite that bothers me when the `max_length` parameter is not None: calling `save` with the same file name and different `max_length` values would result in multiple files being stored:
```python
        name = "test_long_file_name.txt"
        file = ContentFile(b"content")
        stored_name = self.storage.save(name, file, max_length=10)
        self.assertEqual(stored_name, "test_l.txt")
        stored_name = self.storage.save(name, file, max_length=11)
        self.assertEqual(stored_name, "test_lo.txt")
```
I think this result is OK and makes sense, but it feels odd to some extent.
2. I wonder if, in a near future, `_name_is_unavailable` should be a public API for each backend to implement as they need. Perhaps we should mention and consider this in ticket-35607 for `5.2`?

sarahboyce:
> 1. There is something about the backend that allows overwrite that bothers me when the `max_length` parameter is not None: calling `save` with the same file name and different `max_length` values would result in multiple files being stored:
> 
> ```python
>         name = "test_long_file_name.txt"
>         file = ContentFile(b"content")
>         stored_name = self.storage.save(name, file, max_length=10)
>         self.assertEqual(stored_name, "test_l.txt")
>         stored_name = self.storage.save(name, file, max_length=11)
>         self.assertEqual(stored_name, "test_lo.txt")
> ```
>
> I think this result is OK and makes sense, but it feels odd to some extent. 

I believe this is also true when allow overwrite is False, as it's saved with a different name 
I think that `max_length` mostly comes from `FileField.max_length` (so is quite a static value)
Not sure if that makes it feel less odd?

> 2. I wonder if, in a near future, `_name_is_unavailable` should be a public API for each backend to implement as they need. > Perhaps we should mention and consider this in [ticket-35607](https://code.djangoproject.com/ticket/35607) for `5.2`?

Sure, can also be generic that we review the API and what has/should be made public and can be overwritten

tests/file_storage/tests.py
```diff
@@ -729,6 +729,22 @@ def test_save_overwrite_behavior_temp_file(self):
         finally:
             self.storage.delete(name)
 
+    def test_file_name_truncation(self):
+        name = "test_long_file_name.txt"
+        file = ContentFile(b"content")
+        stored_name = self.storage.save(name, file, max_length=10)
+        self.assertEqual(stored_name, "test_l.txt")
+        self.assertEqual(len(stored_name), 10)
+        self.storage.delete(stored_name)
```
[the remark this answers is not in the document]
sarahboyce: Thank you, I updated to
```python
        try:
            self.assertEqual(stored_name, "test_l.txt")
            self.assertEqual(len(stored_name), 10)
        finally:
            self.storage.delete(stored_name)
```
as this follows the structure of the other tests for this test class

> On a second look, the self.storage.delete call may not be needed since in the setUp method from OverwritingStorageTests, the following would take care of that self.addCleanup(shutil.rmtree, self.temp_dir)

Agreed, it's not strictly necessary and we could chose to refactor these tests (but as a separate PR, not for a backport 👍)

django/core/files/storage/base.py
```diff
@@ -51,6 +51,9 @@ def save(self, name, content, max_length=None):
         validate_file_name(name, allow_relative_path=True)
         return name
 
+    def is_name_available(self, name, max_length=None):
+        return not self.exists(name) and not (max_length and len(name) > max_length)
```
nessita, line 55: Perhaps it's a matter of personal taste, and while the proposed code certainly follows the previous code, I think the following is slightly more correct because otherwise for falsey `max_length` values (such as 0)  we are not enforcing any `length`:
```suggestion
        return not self.exists(name) and (max_length is None or len(name) <= max_length)
```

nessita: On a second thought, perhaps we need to leave this as is and create a new ticket to decide whether a `max_length` of `0` means no validation or what does it mean. Specifically, in `main` (and in this branch), this test pass:

```python
    def test_file_name_truncation_max_length_zero(self):
        name = "test_long_file_name.txt"
        file = ContentFile(b"content")
        stored_name = self.storage.save(name, file, max_length=0)
        self.assertEqual(stored_name, "test_long_file_name.txt")
```

I would have expected for it to fail somehow :thinking:

tests/file_storage/tests.py
```diff
@@ -729,6 +729,24 @@ def test_save_overwrite_behavior_temp_file(self):
...
 
+    def test_file_name_truncation(self):
+        name = "test_long_file_name.txt"
+        file = ContentFile(b"content")
+        stored_name = self.storage.save(name, file, max_length=10)
+        try:
+            self.assertEqual(stored_name, "test_l.txt")
+            self.assertEqual(len(stored_name), 10)
+        finally:
+            self.storage.delete(stored_name)
```
nessita, line 740: I know that we discussed this usage vs `addCleanup` but I see plenty of `addCleanup` calls, particularly in newer code. The usages of `try`-`finally` seem to be in older parts of the code:
```
$ grep finally tests.py | wc -l
7
$ grep addCleanup tests.py | wc -l
```
Would you agree to using the more modern pattern?

nessita (approved):
Still looks good, thanks for the multiple improvements! :trophy:

django/core/files/storage/base.py
sarahboyce: I think max_length mostly comes from max_length of FileField which does have validation that it must be a positive integer (check code `fields.E121`)

nessita: Yes, absolutely, I agree that is how it was incorporated (having look at the git history). I still think the `Storages` API should also check or do something with a `max_length` with value `0` since it's a public API "on its own", independently of Django fields. If you agree I'll create the ticket, or perhaps better to post to the forum first? :thinking: 
In any case, nothing within the scope of this PR, which looks amazing. Thank you!

PR #18372, opened 2024-07-16
Refs #10941 -- Reorganized querystring template tag docs.

nessita:
When reading the `querystring` template tag docs for the first time, I couldn't quite grasp how the template tag was expected to be used. Initially, it seemed that a querydict parameter was 1. required and 2. the only accepted param. After some local testing and re-read of the docs, I think some clarification could help future readers, which I'm proposing in this PR.

### DIFF
--- a/docs/ref/templates/builtins.txt
+++ b/docs/ref/templates/builtins.txt
@@ -959,66 +959,99 @@ output (as a string) inside a variable. This is useful if you want to use
 
 .. versionadded:: 5.1
 
-Outputs the query string from a given :class:`~django.http.QueryDict` instance,
-if provided, or ``request.GET`` if not and the
-``django.template.context_processors.request`` context processor is enabled.
-If the ``QueryDict`` is empty, then the output will be an empty string.
-Otherwise, the query string will be returned with a leading ``"?"``.
+Outputs a URL-encoded formatted query string based on the provided parameters.
 
-If not using the ``django.template.context_processors.request`` context
-processor, you must pass either the ``request`` into the template context or a
-``QueryDict`` instance into this tag.
+This tag requires a :class:`~django.http.QueryDict` instance, which defaults to
+:attr:`request.GET <django.http.HttpRequest.GET>` if none is provided.
 
-The following example outputs the current query string verbatim. So if the
-query string is ``?color=green&size=M``, the output would be
-``?color=green&size=M``:
+If the :class:`~django.http.QueryDict` is empty and no additional parameters
+are provided, an empty string is returned. A non-empty result includes a
+leading ``"?"``.
+
+.. admonition:: Using ``request.GET`` as default
+
+    To use ``request.GET`` as the default ``QueryDict`` instance, the
+    ``django.template.context_processors.request`` context processor should be
+    enabled. If it's not enabled, you must either explicitly pass the
+    ``request`` object into the template context, or provide a ``QueryDict``
+    instance to this tag.
+
+Basic usage
+~~~~~~~~~~~
 
 .. code-block:: html+django
 
     {% querystring %}
 
-You can also pass in a custom ``QueryDict`` that will be used instead of
-``request.GET``:
+Outputs the current query string verbatim. So if the query string is
+``?color=green``, the output would be ``?color=green``.
+
+.. code-block:: html+django
+
+    {% querystring size="M" %}
+
+Outputs the current query string with the addition of the ``size`` parameter.
+Following the previous example, the output would be ``?color=green&size=M``.
+
+Custom QueryDict
+~~~~~~~~~~~~~~~~
 
 .. code-block:: html+django
 
     {% querystring my_query_dict %}
 
-Each keyword argument will be added to the query string, replacing any existing
-value for that key. With the query string ``?color=blue``, the following would
-result in ``?color=red&size=S``:
+You can provide a custom ``QueryDict`` to be used instead of ``request.GET``.
+So if ``my_query_dict`` is ``<QueryDict: {'color': ['blue']}>``, this outputs
+``?color=blue``.
+
+Setting items
+~~~~~~~~~~~~~
 
 .. code-block:: html+django
 
     {% querystring color="red" size="S" %}
 
-It is possible to remove parameters by passing ``None`` as a value. With the
-query string ``?color=blue&size=M``, the following would result in ``?size=M``:
+Adds or modifies parameters in the query string. Each keyword argument will be
+added to the query string, replacing any existing value for that key. For
+instance, if the current query string is ``?color=green``, the output will be
+``?color=red&size=S``.
+
+Removing items
+~~~~~~~~~~~~~~
 
 .. code-block:: html+django
 
     {% querystring color=None %}
 
-If the given parameter is a list, the value will remain as a list. For example,
-if ``my_list`` is set to ``["red", "blue"]``, the following would result in
-``?color=red&color=blue``:
+Passing ``None`` as the value removes the parameter from the query string. For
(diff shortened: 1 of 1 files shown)


nessita:
Proposed:

image

knyghty (approved):
I think this reads better than what I wrote, and the sectioning is nice.

docs/ref/templates/builtins.txt
```diff
@@ -959,66 +959,100 @@ output (as a string) inside a variable. This is useful if you want to use
...
 .. code-block:: html+django
 
     {% querystring my_query_dict %}
 
-Each keyword argument will be added to the query string, replacing any existing
-value for that key. With the query string ``?color=blue``, the following would
-result in ``?color=red&size=S``:
+You can provide a custom ``QueryDict`` to be used instead of ``request.GET``.
+So if ``my_query_dict`` is ``<QueryDict: {'color': ['green']}>``, this outputs
+``?color=green``.
```
sarahboyce, line 1006:
```suggestion
So if ``my_query_dict`` is ``<QueryDict: {'color': ['blue']}>``, this outputs
``?color=blue``.
```
I think given the previous examples say green is in the default QueryDict, it might be nice to update the color.

Also this used to say
> Each keyword argument will be added to the query string, replacing any existing
value for that key.

Perhaps we want to keep that sentence and update the example to something like `<QueryDict: {'color': ['blue'], 'fabric': ['silk']}>`

docs/ref/templates/builtins.txt
```diff
@@ -959,66 +959,100 @@ output (as a string) inside a variable. This is useful if you want to use
...
+~~~~~~~~~~~~~~
 
 .. code-block:: html+django
 
     {% querystring color=my_list %}
 
+If ``my_list`` is ``["red", "blue"]``, the output will be
+``?color=red&color=blue``, preserving the list structure in the query string.
+Please note that if the current query string already defines the ``color``
+parameter, this will overwrite it with the given values.
```
sarahboyce, line 1039:
```suggestion
If  ``my_list`` is ``["red", "blue"]`` and the current query string is
``?color=green``, the output will be ``?color=red&color=blue``.
```
I don't think we have to emphasise the overwrite given the previous sections but we can highlight it in the example.
I found "preserving the list structure in the query string" distracting, I initially thought you meant that it preserves the order of the list

docs/ref/templates/builtins.txt
```diff
@@ -959,66 +959,100 @@ output (as a string) inside a variable. This is useful if you want to use
...
-If the ``QueryDict`` is empty, then the output will be an empty string.
-Otherwise, the query string will be returned with a leading ``"?"``.
+Outputs a URL-encoded formatted query string based on the provided parameters.
 
-If not using the ``django.template.context_processors.request`` context
-processor, you must pass either the ``request`` into the template context or a
-``QueryDict`` instance into this tag.
+This tag requires a :class:`~django.http.QueryDict` instance, which defaults to
+:attr:`request.GET <django.http.HttpRequest.GET>` if none is provided, to
+generate the resulting query string from the passed key-value pairs.
```
sarahboyce, line 966:
```suggestion
:attr:`request.GET <django.http.HttpRequest.GET>` if none is provided.
```
I think less is more here as we don't have to pass in key-value pairs explicitly and I think this could be confusing

docs/ref/templates/builtins.txt
```diff
@@ -959,66 +959,100 @@ output (as a string) inside a variable. This is useful if you want to use
...
 
-The following example outputs the current query string verbatim. So if the
-query string is ``?color=green&size=M``, the output would be
-``?color=green&size=M``:
+If the :class:`~django.http.QueryDict` is empty and no additional parameters
+are provided, an empty string is returned. A non-empty result includes a
+leading ``"?"``.
+
+.. admonition:: Using ``request.GET`` as default
```
sarahboyce, line 972: Wondering if we should put this at the very bottom of the querystring docs given `django.template.context_processors.request` is included in the default template context processors and this is probably a niche issue. But not sure 🤔

sarahboyce (commented):
Thank you @nessita ⭐ I like the sectioning, I have some minor comments/suggestions

docs/ref/templates/builtins.txt
nessita: Agreed, removing!

docs/ref/templates/builtins.txt
nessita: I would leave it here because readers may not scroll to the end if they find their "example" or answer early enough.

docs/ref/templates/builtins.txt
nessita: I love your proposal of using a different color here (`blue`). I have made that modification, thank you for the suggestion.

Regarding the "keywork argument" comment, I think it's more appropriate in the following subsection (`Setting items`), and I have moved it there. Perhaps I should add a second example in that subsection updating a given `QueryDict`? :thinking:

docs/ref/templates/builtins.txt
nessita: Fair point! I wanted to highlight that the existing color wasn't going to be included in the resulted list, because initially that was my (incorrect) expectation. But I'm happy to remove :100:

docs/ref/templates/builtins.txt
```diff
@@ -959,66 +959,100 @@ output (as a string) inside a variable. This is useful if you want to use
...
+~~~~~~~~~~~~~
+
 A common example of using this tag is to preserve the current query string when
 displaying a page of results, while adding a link to the next and previous
-pages of results. For example, if the paginator is currently on page 3, and
-the current query string is ``?color=blue&size=M&page=3``, the following code
-would output ``?color=blue&size=M&page=4``:
+pages. For example, if the paginator is currently on page 3, and the current
+query string is ``?color=blue&size=M&page=3``, the following code would output
+``?color=blue&size=M&page=4``:
```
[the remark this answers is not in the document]
nessita: I agree, restored!

docs/ref/templates/builtins.txt
sarahboyce: > Each keyword argument will be added to the query string, replacing any existing
value for that key.

If the current query string (from `request.GET`) was `?color=green` and the passed in query dict is `<QueryDict: {'size': ['M']}>` is the output `?color=green&size=M` or `?size=M` I wonder 🤔

nessita: Is the latter, confirmed by looking at the implementation and IRL testing.

Then there is also this corner case which does not against the docs but I think it's a bit unexpected:

* query string from request is `?color=green`
* an explicit QueryDict is passed that is empty (but not None)
* `{% querystring empty_qs %}` will result in the empty string
* `{% querystring empty_qs something="foo" %}` will result in `?something=foo`

As mentioned, I think this aligns with the docs ("defaults to `request.GET` if none is provided.") but unsure if we should do something else about this case.

docs/ref/templates/builtins.txt
```diff
@@ -959,66 +959,100 @@ output (as a string) inside a variable. This is useful if you want to use
...
 
     {% querystring my_query_dict %}
 
-Each keyword argument will be added to the query string, replacing any existing
-value for that key. With the query string ``?color=blue``, the following would
-result in ``?color=red&size=S``:
+You can provide a custom ``QueryDict`` to be used instead of ``request.GET``.
+So if ``my_query_dict`` is ``<QueryDict: {'color': ['blue']}>``, this outputs
+``?color=blue``.
+
```
sarahboyce, line 1006:
```suggestion
```

sarahboyce (approved):
Thank you ⭐

PR #18374, opened 2024-07-17
Fixed #35606 -- Admin changelist broken for model instances with __html__ method.

Hisham-Pak:
# Trac ticket number

ticket-35606

# Branch description
Fixed a regression in 5.0, where if `__html__()` in model's instance contains unsafe characters, admin changelist might break.

# Checklist

### DIFF
--- a/docs/releases/5.0.8.txt
+++ b/docs/releases/5.0.8.txt
@@ -11,3 +11,7 @@ Bugfixes
 
 * Added missing validation for ``UniqueConstraint(nulls_distinct=False)`` when
   using ``*expressions`` (:ticket:`35594`).
+
+* Fixed a regression in Django 5.0 where ``ModelAdmin.action_checkbox`` could
+  break the admin changelist HTML page when rendering a model instance with a
+  ``__html__`` method (:ticket:`35606`).
--- a/tests/admin_changelist/models.py
+++ b/tests/admin_changelist/models.py
@@ -23,6 +23,12 @@ class GrandChild(models.Model):
     parent = models.ForeignKey(Child, models.SET_NULL, editable=False, null=True)
     name = models.CharField(max_length=30, blank=True)
 
+    def __str__(self):
+        return self.name
+
+    def __html__(self):
+        return f'<h2 class="main">{self.name}</h2>'
+
 
 class Genre(models.Model):
     name = models.CharField(max_length=20)
--- a/tests/admin_changelist/tests.py
+++ b/tests/admin_changelist/tests.py
@@ -364,6 +364,33 @@ class ChangeListTests(TestCase):
             table_output,
         )
 
+    def test_action_checkbox_for_model_with_dunder_html(self):
+        grandchild = GrandChild.objects.create(name="name")
+        request = self._mocked_authenticated_request("/grandchild/", self.superuser)
+        m = GrandChildAdmin(GrandChild, custom_site)
+        cl = m.get_changelist_instance(request)
+        cl.formset = None
+        template = Template(
+            "{% load admin_list %}{% spaceless %}{% result_list cl %}{% endspaceless %}"
+        )
+        context = Context({"cl": cl, "opts": GrandChild._meta})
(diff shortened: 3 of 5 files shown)


tests/admin_changelist/models.py
```diff
@@ -132,3 +132,10 @@ class CharPK(models.Model):
 class ProxyUser(User):
     class Meta:
         proxy = True
+
+
+class Book(models.Model):
+    title = models.CharField(max_length=100)
+
+    def __html__(self):
+        return '"__html__" method called'
```
claudep, line 141: Maybe set a more realistic html output, like `f'<h2 class="main">{self.title}</h2>`?

tests/admin_changelist/tests.py
```diff
@@ -364,6 +366,29 @@ def test_result_list_html(self):
...
+        request.user = self.superuser
+        m = BookAdmin(Book, custom_site)
+        cl = m.get_changelist_instance(request)
+        cl.formset = None
+        template = Template(
+            "{% load admin_list %}{% spaceless %}{% result_list cl %}{% endspaceless %}"
+        )
+        context = Context({"cl": cl, "opts": Book._meta})
+        table_output = template.render(context)
+        self.assertNotInHTML(
```
claudep, line 385: It may be a personal preference, but I don't like `assertNotIn*` methods with large content, as it can easily break for many other reasons that the issue at hand. I would rather use an `assertIn('aria-label="<content>"', table_output)`. Maybe also add a `__str__` method to the Book class?

claudep (commented):
Many thanks for the patch! I added some comments, but maybe some other reviewers can confirm if my suggestions make sense or not.

tests/admin_changelist/models.py
sarahboyce: Let's also avoid defining a new model and admin, we can add `__html__` to one of the existing models

tests/admin_changelist/tests.py
sarahboyce: +1 to adding a `__str__` method
There is a pattern using `build_tbody_html` to create the row html and checking it exsists in the table that we might want to reuse (this is from `test_result_list_html`)
```python
       row_html = build_tbody_html(
            new_child,
            link,
            "name",
            '<td class="field-parent nowrap">%s</td>' % new_parent,
        )
        self.assertNotEqual(
            table_output.find(row_html),
            -1,
            "Failed to find expected row element: %s" % table_output,
        )
```
If we were to add a `__html__` method to `Child`, all of these tests which look for html built in `build_tbody_html` fail (so adding `__html__` and `__str__` to `Child` is an option, but it might be nice to do this to another model and have a separate test case 🤔)

docs/releases/5.0.8.txt
```diff
@@ -9,4 +9,6 @@ Django 5.0.8 fixes several bugs in 5.0.7.
 Bugfixes
 ========
 
-* ...
+* Fixed a regression in Django 5.0 where ``ModelAdmin.action_checkbox`` had the
+  possibility to break admin changelist by returning unsafe characters from
+  model instance's ``__html__`` method. (:ticket:`35606`).
```
sarahboyce, line 14:
```suggestion
* Fixed a regression in Django 5.0 where ``ModelAdmin.action_checkbox`` could
  break the admin changelist HTML page when rendering a model instance with a
  ``__html__`` method (:ticket:`35606`).
```
(maybe 🤔)

sarahboyce (commented):
Thank you for picking this up @Hisham-Pak ⭐

tests/admin_changelist/tests.py
Hisham-Pak: Thank you both for the reviews. Regarding the choice between adding `__html__` to the `Child` model or another model, I opted for the latter. This way, test failures point to one specific test instead of five. If you believe the former option is better, please let me know.

claudep:
A bit late, thanks both for your quick and efficient work :champagne:

PR #18381, opened 2024-07-18
Small fix: fix the result of `TruncHour` example

shipengtaov:
# Trac ticket number

N/A

# Branch description
The result of `TruncHour` in the example should be `'hour': datetime.datetime(2014, 6, 16, 14, 0`, not `'hour': datetime.datetime(2014, 6, 16, 0, 0`


sarahboyce:
Thank you for this @shipengtaov 
I re-ran the output and it is correct because `tzinfo=melb` (which is utc+10 and 15th 14:30 utc = 16th 00:30 in Melbourne)
I am not 100% sure why `tzinfo=melb` is set on some but not others but that is how it has been that way since it was first implemented in https://github.com/django/django/pull/6243
So this isn't a bug 👍

PR #18387, opened 2024-07-18
Fixed #18296 -- Create custom target directory if missing in startproject and startapp

browniebroke:
# Trac ticket number

ticket-18296

# Branch description

Given the following project structure, where `apps` is a directory:

```
.
├── apps
├── manage.py
└── myproject
    ├── __init__.py
    ├── asgi.py
    ├── settings.py
    ├── urls.py
    └── wsgi.py
```

If I run:

```bash
$ django-admin startapp app1 apps/app1
CommandError: Destination directory '/Users/bruno/.../myproj/apps/app1' does not exist, please create it first.
```

This can cause some friction with beginners, and it feels like it would be something that Django could create out of the box. 

The code is shared with `startproject`, so this PR is doing it for both, as I couldn't find a reason why not.

# Checklist

### DIFF
--- a/django/core/management/templates.py
+++ b/django/core/management/templates.py
@@ -46,7 +46,9 @@ class TemplateCommand(BaseCommand):
     def add_arguments(self, parser):
         parser.add_argument("name", help="Name of the application or project.")
         parser.add_argument(
-            "directory", nargs="?", help="Optional destination directory"
+            "directory",
+            nargs="?",
+            help="Optional destination directory, this will be created if needed.",
         )
         parser.add_argument(
             "--template", help="The path or URL to load the template from."
@@ -105,10 +107,10 @@ class TemplateCommand(BaseCommand):
             if app_or_project == "app":
                 self.validate_name(os.path.basename(top_dir), "directory")
             if not os.path.exists(top_dir):
-                raise CommandError(
-                    "Destination directory '%s' does not "
-                    "exist, please create it first." % top_dir
-                )
+                try:
+                    os.makedirs(top_dir)
+                except OSError as e:
+                    raise CommandError(e)
 
         # Find formatters, which are external executables, before input
         # from the templates can sneak into the path.
--- a/docs/ref/django-admin.txt
+++ b/docs/ref/django-admin.txt
@@ -1266,9 +1266,13 @@ By default, :source:`the new directory <django/conf/app_template>` contains a
 ``models.py`` file and other app template files. If only the app name is given,
 the app directory will be created in the current working directory.
 
-If the optional destination is provided, Django will use that existing
-directory rather than creating a new one. You can use '.' to denote the current
-working directory.
+If the optional destination is provided, Django will use that name instead. If
+the directory with the given name doesn't exist, it will be created. You can
+use '.' to denote the current working directory.
+
+.. versionchanged:: 6.0
+
+    Automatic creation of the destination directory was added.
 
 For example:
 
@@ -1378,9 +1382,14 @@ If only the project name is given, both the project directory and project
 package will be named ``<projectname>`` and the project directory
 will be created in the current working directory.
 
-If the optional destination is provided, Django will use that existing
-directory as the project directory, and create ``manage.py`` and the project
-package within it. Use '.' to denote the current working directory.
+If the optional destination is provided, Django will use that name as the
+project directory, and create ``manage.py`` and the project package within it.
+If the directory with the given name doesn't exist, it will be created. Use '.'
+to denote the current working directory.
+
+.. versionchanged:: 6.0
+
+    Automatic creation of the destination directory was added.
 
 For example:
 
--- a/tests/admin_scripts/tests.py
+++ b/tests/admin_scripts/tests.py
@@ -2843,8 +2843,7 @@ class StartProject(LiveServerTestCase, AdminScriptTestCase):
 
     def test_custom_project_destination_missing(self):
         """
-        Make sure an exception is raised when the provided
-        destination directory doesn't exist
+        Create the directory when the provided destination directory doesn't exist.
         """
         template_path = os.path.join(custom_templates_dir, "project_template")
         args = [
@@ -2857,12 +2856,8 @@ class StartProject(LiveServerTestCase, AdminScriptTestCase):
         testproject_dir = os.path.join(self.test_dir, "project_dir2")
         out, err = self.run_django_admin(args)
         self.assertNoOutput(out)
-        self.assertOutput(
-            err,
-            "Destination directory '%s' does not exist, please create it first."
-            % testproject_dir,
-        )
-        self.assertFalse(os.path.exists(testproject_dir))
+        self.assertNoOutput(err)
+        self.assertTrue(os.path.exists(testproject_dir))
 
     def test_custom_project_template_with_non_ascii_templates(self):
         """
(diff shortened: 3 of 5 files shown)


browniebroke:
I _think_ the failing jobs are failing to start due to some infra issues: `java.io.IOException: No space left on device`

Going to assume they're unrelated, but let me know if I miss anything

docs/ref/django-admin.txt
```diff
@@ -1372,9 +1372,10 @@ If only the project name is given, both the project directory and project
...
 
-If the optional destination is provided, Django will use that existing
-directory as the project directory, and create ``manage.py`` and the project
-package within it. Use '.' to denote the current working directory.
+If the optional destination is provided, Django will use that name as the
+project directory, and create ``manage.py`` and the project package within it.
+If the directory with the given name doesn't exist, it will be created. Use '.'
+to denote the current working directory.
 
 For example:
```
ryancheley, line 1380: Can we call out in the example what this new feature is doing? Something like

For example:

    django-admin startapp myapp /Users/jezdez/Code/myapp/newdirectory

Will create a new directory called `newdirectory` where the app will have its code placed.

ryancheley:
@browniebroke this looks great. I added a few comments but overall 💯

docs/ref/django-admin.txt
browniebroke: I'm not sure this is a good example... The example shopws the `startproject` command, and as written currently, it implies that the project is being created at `/Users/jezdez/Code/myproject_repo`. 

I could however add a line below the command:

> It will create the directory `myproject_repo` if needed, where the project will have its code placed.

What do you think?

tests/admin_scripts/tests.py
```diff
@@ -3099,6 +3095,21 @@ def test_template(self):
                 content,
             )
 
+    def test_custom_app_destination_missing(self):
```
sarahboyce, line 3098: Please add more complex examples, such as nested directories which do not exist

django/core/management/templates.py
```diff
@@ -105,10 +105,10 @@ def handle(self, app_or_project, name, target=None, **options):
...
                 self.validate_name(os.path.basename(top_dir), "directory")
             if not os.path.exists(top_dir):
-                raise CommandError(
-                    "Destination directory '%s' does not "
-                    "exist, please create it first." % top_dir
-                )
+                try:
+                    os.makedirs(top_dir)
+                except OSError as e:
+                    raise CommandError(e)
```
sarahboyce, line 111: The error handling is missing a test

sarahboyce (commented):
Thank you! I have initial comments

Note that we need `.. versionchanged:: 6.0` notes for both startapp and startproject.
Also the tutorial currently asks the user to create the directory for the project. We may want to think about updating that

browniebroke:
> Also the tutorial currently asks the user to create the directory for the project. We may want to think about updating that

I've updated it, but now that I think about it, I'm wondering whether **now** is the right time to do so... If/when this is merged, the tutorial will be published at the `dev` version (e.g. https://docs.djangoproject.com/en/dev/intro/tutorial01/#creating-a-project) and I assume that some folks might land on this page before v6.0 is released, and would miss the "create directory" step.

Leaving this step in the tutorial still works in v6.0, but removing it would add friction for earlier versions.

django/core/management/templates.py
browniebroke: > Should I mock `os.makedirs` to raise a specific error to make the test more deterministic?

I've gave this a try

browniebroke: Well, that didn't work because the `self.run_django_admin` helper runs the command in a subprocess. As far as I undertsnad, the mock is applied in the current process, so doesn't trigger. I've tried another approach: by setting the parent directory to be read-only.

LilyFoote (approved):

django/core/management/templates.py
```diff
@@ -105,10 +105,10 @@ def handle(self, app_or_project, name, target=None, **options):
             if app_or_project == "app":
                 self.validate_name(os.path.basename(top_dir), "directory")
             if not os.path.exists(top_dir):
-                raise CommandError(
-                    "Destination directory '%s' does not "
-                    "exist, please create it first." % top_dir
-                )
+                try:
+                    os.makedirs(top_dir)
```
g-nie, line 109: I wonder if it makes sense to rename the `top_dir` variable to something more generic, since we support nested subdirectories now.

Hikari:
@browniebroke I checked the docs, ran the tests and tested it. It looks good to me.

TomHall2020 (changes requested):
Reviewing this:
### Documentation
- [x] Docs build, read ok, changes are reported in the tutorial and cli reference.
- [x] A search on the docs for `django-admin startproject` didn't come up with anything else that needs adjusting in my opinion.
- [x] Release note is included
- [x] I think it might be helpful to include something in the help text on the cli command itself to signal that it can create a missing destination, currently it has no changes and reads as follows:
```
positional arguments:
  name                  Name of the application or project.
  directory             Optional destination directory 
...
```

### Tests
- [x] Test suite is passing
- [x] New tests are written and look suitable.

### Code changes
- [x] Looks good, fairly small changes.

### PR
- [ ] Needs to be [one squashed commit](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/submitting-patches/#contribution-checklist)

django/core/management/templates.py
browniebroke: Humm, sure, I don't really mind... The command line argument is called `target`, so guess it could build on that:`target_dir` or maybe `target_path`? What do you think?

browniebroke:
> * I think it might be helpful to include something in the help text on the cli command itself to signal that it can create a missing destination

Great idea, added. Thanks!

django/core/management/templates.py
g-nie: Either sounds good to me, thanks!

TomHall2020:
All good for me now 👍

TomHall2020 (approved):

g-nie (approved):

browniebroke:
Rebased and squashed all the commits into one as per contribution guidelines

sarahboyce (commented):
Thank you @browniebroke and everyone for all the reviews :star: 

I was going through and when testing locally I tried to create an app within an app, this raised an error:
```
(test-django-package) sarahboyce@sarahboyce-ThinkPad-E15-Gen-4:~/Documents/Django-notes/Projects/test-django-package/mysite$ python manage.py startapp app75 app7/app75
  File "/home/sarahboyce/Documents/Django-notes/Projects/test-django-package/mysite/manage.py", line 22, in <module>
    main()
  File "/home/sarahboyce/Documents/Django-notes/Projects/test-django-package/mysite/manage.py", line 18, in main
  File "/home/sarahboyce/.pyenv/versions/test-django-package/lib/python3.12/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/home/sarahboyce/.pyenv/versions/test-django-package/lib/python3.12/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/sarahboyce/.pyenv/versions/test-django-package/lib/python3.12/site-packages/django/core/management/base.py", line 416, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/sarahboyce/.pyenv/versions/test-django-package/lib/python3.12/site-packages/django/core/management/base.py", line 460, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sarahboyce/.pyenv/versions/test-django-package/lib/python3.12/site-packages/django/core/management/commands/startapp.py", line 14, in handle
    super().handle("app", app_name, target, **options)
  File "/home/sarahboyce/.pyenv/versions/test-django-package/lib/python3.12/site-packages/django/core/management/templates.py", line 207, in handle
    with open(new_path, "w", encoding="utf-8") as new_file:
FileNotFoundError: [Errno 2] No such file or directory: '/home/sarahboyce/Documents/Django-notes/Projects/test-django-package/mysite/app7/app75/migrations/migrations/__init__.py'
```
Can you take a look?

browniebroke:
Nice catch. Quick question: did you rebase from main or did you only split the commit into 2? I'm pretty confident CI was green last night but it's now broken. I _think_ the split when wrong because the variable was renamed from `top_dir` -> `target_path` and then changed to `target_dir`, but `target_dir` is already used in `main` further down in the file:

https://github.com/django/django/blob/fbdbccf27a7d969a92e6ab8d2b7754e21e69b050/django/core/management/templates.py#L164

That might explain why we're getting `migrations` twice in the path `test-django-package/mysite/app7/app75/migrations/migrations/__init__.py`, as we're joining from the wrong base path? 

In the interest of keeping this PR focused, I'm tempted to keep the `top_dir` name? Open to suggestions for better names :)

browniebroke:
> I was going through and when testing locally I tried to create an app within an app

I've added a test to cover this specific case `test_custom_name_with_app_within_other_app` - let me know if you're happy with the approach or if you have any suggestions. Thanks!

sarahboyce (approved):
Thank you! :star:

PR #18390, opened 2024-07-18
Fixed #35601 - Added TelInput widget.

LucasEsposito:
# Trac ticket number

ticket-35601

# Branch description
Despite seeming functionally identical to standard `text` inputs, `tel` input offers some benefits:
* Mobile browsers may opt to present a custom keypad optimized for entering phone numbers.
* Validation for a specific format can be implemented with the `pattern` attribute
* Browsers that don't support type `tel` fall back to a standard `text` input.

# Checklist
- [x] I have attached screenshots in both light and dark modes for any UI changes. (N/A)


docs/ref/forms/widgets.txt
```diff
@@ -558,6 +558,15 @@ These widgets make use of the HTML elements ``input`` and ``textarea``.
...
 
+``TelInput``
+~~~~~~~~~~~~~~
+
+.. class:: TelInput
+
+    * ``input_type``: ``'tel'``
+    * ``template_name``: ``'django/forms/widgets/tel.html'``
+    * Renders as: ``<input type="tel" ...>``
+
```
claudep, line 569: Should we repeat here what MDN tells about phone validation?

`the input value is not automatically validated to a particular format before the form can be submitted, because formats for telephone numbers vary so much around the world.`

We might also point to third-party packages that do (server-side) validation (e.g. django-phonenumber-field).

LucasEsposito: Thanks for your review.

I applied both suggestions to the docs.

docs/releases/5.2.txt
```diff
@@ -165,7 +165,7 @@ File Uploads
 Forms
 ~~~~~
 
-* ...
+* Added :class:`~django.forms.TelInput` form widget that renders as ``<input type="tel" ...>``
```
sarahboyce, line 168:
```suggestion
* The new :class:`~django.forms.TelInput` form widget renders as 
  ``<input type="tel" ...>``. This is to be used for entering a telephone 
  number.
```
This should be in present tense 👍

docs/ref/forms/widgets.txt
```diff
@@ -558,6 +558,28 @@ These widgets make use of the HTML elements ``input`` and ``textarea``.
     * ``template_name``: ``'django/forms/widgets/url.html'``
     * Renders as: ``<input type="url" ...>``
 
+``TelInput``
+~~~~~~~~~~~~~~
+
```
sarahboyce, line 563:
```suggestion

.. versionadded:: 5.2

```

docs/ref/forms/widgets.txt
sarahboyce: I think we generally avoid pointing to third party packages in our docs (I guess due to concerns about those packages being maintained/secure etc).
I can see we do link [`django-statici18n`](https://django-statici18n.readthedocs.io/`) and [`django-debug-toolbar`](https://github.com/jazzband/django-debug-toolbar/) already, I am not 100% sure what our criteria is.
I just want to highlight that we should check django-phonenumber-field before we commit to it being linked in our docs 👍

sarahboyce (commented):
Thank you @lucasesposito ⭐
I have minor comments

docs/ref/forms/widgets.txt
claudep: https://pypi.org/project/django-phonenumber-field/#history shows that the package is currently well maintained and that there are several maintainers.

adamchainz: I also think it's okay to point to this package. It’s also not a huge endorsement, just a “such as”.

PR #18393, opened 2024-07-19
Fixed #30448, Fixed #35618 -- Do not close connection in atomic block.

andersk:
# Trac ticket number

ticket-30448

# Branch description

Previously, the `django.db.close_old_connections` handler for `request_started` and `request_finished` would incorrectly close any connection within a `transaction.atomic` block (such as the one automatically used by `django.db.TestCase`), leading to `InterfaceError: connection already closed`.

The test client and many of our tests have been working around this bug by manually suppressing the `close_old_connections` handler, but the workarounds are incomplete and the bug still affects other projects.  Fix the bug and remove the workarounds.

# Checklist


django/db/backends/base/base.py
```diff
@@ -593,7 +593,7 @@ def close_if_unusable_or_obsolete(self):
         Close the current connection if unrecoverable errors have occurred
         or if it outlived its maximum age.
         """
-        if self.connection is not None:
+        if self.connection is not None and not self.in_atomic_block:
```
charettes, line 596: I think the check should be a bit more complex and ensure that the connection is still usage. The connection might be still be considered in a transaction but actually unusable.

What about

```suggestion
        if self.connection is not None:
            if self.in_atomic_block and self.is_usable():
                self.health_check_done = True
                return
```

andersk: No, it’s not just that part. Within a `transaction.atomic` block, the `if self.get_autocommit() != self.settings_dict["AUTOCOMMIT"]` check fires even though the connection is perfectly fine.

charettes: We should find a way to close the transaction if we `self.in_atomic_block and not self.is_usable()` I don't think the patch can be accepted otherwise as otherwise it might leave dangling connections.

charettes: What about the following

```diff
diff --git a/django/db/backends/base/base.py b/django/db/backends/base/base.py
index cb8afb35ce..04eb7770fa 100644
--- a/django/db/backends/base/base.py
+++ b/django/db/backends/base/base.py
@@ -593,8 +593,15 @@ def close_if_unusable_or_obsolete(self):
         Close the current connection if unrecoverable errors have occurred
         or if it outlived its maximum age.
         """
-        if self.connection is not None and not self.in_atomic_block:
+        if self.connection is not None:
             self.health_check_done = False
+
+            # If the connection is in an atomic block avoid closing
+            # unless it's unusable.
+            if self.in_atomic_block and self.is_usable():
+                self.health_check_done = True
+                return
+
             # If the application didn't restore the original autocommit setting,
             # don't take chances, drop the connection.
             if self.get_autocommit() != self.settings_dict["AUTOCOMMIT"]:
diff --git a/tests/backends/base/test_base.py b/tests/backends/base/test_base.py
index a0fffd924d..6d1303522c 100644
--- a/tests/backends/base/test_base.py
+++ b/tests/backends/base/test_base.py
@@ -381,11 +381,19 @@ def test_set_autocommit_health_checks_enabled(self):
         connection.set_autocommit(True)
         self.assertIs(new_connection, connection.connection)

-    def test_no_close_in_atomic(self):
+    def test_no_close_usable_in_atomic(self):
         with transaction.atomic():
             connection.close_if_unusable_or_obsolete()
             self.run_query()

+    def test_close_unusable_in_atomic(self):
+        with (
+            transaction.atomic(),
+            patch.object(connection, "is_usable", return_value=False),
+        ):
+            connection.close_if_unusable_or_obsolete()
+            self.assertIs(True, connection.closed_in_transaction)
+

 class MultiDatabaseTests(TestCase):
     databases = {"default", "other"}
```

andersk: I still think we should never close a connection in an `atomic` block, even an unusable one. Otherwise we could silently lose data:

```python
with transaction.atomic():
    important_data.save()
    some_long_operation()  # connection might become unusable during this time
    connection.close_if_unusable_or_obsolete()  # closes connection?

# transaction is not committed, but execution continues silently!
# data is gone.
```

We need to leave the connection in the error state so that `transaction.atomic()` will notice the error when it tries to commit the transaction, and raise an exception.

charettes: > I still think we should never close a connection in an atomic block, even an unusable one. Otherwise we could silently lose data

I think you misunderstand what _usable_ mean. It means that we are no longer able to communicate with the server to send the equivalent of a `PING` command. This means that we won't ever be able to commit and thus the data should be considered lost at this point.

andersk: The data is lost either way. But we need to leave the connection in the error state so the data isn’t *silently* lost. We want an *exception* from `transaction.atomic()` in this case.

charettes: > We want an exception from transaction.atomic() in this case.

Is there any tests demonstrating that this wouldn't be the case? AFAIK closing a connection while in a transaction [sets `closed_in_transaction = True`](https://github.com/django/django/blob/9cb8baa0c4fa2c10789c5c8b65f4465932d4d172/django/db/backends/base/base.py#L358) which is then handled properly?

I think the problem we're trying to solve here is usable but possibly obsolete (due to `CONN_MAX_AGE`) that are in an atomic block being closed; we want to avoid that. The space of unusable in atomic block is already well covered.

andersk:
```python
    def test_close_in_atomic(self):
        with transaction.atomic():
                cursor.execute("CREATE TABLE important_table ()")
                cursor.execute("SELECT * FROM important_table")
            connection.close()

            cursor.execute("SELECT * FROM important_table")  # ProgrammingError: relation "important_table" does not exist
```

charettes: This test doesn't relate to what's being changed here though right since you're calling `connection.close` and not `close_if_unusable_or_obsolete`?

The fact that no errors are surfaced here at `Atomic.__exit__` time seems like a different problem altogether.

andersk: `transaction.atomic` “handles” `closed_in_transaction = True` by ignoring it.

https://github.com/django/django/blob/9cb8baa0c4fa2c10789c5c8b65f4465932d4d172/django/db/transaction.py#L237-L240

I’m not sure if that behavior is ideal, but altering that would be a bigger change. So we should at least make sure it isn’t triggered accidentally as a side effect of a cleanup function that runs automatically.

charettes: > I’m not sure if that behavior is ideal, but altering that would be a bigger change.

I completely agree with that, that's a much larger problem.

> So we should at least make sure it isn’t triggered accidentally as a side effect of a cleanup function that runs automatically.

Also checking whether or not the connection is _usable_ seems to check that box without compromising on the _contract_ of `close_if_unusable_or_obsolete`. Given we both agree that within a transaction but unusable is UB / loss of data scenario it seems like the proposed amendments are a safer way to handle callers that might expect that any usuable transaction is closed when calling this method (in a transaction or not).

andersk: This isn’t a contract issue. `close_if_unusable_or_obsolete` is [not a documented part](https://docs.djangoproject.com/en/5.0/search/?q=close_if_unusable_or_obsolete) of the Django API. I don’t expect anyone is manually calling it directly—certainly not inside an `atomic` block. The issue is that it’s called automatically when the `request_started` and `request_finished` signals are fired.

https://github.com/django/django/blob/9cb8baa0c4fa2c10789c5c8b65f4465932d4d172/django/db/__init__.py#L55-L63

I’m willing to go either way on whether it’s okay to silently lose data when someone manually closes a connection, but it’s absolutely not okay to silently lose data automatically.

charettes: > This isn’t a contract issue. close_if_unusable_or_obsolete is [not a documented part](https://docs.djangoproject.com/en/5.0/search/?q=close_if_unusable_or_obsolete) of the Django API.

It is since 5.1 through `close_old_connections` as it was imported and called for years in long running management commands (see ticket-35130 and 184d82d8488b1b31ade5b5a68b0040f1c267f2be). Also it's not because it isn't documented that we should not commit to a form of stability around it. Tons of third party apps rely on undocumented methods of the framework.

> I’m willing to go either way on whether it’s okay to silently lose data when someone manually closes a connection, but it’s absolutely not okay to silently lose data automatically.

By the time `close_if_unusable_or_obsolete` is called which only happens at `request_finished` time all `Atomic` block should have exited so I'm not sure under which scenario your silent data loss would occur in the first place. How can this realistically happen in a normal Django setup?

PR #18397, opened 2024-07-21
Fixed #25889 – organized tests and added summary docstrings.

wassef911:
# Trac ticket number
ticket-25889

# Branch description
codebase contains a lot of tests with random names pointing to Trac issues,
like **def test_ticket_9999** or **class Ticket9999Tests**

Although that's helpful, this makes navigating the tests easier/quicker and hopefully results in tests to be grouped "a little better" in the future...

I tried following an [older attempt](https://github.com/django/django/pull/7993)/ review left on it.
 

# Checklist


tests/defer_regress/tests.py
```diff
@@ -104,16 +104,21 @@ def test_basic(self):
...
         )
         self.assertIsInstance(
             list(SimpleItem.objects.annotate(Count("feature")).only("name")), list
         )
 
-    def test_ticket_23270(self):
+    def test_deferred_model_inheritance(self):
+        """
+        select_related on fields pointing to subclasses works when using defer (#23270).
+        """
```
sarahboyce, line 121:
```suggestion
    def test_defer_with_select_related_inherited_model(self):
```
Maybe we move this down next to the other select related tests for defer

tests/delete_regress/tests.py
```diff
@@ -280,15 +280,15 @@ def test_19187_values(self):
             Image.objects.values_list().delete()
 
 
-class Ticket19102Tests(TestCase):
+class DeletingUsingSubqueriesTests(TestCase):
```
sarahboyce, line 283:
```suggestion
class DeleteUsingSubqueryTests(TestCase):
```

tests/delete_regress/tests.py
```diff
@@ -299,7 +299,10 @@ def setUpTestData(cls):
         cls.l2 = Login.objects.create(description="l2", orgunit=cls.o2)
 
     @skipUnlessDBFeature("update_can_self_select")
-    def test_ticket_19102_annotate(self):
+    def test_delete_subquery_for_annotated(self):
+        """
+        Deletion using subqueries works for annotated queries (#19102).
+        """
```
sarahboyce, line 305:
```suggestion
    def test_annotate(self):
```
As the ticket number is in the class description and due to the name and docstring of the test class, I think I would remove all the doc strings added here and have shorter test names `test_extra`, `test_select_related`, `test_only`

tests/expressions/tests.py
```diff
@@ -476,14 +476,20 @@ def test_object_update_unsaved_objects(self):
         with self.assertRaisesMessage(ValueError, msg):
             acme.save()
 
-    def test_ticket_11722_iexact_lookup(self):
+    def test_lookups_with_f_has_valid_sql(self):
+        """
+        Query lookups that reference an F() expression produce valid sql (#11722).
+        """
```
sarahboyce, line 482:
```suggestion
    def test_filter_lookups_with_f_expression(self):
```

tests/expressions/tests.py
```diff
@@ -476,14 +476,20 @@ def test_object_update_unsaved_objects(self):
...
         test = Employee.objects.create(firstname="Test", lastname="test")
 
         queryset = Employee.objects.filter(firstname__iexact=F("lastname"))
         self.assertSequenceEqual(queryset, [test])
 
-    def test_ticket_16731_startswith_lookup(self):
+    def test_startswith_endswith_works_with_f_expression(self):
+        """
+        startswith endswith and contains works with F expression (#16731).
+        """
```
sarahboyce, line 492:
```suggestion
    def test_filter_startswith_with_f_expression(self):
```
(this test doesn't test endswith)

tests/expressions/tests.py
```diff
@@ -496,7 +502,10 @@ def test_ticket_16731_startswith_lookup(self):
         )
         self.assertSequenceEqual(qs, [e2, e3])
 
-    def test_ticket_18375_join_reuse(self):
+    def test_f_works_with_multiple_relations_1(self):
```
sarahboyce, line 505: I don't like adding 1, 2, 3, 4
Every test tests that something "works" so I would change that naming
I would either say `f_expression` or remove `f` as all tests are for `F` here

tests/fixtures_regress/tests.py
```diff
@@ -506,22 +506,20 @@ def test_loaddata_no_fixture_specified(self):
                 verbosity=0,
             )
 
-    def test_ticket_20820(self):
+    def test_loaddata_m2m(self):
```
sarahboyce, line 509:
```suggestion
    def test_loaddata_with_inherited_m2m(self):
```
I'd not refactor the docstrings in this file

sarahboyce (commented):
Thank you @wassef911 ⭐ 
This is big clean up and I like where it's going 

There's quite a lot to review here, I have left initial comments
I have not reviewed `tests/generic_relations_regress/tests.py`, `tests/prefetch_related/tests.py` or `tests/queries/tests.py` yet but hopefully this gives you a good start and you might be able to anticipate if I'd have any comments there 👍

tests/test_runner/tests.py
```diff
@@ -754,13 +754,13 @@ def run_suite(self, suite, **kwargs):
...
 
-class Ticket17477RegressionTests(AdminScriptTestCase):
+class TestManageHelp(AdminScriptTestCase):
     def setUp(self):
         super().setUp()
         self.write_settings("settings.py")
 
-    def test_ticket_17477(self):
-        """'manage.py help test' works after r16352."""
+    def test_manage_help_test(self):
```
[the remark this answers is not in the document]
wassef911: correct

tests/aggregation_regress/tests.py
```diff
@@ -1950,7 +1954,10 @@ def test_non_nullable_fk_not_promoted(self):
 
 
 class SelfReferentialFKTests(TestCase):
-    def test_ticket_24748(self):
+    def test_self_referential_fk_aggregate(self):
```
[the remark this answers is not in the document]
wassef911: yeah sure since we touched it let's relocate it somewhere that makes more sense. 

I've seen a bunch of cases like this, but I didn't have the courage for bigger diffs... let me know if I should swing at that?

tests/aggregation_regress/tests.py
```diff
@@ -1914,7 +1914,11 @@ def test_aggregate_group_by_unseen_columns_unmanaged(self):
 
 
 class JoinPromotionTests(TestCase):
-    def test_ticket_21150(self):
+    def test_select_related_with_annotation(self):
+        """
+        Select_related with annotation returns the correct results
+        and maintains related objects (#21150).
+        """
```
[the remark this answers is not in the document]
wassef911: true, but I think we should keep it to:
* Do not lose the issue number.
* Do not have a docstring for just the issue number.

tests/defer_regress/tests.py
```diff
@@ -104,16 +104,21 @@ def test_basic(self):
             list(SimpleItem.objects.annotate(Count("feature")).only("name")), list
         )
 
-    def test_ticket_16409(self):
-        # Regression for #16409 - make sure defer() and only() work with annotate()
+    def test_annotation_defer_and_only(self):
+        """
+        Make sure defer() and only() work with annotate() (#16409).
+        """
```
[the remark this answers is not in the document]
wassef911: could you expand on this, I don't get it 
github UI not showing any change, just different colors...

tests/delete_regress/tests.py
wassef911: I usually try to write test names as a full sentence,
I'll try to follow this instead.

tests/expressions/tests.py
wassef911: okay makes more sense.

tests/expressions/tests.py
```diff
@@ -496,7 +496,10 @@ def test_ticket_16731_startswith_lookup(self):
         )
         self.assertSequenceEqual(qs, [e2, e3])
 
-    def test_ticket_18375_join_reuse(self):
+    def test_f_expression_join_reuse(self):
+        """
+        F() works across multijoin relations, join reuse (#18375).
+        """
```
sarahboyce, line 502:
```suggestion
```
ticket number exists in the comment underneath

tests/expressions/tests.py
```diff
@@ -516,7 +522,10 @@ def test_ticket_18375_kwarg_ordering(self):
         )
         self.assertEqual(str(qs.query).count("JOIN"), 1)
 
-    def test_ticket_18375_kwarg_ordering_2(self):
+    def test_f_expression_join_reuse_with_pk(self):
+        """
+        F() works across multijoin relations with PKs (#18375).
+        """
```
sarahboyce, line 528:
```suggestion
```
Ticket number in existing comment

tests/expressions/tests.py
```diff
@@ -505,7 +508,10 @@ def test_ticket_18375_join_reuse(self):
         )
         self.assertEqual(str(qs.query).count("JOIN"), 1)
 
-    def test_ticket_18375_kwarg_ordering(self):
+    def test_f_expression_join_reuse_with_number_evaluation(self):
+        """
+        F() works across multijoin with number evaluation (gt/lt) (#18375).
+        """
```
sarahboyce, line 514:
```suggestion
```
Let's remove and instead update the end of the comment below
```
# other lookups could not reuse (#18375).
```

tests/expressions/tests.py
```diff
@@ -528,7 +537,10 @@ def test_ticket_18375_kwarg_ordering_2(self):
         )
         self.assertEqual(str(qs.query).count("JOIN"), 1)
 
-    def test_ticket_18375_chained_filters(self):
+    def test_f_expression_join_reuse_with_chained_filter(self):
+        """
+        F() works across multijoin relations with chained filters (#18375).
+        """
         # F() expressions do not reuse joins from previous filter.
```
sarahboyce, line 544:
```suggestion
        # F() expressions do not reuse joins from previous filter (#18375).
```

tests/generic_relations_regress/tests.py
```diff
@@ -174,14 +174,20 @@ def test_target_model_bool_false(self):
 
     @skipIfDBFeature("interprets_empty_strings_as_nulls")
     def test_gfk_to_model_with_empty_pk(self):
-        """Test related to #13085"""
+        """
+        GenericForeignKey saves to model instance
+        with an empty CharField PK (#13085).
+        """
```
sarahboyce, line 180: I think we can revert as is the ticket is about poorly named tests

tests/prefetch_related/tests.py
```diff
@@ -1655,15 +1655,16 @@ def setUpTestData(cls):
...
+        """Successful prefetch_related() call (#19607)."""
         list(
             WordEntry.objects.prefetch_related(
                 "lesson_entry", "lesson_entry__wordentry_set"
             )
         )
 
 
-class Ticket21410Tests(TestCase):
+class PrefetchRelatedSuccess2Tests(TestCase):
```
sarahboyce, line 1667: This shouldn't be called 2

I think looking at these, we actually might update to be a single test class `PrefetchRelatedSuccessTests` and have the stuff inside `setUpTestData` be inside each test and make the `list()` call actually assert it's value and give these two tests names as to why they are different
Even `PrefetchRelatedQueryTests` could be combined here as this `setUpTestData` doesn't really make sense if there is only one test for each

tests/aggregation_regress/tests.py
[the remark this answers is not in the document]
sarahboyce: Yes, but we still have issue numbers in commit messages and this is usually how we track the issue number rather than adding them into doc strings.
I don't mind too much if we already have a doc string and we had the number in the test name, but otherwise, the doc string needs to add value beyond a ticket number

tests/defer_regress/tests.py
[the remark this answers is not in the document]
sarahboyce: This is suggesting to delete these lines

sarahboyce (commented):
Thank you for the updates @wassef911 :star: this is really shaping up

I have a few more suggestions. I still haven't looked at `tests/queries/tests.py` at all but progressing with the other files

tests/aggregation_regress/tests.py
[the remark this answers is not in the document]
wassef911: I could loop over all docstring and remove the issue numbers, 
but then we'll have to change the coding style mentioned in the docs? 

https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#python-style

I don't think every character has to add value per say, but keeping more tests consistent with the style, will put pressure on whoever is writing future tests for follow it ?

tests/expressions/tests.py
wassef911: same as the above consistency.

tests/expressions/tests.py
wassef911: same

wassef911:
let's just close this already

PR #18399, opened 2024-07-22
Fixed #35622 -- Made unittest ignore Django asserts in tracebacks.

adamchainz:
# Trac ticket number

ticket-35622

# Branch description
Fixes the issue reported on the ticket, as tested on my example project: the `assertJSONEqual` frame no longer appears in the failure report and the debugger opens within the test. I added a unit test.

# Checklist

### DIFF
--- a/django/contrib/messages/test.py
+++ b/django/contrib/messages/test.py
@@ -1,5 +1,8 @@
 from .api import get_messages
 
+# Make unittest ignore frames in this module when reporting failures.
+__unittest = True
+
 
 class MessagesTestMixin:
     def assertMessages(self, response, expected_messages, *, ordered=True):
--- a/django/test/testcases.py
+++ b/django/test/testcases.py
@@ -67,6 +67,9 @@ __all__ = (
     "skipUnlessDBFeature",
 )
 
+# Make unittest ignore frames in this module when reporting failures.
+__unittest = True
+
 
 if not PY311:
     # Backport of unittest.case._enter_context() from Python 3.11.
--- a/tests/messages_tests/tests.py
+++ b/tests/messages_tests/tests.py
@@ -1,5 +1,7 @@
 import importlib
 import sys
+import traceback
+import unittest
 from unittest import mock
 
 from django.conf import settings
@@ -185,3 +187,17 @@ class AssertMessagesTest(MessagesTestMixin, SimpleTestCase):
         )
         with self.assertRaisesMessage(AssertionError, msg):
             self.assertMessages(response, [])
+
+    def test_method_frames_ignored_by_unittest(self):
+        response = FakeResponse()
+        try:
+            self.assertMessages(response, [object()])
+        except AssertionError:
+            exc_type, exc, tb = sys.exc_info()
+
+        result = unittest.TestResult()
+        result.addFailure(self, (exc_type, exc, tb))
+        stack = traceback.extract_tb(exc.__traceback__)
+        self.assertEqual(len(stack), 1)
+        # Top element in the stack is this method, not assertMessages.
+        self.assertEqual(stack[-1].name, "test_method_frames_ignored_by_unittest")
--- a/tests/test_utils/tests.py
+++ b/tests/test_utils/tests.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import threading
+import traceback
 import unittest
 import warnings
 from io import StringIO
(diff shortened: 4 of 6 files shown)


tests/test_utils/tests.py
```diff
@@ -1113,6 +1114,19 @@ def test_not_equal_parsing_errors(self):
...
 
+    def test_method_frames_ignored_by_unittest(self):
+        try:
+            self.assertJSONEqual("1", "2")
+        except AssertionError:
+            exc_type, exc, tb = sys.exc_info()
+
+        result = unittest.TestResult()
+        result.addFailure(self, (exc_type, exc, tb))
+        stack_summary = traceback.extract_tb(exc.__traceback__)
```
adamchainz, line 1125: https://docs.python.org/3.12/library/traceback.html#traceback.extract_tb

David-Wobrock (approved):

ngnpope (approved):

tests/messages_tests/tests.py
```diff
@@ -185,3 +187,20 @@ def test_mismatching_length(self):
         )
         with self.assertRaisesMessage(AssertionError, msg):
             self.assertMessages(response, [])
+
+    def test_method_frames_ignored_by_unittest(self):
+        response = FakeResponse()
+        add_message(response.wsgi_request, constants.INFO, "INFO message.")
+        try:
+            self.assertMessages(response, [])
```
nessita, line 195: For the sake of simplicity, I would suggest to do the following instead, to not distract the reader with a call (`add_message`) that is not strictly necessary for the correctness of the test:
```suggestion
        try:
            self.assertMessages(response, [object()])
```

django/contrib/messages/test.py
```diff
@@ -1,5 +1,8 @@
 from .api import get_messages
 
+# Make unittest ignore frames in this module when reporting failures
```
nessita, line 3:
```suggestion
# Make unittest ignore frames in this module when reporting failures.
```

django/test/testcases.py
```diff
@@ -67,6 +67,9 @@
     "skipUnlessDBFeature",
 )
 
+# Make unittest ignore frames in this module when reporting failures
```
nessita, line 70:
```suggestion
# Make unittest ignore frames in this module when reporting failures.
```

tests/messages_tests/tests.py
```diff
@@ -185,3 +187,20 @@ def test_mismatching_length(self):
...
+            exc_type, exc, tb = sys.exc_info()
+
+        result = unittest.TestResult()
+        result.addFailure(self, (exc_type, exc, tb))
+        stack_summary = traceback.extract_tb(exc.__traceback__)
+        self.assertEqual(len(stack_summary), 1)
+        # This method, not assertMessages
+        self.assertEqual(
+            stack_summary[-1].name, "test_method_frames_ignored_by_unittest"
+        )
```
nessita, line 206: Could we just call this `stack` and avoid line wrapping down below? :smile: 
```suggestion
        stack = traceback.extract_tb(exc.__traceback__)
        self.assertEqual(len(stack), 1)
        # Top element in the stack is this method, not assertMessages.
        self.assertEqual(stack[-1].name, "test_method_frames_ignored_by_unittest")
```

tests/test_utils/tests.py
```diff
@@ -1113,6 +1114,21 @@ def test_not_equal_parsing_errors(self):
...
+        self.assertEqual(len(stack_summary), 1)
+        # This method, not assertJSONEqual
```
nessita, line 1127: I would also adjust this comment as suggested above.

tests/test_utils/tests.py
```diff
@@ -1113,6 +1114,21 @@ def test_not_equal_parsing_errors(self):
...
 
+    def test_method_frames_ignored_by_unittest(self):
+        try:
+            self.assertJSONEqual("1", "2")
+        except AssertionError:
+            exc_type, exc, tb = sys.exc_info()
+
+        result = unittest.TestResult()
+        result.addFailure(self, (exc_type, exc, tb))
+        stack_summary = traceback.extract_tb(exc.__traceback__)
```
nessita, line 1125: If you agree to rename `stack` in the messages app, I would suggest we also rename here.

nessita (commented):
Thank you @adamchainz, this looks great! I learnt quite a few things while reviewing this work :star2: 

I have suggested a few nitpicks, let me know what you think. Also, did you consider to also define `__unitest = True` in `django/contrib/admin/tests.py`? Despite that module not containing publicly documented assertions, it does define a few helpers used within Django itself, such as `assertCountSeleniumElements`. On failure, I do get "distracting" frames in the traceback, e.g.:

* without `__unittest = True`:
```
FAIL: test_select (admin_views.test_autocomplete_view.SeleniumTests.test_select)
  File "/home/nessita/fellowship/django/tests/admin_views/test_autocomplete_view.py", line 457, in test_select
    self.assertCountSeleniumElements(
  File "/home/nessita/fellowship/django/django/contrib/admin/tests.py", line 181, in assertCountSeleniumElements
    self.assertEqual(
AssertionError: 21 != 22
```

* with `__unittest = True`:
```
FAIL: test_select (admin_views.test_autocomplete_view.SeleniumTests.test_select)
  File "/home/nessita/fellowship/django/tests/admin_views/test_autocomplete_view.py", line 457, in test_select
    self.assertCountSeleniumElements(
AssertionError: 21 != 22
```

I also think we may need a small release note for `5.2` since I can imagine Django users out there relying on the printed traceback in a given scenario to take further actions in their code base. IMHO, it will also be nice to call this cleanup out.

adamchainz:
Thanks for the review @nessita , adjusting now.

> Also, did you consider to also define `__unitest = True` in `django/contrib/admin/tests.py`?

I skipped it because it’s not public, but without much real consideration. I will add it.

nessita (approved):
Amazing, this looks great @adamchainz, thank you! :star2: 

I pushed a minor release note edit, will land when CI is green.

PR #18403, opened 2024-07-23
Fixed #35627 -- Raised a LookupError rather than an unhandled ValueError in get_supported_language_variant().

lorinkoz:
# Trac ticket number

ticket-35627

# Branch description
`get_supported_language_variant` can raise a `ValueError` when the language code is too long under certain circumstances. This error is not handled by any caller and can make e.g. `LocaleMiddleware` end up in a server error by constructing a URL that contains a thousand characters on the first path element.

This PR modifies the callers of `get_supported_language_variant` to not only catch `LookupError` but also `ValueError`.

# Checklist

### DIFF
--- a/docs/releases/5.0.8.txt
+++ b/docs/releases/5.0.8.txt
@@ -19,3 +19,7 @@ Bugfixes
 * Fixed a crash when creating a model with a ``Field.db_default`` and a
   ``Meta.constraints`` constraint composed of ``__endswith``, ``__startswith``,
   or ``__contains`` lookups (:ticket:`35625`).
+
+* Fixed a regression in Django 5.0.7 that caused a crash in
+  ``LocaleMiddleware`` when processing a language code over 500 characters
+  (:ticket:`35627`).
--- a/django/utils/translation/trans_real.py
+++ b/django/utils/translation/trans_real.py
@@ -517,7 +517,7 @@ def get_supported_language_variant(lang_code, strict=False):
                 # There is a generic variant under the maximum length accepted length.
                 lang_code = lang_code[:index]
             else:
-                raise ValueError("'lang_code' exceeds the maximum accepted length")
+                raise LookupError(lang_code)
         # If 'zh-hant-tw' is not supported, try special fallback or subsequent
         # language codes i.e. 'zh-hant' and 'zh'.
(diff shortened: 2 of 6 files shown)


sarahboyce:
Thank you for picking this up @lorinkoz ⭐ 

We're going to need a release note for Django 5.0 and 4.2 as we will do a backport (see this as maybe a similar-ish idea: 38e577993872da70c43468271d04d329aee7ec13)
We don't have release notes for Django 4.2.15 yet so you will need to make a new file starting with:
```
===========================
Django 4.2.15 release notes
===========================

*Expected August 6, 2024*

Django 4.2.15 fixes a regression in 4.2.14.

Bugfixes
========
```
and add `4.2.15` to `docs/releases/index.txt`

docs/releases/5.0.8.txt
```diff
@@ -15,3 +15,6 @@ Bugfixes
 * Fixed a regression in Django 5.0 where ``ModelAdmin.action_checkbox`` could
   break the admin changelist HTML page when rendering a model instance with a
   ``__html__`` method (:ticket:`35606`).
+
+* Fixed a regression in Django 5.0.7 where a long initial path fragment in the
+  URL would raise ``ValueError`` from the ``LocaleMiddleware``.
```
sarahboyce, line 20:
```suggestion
* Fixed a regression in Django 5.0.7 that caused a crash in
  ``LocaleMiddleware`` when processing a language code over 500 characters
  (:ticket:`35627`).
```
We need a ticket link (similar comment for 4.2.15)

sarahboyce (commented):
Thank you for the updates @lorinkoz 👍 
I discussed this with @nessita and we should update the error raised to be a `LookupError`
This would also mean that (looking at 9e9792228a6bb5d6402a5d645bc3be4cf364aefb) we need to update:
- `docs/ref/utils.txt` to say `LookupError` and `.. versionchanged:: 4.2.15`
- the `ValueError` asserts in `i18n.tests.MiscTests.test_get_supported_language_variant_real`

I don't think the original release notes for the CVE need to be updated so let's leave them
The tests you've added are great, please keep these ⭐

sarahboyce (approved):
Thank you @lorinkoz this looks good to me
Welcome on board ⛵

PR #18407, opened 2024-07-24
Fixed #35628 -- Allowed compatible GeneratedFields for ModelAdmin.date_hierarchy.

john-parton:
Fixed #35628, allow GeneratedField with appropriate output_field for date_hierarchy.

# Trac ticket number

This is the basic fix for https://code.djangoproject.com/ticket/35628# , but without tests/docs. Will add tests/docs if approach is deemed appropriate.

# Branch description
Essentially, there's a check to make sure that `date_hierarchy` is a `DateField` or `DateTimeField`, but there is no technical reason to prohibit a `GeneratedField` with `output_field` set to `DateField` or `DateTimeField`.

This adjusts the check to allow that specific case through.

# Checklist

### DIFF
--- a/django/contrib/admin/checks.py
+++ b/django/contrib/admin/checks.py
@@ -1184,7 +1184,7 @@ class ModelAdminChecks(BaseModelAdminChecks):
                     )
                 ]
             else:
-                if not isinstance(field, (models.DateField, models.DateTimeField)):
+                if field.get_internal_type() not in {"DateField", "DateTimeField"}:
                     return must_be(
                         "a DateField or DateTimeField",
                         option="date_hierarchy",
--- a/tests/modeladmin/test_checks.py
+++ b/tests/modeladmin/test_checks.py
@@ -4,16 +4,17 @@ from django.contrib.admin import BooleanFieldListFilter, SimpleListFilter
 from django.contrib.admin.options import VERTICAL, ModelAdmin, TabularInline
 from django.contrib.admin.sites import AdminSite
 from django.core.checks import Error
+from django.db import models
 from django.db.models import CASCADE, F, Field, ForeignKey, ManyToManyField, Model
 from django.db.models.functions import Upper
 from django.forms.models import BaseModelFormSet
-from django.test import SimpleTestCase
+from django.test import TestCase, skipUnlessDBFeature
 from django.test.utils import isolate_apps
 
 from .models import Band, Song, User, ValidationTestInlineModel, ValidationTestModel
 
 
-class CheckTestCase(SimpleTestCase):
+class CheckTestCase(TestCase):
     def assertIsInvalid(
         self,
         model_admin,
@@ -96,6 +97,29 @@ class RawIdCheckTests(CheckTestCase):
             raw_id_fields = ("users",)
 
         self.assertIsValid(TestModelAdmin, ValidationTestModel)
+
+    @isolate_apps("modeladmin")
+    def assertGeneratedDateTimeFieldIsValid(self, *, db_persist):
+        class TestModel(Model):
+            date = models.DateTimeField()
+            date_copy = models.GeneratedField(
+                expression=F("date"),
+                output_field=models.DateTimeField(),
+                db_persist=db_persist,
+            )
+
+        class TestModelAdmin(ModelAdmin):
+            date_hierarchy = "date_copy"
+
+        self.assertIsValid(TestModelAdmin, TestModel)
+
+    @skipUnlessDBFeature("supports_stored_generated_columns")
+    def test_valid_case_stored_generated_field(self):
+        self.assertGeneratedDateTimeFieldIsValid(db_persist=True)
+
+    @skipUnlessDBFeature("supports_virtual_generated_columns")
+    def test_valid_case_virtual_generated_field(self):
+        self.assertGeneratedDateTimeFieldIsValid(db_persist=False)
 
     def test_field_attname(self):
         class TestModelAdmin(ModelAdmin):
@@ -1028,6 +1052,33 @@ class DateHierarchyCheckTests(CheckTestCase):
             "The value of 'date_hierarchy' must be a DateField or DateTimeField.",
             "admin.E128",
         )
+
+    @isolate_apps("modeladmin")
+    def assertGeneratedIntegerFieldIsInvalid(self, *, db_persist):
+        class TestModel(Model):
+            generated = models.GeneratedField(
+                expression=models.Value(1),
+                output_field=models.IntegerField(),
+                db_persist=db_persist,
+            )
+
+        class TestModelAdmin(ModelAdmin):
+            date_hierarchy = "generated"
+
+        self.assertIsInvalid(
+            TestModelAdmin,
+            TestModel,
+            "The value of 'date_hierarchy' must be a DateField or DateTimeField.",
+            "admin.E128",
+        )
+
+    @skipUnlessDBFeature("supports_stored_generated_columns")
+    def test_related_invalid_field_type_stored_generated_field(self):
+        self.assertGeneratedIntegerFieldIsInvalid(db_persist=True)
+
+    @skipUnlessDBFeature("supports_virtual_generated_columns")
+    def test_related_invalid_field_type_virtual_generated_field(self):
+        self.assertGeneratedIntegerFieldIsInvalid(db_persist=False)
 
     def test_valid_case(self):
         class TestModelAdmin(ModelAdmin):
--- a/docs/releases/5.0.8.txt
+++ b/docs/releases/5.0.8.txt
@@ -24,3 +24,7 @@ Bugfixes
 * Fixed a regression in Django 5.0.7 that caused a crash in
   ``LocaleMiddleware`` when processing a language code over 500 characters
   (:ticket:`35627`).
+
+* Fixed a bug in Django 5.0 that caused a system check crash when
+  ``ModelAdmin.date_hierarchy`` was a ``GeneratedField`` with an
+  ``output_field`` of ``DateField`` or ``DateTimeField`` (:ticket:`35628`).


john-parton:
As an alternate implementation, we could do something like

```python
if field.get_internal_type() in {"DateField", "DateTimeField"}:
    ...
 ```
 
 This would also permit user defined date fields that don't explicitly subclass `DateField` or `DateTimeField` so long as the user defines the `get_internal_type` method.

File: django/contrib/admin/checks.py [hunk 1]
```diff
@@ -1184,16 +1184,21 @@ def _check_date_hierarchy(self, obj):
...
-                        obj=obj,
-                        id="admin.E128",
-                    )
-                else:
+                supported_fields = (models.DateField, models.DateTimeField)
+
+                if isinstance(field, supported_fields) or (
+                    isinstance(field, models.GeneratedField)
+                    and isinstance(field.output_field, supported_fields)
+                ):
```
charettes, line 1192: I think that checking for `output_field` like you did over `get_internal_type` is the better solution.

john-parton: Sounds good. I'll see if I can add tests/docs later today. Docs should be trivial. I'll reach out if I get stuck on tests.

sarahboyce: Sorry to follow up on this, as this is a release blocker we will need to have this merged by August 6th (Tuesday)
<details><summary>Here are some tests in case this helps</summary>

```diff
diff --git a/tests/modeladmin/test_checks.py b/tests/modeladmin/test_checks.py
index f767a6c92b..1206da2892 100644
--- a/tests/modeladmin/test_checks.py
+++ b/tests/modeladmin/test_checks.py
@@ -4,16 +4,17 @@ from django.contrib.admin import BooleanFieldListFilter, SimpleListFilter
 from django.contrib.admin.options import VERTICAL, ModelAdmin, TabularInline
 from django.contrib.admin.sites import AdminSite
 from django.core.checks import Error
+from django.db import models
 from django.db.models import CASCADE, F, Field, ForeignKey, ManyToManyField, Model
 from django.db.models.functions import Upper
 from django.forms.models import BaseModelFormSet
-from django.test import SimpleTestCase
+from django.test import TestCase, skipUnlessDBFeature
 from django.test.utils import isolate_apps
 
 from .models import Band, Song, User, ValidationTestInlineModel, ValidationTestModel
 
 
-class CheckTestCase(SimpleTestCase):
+class CheckTestCase(TestCase):
     def assertIsInvalid(
         self,
         model_admin,
@@ -1035,6 +1036,22 @@ class DateHierarchyCheckTests(CheckTestCase):
 
         self.assertIsValid(TestModelAdmin, ValidationTestModel)
 
+    @skipUnlessDBFeature("supports_stored_generated_columns")
+    @isolate_apps("modeladmin")
+    def test_valid_case_generated_field(self):
+        class TestModel(Model):
+            date = models.DateTimeField()
+            date_copy = models.GeneratedField(
+                expression=F("date"),
+                output_field=models.DateTimeField(),
+                db_persist=True,
+            )
+
+        class TestModelAdmin(ModelAdmin):
+            date_hierarchy = "date_copy"
+
+        self.assertIsValid(TestModelAdmin, TestModel)
+
     def test_related_valid_case(self):
         class TestModelAdmin(ModelAdmin):
             date_hierarchy = "band__sign_date"
@@ -1052,6 +1069,26 @@ class DateHierarchyCheckTests(CheckTestCase):
             "admin.E128",
         )
 
+    @skipUnlessDBFeature("supports_stored_generated_columns")
+    @isolate_apps("modeladmin")
+    def test_related_invalid_field_type_generated_field(self):
+        class TestModel(Model):
+            generated = models.GeneratedField(
+                expression=models.Value(1),
+                output_field=models.IntegerField(),
+                db_persist=True,
+            )
+
+        class TestModelAdmin(ModelAdmin):
+            date_hierarchy = "generated"
+
+        self.assertIsInvalid(
+            TestModelAdmin,
+            TestModel,
+            "The value of 'date_hierarchy' must be a DateField or DateTimeField.",
+            "admin.E128",
+        )
+
 
 class OrderingCheckTests(CheckTestCase):
```
</details>

john-parton: I believe this should work even with a 'virtual' column

Should I add tests for `supports_virtual_generated_columns` as well?

john-parton: @sarahboyce I took the example test cases you provided and fleshed them out so that it will run for both the db features `supports_stored_generated_columns` and `supports_virtual_generated_columns`, setting `db_persist` to the appropriate value.

In the future, if there's a database backend that supports BOTH features, then the test will only be run for the db_persist == False case.

If we would rather, I can create separate methods for each case and have them be thin wrappers around another assert statement.

Maybe something like this

```python
def assertGeneratedFieldIsValid(self, *, db_persist):

    class TestModel(Model):
        date = models.DateTimeField()
        date_copy = models.GeneratedField(
            expression=F("date"),
            output_field=models.DateTimeField(),
            db_persist=db_persist,
        )

    class TestModelAdmin(ModelAdmin):
        date_hierarchy = "date_copy"

    self.assertIsValid(TestModelAdmin, TestModel)

@skipUnlessDBFeature("supports_stored_generated_columns")
@isolate_apps("modeladmin")
def test_valid_case_virtual_generated_field(self):
    self.assertGeneratedFieldIsValid(db_persist=True)

@skipUnlessDBFeature("supports_virtual_generated_columns")
@isolate_apps("modeladmin")
def test_valid_case_stored_generated_field(self):
    self.assertGeneratedFieldIsValid(db_persist=False)
```

Let me know if that general approach is preferred and I'll update the pull request.

john-parton:
I added a release note.

Is there a way that I can render out those pages locally so that I can make sure they're formatted nicely?

I don't think it's necessary to add a 'version changed' note to the date_hierarchy docs. I think a typical user of the `GeneratedField` will just assume it's going to work.

django/contrib/admin/checks.py
sarahboyce: I like the `assertGeneratedFieldIsValid` pattern :heart_eyes: 
This has been something I've found quite difficult to manage in tests in general

We might be ok just having the one though as it is enough to give us coverage for the `GeneratedField` case (Simon is more familiar with ORM test handling than me so he might have an opinion also)

sarahboyce:
> I added a release note.

Thank you! This is a special case where the release note belongs in the 5.0.8 release notes (as this is a release blocker for a monthly release of 5.0). Something like
```
* Fixed a bug in Django 5.0 where a system check would crash when
  ``ModelAdmin.date_hierarchy`` was set to a ``GeneratedField`` with an
  ``output_field`` of a ``DateField`` or a ``DateTimeField``
  (:ticket:`35628`).
```
You might need to rebase with the latest main

> Is there a way that I can render out those pages locally so that I can make sure they're formatted nicely?

Yes, see [this guide](https://docs.djangoproject.com/en/5.0/internals/contributing/writing-documentation/#build-the-documentation-locally) also have the requirements of `docs/requirements.txt` installed :+1: 

> I don't think it's necessary to add a 'version changed' note to the date_hierarchy docs. I think a typical user of the `GeneratedField` will just assume it's going to work.

Correct :+1:

john-parton:
- Rewrote the tests using the assertion pattern discussed previously
- Moved release note to 5.0.8 bugfixes section
- Squashed my commit and cherry-picked it against main

django/contrib/admin/checks.py
john-parton: I think this version is actually more readable and technically more "correct". Having the random "assert" in the previous version was a bad smell.

john-parton:
Fixed linter errors, squashed, force-pushed.

sarahboyce (approved):
Thank you @john-parton :star: this looks good to me

pauloxnet (approved):

pauloxnet:
LGTM

tests/modeladmin/test_checks.py
```diff
@@ -97,6 +98,30 @@ class TestModelAdmin(ModelAdmin):
 
         self.assertIsValid(TestModelAdmin, ValidationTestModel)
 
+    def assertGeneratedDateTimeFieldIsValid(self, *, db_persist):
```
charettes, line 101: Any thoughts about decorating `assertGeneratedDateTimeFieldIsValid` with `isolate_apps` instead of the callers since it's the one that actually creates and registers a model?

```suggestion
    @isolate_apps("modeladmin")
    def assertGeneratedDateTimeFieldIsValid(self, *, db_persist):
```

tests/modeladmin/test_checks.py
```diff
@@ -97,6 +98,30 @@ class TestModelAdmin(ModelAdmin):
...
+                db_persist=db_persist,
+            )
+
+        class TestModelAdmin(ModelAdmin):
+            date_hierarchy = "date_copy"
+
+        self.assertIsValid(TestModelAdmin, TestModel)
+
+    @skipUnlessDBFeature("supports_stored_generated_columns")
+    @isolate_apps("modeladmin")
```
charettes, line 116:
```suggestion
```

tests/modeladmin/test_checks.py
```diff
@@ -97,6 +98,30 @@ class TestModelAdmin(ModelAdmin):
...
+    def test_valid_case_stored_generated_field(self):
+        self.assertGeneratedDateTimeFieldIsValid(db_persist=True)
+
+    @skipUnlessDBFeature("supports_virtual_generated_columns")
+    @isolate_apps("modeladmin")
```
charettes, line 121:
```suggestion
```

tests/modeladmin/test_checks.py
```diff
@@ -1029,6 +1054,34 @@ class TestModelAdmin(ModelAdmin):
...
+
+        self.assertIsInvalid(
+            TestModelAdmin,
+            TestModel,
+            "The value of 'date_hierarchy' must be a DateField or DateTimeField.",
+            "admin.E128",
+        )
+
+    @skipUnlessDBFeature("supports_stored_generated_columns")
+    @isolate_apps("modeladmin")
```
charettes, line 1076:
```suggestion
```

tests/modeladmin/test_checks.py
```diff
@@ -1029,6 +1054,34 @@ class TestModelAdmin(ModelAdmin):
...
+    def test_related_invalid_field_type_stored_generated_field(self):
+        self.assertGeneratedIntegerFieldIsInvalid(db_persist=True)
+
+    @skipUnlessDBFeature("supports_virtual_generated_columns")
+    @isolate_apps("modeladmin")
```
charettes, line 1081:
```suggestion
```

tests/modeladmin/test_checks.py
```diff
@@ -1029,6 +1054,34 @@ class TestModelAdmin(ModelAdmin):
             "admin.E128",
         )
 
+    def assertGeneratedIntegerFieldIsInvalid(self, *, db_persist):
```
charettes, line 1057:
```suggestion
    @isolate_apps("modeladmin")
    def assertGeneratedIntegerFieldIsInvalid(self, *, db_persist):
```

charettes (approved):

tests/modeladmin/test_checks.py
sarahboyce: Good idea :+1:

PR #18411, opened 2024-07-25
Updated Sphinx source_suffix setting to use dictionary mapping.

smithdc1:
# Trac ticket number

N/A

# Branch description
Since Sphinx 1.8 this setting should be a mapping of file extensions to file types.

https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-source_suffix

# Checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -182,9 +182,9 @@ pygments_style = "trac"
 # Links to Python's docs should reference the most recent version of the 3.x
 # branch, which is located at this URL.
 intersphinx_mapping = {
-    "python": ("https://docs.python.org/3/", None),
-    "sphinx": ("https://www.sphinx-doc.org/en/master/", None),
-    "psycopg": ("https://www.psycopg.org/psycopg3/docs/", None),
+    "python": ("https://docs.python.org/3", None),
+    "sphinx": ("https://www.sphinx-doc.org/en/master", None),
+    "psycopg": ("https://www.psycopg.org/psycopg3/docs", None),
 }
 
 # Python's docs don't change every week.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -94,7 +94,7 @@ spelling_warning = True
 # templates_path = []
 
 # The suffix of source filenames.
-source_suffix = ".txt"
+source_suffix = {".txt": "restructuredtext"}
 
 # The encoding of source files.
 # source_encoding = 'utf-8-sig'


nessita:
Hi @smithdc1, thank you for this! I have read the docs and I can see how this change makes sense :star2: 

I'm happy to approve and merge, but I'd like to understand the following: from the docs, it seems that a plain `str` value is still supported and I couldn't find a deprecation note or similar. May I ask what is the motivation for this change? Is it preemptively updating to avoid a future missed deprecation?

smithdc1:
> May I ask what is the motivation for this change?

Sure!

When using Sphinx 8+, a new message is printed to the console. This patch means the message is no longer shown. 

```
Converting `source_suffix = '.txt'` to `source_suffix = {'.txt': 'restructuredtext'}`.
```

nessita:
> > May I ask what is the motivation for this change?
> 
> Sure!
> 
> When using Sphinx 8+, a new message is printed to the console. This patch means the message is no longer shown.
> 
> ```
> Converting `source_suffix = '.txt'` to `source_suffix = {'.txt': 'restructuredtext'}`.
> ```

Thank you! I reproduced by upgrading to Sphinx 8.0.2. I also see these new messages:
```
loading intersphinx inventory 'python' from https://docs.python.org/3//objects.inv ...
loading intersphinx inventory 'sphinx' from https://www.sphinx-doc.org/en/master//objects.inv ...
loading intersphinx inventory 'psycopg' from https://www.psycopg.org/psycopg3/docs//objects.inv ...
intersphinx inventory has moved: https://www.sphinx-doc.org/en/master//objects.inv -> https://www.sphinx-doc.org/en/master/objects.inv
```
which are harmless but made me notice that we could apply this cleanup as well:
```diff
diff --git a/docs/conf.py b/docs/conf.py
index 0c633397ac..2c578d5a8d 100644
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -182,9 +182,9 @@ pygments_style = "trac"
 # Links to Python's docs should reference the most recent version of the 3.x
 # branch, which is located at this URL.
 intersphinx_mapping = {
-    "python": ("https://docs.python.org/3/", None),
-    "sphinx": ("https://www.sphinx-doc.org/en/master/", None),
-    "psycopg": ("https://www.psycopg.org/psycopg3/docs/", None),
+    "python": ("https://docs.python.org/3", None),
+    "sphinx": ("https://www.sphinx-doc.org/en/master", None),
+    "psycopg": ("https://www.psycopg.org/psycopg3/docs", None),
 }
 
 # Python's docs don't change every week.
```
Would you fancy making a second commit with that cleanup?

nessita:
> Would you fancy making a second commit with that cleanup?

Actually since I had this diff already in my working copy, I pushed it as a new commit. Will merge when CI is green, thank you!

PR #18412, opened 2024-07-25
Fixed #35586 -- Added support for set-returning database functions.

devin13cox:
# Trac ticket number
ticket-35586

# Branch description
This PR introduces the `set_returning` parameter for Expressions that will enforce subquery evaluation for instances that aggregation optimization would early-exit, not anticipating added rows. This will specifically resolve issues like Postgres "jsonb_path_query" function returning additional items that do not get picked up and result in problems like incorrect counts. This code is based on recommendations by @charettes.

# Checklist

### DIFF
--- a/django/db/models/sql/query.py
+++ b/django/db/models/sql/query.py
@@ -491,6 +491,11 @@ class Query(BaseExpression):
             )
             or having
         )
+        set_returning_annotations = {
+            alias
+            for alias, annotation in self.annotation_select.items()
+            if getattr(annotation, "set_returning", False)
+        }
         # Decide if we need to use a subquery.
         #
         # Existing aggregations would cause incorrect results as
@@ -510,6 +515,7 @@ class Query(BaseExpression):
             or qualify
             or self.distinct
             or self.combinator
+            or set_returning_annotations
         ):
             from django.db.models.sql.subqueries import AggregateQuery
 
@@ -551,6 +557,9 @@ class Query(BaseExpression):
                         if annotation.get_group_by_cols():
                             annotation_mask.add(annotation_alias)
                     inner_query.set_annotation_mask(annotation_mask)
+                    # Annotations that possibly return multiple rows cannot
+                    # be masked as they might have an incidence on the query.
+                    annotation_mask |= set_returning_annotations
 
             # Add aggregates to the outer AggregateQuery. This requires making
             # sure all columns referenced by the aggregates are selected in the
--- a/docs/ref/models/expressions.txt
+++ b/docs/ref/models/expressions.txt
@@ -1095,6 +1095,16 @@ calling the appropriate methods on the wrapped expression.
         :py:data:`NotImplemented` which forces the expression to be computed on
         the database.
 
+    .. attribute:: set_returning
+
+    .. versionadded:: 5.2
+
+        Tells Django that this expression contains a set-returning function,
+        enforcing subquery evaluation. It's used, for example, to allow some
+        Postgres set-returning functions (e.g. ``JSONB_PATH_QUERY``,
+        ``UNNEST``, etc.) to skip optimization and be properly evaluated when
+        annotations spawn rows themselves. Defaults to ``False``.
+
     .. method:: resolve_expression(query=None, allow_joins=True, reuse=None, summarize=False, for_save=False)
 
         Provides the chance to do any preprocessing or validation of
--- a/docs/releases/5.2.txt
+++ b/docs/releases/5.2.txt
@@ -218,6 +218,10 @@ Models
 * Added support for validation of model constraints which use a
   :class:`~django.db.models.GeneratedField`.
 
+* The new :attr:`.Expression.set_returning` attribute specifies that the
+  expression contains a set-returning function, enforcing subquery evaluation.
+  This is necessary for many Postgres set-returning functions.
+
 Requests and Responses
 ~~~~~~~~~~~~~~~~~~~~~~
 
--- a/tests/annotations/models.py
+++ b/tests/annotations/models.py
@@ -58,3 +58,11 @@ class Company(models.Model):
 class Ticket(models.Model):
     active_at = models.DateTimeField()
     duration = models.DurationField()
+
+
+class JsonModel(models.Model):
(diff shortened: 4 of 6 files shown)


tests/annotations/models.py
```diff
@@ -58,3 +58,8 @@ class Company(models.Model):
 class Ticket(models.Model):
     active_at = models.DateTimeField()
     duration = models.DurationField()
+
+
+class JsonModel(models.Model):
+    data = models.JSONField(default=dict, blank=True)
+    id = models.IntegerField(primary_key=True)
```
sarahboyce, line 65:
```suggestion
    id = models.IntegerField(primary_key=True)
    
    class Meta:
        required_db_features = {"supports_json_field"}
```

tests/annotations/tests.py
```diff
@@ -1167,6 +1170,27 @@ def test_alias_forbidden_chars(self):
                 with self.assertRaisesMessage(ValueError, msg):
                     Book.objects.annotate(**{crafted_alias: Value(1)})
 
+    @skipUnless(connection.vendor == "postgresql", "PostgreSQL tests")
+    def test_set_returning_functions(self):
+        """Test that utilizing the set returning flag returns the expected count"""
```
sarahboyce, line 1175:
```suggestion
    @skipUnlessDBFeature("supports_json_field")
    def test_set_returning_functions(self):
```

docs/ref/models/expressions.txt
```diff
@@ -1094,6 +1094,14 @@ calling the appropriate methods on the wrapped expression.
         to apply a function over an empty result set. Defaults to
         :py:data:`NotImplemented` which forces the expression to be computed on
         the database.
+    
+    .. attribute:: set_returning
```
sarahboyce, line 1098:
```suggestion
    .. attribute:: set_returning

    .. versionadded:: 5.2
```

docs/releases/5.2.txt
```diff
@@ -70,6 +70,9 @@ Minor features
 :mod:`django.contrib.postgres`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+* Expressions now contain a ``set_returning`` parameter that enforces
+  subquery evaluation, necessary for many Postgres set-returning functions.
+
```
sarahboyce, line 75: This should be in the `Models` section and maybe
```
* The new :attr:`.Expression.set_returning` attribute specifies that the
  expression contains a set-returning function, enforcing subquery evaluation.
  This is necessary for many Postgres set-returning functions.
```

django/db/models/sql/query.py
```diff
@@ -551,6 +557,9 @@ def get_aggregation(self, using, aggregate_exprs):
                         if annotation.get_group_by_cols():
                             annotation_mask.add(annotation_alias)
                     inner_query.set_annotation_mask(annotation_mask)
+                    # Annotations that possibly return multiple rows cannot
+                    # be masked as they might have an incidence on the query
```
sarahboyce, line 561:
```suggestion
                    # Annotations that possibly return multiple rows cannot
                    # be masked as they might have an incidence on the query.
```
(nit)

devin13cox:
> Thank you @devin13cox ⭐ I have initial comments

Thanks @sarahboyce! Went ahead and implemented your suggestions.

tests/annotations/tests.py
devin13cox: @sarahboyce I think this is causing an issue since `jsonb_path_query` is a postgres-specific (to my understanding) function. So even if other databases support JSONField, they won't necessarily support the `jsonb_path_query`. Would it make sense to revert to the initial `skipUnless` or do you prefer adding onto the `skipUnlessDBFeature` to add in something specific to that function. Thanks!

sarahboyce: Sorry I wasn't trying to remove the initial `skipUnless`, I think we want both

docs/releases/5.2.txt
```diff
@@ -70,6 +70,7 @@ Minor features
 :mod:`django.contrib.postgres`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+
```
sarahboyce, line 73:
```suggestion
```

tests/annotations/tests.py
devin13cox: @sarahboyce my bad, misread the suggestion and though it included the removal. I'll tack it back on, thanks!

docs/ref/models/expressions.txt
```diff
@@ -1094,6 +1094,16 @@ calling the appropriate methods on the wrapped expression.
...
+
+    .. versionadded:: 5.2
+
+        Tells Django that this expression contains a set-returning function, 
+        enforcing subquery evaluation. It's used, for example, to allow some
+        Postgres set-returning functions (e.g. JSONB_PATH_QUERY, UNNEST, etc.)
```
charettes, line 1104:
```suggestion
        Postgres set-returning functions (e.g. ``JSONB_PATH_QUERY``, ``UNNEST``, etc.)
```

tests/annotations/tests.py
```diff
@@ -1167,6 +1170,28 @@ def test_alias_forbidden_chars(self):
                 with self.assertRaisesMessage(ValueError, msg):
                     Book.objects.annotate(**{crafted_alias: Value(1)})
 
+    @skipUnless(connection.vendor == "postgresql", "PostgreSQL tests")
+    @skipUnlessDBFeature("supports_json_field")
+    def test_set_returning_functions(self):
+        """Test that utilizing the set returning flag returns the expected count"""
```
charettes, line 1176: The doc string is redundant with the test itself

```suggestion
```

tests/annotations/tests.py
```diff
@@ -1167,6 +1170,28 @@ def test_alias_forbidden_chars(self):
...
+        from django.db.models import JSONField
+
```
charettes, line 1178: Move to the top, it can always be imported even when the backend doesn't support using it.

charettes (approved):
Left some comments about cosmetic changes but otherwise LGTM!

sarahboyce (approved):
Thank you @devin13cox :star:

PR #18341, opened 2024-07-04
Fixed #35546 -- Updated  contributing pages to align with current requirement of having an accepted ticket

mharyam:
# Trac ticket number
ticket-35546

# Branch description
Following some feedback from PR https://github.com/django/django/pull/17389 mentioned in the TRAC ticket, I have checked and updated the places where we need to be more explicit about the need to have an accepted ticket when working on changes.

Most of the updates involve rewording paragraphs to ensure this requirement is clearly stated.

I am happy to remove any misleading paragraphs if they are unnecessary, as we now explicitly mention the need to add a TRAC ticket for any kind of patches (whether trivial or not) in several places.

# Checklist
- [x] This PR targets the `main` branch. 
- [x] The commit message is written in past tense, mentions the ticket number, and ends with a period.
- [x] I have checked the "Has patch" ticket flag in the Trac system.
- [x] I have added or updated relevant tests.
- [x] I have added or updated relevant docs, including release notes if applicable.
- [ ] I have attached screenshots in both light and dark modes for any UI changes.

PR #18418, opened 2024-07-27
Referenced joining the triage and review team as motivation to do PR reviews.

sarahboyce:
Thank you @mharyam for this (pulled from #18341) - this was a proposed update after reviewing the contributing docs

The role of "merger" is not often awarded to volunteers, this was suggested to be updated to "Review and Triage team" member.

We could also say something very generic like "If you're looking to be awarded a role of responsibility". This would still include the original role of merger but also things like the "Accessibility Team" - as I think PR reviews would also be a good way to be noticed by the accessibility team also.

I wouldn't want to delete the sentence as it is the only place I think we encourage PR reviews.

What do you think? Are you in favor of updating this or shall we leave it?

### DIFF
--- a/docs/internals/contributing/writing-code/submitting-patches.txt
+++ b/docs/internals/contributing/writing-code/submitting-patches.txt
@@ -288,8 +288,11 @@ appropriate flags on the Trac ticket based on the results of your review:
 "Patch needs improvement", "Needs documentation", and/or "Needs tests". As time
 and interest permits, mergers do final reviews of "Ready for checkin" tickets
 and will either commit the changes or bump it back to "Accepted" if further
-work needs to be done. If you're looking to become a merger, doing thorough
-reviews of contributions is a great way to earn trust.
+work needs to be done.
+
+If you're looking to become a member of the `triage & review team
+<https://www.djangoproject.com/foundation/teams/#triage-review-team>`_, doing
+thorough reviews of contributions is a great way to earn trust.
 
 Looking for a patch to review? Check out the "Patches needing review" section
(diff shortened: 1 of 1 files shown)


docs/internals/contributing/writing-code/submitting-patches.txt
```diff
@@ -288,8 +288,10 @@ appropriate flags on the Trac ticket based on the results of your review:
 "Patch needs improvement", "Needs documentation", and/or "Needs tests". As time
 and interest permits, mergers do final reviews of "Ready for checkin" tickets
 and will either commit the changes or bump it back to "Accepted" if further
-work needs to be done. If you're looking to become a merger, doing thorough
-reviews of contributions is a great way to earn trust.
+work needs to be done.
+
+If you're looking to become a member of the Review and Triage team, doing
```
jacobtylerwalls, line 293: minor nit: seems to be "Triage and Review" everywhere else

RealOrangeOne (approved):
This change makes sense to me. 

The "a role of responsibility" extension might cause a little more confusion, as there are likely roles which have nothing to do with code (eg forum / discord moderator), where reviewing PRs doesn't gain much. Similarly, some roles (read: [teams](https://www.djangoproject.com/foundation/teams)) aren't always open to new members by design (eg security and ops).

docs/internals/contributing/writing-code/submitting-patches.txt
sarahboyce: Thank you! Updated :star:

jacobtylerwalls (approved):

nessita (approved):
Looks good, thanks! :star2:

PR #18419, opened 2024-07-27
Fixed #35459 Refactor Subquery and Case Classes: Remove **extra Argum…

priyank-panchal:
# Trac ticket number
[ [ticket-35459
](url)](https://code.djangoproject.com/ticket/35459)

# Branch description
Refactor Expression Classes: Remove **extra and Add Validation for Subquery and Case.

# Checklist


tests/expressions_case/tests.py
```diff
@@ -722,6 +722,22 @@ def test_filter_with_aggregation_in_predicate(self):
...
+            template="CASE %(cases)s ELSE %(default)s + 5 END",
+        )
+        self.assertListEqual(
+            list(
+                CaseTestModel.objects.annotate(values=case_expression).values_list(
+                    "values", flat=True
+                )
+            ),
+            [10, 20, 5, 20, 5, 5, 5],
+        )
```
jacobtylerwalls, line 739: The other tests in this file seem to favor `assertQuerySetEqual`, so I suggest:
```suggestion
        self.assertQuerySetEqual(
            CaseTestModel.objects.annotate(values=case_expression).order_by("pk"),
            [10, 20, 5, 20, 5, 5, 5],
            transform=attrgetter("values"),
        )
```

tests/expressions/tests.py
```diff
@@ -855,6 +855,24 @@ def test_subquery_filter_by_lazy(self):
...
         self.assertEqual(qs.get(), self.gmbh)
 
+    def test_subquery_with_custom_template(self):
+        companies = Company.objects.annotate(
+            ceo_manager_count=Subquery(
+                Employee.objects.filter(
+                    lastname=OuterRef("ceo__lastname"),
+                ).values("manager"),
+                template="(SELECT count(*) FROM (%(subquery)s) _count)",
+            )
```
jacobtylerwalls, line 865: I'm not seeing the use of `**extra` here?

tests/expressions_case/tests.py
```diff
@@ -722,6 +722,22 @@ def test_filter_with_aggregation_in_predicate(self):
             transform=itemgetter("integer", "integer2", "max"),
         )
 
+    def test_case_with_extra_kwargs(self):
+        case_expression = Case(
+            When(integer=1, then=Value(10)),
+            When(integer=2, then=Value(20)),
+            default=Value(0),
```
jacobtylerwalls, line 729: `default` is already defined as a keyword argument, so to really test `**extra`, let's use an example that instead of `default` uses a custom keyword.

priyank-panchal:
Thank you for the guidance, @jacobtylerwalls

tests/expressions/tests.py
```diff
@@ -855,6 +855,22 @@ def test_subquery_filter_by_lazy(self):
...
+                salary=20,
+                template="(SELECT salary FROM (%(subquery)s) _subquery "
+                "WHERE salary = %(salary)s)",
+            ),
+        )
+        expected_companies = Company.objects.filter(name="Foobar Ltd.")
+        self.assertQuerySetEqual(
+            custom_subquery.order_by("name"),
+            expected_companies.order_by("name"),
+            transform=lambda x: x,
```
jacobtylerwalls, line 871:
```suggestion
```

jacobtylerwalls (commented):
Thanks for the updates! I'll unset the "patch needs improvement" flag on Trac although there is one cleanup opportunity I noticed from the latest push:

sarahboyce:
Thank you for taking this on @priyank-panchal :star: 

Have you seen [Tim's comment in the ticket](https://code.djangoproject.com/ticket/35459#comment:13)?
I think we need to first decide if we should deprecate `extra` given that the more we point it out in the docs, the more likely that we will encourage folks to use it and a decision to remove it might be harder

I think the next step would be starting a discussion in the [Django forum](https://forum.djangoproject.com/c/internals/orm/9) which explains the ticket and ask folks if there are any strong use cases for why folks might want to use `extra`, and add your opinion
If there's agreement that we should deprecate, you can look at our [deprecation guide](https://docs.djangoproject.com/en/5.1/internals/contributing/writing-code/submitting-patches/#deprecating-a-feature) for how to update this PR

PR #18424, opened 2024-07-29
Fixed #35440 -- Simplified parse_header_parameters by leveraging stdlid's Message.

chudilo:
Updated parse_header_parameters to leverage the parsing logic from (stdlib) email Message implementation. Limited the number of parameters that parsed by default to two.

# Trac ticket number

ticket-35440

# Branch description

In this branch I've changed code of the `parse_header_parameters` function with logic elements from python `email.utils` module implementation. This is mainly done through the use of the `decode_params` function from email.utils, and then converted to the old format.

:exclamation:  There is one difference from the current implementation: logic of `decode_params` functions [always](https://github.com/python/cpython/blob/main/Lib/email/utils.py#L432) unquote  in encoding in params values, even if it doesn't correct due to "wrongly formatted RFC 2231 headers" (shown in tests):

```python
#cgi and old implementations
>>> from django.utils.http import parse_header_parameters
>>> parse_header_parameters("Content-Type: application/x-stuff; title*='This%20is%20%2A%2A%2Afun%2A%2A%2A")
('content-type: application/x-stuff', {'title': "'This%20is%20%2A%2A%2Afun%2A%2A%2A"})
>>> from cgi import parse_header
parse_header("Content-Type: application/x-stuff; title*='This%20is%20%2A%2A%2Afun%2A%2A%2A")

#email and new implementations
>>> from email.message import Message
>>> m = Message()
>>> m["Content-Type"] = "application/x-stuff; title*='This%20is%20%2A%2A%2Afun%2A%2A%2A"
>>> m.get_params()
[('application/x-stuff', ''), ('title', (None, None, "'This is ***fun***"))]
>>> from django.utils.http import parse_header_parameters
>>> parse_header_parameters("Content-Type: application/x-stuff; title*='This%20is%20%2A%2A%2Afun%2A%2A%2A")
('content-type: application/x-stuff', {'title': "'This is ***fun***"})
```

And I do not see it possible to revert such a conversion of the param value to the initial state when using the `decode_params` function. 

```python
>>> s = "'This%20is%20%2A%2A%2Afun%2A%2A%2A"
>>> s1 = urllib.parse.quote(urllib.parse.unquote(s, encoding="latin-1") , encoding="latin-1")
>>> s1
'%27This%20is%20%2A%2A%2Afun%2A%2A%2A'
>>> s1 == s
False
```

So this can be fixed by copying functionality of decode_params into the django code base, with change of the unqote part, to lead current logic to the old implementation, but I'm not sure this is necessary.

### Parameters limitations

> The Security Team also agreed that it's worth adding some early checks in the `parse_header_parameters` function to limit the amount of provided semicolons. This would require some investigation as to what would be a good threshold, considering that it's likely that more than one semicolon may not be necessary in valid HTTP headers.

In this branch I've added an optional parameter `limit` to the function `parse_header_parameters` with default value=`2`. That means that current usages of this function in django project will return only first `limit` parameters for the value, for example: 

```python
>>> s = 'attachment; one=one; two=two; three=three'
>>> from cgi import parse_header
>>> parse_header(s)
('attachment', {'one': 'one', 'two': 'two', 'three': 'three'})

>>> from django.utils.http import parse_header_parameters
>>> parse_header_parameters(s)
('attachment', {'one': 'one', 'two': 'two'})
>>> parse_header_parameters(s, limit=3)
('attachment', {'one': 'one', 'two': 'two', 'three': 'three'})
>>> parse_header_parameters(s, 0)
('attachment', {})
```
The choice of the value 2 was justified by existing examples of using several parameters, like in [Content-Disposition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) header. Despite the  [resolution](https://datatracker.ietf.org/doc/html/rfc7231) that allows Content-Type header has multiple parameters without limitations `media-type = type "/" subtype *( OWS ";" OWS parameter )`, it's really hard to find any real usage examples.

It is also worth noting that the task description talks about pre-counting semicolons. But since there may be semicolons inside the parameters, it is not possible to determine in advance which of them are inside the parameters. Therefore, the checks are based on the number of parsed parameters.

p.s. This is my first pull request, so any comments or suggestions are greatly welcomed.

# Checklist

### DIFF
--- a/django/utils/http.py
+++ b/django/utils/http.py
@@ -3,8 +3,9 @@ import re
 import unicodedata
 from binascii import Error as BinasciiError
 from datetime import UTC, datetime
-from email.utils import formatdate
-from urllib.parse import quote, unquote
+from email.message import Message
+from email.utils import collapse_rfc2231_value, formatdate
+from urllib.parse import quote
 from urllib.parse import urlencode as original_urlencode
 from urllib.parse import urlsplit
 
@@ -24,6 +25,7 @@ ETAG_MATCH = _lazy_re_compile(
     re.X,
 )
 
+MAX_HEADER_LENGTH = 10_000
 MONTHS = "jan feb mar apr may jun jul aug sep oct nov dec".split()
 __D = r"(?P<day>[0-9]{2})"
 __D2 = r"(?P<day>[ 0-9][0-9])"
@@ -310,46 +312,28 @@ def escape_leading_slashes(url):
     return url
 
 
-def _parseparam(s):
-    while s[:1] == ";":
-        s = s[1:]
-        end = s.find(";")
-        while end > 0 and (s.count('"', 0, end) - s.count('\\"', 0, end)) % 2:
-            end = s.find(";", end + 1)
-        if end < 0:
-            end = len(s)
-        f = s[:end]
-        yield f.strip()
-        s = s[end:]
-
-
-def parse_header_parameters(line):
+def parse_header_parameters(line, max_length=MAX_HEADER_LENGTH):
     """
     Parse a Content-type like header.
     Return the main content-type and a dictionary of options.
+
+    If `line` is longer than `max_length`, `ValueError` is raised.
     """
-    parts = _parseparam(";" + line)
-    key = parts.__next__().lower()
+    if max_length is not None and line and len(line) > max_length:
+        raise ValueError("Unable to parse header parameters (value too long).")
+
+    m = Message()
+    m["content-type"] = line
+    params = m.get_params()
+
     pdict = {}
-    for p in parts:
-        i = p.find("=")
-        if i >= 0:
-            has_encoding = False
-            name = p[:i].strip().lower()
-            if name.endswith("*"):
-                # Lang/encoding embedded in the value (like "filename*=UTF-8''file.ext")
-                # https://tools.ietf.org/html/rfc2231#section-4
-                name = name[:-1]
-                if p.count("'") == 2:
-                    has_encoding = True
-            value = p[i + 1 :].strip()
-            if len(value) >= 2 and value[0] == value[-1] == '"':
-                value = value[1:-1]
-                value = value.replace("\\\\", "\\").replace('\\"', '"')
-            if has_encoding:
-                encoding, lang, value = value.split("'")
-                value = unquote(value, encoding=encoding)
-            pdict[name] = value
+    key = params.pop(0)[0].lower()
+    for name, value in params:
+        if not name:
+            continue
+        if isinstance(value, tuple):
+            value = collapse_rfc2231_value(value)
+        pdict[name] = value
     return key, pdict
 
 
--- a/docs/releases/6.0.txt
+++ b/docs/releases/6.0.txt
@@ -311,6 +311,10 @@ Miscellaneous
 * The :ref:`JSON <serialization-formats-json>` serializer now writes a newline
   at the end of the output, even without the ``indent`` option set.
 
+* The undocumented ``django.utils.http.parse_header_parameters()`` function is
+  refactored to use Python's :py:class:`email.message.Message` for parsing.
+  Input headers exceeding 10000 characters will now raise :exc:`ValueError`.
+
 .. _deprecated-features-6.0:
 
 Features deprecated in 6.0
--- a/tests/utils_tests/test_http.py
+++ b/tests/utils_tests/test_http.py
@@ -6,6 +6,7 @@ from unittest import mock
 from django.test import SimpleTestCase
 from django.utils.datastructures import MultiValueDict
 from django.utils.http import (
+    MAX_HEADER_LENGTH,
     base36_to_int,
     content_disposition_header,
     escape_leading_slashes,
@@ -424,6 +425,8 @@ class EscapeLeadingSlashesTests(unittest.TestCase):
 class ParseHeaderParameterTests(unittest.TestCase):
     def test_basic(self):
         tests = [
+            ("", ("", {})),
+            (None, ("none", {})),
             ("text/plain", ("text/plain", {})),
             ("text/vnd.just.made.this.up ; ", ("text/vnd.just.made.this.up", {})),
             ("text/plain;charset=us-ascii", ("text/plain", {"charset": "us-ascii"})),
@@ -448,8 +451,16 @@ class ParseHeaderParameterTests(unittest.TestCase):
                 ("attachment", {"filename": "strange;name", "size": "123"}),
             ),
             (
+                'attachment; filename="strange;name";;;;size=123;;;',
+                ("attachment", {"filename": "strange;name", "size": "123"}),
+            ),
+            (
                 'form-data; name="files"; filename="fo\\"o;bar"',
                 ("form-data", {"name": "files", "filename": 'fo"o;bar'}),
+            ),
+            (
+                'form-data; name="files"; filename="\\"fo\\"o;b\\\\ar\\""',
+                ("form-data", {"name": "files", "filename": '"fo"o;b\\ar"'}),
             ),
         ]
         for header, expected in tests:
@@ -480,12 +491,13 @@ class ParseHeaderParameterTests(unittest.TestCase):
         """
         Test wrongly formatted RFC 2231 headers (missing double single quotes).
         Parsing should not crash (#24209).
+        But stdlib email still decodes (#35440).
         """
         test_data = (
             (
                 "Content-Type: application/x-stuff; "
                 "title*='This%20is%20%2A%2A%2Afun%2A%2A%2A",
-                "'This%20is%20%2A%2A%2Afun%2A%2A%2A",
+                "'This is ***fun***",
             ),
             ("Content-Type: application/x-stuff; title*='foo.html", "'foo.html"),
             ("Content-Type: application/x-stuff; title*=bar.html", "bar.html"),
@@ -493,6 +505,37 @@ class ParseHeaderParameterTests(unittest.TestCase):
         for raw_line, expected_title in test_data:
             parsed = parse_header_parameters(raw_line)
             self.assertEqual(parsed[1]["title"], expected_title)
+
+    def test_header_max_length(self):
+        base_header = "Content-Type: application/x-stuff; title*="
+        base_header_len = len(base_header)
+
+        test_data = [
+            (MAX_HEADER_LENGTH, {}),
+            (MAX_HEADER_LENGTH, {"max_length": None}),
+            (MAX_HEADER_LENGTH + 1, {"max_length": None}),
+            (100, {"max_length": 100}),
+        ]
+        for line_length, kwargs in test_data:
+            with self.subTest(line_length=line_length, kwargs=kwargs):
+                title = "x" * (line_length - base_header_len)
+                line = base_header + title
+                assert len(line) == line_length
+
+                parsed = parse_header_parameters(line, **kwargs)
+
+                expected = ("content-type: application/x-stuff", {"title": title})
+                self.assertEqual(parsed, expected)
+
+    def test_header_too_long(self):
+        test_data = [
+            ("x" * (MAX_HEADER_LENGTH + 1), {}),
+            ("x" * 101, {"max_length": 100}),
+        ]
+        for line, kwargs in test_data:
+            with self.subTest(line_length=len(line), kwargs=kwargs):
+                with self.assertRaises(ValueError):
+                    parse_header_parameters(line, **kwargs)
 
 
 class ContentDispositionHeaderTests(unittest.TestCase):


nessita:
Hello @chudilo! Thank you so much for your work on this PR :star2: 

When we created ticket-35440, the intention was to fully replace the existing parsing logic in the Django codebase with calls to the Message class and its methods (specifically, to remove most of the custom parsing logic and use `Message()` and `get_params()` instead). The goal was to avoid potential security issues and duplication of logic from the Python standard library.

Have you considered this approach? Did you encounter any blockers?

chudilo:
There are no blocks, just kind of overthinking while figuring out how stdlib works.

I proposed an implementation of `parse_header_parameters` with usage of `Message()`, but there are some questions here, that can help if any fixes are going to be needed:

-  Is current implementation of converting `get_params` result to old format ok, or intention of this ticket was to remove `parse_header_parameters` completely? 
- As mentioned in description, it's hard to make _early checks_ in the `parse_header_parameters` function to limit the amount of provided semicolons (because of possible inner semicolons), without duplicating old parsing part especially. If current implementation needs rework, please provide some details of expected result.

p.s. Should I rewrite the "Branch description" section due to fixes?

chudilo:
Hello @nessita! In case this task is not on the review radar, should I do additional actions in order to get this ticket reviewed?

nessita:
> Hello @nessita!

Hello @chudilo! Thank you for your fixes and the further explanations. Answers to your questions:

> In case this task is not on the review radar, should I do additional actions in order to get this ticket reviewed?

This is in the review radar, thank you for asking. You can double-check by seeing the "patches needing review" section of the [Django Developer Dahsboard](https://dashboard.djangoproject.com/). Your ticket/PR are there so we'll get to it as soon as possible.

Please take into consideration that Django has only 1.5 (paid) Fellows doing reviews (and other priority work such as handling security reports and fixing release blockers bugs). There are also community members (volunteers) that contribute their time and energy when they can, so sometimes the process can feel a bit slow. But we'll get there!

>     * Is current implementation of converting `get_params` result to the old format ok, or intention of this ticket was to remove `parse_header_parameters` completely?

I can't answer this without first reviewing the PR in depth and experimenting with the code. But, I can say that the goal of the ticket is to remove as much parsing logic as realistically possible without breaking or changing the semantics for the cases that make sense. What I mean is that, ideally, we would remove as much as custom parsing as possible without introducing breaking changes.

>     * As mentioned in description, it's hard to make _early checks_ in the `parse_header_parameters` function to limit the amount of provided semicolons (because of possible inner semicolons), without duplicating old parsing part especially. If current implementation needs rework, please provide some details of expected result.

Understood, thanks for the clarification. Me or any other reviewer will keep this in mind when reviewing this work.

> p.s. Should I rewrite the "Branch description" section due to fixes?

A small update could be beneficial for future readers if there is something in there that no longer holds.

I hope the above helps!
Thank you for your patience.

bcail:
@chudilo What do you think of making a change like this to your PR?
```
$ gitd
diff --git a/django/utils/http.py b/django/utils/http.py
index 991c5d1235..d6901c66db 100644
--- a/django/utils/http.py
+++ b/django/utils/http.py
@@ -311,18 +311,10 @@ def escape_leading_slashes(url):
     return url
 
 
-def _parseparam(s):
-    end = s.find(";")
-    while end > 0 and (s.count('"', 0, end) - s.count('\\"', 0, end)) % 2:
-        end = s.find(";", end + 1)
-    if end < 0:
-        end = len(s)
-    return s[:end].strip().lower(), s[end:]
-
-
 def _cgi_compatible_params(params, limit):
     pdict = {}
-    params = list(filter(lambda x: x[0], params))[:limit]
+    key = params[0][0].strip().lower()
+    params = list(filter(lambda x: x[0], params[1:]))[:limit]
     for name, value in params:
         if isinstance(value, tuple):
             encoding, lang, value = value
@@ -331,7 +323,7 @@ def _cgi_compatible_params(params, limit):
                 value = unquote(quote(value, encoding="latin-1"), encoding=encoding)
         pdict[name] = value
 
-    return pdict
+    return key, pdict

 def parse_header_parameters(line, limit=2):
@@ -339,11 +331,10 @@ def parse_header_parameters(line, limit=2):
     Parse a Content-type like header.
     Return the main content-type and a dictionary of options.
     """
-    key, params = _parseparam(line)
     m = Message()
-    m["content-type"] = params
+    m["content-type"] = line
 
-    pdict = _cgi_compatible_params(m.get_params(), limit)
+    key, pdict = _cgi_compatible_params(m.get_params(), limit)
 
     return key, pdict
```

chudilo:
Hello @bcail! Your suggestion is good and I added it to this PR. Thank you for your review.

rtpg:
Here's a Hypothesis test script that tests the current code against the old code, finding a good amount of changes:
```
# /// script
# dependencies = [
#   "hypothesis",
# ]
# ///
from hypothesis import assume, given

from email.message import Message
from urllib.parse import quote, unquote

from hypothesis.strategies import text


def _cgi_compatible_params(params, limit):
    pdict = {}
    key = params.pop(0)[0].strip().lower()
    params = list(filter(lambda x: x[0], params))[:limit]
    for name, value in params:
        if isinstance(value, tuple):
            encoding, lang, value = value
            # email.message params always unquoted with "latin-1"
            if encoding:
                value = unquote(quote(value, encoding="latin-1"), encoding=encoding)
        pdict[name] = value

    return key, pdict


def new_parse_header_parameters(line, limit=2):
    """
    Parse a Content-type like header.
    Return the main content-type and a dictionary of options.
    """
    m = Message()
    m["content-type"] = line

    key, pdict = _cgi_compatible_params(m.get_params(), limit)
    return key, pdict


def _parseparam(s):
    while s[:1] == ";":
        s = s[1:]
        end = s.find(";")
        while end > 0 and (s.count('"', 0, end) - s.count('\\"', 0, end)) % 2:
            end = s.find(";", end + 1)
        if end < 0:
            end = len(s)
        f = s[:end]
        yield f.strip()
        s = s[end:]


def parse_header_parameters(line):
    """
    Parse a Content-type like header.
    Return the main content-type and a dictionary of options.
    """
    parts = _parseparam(";" + line)
    key = parts.__next__().lower()
    pdict = {}
    for p in parts:
        i = p.find("=")
        if i >= 0:
            has_encoding = False
            name = p[:i].strip().lower()
            if name.endswith("*"):
                # Lang/encoding embedded in the value (like "filename*=UTF-8''file.ext")
                # https://tools.ietf.org/html/rfc2231#section-4
                name = name[:-1]
                if p.count("'") == 2:
                    has_encoding = True
            value = p[i + 1 :].strip()
            if len(value) >= 2 and value[0] == value[-1] == '"':
                value = value[1:-1]
                value = value.replace("\\\\", "\\").replace('\\"', '"')
            if has_encoding:
                encoding, lang, value = value.split("'")
                value = unquote(value, encoding=encoding)
            pdict[name] = value
    return key, pdict


@given(text())
def test_parameter_parsing(s):
    # I can write out assume to eliminate "known" isses
    assume(s != "=")
    assume(len(s) > 0)
    assume(s[0] != ";")
    assume(s[-1] != "=")
    assert parse_header_parameters(s) == new_parse_header_parameters(s)


if __name__ == "__main__":
    test_parameter_parsing()
```

(the above can be run with `uv run`, like `uv run regression_test.py`, or `uv run pytest` if you have hypothesis + pytest installed)

rtpg:
With the above found the following kind of errors:
- `;0` parsed as `('', {})` in the old, `('', {'0': {}})` in the new
- `=` parsed  as `('=', {})` in the old, `('', {})` in the new
- `0;0` parsed as `('0', {})` in the old, `('0', {'0': ''})` in the new

nessita:
> With the above found the following kind of errors:
> 
>     * `;0` parsed as `('', {})` in the old, `('', {'0': {}})` in the new
> 
>     * `=` parsed  as `('=', {})` in the old, `('', {})` in the new
> 
>     * `0;0` parsed as `('0', {})` in the old, `('0', {'0': ''})` in the new

Thank you so much @rtpg for the extra testing. I believe these changes are acceptable considering that they are corner cases (and arguably invalid `Content-Type` headers), and also because this method is undocumented.

django/utils/http.py
```diff
@@ -310,46 +311,31 @@ def escape_leading_slashes(url):
...
+def _cgi_compatible_params(params, limit):
+    pdict = {}
+    key = params.pop(0)[0].strip().lower()
+    params = list(filter(lambda x: x[0], params))[:limit]
+    for name, value in params:
+        if isinstance(value, tuple):
+            encoding, lang, value = value
+            # email.message params always unquoted with "latin-1"
+            if encoding:
+                value = unquote(quote(value, encoding="latin-1"), encoding=encoding)
```
nessita, line 323: I was reading the documentation for Message.get_param, and I noticed that it shows:

> If your application doesn't care whether the parameter was RFC 2231
> encoded, it can turn the return value into a string as follows:
> ...
> param = email.utils.collapse_rfc2231_value(rawparam)

I tried replacing this if with a call like this:
```suggestion
        value = collapse_rfc2231_value(value)
```

No test failed and in fact the values are properly decoded. Do you see a reason not to use this instead of doing the manual quote/unquote calls?

nessita (changes requested):
Thank you for working on this @chudilo! And thank you for your patience waiting for a review :star2: 

I have reviewed the changes, and overall they look good. I have added an inline question but also I would like the changes to be refactored into something like this:

```python
def parse_header_parameters(line, limit=2):
    """
    Parse a Content-type like header.
    Return the main content-type and a dictionary of options.
    """
    m = Message()
    m["content-type"] = line
    params = m.get_params()

    # XXX: could params be None?

    key = params.pop(0)[0].lower()
    pdict = {}
    for name, value in params:
        if not name:
            continue
        if len(pdict) >= limit:
            break
        pdict[name] = collapse_rfc2231_value(value)
    return key, pdict
```

Also, I think the limiting of the amount of values should be done in a separated commit, presumably with extra feedback from the security team (which I can reach out to).

Also, this branch needs a small release note in the `Miscellaneous` section of the 5.2.txt file. For inspiration, we should add something similar to what was [added for 4.2.txt](https://docs.djangoproject.com/en/dev/releases/4.2/#miscellaneous):

```
* The undocumented ``django.utils.http.parse_header_parameters()`` function was
  refactored to leverage its implementation to the
  :py:class:`email.message.Message` Python class. 
```

So, in summary:
* inquiry on using `collapse_rfc2231_value` needs answering 
* code needs a slight refactor
* a release note is needed

nessita:
Hello @chudilo, would you have any availability to keep working on this? It's OK if you can't, but I wanted to ask before closing the PR. Thanks!

chudilo:
Hello! Sorry for such a long delay. I will try to keep working on this at this week, otherwise it fair to close this pr. Thanks for understanding

chudilo:
Hello, @nessita! I made the requested changes, with following notes:

> inquiry on using `collapse_rfc2231_value`

Use of this function is helpfull and I added it. 
But for compatibility with previous versions of code, I use `if isinstance(value, tuple):` check, because suggested functions unquotes non-tuple values, and cgi's `parse_header` doesn't. You can see this in the added test:

```
('form-data', {'name': 'files', 'filename': 'fo"o;b\\ar'}) # only collapse_rfc2231_value 
('form-data', {'name': 'files', 'filename': '"fo"o;b\\ar"'}) # with check
```

> code needs a slight refactor
 
Refactor was done, code was rebased from upstream django repository. 

>  could `params` be None?

It's never None and always has at least one parameter in array because of `m["content-type"] = line`. 

There is another question: old code always throw and exception if the `line` parameter is not a string. New code can work with everything (that is shown in the new test with `none`). **Should we keep it like this?**

> a release note is needed 

Release note was added to [5.2.txt](https://docs.djangoproject.com/en/dev/releases/5.2/#miscellaneous). In case it should be in another version or wording, I will fix it.

> presumably with extra feedback from the security team (which I can reach out to)

I will be glad if you can help!

Thank you for review!

nessita:
@chudilo Hello! This is in great shape, thank you so much for working on it. I will ask for further review from the security team, as mentioned before.

I will give a final review to this PR on March, since I'll off for two weeks on holiday. Thank you again! :star2:

django/utils/http.py
```diff
@@ -3,8 +3,9 @@
 import unicodedata
 from binascii import Error as BinasciiError
 from datetime import datetime, timezone
-from email.utils import formatdate
-from urllib.parse import quote, unquote
+from email.message import Message
+from email.utils import collapse_rfc2231_value, formatdate
+from urllib.parse import quote
```
nessita, line 8: I would move this import along with the existing `from urllib.parse import urlsplit`, in general we rather not have multiple lines importing from the same module.

chudilo: It was actually a fix of pre-commit, and pipeline fails when I try to get imports in one line. Three imports are listed in alphabetical order.

I do not think I can actually remove `urlencode as original_urlencode` construction, so it may be the best version of imports.

adamchainz: Yes, if you drop the `as` import then isort (under pre-commit) will re-combine the imports.

shaib:
Hi @chudilo ,

I am a member of the Security Team, I came here (relatively late in the life of the PR...) at @nessita's request.

The PR in general looks good, thanks for your work! However, there's a bit of an issue with the Parameters Limitation part. It is not your fault, see below -- but it does need to be addressed.

> ### Parameters limitations
> 
> > The Security Team also agreed that it's worth adding some early checks in the `parse_header_parameters` function to limit the amount of provided semicolons. This would require some investigation as to what would be a good threshold, considering that it's likely that more than one semicolon may not be necessary in valid HTTP headers.
> 
> [...]
> It is also worth noting that the task description talks about pre-counting semicolons. But since there may be semicolons inside the parameters, it is not possible to determine in advance which of them are inside the parameters. Therefore, the checks are based on the number of parsed parameters.

It was not explained before, but the reason the team wanted these checks, is that the time to parse the headers grows quadratically with the number of semicolons in the header string (this was true in the old code, and still holds with the new code). In theory, this quadratic growth could be the subject of a DOS attack -- sending a request with a bogus header which will take a long time to process. In practice, the factors are such that it takes a header whose length is at least tens of thousands, to make a real dent. Such headers should normally be caught by a web server sitting in front of Django, so there is no real security issue here; but we would like to harden Django against such attacks, in the name of "defense in depth", if it doesn't cost too much.

Since the issue is with the time of parsing the header, there is little use in limiting the number of fields extracted after the parsing is already completed. Since, as you noted, it is hard to count the semicolons correctly without parsing, we suggest instead to limit the total length of the header to 10_000.

Thanks again for your good work!

django/utils/http.py
nessita: Dropping the `as` would be possible with some refactor but that would be a separated change that the one in this branch. For now I think we will have to live with the multiple imports for the same lib. Thanks both!

nessita:
@bcail @rtpg @adamchainz @shaib Tagging you in case you want to take another look. I'll merge tomorrow morning, and we can always land tweaks with future PRs given that we have plenty of time until 6.0 feature freeze.

rtpg:
looks good to me 👍

nessita:
> LGTM. The only thing I'd fix is the typo in the commit message :)

Sorry about that :see_no_evil:

PR #18425, opened 2024-07-30
Fixed #35638 -- Updated validate_constraints to consider db_default.

shangxiao:
ticket-35638

 - Rename & update `_get_field_value_map() -> _get_field_expression_map()` to avoid wrapping expressions in `Value`
 - ~Update constraint validation to make consideration for expressions like `DatabaseDefault`~
   - Prefer using Simon's suggestion of updating `DatabaseDefault` to selectively use either the keyword `DEFAULT` or the underlying default expression depending on whether it's an insert or otherwise
 - Add tests for:
   - check constraints
   - unique with fields defined 
   - unique with expression defined
   - exclusion constraints

checklist

### DIFF
--- a/django/db/models/base.py
+++ b/django/db/models/base.py
@@ -1333,12 +1333,17 @@ class Model(AltersData, metaclass=ModelBase):
             setattr(self, cachename, obj)
         return getattr(self, cachename)
 
-    def _get_field_value_map(self, meta, exclude=None):
+    def _get_field_expression_map(self, meta, exclude=None):
         if exclude is None:
             exclude = set()
         meta = meta or self._meta
         field_map = {
-            field.name: Value(getattr(self, field.attname), field)
+            field.name: (
+                value
+                if (value := getattr(self, field.attname))
+                and hasattr(value, "resolve_expression")
+                else Value(value, field)
+            )
             for field in meta.local_concrete_fields
             if field.name not in exclude and not field.generated
         }
--- a/django/db/models/expressions.py
+++ b/django/db/models/expressions.py
@@ -1250,9 +1250,41 @@ class Star(Expression):
 
 
 class DatabaseDefault(Expression):
-    """Placeholder expression for the database default in an insert query."""
+    """
+    Expression to use DEFAULT keyword during insert otherwise the underlying expression.
+    """
+
+    def __init__(self, expression, output_field=None):
+        super().__init__(output_field)
+        self.expression = expression
+
+    def get_source_expressions(self):
+        return [self.expression]
+
+    def set_source_expressions(self, exprs):
+        (self.expression,) = exprs
+
+    def resolve_expression(
+        self, query=None, allow_joins=True, reuse=None, summarize=False, for_save=False
+    ):
+        resolved_expression = self.expression.resolve_expression(
+            query=query,
+            allow_joins=allow_joins,
+            reuse=reuse,
+            summarize=summarize,
+            for_save=for_save,
+        )
+        # Defaults used outside an INSERT context should resolve to their
+        # underlying expression.
+        if not for_save:
+            return resolved_expression
+        return DatabaseDefault(
+            resolved_expression, output_field=self._output_field_or_none
+        )
 
     def as_sql(self, compiler, connection):
+        if not connection.features.supports_default_keyword_in_insert:
+            return compiler.compile(self.expression)
         return "DEFAULT", []
 
 
--- a/docs/releases/5.0.8.txt
+++ b/docs/releases/5.0.8.txt
@@ -28,3 +28,7 @@ Bugfixes
 * Fixed a bug in Django 5.0 that caused a system check crash when
   ``ModelAdmin.date_hierarchy`` was a ``GeneratedField`` with an
   ``output_field`` of ``DateField`` or ``DateTimeField`` (:ticket:`35628`).
+
+* Fixed a bug in Django 5.0 which caused constraint validation to either crash
+  or incorrectly raise validation errors for constraints referring to fields
+  using ``Field.db_default`` (:ticket:`35638`).
--- a/django/contrib/postgres/constraints.py
+++ b/django/contrib/postgres/constraints.py
@@ -178,7 +178,7 @@ class ExclusionConstraint(BaseConstraint):
 
     def validate(self, model, instance, exclude=None, using=DEFAULT_DB_ALIAS):
         queryset = model._default_manager.using(using)
-        replacement_map = instance._get_field_value_map(
+        replacement_map = instance._get_field_expression_map(
             meta=model._meta, exclude=exclude
         )
         replacements = {F(field): value for field, value in replacement_map.items()}
--- a/django/db/models/constraints.py
+++ b/django/db/models/constraints.py
@@ -241,7 +241,7 @@ class CheckConstraint(BaseConstraint):
         return schema_editor._delete_check_sql(model, self.name)
 
     def validate(self, model, instance, exclude=None, using=DEFAULT_DB_ALIAS):
-        against = instance._get_field_value_map(meta=model._meta, exclude=exclude)
+        against = instance._get_field_expression_map(meta=model._meta, exclude=exclude)
         try:
             if not Q(self.condition).check(against, using=using):
                 raise ValidationError(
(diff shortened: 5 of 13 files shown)


django/db/models/base.py
```diff
@@ -1333,12 +1333,19 @@ def _get_next_or_previous_in_order(self, is_next):
             setattr(self, cachename, obj)
         return getattr(self, cachename)
 
-    def _get_field_value_map(self, meta, exclude=None):
+    def _wrap_value(self, value, field):
+        if hasattr(value, "resolve_expression"):
+            value.output_field = field
```
shangxiao, line 1338: Instead of doing this here we could set `output_field` when the `DatabaseDefault` is first instantiated

charettes: Were you thinking of something like

```diff
diff --git a/django/db/models/expressions.py b/django/db/models/expressions.py
index ffb9f3c816..5c33f6c6b2 100644
--- a/django/db/models/expressions.py
+++ b/django/db/models/expressions.py
@@ -1252,7 +1252,37 @@ def as_sql(self, compiler, connection):
 class DatabaseDefault(Expression):
     """Placeholder expression for the database default in an insert query."""

+    def __init__(self, expression, output_field=None):
+        super().__init__(output_field)
+        self.expression = expression
+
+    def get_source_expressions(self):
+        return [self.expression]
+
+    def set_source_expressions(self, exprs):
+        (self.expression,) = exprs
+
+    def resolve_expression(
+        self, query=None, allow_joins=True, reuse=None, summarize=False, for_save=False
+    ):
+        resolved_expression = self.expression.resolve_expression(
+            query=query,
+            allow_joins=allow_joins,
+            reuse=reuse,
+            summarize=summarize,
+            for_save=for_save,
+        )
+        # Defaults used outside of INSERTs context should resolved to
+        # their underlying expression.
+        if not for_save:
+            return resolved_expression
+        return DatabaseDefault(
+            resolved_expression, output_field=self._output_field_or_none
+        )
+
     def as_sql(self, compiler, connection):
+        if not connection.features.supports_default_keyword_in_insert:
+            return compiler.compile(self.expression)
         return "DEFAULT", []


diff --git a/django/db/models/fields/__init__.py b/django/db/models/fields/__init__.py
index 796c4d23c4..d1f31f0211 100644
--- a/django/db/models/fields/__init__.py
+++ b/django/db/models/fields/__init__.py
@@ -983,13 +983,7 @@ def get_internal_type(self):

     def pre_save(self, model_instance, add):
         """Return field's value just before saving."""
-        value = getattr(model_instance, self.attname)
-        if not connection.features.supports_default_keyword_in_insert:
-            from django.db.models.expressions import DatabaseDefault
-
-            if isinstance(value, DatabaseDefault):
-                return self._db_default_expression
-        return value
+        return getattr(model_instance, self.attname)

     def get_prep_value(self, value):
         """Perform preliminary non-db specific value checks and conversions."""
@@ -1031,7 +1025,9 @@ def _get_default(self):
         if self.db_default is not NOT_PROVIDED:
             from django.db.models.expressions import DatabaseDefault

-            return DatabaseDefault
+            return lambda: DatabaseDefault(
+                self._db_default_expression, output_field=self
+            )

         if (
             not self.empty_strings_allowed
```

shangxiao: I wasn't thinking of that but it's worth considering.  Because this overloads the responsibility of the class we just need to make sure we're ok the adaptive behaviour? 🤔 My initial thoughts are that it seems to fit nicely.

shangxiao: Also, I've been using the raw `db_default` when I should've been using `_db_default_expression` - I completely forgot about this 😅

shangxiao: Ok I tried this approach and it's better because we're consolidating into single-point-of-control 👍

django/db/models/base.py
```diff
@@ -1333,12 +1333,17 @@ def _get_next_or_previous_in_order(self, is_next):
             setattr(self, cachename, obj)
         return getattr(self, cachename)
 
-    def _get_field_value_map(self, meta, exclude=None):
+    def _get_field_expression_map(self, meta, exclude=None):
```
shangxiao, line 1336: @charettes we probably no longer need to do this since DatabaseDefault is handling the logic… or do we prefer to leave it in?

charettes: I haven't run the branch locally but from my understanding the logic below was still required as we wanted

1. To have literal value wrapped in `Value`
2. To have non-literal expressions, such as `DatabaseDefault`, be returned as is

shangxiao: Ah yes I see now `Value` does its own `as_sql()` so we do still need it 👍

shangxiao: As a side note I feel like I've seen this optional wrapping pattern a few times I wonder if there's any benefit to making a factory method along the lines of `Value.wrap_if_not_expression(value_maybe_expression)`

sarahboyce (approved):
Thank you @shangxiao! 
Did some testing this morning and works well - this looks good to me :star:

django/db/models/base.py
charettes: Possibly but since this will be backported it should be done in another PR.

django/db/models/expressions.py
```diff
@@ -1250,9 +1250,41 @@ def as_sql(self, compiler, connection):
...
+    ):
+        resolved_expression = self.expression.resolve_expression(
+            query=query,
+            allow_joins=allow_joins,
+            reuse=reuse,
+            summarize=summarize,
+            for_save=for_save,
+        )
+        # Defaults used outside of INSERTs context should resolved to
+        # their underlying expression.
```
charettes, line 1278:
```suggestion
        # Defaults used outside of INSERTs context should resolve to
        # their underlying expression.
```

docs/releases/5.0.8.txt
```diff
@@ -23,3 +23,7 @@ Bugfixes
 * Fixed a regression in Django 5.0.7 that caused a crash in
   ``LocaleMiddleware`` when processing a language code over 500 characters
   (:ticket:`35627`).
+
+* Fixed a bug in Django 5.0 which caused ``Model.validate_constraints()`` to
+  incorrectly raise validation errors for constraints referring to fields using
+  ``Field.db_default`` (:ticket:`35638`).
```
charettes, line 29: The issue was more wide spread than that as `validate_unique` which powers `Field.unique=True`, `Meta.unique_together`, and `Date(Time?)Field.unique_for_(date, month, year)` was also affected. Maybe something along the lines of 

```suggestion
* Fixed a bug in Django 5.0 which caused constraint validation to either crash
  or incorrectly raise validation errors for constraints referring to fields using
  ``Field.db_default`` (:ticket:`35638`).
```

charettes (commented):
Patch is looking good I think we should test the resolved issue in `Model.validate_unique` as well

```
diff --git a/tests/validation/models.py b/tests/validation/models.py
index 653be4a239..ed88750364 100644
--- a/tests/validation/models.py
+++ b/tests/validation/models.py
@@ -48,7 +48,7 @@ def clean(self):

 class UniqueFieldsModel(models.Model):
     unique_charfield = models.CharField(max_length=100, unique=True)
-    unique_integerfield = models.IntegerField(unique=True)
+    unique_integerfield = models.IntegerField(unique=True, db_default=42)
     non_unique_field = models.IntegerField()


diff --git a/tests/validation/test_unique.py b/tests/validation/test_unique.py
index 4a8b3894f0..36ee6e9da0 100644
--- a/tests/validation/test_unique.py
+++ b/tests/validation/test_unique.py
@@ -146,6 +146,20 @@ def test_primary_key_unique_check_not_performed_when_not_adding(self):
             mtv = ModelToValidate(number=10, name="Some Name")
             mtv.full_clean()

+    def test_unique_db_default(self):
+        UniqueFieldsModel.objects.create(unique_charfield="foo", non_unique_field=42)
+        um = UniqueFieldsModel(unique_charfield="bar", non_unique_field=42)
+        with self.assertRaises(ValidationError) as cm:
+            um.full_clean()
+        self.assertEqual(
+            cm.exception.message_dict,
+            {
+                "unique_integerfield": [
+                    "Unique fields model with this Unique integerfield already exists."
+                ]
+            },
+        )
+
     def test_unique_for_date(self):
         Post.objects.create(
             title="Django 1.0 is released",
```

PR #18426, opened 2024-07-30
Fixed #35640 : ModelMultipleChoiceField validation of inherited models.

omarbenhamid:
# Trac ticket number

ticket-35640

# Branch description
ModelMultipleChoiceField was failing to validate inherited models due to a difference between choice value generation and check. Made both consistant.

# Checklist
- [X] I have attached screenshots in both light and dark modes for any UI changes.

### DIFF
--- a/django/forms/models.py
+++ b/django/forms/models.py
@@ -1647,7 +1647,7 @@ class ModelMultipleChoiceField(ModelChoiceField):
                     params={"pk": pk},
                 )
         qs = self.queryset.filter(**{"%s__in" % key: value})
-        pks = {str(getattr(o, key)) for o in qs}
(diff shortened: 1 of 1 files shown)


shangxiao:
The change proposed doesn't appear to change behaviour because both `_check_values()` and `prepare_value()` take into consideration having a `to_field_name`.

I'm not entirely sure there is an issue here.  If you create a unit test or a sample project to demonstrate a bug then we can reopen the ticket.

I'd originally closed the ticket because it was reporting an issue with 3.2 which isn't supported but also not sure this patch is fixing anything.

Feel free to reopen if you can confirm there is an issue👍

PR #18429, opened 2024-07-30
Extended script to manage translations to support fetching new translations since a given date.

nessita:
# Branch description

Fetching translations from Transifex in preparation for a final release is cumbersome and time-consuming. This work aims to minimize manual effort by reducing the amount of outdated translations fetched. It now only retrieves translations from Transifex for a language and resource if they have been updated since a specified `since` date.

This allows translation and release managers to fetch only new translations since the last update (which occurred likely during the previous final release, or in a following patch release), rather than downloading and manually reviewing over 1200 files as was previously necessary.

Related to https://forum.djangoproject.com/t/improving-the-process-of-incorporating-translations-to-django/26140

### DIFF
--- a/scripts/manage_translations.py
+++ b/scripts/manage_translations.py
@@ -20,13 +20,86 @@
 
 import os
 from argparse import ArgumentParser
+from collections import defaultdict
+from configparser import ConfigParser
+from datetime import datetime
 from subprocess import run
+
+import requests
 
 import django
 from django.conf import settings
 from django.core.management import call_command
 
 HAVE_JS = ["admin"]
+LANG_OVERRIDES = {
+    "zh_CN": "zh_Hans",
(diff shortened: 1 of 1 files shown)


scripts/manage_translations.py
```diff
@@ -20,13 +20,86 @@
...
+    "zh_CN": "zh_Hans",
+    "zh_TW": "zh_Hant",
+}
+
+
+def list_resources_with_updates(date_since, date_skip=None, verbose=False):
+    resource_lang_changed = defaultdict(list)
+    resource_lang_unchanged = defaultdict(list)
+
+    # Read token from ENV, otherwsie read from the ~/.transifexrc file.
```
claudep, line 45: otherwsie => otherwise

PR #18433, opened 2024-07-31
Fixed #35646 -- Extended SafeExceptionReporterFilter.hidden_settings to treat `AUTH` as a sensitive match.

nessita:
# Trac ticket number
ticket-35646

# Branch description
Harden the filtering of settings and meta values during error reporting.

# Checklist

### DIFF
--- a/docs/howto/error-reporting.txt
+++ b/docs/howto/error-reporting.txt
@@ -282,7 +282,11 @@ following attributes and methods:
 
             import re
 
-            re.compile(r"API|TOKEN|KEY|SECRET|PASS|SIGNATURE|HTTP_COOKIE", flags=re.IGNORECASE)
+            re.compile(r"API|AUTH|TOKEN|KEY|SECRET|PASS|SIGNATURE|HTTP_COOKIE", flags=re.IGNORECASE)
+
+        .. versionchanged:: 5.2
+
+            The term ``AUTH`` was added.
 
     .. method:: is_active(request)
 
--- a/docs/releases/5.2.txt
+++ b/docs/releases/5.2.txt
@@ -150,7 +150,8 @@ Email
 Error Reporting
 ~~~~~~~~~~~~~~~
 
-* ...
+* The attribute :attr:`.SafeExceptionReporterFilter.hidden_settings` now
+  treats values as sensitive if their name includes ``AUTH``.
 
 File Storage
 ~~~~~~~~~~~~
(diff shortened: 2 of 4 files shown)


claudep (approved):
Looks good!

docs/releases/5.2.txt
```diff
@@ -150,7 +150,8 @@ Email
 Error Reporting
 ~~~~~~~~~~~~~~~
 
-* ...
+* The attribute :attr:`.SafeExceptionReporterFilter.hidden_settings` now
+  treats values as sensitive if their key name include ``AUTH``.
```
jacobtylerwalls, line 154:
```suggestion
  treats values as sensitive if their key name includes ``AUTH``.
```

jacobtylerwalls: Or keep everything singular:  "treats a value as sensitive if its key name includes ``AUTH``."

Also, we might improve on "key name", since it's not always the case that users interact with their settings as a dictionary.

Maybe:

"Settings including ``AUTH`` in their name are now treated as sensitive by the attribute ...`"

docs/howto/error-reporting.txt
```diff
@@ -282,7 +282,7 @@ following attributes and methods:
 
             import re
 
-            re.compile(r"API|TOKEN|KEY|SECRET|PASS|SIGNATURE|HTTP_COOKIE", flags=re.IGNORECASE)
+            re.compile(r"API|AUTH|TOKEN|KEY|SECRET|PASS|SIGNATURE|HTTP_COOKIE", flags=re.IGNORECASE)
```
jacobtylerwalls, line 285: Should we add .. versionchanged?

nessita: I pondered about this, I was undecided because our docs say that `versionchanged` is [for new features](https://docs.djangoproject.com/en/dev/internals/contributing/writing-documentation/#documenting-new-features), but given that you are also highlighting this, it may be worth adding it.

Will push this change shortly. Thank you!

docs/releases/5.2.txt
nessita: @jacobtylerwalls I opted for using "if their name includes..." (removing `key`) because I specifically want to avoid mentioning `settings` since a common use case for this filter is to sanitize request.META which is not about settings at all (but header names).

sarahboyce (approved):
:rocket:

PR #18434, opened 2024-07-31
Fixed #35648 -- Extended SafeString to support __radd__.

matthiask:
# Trac ticket number

ticket-35648

# Branch description

`SafeString.__add__` should return `NotImplemented` if it isn't sure it can handle the RHS so that the RHS gets a chance at running its `__radd__` method.

# Checklist

### DIFF
--- a/django/utils/safestring.py
+++ b/django/utils/safestring.py
@@ -35,10 +35,16 @@ class SafeString(str, SafeData):
         Concatenating a safe string with another safe bytestring or
         safe string is safe. Otherwise, the result is no longer safe.
         """
-        t = super().__add__(rhs)
-        if isinstance(rhs, SafeData):
-            return SafeString(t)
-        return t
+        if isinstance(rhs, str):
+            t = super().__add__(rhs)
+            if isinstance(rhs, SafeData):
+                t = SafeString(t)
+            return t
+
+        # Give the rhs object a chance to handle the addition, for example if
+        # the rhs object's class implements `__radd__`. More details:
+        # https://docs.python.org/3/reference/datamodel.html#object.__radd__
+        return NotImplemented
 
     def __str__(self):
         return self
--- a/docs/releases/5.2.txt
+++ b/docs/releases/5.2.txt
@@ -262,7 +262,10 @@ URLs
 Utilities
 ~~~~~~~~~
 
-* ...
+* :class:`~django.utils.safestring.SafeString` now raises
+  :exc:`NotImplementedError` in ``__add__`` for non-string right-hand side
+  values. This aligns with the :py:class:`str` addition behavior and allows
+  ``__radd__`` to be used if available.
 
 Validators
 ~~~~~~~~~~
--- a/tests/utils_tests/test_safestring.py
+++ b/tests/utils_tests/test_safestring.py
@@ -132,3 +132,54 @@ class SafeStringTest(SimpleTestCase):
         for case, expected in cases:
             with self.subTest(case=case):
                 self.assertRenderEqual("{{ s }}", expected, s=s + case)
+
+    def test_add_obj(self):
+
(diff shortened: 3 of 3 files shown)


nessita (commented):
Hello @matthiask! Thank you for this work, it looks like a great start! :star2: 

A few initial comments:

* It would be great to have an extra commit with some extra tests for the add operation which is not fully covered by tests, I was thinking on something like this:
```python
    def test_add_str(self):
        s = SafeString("a&b")
        cases = [
            ("test", "a&amp;btest"),
            ("<p>unsafe</p>", "a&amp;b&lt;p&gt;unsafe&lt;/p&gt;"),
            (SafeString("<p>safe</p>"), SafeString("a&b<p>safe</p>")),
        ]
        for case, expected in cases:
            with self.subTest(case=case):
                self.assertRenderEqual("{{ s }}", expected, s=s + case)
```
* I think we could grow your test with this:
```python
    def test_add_obj(self):
        class Strange:
            def __str__(self):
                return "strange"

            def __add__(self, other):
                return str(self) + other

            def __radd__(self, other):
                return other + str(self)

        s = Strange()
        self.assertEqual("hello_" + s, "hello_strange")
        self.assertEqual(s + "_result", "strange_result")
        self.assertEqual(mark_safe("hello_") + s, "hello_strange")
```
* We would need a small entry in the release notes for 5.2 mentioning this change (in present tense).

matthiask:
Thank you! I have added the changes to the tests and a release note.

I thought a bit about merging your `test_add_str` with https://github.com/django/django/blob/8cf931dd2fde2561a615b96d5f709c15b672c4ba/tests/utils_tests/test_safestring.py#L71-L78 but I don't it makes the test case clearer, so I left both of them as they are now.

django/utils/safestring.py
```diff
@@ -35,10 +35,12 @@ def __add__(self, rhs):
         Concatenating a safe string with another safe bytestring or
         safe string is safe. Otherwise, the result is no longer safe.
         """
-        t = super().__add__(rhs)
         if isinstance(rhs, SafeData):
-            return SafeString(t)
-        return t
+            return SafeString(super().__add__(rhs))
+        if isinstance(rhs, str):
+            return super().__add__(rhs)
```
nessita, line 41: I think a slightly more clear layout for this logic could be:

```suggestion
        if isinstance(rhs, str):
            t = super().__add__(rhs)
            if isinstance(rhs, SafeData):
                t =  SafeString(t)
            return t
```

nessita (commented):
Thank you @matthiask! This is looking almost ready! :star2: 

Besides the code suggestion I added, I have two requests:
1. Can you add a new test ensuring that if a class implements an `__radd__` method such that the result *is* a `SafeString`, the addition with a `SafeString` also results in a `SafeString`?
2. Could you please restructure the new tests so they use `subTest`, similar to the test I provided for `test_add_str`?

Lastly, if you could cleanup the commits to have 2 isolated commits, a first commit with the new `test_add_str`, and a second commit with the ticket fix + texts.

matthiask:
Hi @nessita 

Thanks for your comments! I have applied the requested change to the `SafeString` implementation.

I have also worked on the tests. The class `Strange` is again splitted into a selection of classes which all only ever override `__add__` or `__radd__` -- I think it's clearer what's going on because it's immediately obvious which of the methods is executed. That isn't the case when overriding both in the same class.

I have also added two subtests showing that the `SafeData` marker is preserved when safe strings are involved, and nothing else.

Thanks again!

nessita (approved):
Thank you @matthiask , stellar work! :star2: 

I pushed minor edits to release notes and the test, I'll merge when CI is green. Well done!

matthiask:
Thank you! That was a great experience :)

PR #18438, opened 2024-08-01
Fixed #35650 - Added multi select checkboxes with shift in TabularInlineAdmin

tanaydin:
Added multi select checkboxes with shift in TabularInlineAdmin

# Trac ticket number

ticket-35650

# Branch description
This changes are adding functionally for selecting multi checkboxes by pressing shift key with adding a class to TabularInlineAdmin form table tag, and a javascript function.

# Checklist


nessita:
@tanaydin Thank you for your work on this! :star2: 

(Repeating my comment from the ticket below.)

I took a quite look at this PR, and I was thinking that instead of providing new javascript, we should try to re-use/generalize what the changelist page is already using. Could you look into that?

tanaydin:
sure I'll check that way.

sarahboyce:
@tanaydin this will need a test to be accepted :+1: you may need to consider adding to the JS tests in `js_tests/admin/actions.test.js` and/or writing a selenium test  (see `tests.admin_changelist.tests.SeleniumTests` for some examples)

tanaydin:
Hi @sarahboyce,

Thanks for explanation and checks. I'll work on this.

django/contrib/admin/static/admin/js/actions.js
```diff
@@ -51,29 +51,33 @@
...
+                    sel: sel,
+                    cnt: actions_icnt
+                }, true);
+            const allToggle = document.getElementById(options.allToggleId);
+            allToggle.checked = sel === actionCheckboxes.length;
+            if (allToggle.checked) {
+                showQuestion(options);
+            } else {
+                clearAcross(options);
+            }
```
sarahboyce, line 80:
```suggestion
        const sel = Array.from(actionCheckboxes).filter(function(el) {
            return el.checked;
        }).length;
        const counter = document.querySelector(options.counterContainer);
        // data-actions-icnt is defined in the generated HTML
        // and contains the total amount of objects in the queryset
        const actions_icnt = Number(counter.dataset.actionsIcnt);
        counter.textContent = interpolate(
            ngettext('%(sel)s of %(cnt)s selected', '%(sel)s of %(cnt)s selected', sel), {
                sel: sel,
                cnt: actions_icnt
            }, true);
        const allToggle = document.getElementById(options.allToggleId);
        allToggle.checked = sel === actionCheckboxes.length;
        if (allToggle.checked) {
            showQuestion(options);
        } else {
            clearAcross(options);
```
As `updateCounter` is not called for inlines, we should be able to remove the if

django/contrib/admin/static/admin/js/actions.js
```diff
@@ -142,28 +142,70 @@
...
+            const handleCheckboxChange = (event) => {
                 const target = event.target;
-                if (target.classList.contains('action-select')) {
+
+                if (!lastChecked || !target.name.endsWith('-DELETE')) {
+                    lastChecked = target;
+                    return;
+                }
+
+                if (lastChecked.name.slice(0, -9) === target.name.slice(0, -9)) {
```
sarahboyce, line 188: This is a bit magic, can you explain what you're doing here?

django/contrib/admin/static/admin/js/actions.js
```diff
@@ -142,28 +142,70 @@
...
+                if (!element) {
+                    return;
                 }
-            }
-        });
+                element.addEventListener('change', handleCheckboxChange);
+            };
+
+            // Handle tabular inline tables
+            document.querySelectorAll('.tabular_inline_table tbody').forEach(attachChangeListener);
```
sarahboyce, line 204:
```suggestion
            document.querySelectorAll('.tabular tbody').forEach(attachChangeListener);
```
Then we can remove adding `class="tabular_inline_table`

sarahboyce (commented):
Thank you @tanaydin :+1: 
Have a few comments

django/contrib/admin/static/admin/js/actions.js
tanaydin: It's checking if both check boxes are in the same form. Without this, it's possible to toggle check boxes with different inline admins which shouldn't be.

jacobtylerwalls:
Hi @tanaydin. If you're caught up responding to Sarah's feedback, can you please uncheck "Patch needs improvement" on the ticket so it goes back in the review queue? Thanks!

PR #18445, opened 2024-08-04
Fixed #35656 -- Added support for changing migration autodetector.

LeOndaz:
# Trac ticket number

ticket-35656

# Branch description
Allow changing autodetector class for migration commands

# Checklist


LeOndaz:
I missed naming my branch to the ticket number :(

LeOndaz:
I will PR again from a different branch

LeOndaz:
Any clue why some checks are failing?

LeOndaz:
@nessita thanks!

sarahboyce:
Thank you for this @LeOndaz :star: 
I think as we are doing this to allow people to extend this, we should treat this as a new feature and so this needs tests, docs and a release note :+1:

LeOndaz:
@sarahboyce I agree, I'm on it

LeOndaz:
@sarahboyce I should be documenting the Autodetector generally, not just the feature, since it's not documented afaik, right?

I'm tempted to document the migration execution flow here (which includes the autodetector), what do you think?

sarahboyce:
> @sarahboyce I should be documenting the Autodetector generally, not just the feature, since it's not documented afaik, right?
> 
> I'm tempted to document the migration execution flow here (which includes the autodetector), what do you think?

Oooh I don't know actually :thinking: by documenting it, it becomes part of the public API
Certainly willing to see a PR, but perhaps separate to this?

sarahboyce:
> I guess 🤔 so for now tests without docs ? maybe we should ask for a third opinion, what do you think

Let's add tests and a release note and then can have a look at whether further docs make sense

LeOndaz:
> > I guess 🤔 so for now tests without docs ? maybe we should ask for a third opinion, what do you think
> 
> Let's add tests and a release note and then can have a look at whether further docs make sense

okay, you can count on me

LeOndaz:
@sarahboyce Hey, I figured out that custom migration commands ARE NOT testable,

Code:

```
from django.core.management.migration import MigrationCommand
from django.db.migrations import MigrationAutodetector


class CustomAutodetector(MigrationAutodetector):
    def changes(self, *args, **kwargs):
        print("HELLO? ANYONE HERE?")
        print(args, kwargs)
        return super().changes(*args, **kwargs)


class CustomMigrateCommand(MigrationCommand):
    autodetector_class = CustomAutodetector
```


1. Can't add to MigrateTests since this is a different command2
2. Can't test in `user_commands` because user commands because they can't receive `database`, `run_syncdb`
```
TypeError: Unknown option(s) for migrate command: database, run_syncdb. Valid options are: force_color, help, interactive, no_color, no_input, pythonpath, settings, skip_checks, stderr, stdout, traceback, verbosity, version.
```

I'm not sure if there're ways to test it as of now, my opinion is that we should think of this ticket as a beginning for supporting custom migration commands, and the tests should  happen in the same ticket that implements `settings.DJANGO_DEFAULT_AUTODETECTOR` for example

What do you think ?

sarahboyce:
See `tests.admin_scripts.tests.CommandTypes.test_custom_stdout` where a command is defined in the test and use `call_command` on it directly like...
```python
out = StringIO()
command = Command(stdout=out)
call_command(command)
```

LeOndaz:
Thanks @sarahboyce , I'm still wrapping my head around the test

I will suggest an approach and please, tell me your opinion

Instead of writing a new test, I can parameterize the `MigrateTests.all_methods` and `MakeMigrationsTests.all_methods` so that they all get tested with the builtin migrate/makemigrations commands and custom ones, and both should pass, am I correct?

LeOndaz:
Note the repetitiveness, I couldn't find a way to parameterize the whole `TestCase`

sarahboyce:
As an example in the `makemigrations` case, I was thinking something like:
```diff
--- a/tests/migrations/test_commands.py
+++ b/tests/migrations/test_commands.py
@@ -2801,6 +2801,32 @@ class MakeMigrationsTests(MigrationTestBase):
             with self.assertRaisesMessage(CommandError, msg):
                 call_command("makemigrations", "book_app", update=True)
 
+    def test_makemigrations_custom_autodetector(self):
+        from django.core.management.commands.makemigrations import Command
+        from django.db.migrations.autodetector import MigrationAutodetector
+
+        class CustomAutodetector(MigrationAutodetector):
+            def changes(self, graph, trim_to_apps=None, convert_apps=None, migration_name=None):
+                # Detects no changes.
+                return {}
+
+        class CustomMakeMigrationCommand(Command):
+            autodetector_class = CustomAutodetector
+
+        self.assertTableNotExists("migrations_unicodemodel")
+        apps.register_model("migrations", UnicodeModel)
+        out = io.StringIO()
+        command = CustomMakeMigrationCommand(stdout=out)
+        with self.temporary_migration_module() as migration_dir:
+            call_command(command, "migrations", verbosity=0)
+
+            # No empty __init__.py file in migrations folder.
+            init_file = os.path.join(migration_dir, "__init__.py")
+            self.assertFalse(os.path.exists(init_file))
+            # No initial migration file created.
+            initial_file = os.path.join(migration_dir, "0001_initial.py")
+            self.assertFalse(os.path.exists(initial_file))
+
 
 class SquashMigrationsTests(MigrationTestBase):
```
I also think this design doesn't quite handle Claude's consern that:
> the autodetector class should be the same in makemigrations as the one in migrate command, so it might not be trivial to define it so it doesn't require the user to customize the class separately in both commands.

We still need to overwrite both commands - `MigrationCommand` makes it _look_ like we can define it in one place (and for Django it would be) but actually a project would need to override both - maybe that's ok (we can have docs and a system check to make sure they are the same, this is too niche for a setting)

LeOndaz:
Your idea looks good, and yes, Claude's wasn't addressed in the solutions introduced, isn't adding it to settings an option to be considered? Maybe this solves Claude's concern, not necessarily in the same ticket

Advantages:

1. People won't need to subclass the commands
2. It will be the same autodetector for both classes
3. imo, it makes sense to have a setting for it

sarahboyce:
I agree it _feels_ like a setting, but the issue is a lot of things feel like a setting and we try to prevent this from becoming a dumping ground. A new setting will need a discussion with the community to agree that it's a good idea (and community discussions can be a lot of work)
Personally, I'd prefer we add this without a setting, and then see if this gets used enough to justify the setting, which could be added in a future release

As system checks are a built-in way to alert people that they have done something wrong - I think this is worth investigating - then a setting is probably not required. They still need to override the two commands but we mitigate against them overriding one and not the other

LeOndaz:
@sarahboyce let me know what you think when you have time

LeOndaz:
@sarahboyce I can see some tests are failing, but I didn't touch them, can you check when you have the time?

sarahboyce:
> @sarahboyce I can see some tests are failing, but I didn't touch them, can you check when you have the time?

The test failures are related. Looks like there is an issue that we are now registering UnicodeModel twice. We might want to create our own model for this test

LeOndaz:
@sarahboyce I think the failing doc test shouldn't be fixed, it's because I referenced the autodetector, but it doesn't yet have something to reference, but having it now is better than forgetting it later imo, what do you think?

sarahboyce:
> @sarahboyce I think the failing doc test shouldn't be fixed, it's because I referenced the autodetector, but it doesn't yet have something to reference, but having it now is better than forgetting it later imo, what do you think?

If you add docs for `MigrationAutodetector.autodetector_class` then you can ignore it for now. We won't merge anything with failing checks (so it will need to be fixed eventually :+1: )

LeOndaz:
@sarahboyce thanks for responding, I removed `:attr:` 

I will write a migration execution flow reference (I actually started weeks ago, but stopped mid because I'm not sure if it will be merged). I will try to build consent to make sure everyone agrees that we should have that as a part of the docs, and if it's fine, that would be my ticket, for now, I'll create a ticket for documenting the migration execution flow on the tracker, does this sound good for you?

LeOndaz:
wops, I missed up so bad, I will need to resubmit this in a separate PR, I will mention you there

PR #18448, opened 2024-08-05
Renamed BaseConstraint._check() to check()

Sourav6971:
Renamed the BaseConstraint._check() to check()


jacobtylerwalls:
Thanks for your interest in contributing to Django, but we discuss the motivations for changes in the [ticket tracker](https://code.djangoproject.com).

Sourav6971:
Yeah got a comment there too. So sorry, just got carried away with the excitement of solving my first issue

PR #18449, opened 2024-08-05
Fixed #35657 -- Made FileField handle db_default values.

sarahboyce:
# Trac ticket number

ticket-35657

# Branch description

# Checklist

### DIFF
--- a/django/db/models/fields/files.py
+++ b/django/db/models/fields/files.py
@@ -9,6 +9,7 @@ from django.core.files.images import ImageFile
 from django.core.files.storage import Storage, default_storage
 from django.core.files.utils import validate_file_name
 from django.db.models import signals
+from django.db.models.expressions import DatabaseDefault
 from django.db.models.fields import Field
 from django.db.models.query_utils import DeferredAttribute
 from django.db.models.utils import AltersData
@@ -195,6 +196,12 @@ class FileDescriptor(DeferredAttribute):
         # handle None.
         if isinstance(file, str) or file is None:
             attr = self.field.attr_class(instance, self.field, file)
+            instance.__dict__[self.field.attname] = attr
+
+        # If this value is a DatabaseDefault, initialize the attribute class
+        # for this field with its db_default value.
+        elif isinstance(file, DatabaseDefault):
+            attr = self.field.attr_class(instance, self.field, self.field.db_default)
             instance.__dict__[self.field.attname] = attr
 
         # Other types of files may be assigned as well, but they need to have
--- a/docs/releases/5.0.8.txt
+++ b/docs/releases/5.0.8.txt
@@ -32,3 +32,6 @@ Bugfixes
 * Fixed a bug in Django 5.0 which caused constraint validation to either crash
   or incorrectly raise validation errors for constraints referring to fields
   using ``Field.db_default`` (:ticket:`35638`).
+
+* Fixed a crash in Django 5.0 when saving a model containing a ``FileField``
+  with a ``db_default`` set (:ticket:`35657`).
--- a/tests/file_storage/models.py
+++ b/tests/file_storage/models.py
@@ -72,6 +72,9 @@ class Storage(models.Model):
     default = models.FileField(
         storage=temp_storage, upload_to="tests", default="tests/default.txt"
     )
+    db_default = models.FileField(
+        storage=temp_storage, upload_to="tests", db_default="tests/db_default.txt"
+    )
     empty = models.FileField(storage=temp_storage)
     limited_length = models.FileField(
         storage=temp_storage, upload_to="tests", max_length=20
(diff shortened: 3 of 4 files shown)


docs/releases/5.0.8.txt
```diff
@@ -28,3 +28,6 @@ Bugfixes
 * Fixed a bug in Django 5.0 that caused a system check crash when
   ``ModelAdmin.date_hierarchy`` was a ``GeneratedField`` with an
   ``output_field`` of ``DateField`` or ``DateTimeField`` (:ticket:`35628`).
+
+* Fixed a crash when saving a model with a ``FileField`` with a ``db_default``
```
nessita, line 32: Perhaps "including" or "defining" to avoid the "with a" repetition?
```suggestion
* Fixed a crash when saving a model with a ``FileField`` including a ``db_default``
```

tests/file_storage/models.py
```diff
@@ -72,6 +72,9 @@ def pathlib_upload_to(self, filename):
     default = models.FileField(
         storage=temp_storage, upload_to="tests", default="tests/default.txt"
     )
+    db_default = models.FileField(
+        storage=temp_storage, upload_to="tests", db_default="tests/db_default.txt"
+    )
```
nessita, line 77: I was wondering if we should add a test for an empty db default.. From the docs I'm unclear about the semantics of `db_default=""` which I think it's a valid value for this param?

tests/file_storage/tests.py
```diff
@@ -944,6 +944,18 @@ def test_filefield_default(self):
         self.assertEqual(obj.default.read(), b"default content")
         obj.default.close()
 
+    def test_filefield_db_default(self):
+        temp_storage.save("tests/db_default.txt", ContentFile("default content"))
+        obj = Storage.objects.create()
+        self.assertEqual(obj.db_default.name, "tests/db_default.txt")
+        self.assertEqual(obj.db_default.read(), b"default content")
+        obj.db_default.close()
```
nessita, line 952: Do we need this call to be a cleanup call?
```suggestion
        self.addCleanup(obj.db_default.close)
```

tests/file_storage/tests.py
```diff
@@ -944,6 +944,18 @@ def test_filefield_default(self):
...
+
+        obj.delete()
+        obj = Storage()
+        self.assertEqual(obj.db_default.read(), b"default content")
```
nessita, line 956: I'm missing why we are not testing the `name` attr here :thinking:

tests/file_storage/tests.py
```diff
@@ -944,6 +944,18 @@ def test_filefield_default(self):
...
 
+    def test_filefield_db_default(self):
+        temp_storage.save("tests/db_default.txt", ContentFile("default content"))
+        obj = Storage.objects.create()
+        self.assertEqual(obj.db_default.name, "tests/db_default.txt")
+        self.assertEqual(obj.db_default.read(), b"default content")
+        obj.db_default.close()
+
+        obj.delete()
+        obj = Storage()
```
nessita, line 955: Do you think it'll be worth making this either two tests or a subTest-formatted test?

nessita (commented):
Thank you for building this PR so quickly!

tests/file_storage/models.py
sarahboyce: 7bcbf8d4adad720215058e0156da387e05ec744e I'm not sure the blank case makes sense, so I would be tempted to drop this commit. However, the behavior between `default=""` and `db_default=""` is the same so I think we're ok for this ticket/backport

tests/file_storage/tests.py
sarahboyce: It was copying the test from above but I've added this in now

tests/file_storage/models.py
nessita: Agreed we can drop this commit for this fix, though I wonder if there is any follow up (no release blocker) necessary? Feels odd to allow for an empty name when saving a file.

django/db/models/fields/files.py
```diff
@@ -197,6 +198,10 @@ def __get__(self, instance, cls=None):
             attr = self.field.attr_class(instance, self.field, file)
             instance.__dict__[self.field.attname] = attr
 
+        elif isinstance(file, DatabaseDefault):
```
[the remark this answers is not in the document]
sarahboyce: Added a comment, let me know what you think

tests/file_storage/models.py
sarahboyce: Yeah I think a system check so that you don't have blank default values might make sense, probably needs a deprecation warning, they should set null=True I think
Probably needs a forum discussion

tests/file_storage/tests.py
sarahboyce: This caused some failures here: https://github.com/django/django/actions/runs/10251452778/job/28359533468
I might have been doing it wrong but reverting for now

tests/file_storage/tests.py
nessita: I originally thought these were 2 independent test conditions/assertions but now with the added comment, I think I understand how the deletion is part of the tested flow.

nessita (approved):

I pushed small edits, will merge when CI is green.

